PR #25, opened 2019-03-27
Add sample of vev configuration in docs.configuration

tamuhey:
I added some lines to docs.configuration because there is no sample configuration of venv.

I was stuck at the configuration of venv, so I think this sample will help the users.


erictraut:
Thanks for the contribution.

PR #35, opened 2019-03-29
Update README.md syntax

dmey:
Minimal change but I found the sentence confusing at first...


erictraut:
Yeah, I agree the sentence was confusing and was a bit long-winded as well. I've taken a pass over the introductory portion of the README and streamlined it.

PR #55, opened 2019-04-05
keep undefine_check style consistent

ChrisLinn:


erictraut:
Is this addressing any functional problem, or is it just a stylistic preference? If it's just a stylistic preference, I'm inclined to not make the change because the code is currently all written by me, and I'm happy with the style. :)

PR #56, opened 2019-04-05
Add logging for TypeAnalyzer._applyDecorator

ChrisLinn:


erictraut:
I'm actively working on this code currently. Logging output to the console isn't the right approach here anyway. It should emit diagnostics that appear in VS Code's Problems panel.

ChrisLinn:
u r right

ChrisLinn:
oh just becoz i find  that   we compare `watch` with `undefined` in _server/src/pyright.ts_ and guessed we'd better follow this when it comes to an undefined obj (excluding arrayy, boolean, string...).

erictraut:
That's a different circumstance because `watch` is allowed to be defined as `undefined` or `null`, and it's important to distinguish between these two cases.

ChrisLinn:
thx for ur explanation

PR #60, opened 2019-04-07
Support for subclassing Namedtuple

mhzed:
Add type checking for classes extend from typing.NamedTuple.

### DIFF
--- a/server/src/analyzer/expressionEvaluator.ts
+++ b/server/src/analyzer/expressionEvaluator.ts
@@ -843,6 +843,30 @@ export class ExpressionEvaluator {
         let validatedTypes = false;
         let returnType: Type | undefined;
 
+        if (TypeUtils.isDataClass(type)) {
+            let constructorMethodType: FunctionType = new FunctionType(FunctionTypeFlags.InstanceMethod);
+            constructorMethodType.getParameters().push({
+                category: ParameterCategory.Simple,
+                name: 'self',
+                hasDefault: false,
+                type: type
+            });
+            type.getClassFields().forEach((s, k) => {
+                if (TypeUtils.isFunctionType(s.currentType) || k[0] === '_') {
+                    return;
+                }
+                constructorMethodType.getParameters().push({
+                    category: ParameterCategory.Simple,
+                    name: k,
+                    hasDefault: !type.getDataFields().get(k),
+                    type: s.currentType
+                });
+            });
+            let boundType = this._bindFunctionToClassOrObject(new ObjectType(type), constructorMethodType);
+            this._validateCallArguments(errorNode, argList, boundType, new TypeVarMap());
+            validatedTypes = true;
+        }
+
         // See if there's a "__new__" defined within the class (but not its base classes).
         let constructorMethodType = this._getTypeFromClassMemberName('__new__', type,
             MemberAccessFlags.SkipGetAttributeCheck | MemberAccessFlags.SkipInstanceMembers |
--- a/server/src/analyzer/typeAnalyzer.ts
+++ b/server/src/analyzer/typeAnalyzer.ts
@@ -165,6 +165,9 @@ export class TypeAnalyzer extends ParseTreeWalker {
             TypeUtils.addTypeVarsToListIfUnique(typeParameters,
                 TypeUtils.getTypeVarArgumentsRecursive(argType));
         });
+        if (TypeUtils.isDataClass(classType)) {
+            classType.getClassFields().merge(classType.getDataFields());
+        }
 
         // Update the type parameters for the class.
         if (classType.setTypeParameters(typeParameters)) {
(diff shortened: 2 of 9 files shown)


server/src/analyzer/expressionEvaluator.ts
```diff
@@ -849,6 +849,28 @@ export class ExpressionEvaluator {
         let validatedTypes = false;
         let returnType: Type | undefined;
 
+        if (TypeUtils.isDataClass(type)) {
+          let constructorMethodType: FunctionType = new FunctionType(FunctionTypeFlags.InstanceMethod);
```
erictraut: Nit: No need to specify type here because typescript can infer it correctly. Linter will complain about this because it's an unnecessary type annotation.

server/src/analyzer/expressionEvaluator.ts
```diff
@@ -849,6 +849,28 @@ export class ExpressionEvaluator {
...
+          constructorMethodType.getParameters().push({
```
erictraut: Please use four-space tabbing for consistency with remainder of source base.

server/src/analyzer/expressionEvaluator.ts
```diff
@@ -849,6 +849,28 @@ export class ExpressionEvaluator {
...
+            category: ParameterCategory.Simple,
+            name: 'self',
+            hasDefault: false,
+            type: type
+          });
+          type.getClassFields().forEach((s, k) => {
+            if (TypeUtils.isFunctionType(s.currentType) || k[0] === '_') { return; }
```
erictraut: Minor nit: For consistency with rest of code, please move curly brackets and return statements to their own lines.

server/src/analyzer/typeAnalyzer.ts
```diff
@@ -175,6 +178,14 @@ export class TypeAnalyzer extends ParseTreeWalker {
             this.walk(node.suite);
         });
 
+        if (TypeUtils.isDataClass(classType)) {
+          classType.getClassFields().forEach((s, k) => {
+            if (!TypeUtils.isFunctionType(s.currentType) && /^_/.test(k) && s.declarations) {
+              this._addError(`Data field name can not start with _`, s.declarations[0].node);
```
erictraut: Is this just a convention, or is it enforced by the python interpreter? If it's just a convention, we may want to add a switch to enable/disable this. For now, it's fine to add a TODO comment.

server/src/analyzer/typeUtils.ts
```diff
@@ -1143,4 +1143,13 @@ export class TypeUtils {
 
         return unionType;
     }
+
+    static isDataClass(type: ClassType): boolean {
+      return (type.getBaseClasses().length === 1 && type.getBaseClasses()[0].type.asString() === 'Type[NamedTuple]');
```
erictraut: This looks like a very fragile approach. What if NamedTuple isn't an immediate base class? Also, we shouldn't use the string representation (which is meant only as a human-readable form) for internal logic.

server/src/tests/namedtuple.test.ts
```diff
@@ -0,0 +1,27 @@
+/*
```
erictraut: Please add these test cases to the existing testAnalyzer.test.ts rather than creating a new test source file.

server/src/tests/samples/namedtuple.py
```diff
@@ -0,0 +1,15 @@
+from typing import NamedTuple
```
erictraut: Please rename this file to namedtuple1.py for consistency.

erictraut: Also, add a header comment that indicates the use of the test sample and internal comments that indicate where errors or warnings should be emitted by the type analyzer.

server/src/tests/samples/namedtuple2.py
```diff
@@ -0,0 +1,4 @@
+from typing import NamedTuple
```
erictraut: Same comments here.

erictraut:
Did you mean to close the PR?

server/src/analyzer/typeAnalyzer.ts
mhzed: This is enforced in Python 3.7, results in runtime error.  Couldn't find official reference to this.  I've added a TODO regardless.

server/src/analyzer/typeUtils.ts
mhzed: I agree, but am out of ideas.  Preferably a python like 'isinstance()' type of check needs to be performed here.  If there are any internal apis that perform this, please point them out.

mhzed: On further check, Python 3.7 accepts this syntax if only if NamedTuple is one of the direct base classes, there is no need to scan ancestor tree. The code should be:
```ts
    static isDataClass(type: ClassType): boolean {
        return type.getBaseClasses().some( v => v.type.asString() === 'Type[NamedTuple]');
    }
```
Now regarding the string literal check "v.type.asString() === 'Type[NamedTuple]'",  if there is a better check, please let me know.

mhzed:
Reopen

server/src/analyzer/typeUtils.ts
mhzed: The check is changed to:
```ts
        return (type.getBaseClasses().some( v => v.type instanceof ClassType && v.type.isBuiltIn()
            && v.type.getClassName() === 'NamedTuple' ));
```

erictraut:
@mhzed, thanks again for the contribution. I merged your changes, and I made a few modifications. Please review and let me know if I broke anything in the process.

mhzed:
Looks all good.  Glad to contribute and thank you for the help.  Now my python environment just got a little better.

mhzed:
Actually the tests failed, likely due to the NamedTuple check.    Note 'NamedTuple' is not the same as older 'namedtuple'.  I will dig a bit deeper...

mhzed:
I've created PR #65 that contains the fix with additional test case.

PR #63, opened 2019-04-09
Suggestions to update readme for VS Code clarifications

qubitron:
Clarifying differences between the Python extension and Pyright extension


erictraut:
Thanks Dan, looks good. Merging.

PR #65, opened 2019-04-10
Namedtuple fix

mhzed:
Fix the 'typing.NamedTuple' type check.


erictraut:
Thanks @mhzed. I thought I had run the tests before checking in my previous changes, but clearly I hadn't. Thanks for the fix. I needed to manually merge it because an interim change that I was working on created a merge conflict with your PR. So I'm going to close this PR.

PR #69, opened 2019-04-11
Fixed print compatible in Python 3

zhangkaizhao:


erictraut:
Thanks for the contribution.

PR #87, opened 2019-04-25
Allow --watch usage with filespec

grigoryvp:
If 'Boolean' is not specified, 'command-line-args' will set files list as a value of 'watch' property, not the 'files' property the code expects. Now 'pyright --watch *.py' correctly pass python files into the default 'files' property.


erictraut:
Good fix! Thanks for the contribution.

PR #115, opened 2019-05-07
Add support for multi-root workspace

mhzed:
Support for multiple workspace folders in vscode.

### DIFF
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -10,43 +10,79 @@
 */
 
 import * as path from 'path';
-import { ExtensionContext } from 'vscode';
+import { ExtensionContext, workspace as Workspace, TextDocument, OutputChannel, window as Window } from 'vscode';
 import { LanguageClient, LanguageClientOptions, ServerOptions, TransportKind } from 'vscode-languageclient';
 import { ProgressReporting } from './progress';
 
+let clients: Map<string, LanguageClient> = new Map();
+
 export function activate(context: ExtensionContext) {
+	let outputChannel: OutputChannel = Window.createOutputChannel('pyright');
 	let serverModule = context.asAbsolutePath(path.join('server', 'server.js'));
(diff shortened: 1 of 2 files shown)


erictraut:
Thanks for the contribution! I didn't have a change to review the change today, but I'll try to get to it tomorrow.

erictraut:
After doing some testing, I noticed that this feature breaks the ability to open and analyze files that are not within the folder hierarchy of the current project. That's way more important to me than supporting multi-root workspaces. I spent some time trying to get both features to work, but the solution will require more investigation. I don't have time to look into it now, so I'm going to revert this change until I do.

mhzed:
It worked fine for me locally.  I will look into it if you don't mind giving me a bit more details on how to reproduce your problems.

erictraut:
Try opening a python file that is not within any of the workspace roots. The language server isn't loaded for that file, so you don't get any type checking or other language service features.

mhzed:
Suppose:
1. Old behaviour if there is only one root: pyright loads for any python files.
2. Multiple roots: pyright loads only for python files inside one of the root dirs, since it will be impossible to tell which root the file belongs to otherwise.
Will this solution be satisfactory?  If so, I will take a stab.

erictraut:
I think it should still load and analyze files even if it's not within a root directory. As far as I know, other language tools for VS Code do this. Perhaps it should use a separate server instance specifically for files that are not within a root dir, using default configuration settings?

mhzed:
Took a deeper look, can't seem to find a way to achieve this without changing the library 'vscode-languageserver-node'.  An alternative would be to change pyright server: one instance for multiple roots, but this change will be considerably more extensive and time consuming.

How about disable loading pyright when multiple roots are detected as a workaround?  This is still considerably better than having hundreds of errors.

erictraut:
I will open an issue to investigate this further. 

In the meantime, you can disable pyright on a per-workspace basis. I don't think multi-root workspaces are very common, so I'm assuming this doesn't affect many people.

PR #137, opened 2019-05-27
Fix small typo that made list display incorrectly

ldorigo:


erictraut:
Thanks for the contribution.

PR #237, opened 2019-09-07
Add con-pyright

fannheyward:


erictraut:
Nice! Thanks for the contribution.

PR #288, opened 2019-10-15
Use built-in `json` module to format library search paths, when getting them from the Python interpreter

necaris:
When using the fallback code to get the Python library search paths
from the interpreter (i.e. `sys.path`), use the `json` module to
dump the list of paths as output. This enables us to use
Javascript's built-in parser on the result, rather than an ad-hoc
parsing routine.

Note that the `json` module is available in all versions of Python 3
so backward compatibility should not be an issue.


erictraut:

PR #357, opened 2019-11-24
type: ignore

cshesse:
In my tests `# type ignore` did not seem to work, but `# type: ignore` did.


erictraut:
Yes, this was a bug in the documentation. Thanks for the fix.

PR #384, opened 2019-12-02
fixed typo in configoption file

heejaechang:
added .vs as one of git ignore directory and fixed typo in configOptions.ts

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,9 @@ node_modules
 # Optional npm cache directory
 .npm
 
+# vs cache directory
+.vs
+
 # Output files
 dist
 server/out/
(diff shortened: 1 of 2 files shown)


heejaechang:
tagging @msfterictraut 

I would like to add test but not sure how I can run tests in vscode. if I do debug (Pyright Tests) in vscode, break point doesn't get hit with message "sourcemap might be missing" which is weird since I can see js and js.map file in out/server dir

erictraut:
Yeah, running the tests in with the VS Code debugger is a little weird because of the way the client and server are packaged within the VS Code extension. You need to open another VS Code for the "server" directory and run it there. At least that's the best solution I've come up with so far.

PR #436, opened 2019-12-19
Introducing customBuiltins in pyright config file

willydouhard:
In specific cases it is useful to add custom builtins to get auto completion and type checking without importing a module.

### DIFF
--- a/client/schemas/pyrightconfig.schema.json
+++ b/client/schemas/pyrightconfig.schema.json
@@ -72,6 +72,17 @@
       ],
       "pattern": "^(.*)$"
     },
+    "customBuiltins": {
+      "$id": "#/properties/customBuiltins",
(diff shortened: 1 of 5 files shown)


c3-willydouhard:
Thank you for your quick reply!
I will try to build the latest master and see if everything works for me.
Last question, how does the vscode extension release cycle work ? In other words, how much time will it take for this change to be in the vscode extension ?

erictraut:
I typically release the VS Code extension two or three times a week, so you won't have to wait for long. Here's the [release history](https://github.com/microsoft/pyright/releases).

c3-willydouhard:
Okay thank you for the clarification.
I confirm that your last commit fix the issue 👍. Feel free to close this PR.
Thank you very much!

PR #467, opened 2020-01-02
Correction of README.md

rubenbelem:
There was two "VS Code Integration" sessions, and the first had an incomplete content similar to the content of the second one.


erictraut:
Thanks for the contribution.

rubenbelem:
You're welcome =)!

PR #481, opened 2020-01-10
Refactor server code into a class

MikhailArkhipov:
- Move server code into a base class and derive specific server from it.
- Add debugging instructions


erictraut:
Looks good to me. Feel free to merge when you think it's ready to go.

PR #487, opened 2020-01-16
Add parameter to language server base to pass root directory

MikhailArkhipov:
This is to resolve typeshed fallback when bundle setup is different from standard client.


server/src/languageServerBase.ts
```diff
@@ -62,8 +59,10 @@ export class LanguageServerBase {
     private _isDisplayingProgress = false;
     private _defaultWorkspacePath = '<default>';
 
-    constructor(private _productName: string) {
+    constructor(private _productName: string, rootDirectory?: string) {
         this._connection.console.log(`${_productName} language server starting`);
+        // Stash the base directory into a global variable.
+        (global as any).__rootDirectory = rootDirectory ? rootDirectory : getDirectoryPath(__dirname);
```
erictraut: This is fine, but a more typical (and compact) way of expressing this in TypeScript is:
rootDirectory || getDirectoryPath(__dirname)

MikhailArkhipov: Thanks!

PR #488, opened 2020-01-18
[Draft] [WIP] porting fourslash unit test framework from typescript

heejaechang:
pretty much work in progress. most of code is straight copy from typescript with some tweaks to make it work in pyright setup.

added unit tests to cover code ported. 

this only includes ones that can be built and pass tests. fixed some existing unit tests to pass on windows over path separator. also fixed/workaround an issue where process.cwd() is used in test when some test do process.changecwd which breaks global state for other tests. 

* work items - open issues to audit usage of normalizeSlashes and process.cwd


heejaechang:
working adding unit tests for fourslash test reader.

PR #493, opened 2020-01-29
Allow custom configuration settings

MikhailArkhipov:


server/src/analyzer/commentUtils.ts
```diff
@@ -59,10 +59,20 @@ function _applyStrictSettings(settings: DiagnosticSettings) {
 }
 
 function _parsePyrightComment(commentValue: string, settings: DiagnosticSettings) {
-    // Is this a pyright-specific comment?
+    let prefixLength = 0;
+    // Is this a pyright or mspython-specific comment?
     const pyrightPrefix = 'pyright:';
```
erictraut: No need to change this if you don't want, but there's a bit cleaner way to do this.
```
const validPrefixes = ['pyright:', 'mspython:'];
const prefix = validPrefixes.find(p => commentValue.startsWith(p));
if (prefix) {
   ...
}
```

server/src/analyzer/service.ts
```diff
@@ -69,15 +68,17 @@ export class AnalyzerService {
     private _analyzeTimer: any;
     private _requireTrackedFileUpdate = true;
     private _lastUserInteractionTime = Date.now();
+    private _configFileName: string;
 
-    constructor(instanceName: string, console?: ConsoleInterface) {
+    constructor(instanceName: string, console?: ConsoleInterface, configFileName?: string) {
```
erictraut: It's important for us to support both "pyrightconfig.json" and "mspythonconfig.json" in pyrx. Otherwise there will be no way for users to use the pyright command-line tool along with the Python extension (with pyrx). It looks like the current code supports only one or the other. I think it needs to support both. I'm fine if it looks first for "mspythonconfig.json" and falls back to "pyrightconfig.json".

server/src/common/diagnostic.ts
```diff
@@ -1,11 +1,13 @@
 /*
-* diagnostics.ts
```
erictraut: The header comments (both content and formatting) are consistent across the entire pyright code base. I'd prefer that we leave this as is to maintain that consistency. If you want to change it, let's discuss and then push the change throughout the entire code base.

server/src/common/extensions.ts
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
erictraut: See note above about header comments. Please stay consistent with the existing pyright header comment contents and format. I feel very strongly that every code file should have a short description of its contents — the reason for the file's existence. It's really important for understanding a code base and helps encourage good thinking about modularity.

server/src/common/extensions.ts
```diff
@@ -0,0 +1,28 @@
...
+// Licensed under the MIT License.
+
+import { URI } from 'vscode-uri';
+import { normalizePath } from './pathUtils';
+
+/* eslint-disable @typescript-eslint/no-empty-function */
+
+String.prototype.uriToPath = function(this: string): string {
```
erictraut: Yikes. I really don't think we should be extending "string" in this manner. These should be helper functions that take a string as a parameter. Otherwise we're polluting the namespace for a built-in type.

server/src/languageServerBase.ts
```diff
@@ -58,30 +53,43 @@ export class LanguageServerBase {
     // Tracks whether we're currently displaying progress.
     private _isDisplayingProgress = false;
     private _defaultWorkspacePath = '<default>';
+    private _configFileName: string | undefined;
```
erictraut: See note above. We need to support multiple config file names, not a single name.

server/src/languageServerBase.ts
```diff
@@ -247,9 +231,6 @@ export class LanguageServerBase {
                 });
             }
 
-            this._connection.console.log(`Fetching settings for workspace(s)`);
-            this.updateSettingsForAllWorkspaces();
```
erictraut: I don't see how this will work if we don't call updateSettingsForAllWorkspaces here. Does the LS protocol guarantee that we'll receive an onDidChangeConfiguration callback when we start?

server/src/languageServerBase.ts
```diff
@@ -805,43 +768,12 @@ export class LanguageServerBase {
...
-    }
-
-    private convertPosition(position?: DiagnosticTextPosition): Position {
-        if (!position) {
-            return Position.create(0, 0);
-        }
-        return Position.create(position.line, position.column);
-    }
-
-    private convertUriToPath(uriString: string): string {
```
erictraut: If you're replacing the implementation of convertUriToPath, please make sure to exercise the functionality on a Windows machine where paths contain DOS-style paths. I don't currently have unit tests in place for this case because. (Yes, I really should.)

server/src/server.ts
```diff
@@ -1,14 +1,48 @@
 /*
-* server.ts
```
erictraut: See note above about header comment consistency.

server/src/analyzer/service.ts
MikhailArkhipov: OK.

server/src/analyzer/commentUtils.ts
MikhailArkhipov: Sure

server/src/common/diagnostic.ts
MikhailArkhipov: I'll reformat with `prettier` formatter that Python extension uses if you don't mind.

erictraut: OK, let's do that as part of a separate PR since it will touch pretty much every file.

server/src/common/extensions.ts
MikhailArkhipov: Up to you, but this is TS extension methods thing. Pretty common...

erictraut: Just because a language allows you to do something doesn't mean it's a good idea. This is really fragile. I also see no benefit to doing it this way — only downsides.

server/src/common/diagnostic.ts
MikhailArkhipov: I did it manually. Extension settings is 2 spaces for indent and there is more. Too many changes.

server/src/languageServerBase.ts
MikhailArkhipov: Sure, working on it. BTW, stub creation fails if `typingsPath` is not specified, although doc says default is `./typings`. I can make it `projectRoot/typings` if you want.

MikhailArkhipov: Looks like `typingsPath` is set to default only if `pyrightconfig.json` exists. Otherwise config watcher is not created and settings are not loaded from JSON (and that's where the default is set).

server/src/languageServerBase.ts
MikhailArkhipov: Yes, AFAIK client always calls it passing current client configuration.

server/src/languageServerBase.ts
MikhailArkhipov: Looks like unrelated existing issues. Filed https://github.com/microsoft/pyright/issues/494.

server/src/analyzer/service.ts
```diff
@@ -48,6 +47,8 @@ export interface AnalysisResults {
 
 export type AnalysisCompleteCallback = (results: AnalysisResults) => void;
 
+const _configFileNames = ['pyrightconfig.json', 'mspython.json'];
```
erictraut: Do you think this should be "mspython.json" or "mspythonconfig.json"? I guess either one works. The former is more terse but also potentially a bit more ambiguous. I guess I don't have a strong opinion.

Surveying other similar config files, it looks like there's no consistent standard.
TypeScript: "tsconfig.json"
TSLint: "tslint.json"
Webpack: "webpack.config.json"
NPM: ".npmrc"
ESLint: ".eslintrc"

PR #501, opened 2020-02-05
ported typescript fourslash test framework to pyright

heejaechang:
ported typescript fourslash test framework to pyright

this PR consists of 3 main changes.

refactoring on pyright, 2. bunch of helper methods in src/common, 3. fourslash related codes in src/test
...

for 1. refactoring.

it has 2 refactorings on pyright and that affects many pyright code. which made file counts of this PR high.

first refactoring is rename of DiagnosticTextPosition and related types to Position/Range and move them out from diagnostics.ts to textRange.ts.

this is done since I needed line column and it is not diagnostic specific concept.

second one is introducing virtual file system to pyright (server/src/common/vfs.ts).

this is needed since I needed pyright to run over virtual file system constructed from fourslash files.
also, after this, nobody in pyright should interact with file system directly (no import * from "fs").

...

for 2. bunch of helper methods in src/common

it is just bunch of helper methods over path, collections, strings and etc fourslash test framework was using. mostly about string, path comparisons, adding/removing collections without caring about corner cases and etc.

...

for 3. fourslash related codes in src/test

all code related to it should be under src/test/harness and actual tests will be in src/test/fourslash

fourslash test framework is consist of 5 components

virtual file system - this is a file system (server/src/tests/harness/vfs/filesystem.ts) emulating real one. it is a standalone lib people can use for other purpose as well. see (server/src/tests/filesystem.test.ts) on how to use it.

fourslash test reader/parser
it reads fourslash file like (server/src/tests/fourslash/dataclass1.fourslash.ts) or (server/src/tests/fourSlashParser.test.ts) and return FourSlashData that contains source + markup information.
this is also a standalone library one can use on other scenarios just for markups.

testState
it maintain current test states. such as what file is opened, where the caret position is, what is current active files, selection is. and hold pyright program, virtual file system and etc. this will be given to a fourslash test file when the test run. and people can interact with it to manipulate or verify test results. see (server/src/tests/testState.test.ts) for what it currently supports. this should keep updated as we add new tests and new test ability.
one thing to remember is that, TestState is what run at runtime (implementation), but for editting, fourslash.ts (declaration) (server/src/tests/fourslash/fourslash.ts) is the one that is used.
that is why there is reference path="fourslash.ts" at the top of each fourslash test file (ex, server/src/tests/fourslash/dataclass1.fourslash.ts). because of that, it is important to put declaration in fourslash.ts if one added new functionality in TestState.

fourslash test runner
this puts everything together and run given fourslash markup strings. to learn how it actually run code. see (server/src/tests/harness/fourslash/runner.ts).
basically, it gets sources (python code) out of fourslash string, and construct virtual file system and connect it to pyright using TestState and then compile the given fourslash markup string since it is actually pure typescript code and then wrapped the compiled javascript code under function and eval the code and run it with TestState as argument. the reference path=".." trick works since it gets ignored within method body as a regular comment.
this is also a standalone lib people can use in other way if wants. see (

server/src/tests/testState.test.ts

 test('VerifyDiagnosticsTest1', () => { 
) on how to use it.

jest test suite
this connects fourslash test framework to jest. it basically read fourslash test files in **/test/fourslahs/*.fourslash.ts and register it to jest and run them with the fourslash test runner.
(server/src/tests/fourSlashRunner.test.ts)
...

PR also includes some misc changes such as ignoring eslint on fourslash.ts files. or js/ts formatting rule such as putting space after template string since that seems what pyright currently has as a code style. and this also includes some of checker.test.ts converted to fourslahst test to show differences. and debugger launch settings for fourslash tests.

...

how to run fourslash tests.

on command line, if one runs jest tests (ex, npm run test), it will run fourslash tests as well. in vscode, if users choose vscode-jest-tests, it will run fourslash tests as well. if user opens fourslash.ts file and then select fourslash current file in debugger configuration and F5, vscode will run only that fourslash test file.


heejaechang:
@erictraut can you take a look? it is same code you reviewed + other change made to pyright and squashed.

PR #507, opened 2020-02-11
made reportMissingTypeStub diagnostics warning by default and some refactoring around code actions.

heejaechang:
user will now see squiggle on third party library on import statement if it doesn't have a typestub.

this will also enable corresponding "create type stub" code action to light up. when invoked, "generated typestub" can improve user's editing experience.


heejaechang:
failure is due to typo in CI pipeline which I just fixed. so it can be ignored.

PR #509, opened 2020-02-11
previously added assert was using wrong path (__dirname rather than _…

heejaechang:
…_dirname/..)

fixed the assert and added comments indicating what is supposed to point what.


heejaechang:
@erictraut can you take a look?

heejaechang:
close and reopen for CI to pick up.

heejaechang:
thank you!

PR #515, opened 2020-02-13
added a hook to ImportResolver for custom importing logic.

heejaechang:
sending PR on behalf of @huguesv 

@erictraut already looked and signed off. this PR is just to make sure CI passes.


heejaechang:
thank you

PR #534, opened 2020-02-20
added code action tests

heejaechang:
there are 2 kinds of code actions, one that introduce new files or one that returns text edits. this introduce test for one that adds files. there will be another PR for ones that return text edits.

this also includes misc changes such as making del-cli works on windows. and some refactoring to remove language service dependencies on command handler and etc.


heejaechang:
@erictraut already took a look at this PR. this is just to make sure that CI is clean.

PR #541, opened 2020-02-24
changing the connection type from IPC to be based on the process's co…

heejaechang:
…mmandline arguments so that the benchmark tool can set stdin/stdout


heejaechang:
creating PR on behalf of @bschnurr

heejaechang:
this is just to run CI. code is already reviewed by eric.

PR #552, opened 2020-03-03
made LS to automatically add "src" as one of extra path to find imports.

heejaechang:
it is common pratice where user start vscode on repo root but sources are under src. this change introduces a new option called "autoSearchPaths" (on by default) that will let LS to detect such case and add "src" as extra path to search packages.

this PR also includes misc fix on test framework where all file path is normalized to absolute path, let verify diagnostics to handle empty error case and etc.


heejaechang:
tagging @huguesv

PR #553, opened 2020-03-03
refactored on analysis completed handler

heejaechang:
this should let derived class to override behavior on analysis complete


heejaechang:
tagging @bschnurr

PR #570, opened 2020-03-19
added auto venv detection, LS extension points, auto detection on lib…

heejaechang:
…rary changes and more tests added.

auto venv detection makes sure pyright automatically exlude virtual environment from user codes
LS extension points lets plug in such as intellicode to extend what shows up in completion list
auto detection on library let pyright to refresh automatically when new packages are installed/removed.

more test coverage on code fix, new features


heejaechang:
tagging @huguesv @RaymonGulati1

PR #579, opened 2020-03-23
added completion test coverage

heejaechang:


heejaechang:
checking in since it is just test changes.

PR #583, opened 2020-03-25
added cancellation support and more tests.

heejaechang:
cancellation support is added through vscode-languageserver's new cancellation strategy API.

more tests on hover and completion.


heejaechang:
tagging @huguesv

PR #596, opened 2020-03-28
Implemented support for incremental text changes LSP

erictraut:
I implemented support for incremental text changes. Unfortunately, I think this is going to be slower and more expensive than the alternative because incremental changes are specified in terms of lines/columns rather than file offsets. That means the code needs to scan the old file contents to break it into lines, then insert the text change.

My inclination is to discard this change and stick with "full" text updates. Python source files are typically are, at most, several K in size. Passing the entire contents is cheap.

### DIFF
--- a/server/src/analyzer/program.ts
+++ b/server/src/analyzer/program.ts
@@ -13,7 +13,8 @@ import {
     CompletionItem,
     CompletionList,
     DocumentSymbol,
-    SymbolInformation
+    SymbolInformation,
+    TextDocumentContentChangeEvent
 } from 'vscode-languageserver';
 
 import { throwIfCancellationRequested } from '../common/cancellationUtils';
@@ -202,7 +203,7 @@ export class Program {
         return sourceFile;
     }
 
-    setFileOpened(filePath: string, version: number | null, contents: string) {
+    setFileOpened(filePath: string, version: number | null, contents: TextDocumentContentChangeEvent[]) {
         let sourceFileInfo = this._sourceFileMap.get(filePath);
         if (!sourceFileInfo) {
             const sourceFile = new SourceFile(this._fs, filePath, false, false, this._console);
@@ -235,7 +236,7 @@ export class Program {
         const sourceFileInfo = this._sourceFileMap.get(filePath);
         if (sourceFileInfo) {
             sourceFileInfo.isOpenByClient = false;
-            sourceFileInfo.sourceFile.setClientVersion(null, '');
+            sourceFileInfo.sourceFile.setClientVersion(null, [{ text: '' }]);
 
             if (this._configOptions.checkOnlyOpenFiles) {
                 // Reset the diagnostic version so we force an update
--- a/server/src/analyzer/service.ts
+++ b/server/src/analyzer/service.ts
@@ -13,7 +13,8 @@ import {
     CompletionItem,
     CompletionList,
     DocumentSymbol,
-    SymbolInformation
+    SymbolInformation,
+    TextDocumentContentChangeEvent
 } from 'vscode-languageserver';
 
 import { getGlobalCancellationToken, OperationCanceledException } from '../common/cancellationUtils';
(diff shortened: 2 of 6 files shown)


jakebailey:
Is there a reason to implement all of this by hand as opposed to using the LSP library's document type? I believe it has a datatype that would handle all of this behavior for you. See TextDocument in this subpackage of the LSP lib: https://github.com/microsoft/vscode-languageserver-node/tree/master/textDocument

Many common language servers use diffs (and even convert them to "standard" diffs as that turns out to be more efficient, probably due to fancy black-box methods to reduce copies).

I've seen some pretty massive Python files, thanks to the fact that most modules are a single file and only grow.

erictraut:
I don't want to pull in a dependency on vscode library types within the core of pyright. It's fine to add such dependencies to the languageServerBase, but they shouldn't be used in other core modules. The CLI version of pyright shouldn't depend on vscode libraries.

The published version of vscode-languagserver-textdocument (1.0.1) doesn't contain the FullTextDocument class. In other words, it doesn't have the support we need. It looks like the latest code checked in [to the repo](https://github.com/microsoft/vscode-languageserver-node/blob/master/textDocument/src/main.ts) does have this support, but it hasn't been published yet.

If a new version of vscode-languageserver-textdocument were published, we could apply the diffs within languageServerBase and then pass the resulting contents to the lower-level modules. That would remove about 20 lines of code from this PR (at the cost of taking a dependency on another module). I looked at the implementation of FullTextDocument, and it's doing the same thing I am in the current PR, so it won't be any more efficient than what I've coded here.

I guess there might be some small perf improvement over the "full" update mechanism because of the need to serialize and deserialize the file contents in JSON when it gets passed from the client to the server. If you agree with this logic, then I'll probably just go ahead and check in this change.

jakebailey:
I don't think the efficiency is exactly the same between this PR and that library, though. Just skimming, it appears to be amortizing the cost of finding the newline offsets by storing them with the text, with the expectation that you're keeping the document around rather than recreating it. Given most editors send a change per character update, I'd think this would end up mattering a lot on large files. (Perhaps so much so that this PR ends up slower due to not going that extra step... Unsure.)

This PR seems to have a bunch of unrelated changes to do with `Dict`; any chance you can merge from master so the diff shows just the incremental text changes?

I haven't personally gone and reviewed or tested this code thoroughly, either way. I'm happy to test on Monday since I know diffing is a really fiddly thing to get right without desyncs. (It's a place we're going to end up covering in some tests, I'm sure.)

erictraut:
I'm seeing only one text-change entry passed with each message when typing. The only time multiple entries come through is when I do a "find and replace all" operation. So in that case, you're correct that the perf could be somewhat faster. The library recalculates line endings only on the portion of the file after each change, not the lines before.

I rebased the branch to the latest master.

jakebailey:
> I'm seeing only one text-change entry passed with each message when typing.

I meant that every character is its own `textDocument/didChange` notification, not that the editor is sending batching multiple changes in a single `textDocument/didChange` notification.

The LSP library stores line info within the document (along side the text), so between `textDocument/didChange` notifications, this info is preserve and can be reused/maintained versus needing to be recalculated again each time. That was the difference I was speaking of, not neccesarily multiple changes within a notification (which would be equivalent to sending each as their own `didChange`).

Thanks for rebasing, the change is clearer (though I'm not certain how that 1.1.30 thing got in there, strange).

erictraut:
Hmm, I'm not sure how the 1.1.30 changes got mixed in. I did another rebase to remove them.

erictraut:
I'm not opposed to using the vscode-languageserver-textdocument library if we can get that team to publish the latest code.

jakebailey:
Are you sure 1.0.1 doesn't contain that type? Looking at the changes for the subfolder (https://github.com/microsoft/vscode-languageserver-node/commits/master/textDocument), the most recent changes were the 1.0.1 release and a badge change, which would imply that the type is there. This would make sense given v6+ of the server library has a breaking change where you need to explicitly provide `TextDocument` as the document type when using the built-in document table (https://github.com/microsoft/vscode-languageserver-node#3150-protocol-600-client--600-server). I'm not sure they would have released the LSP server library without the capability to process diffs as effectively as in v5.

I would agree that uses of this type would make sense above at the LSP server level rather than pushed all the way down from a library-usage point of view (and that's how we did it in MPLS, in memory files are just `StringBuffer` instances we modify whereas modules themselves only have access to the `string` and not the buffer, providing some semblance of snapshotting).

But, I don't know how large of a change that'd be; I'm picturing needing two have the data in two places which seems like a bad idea. You'd know how invasive this'd be better than me.

PR #608, opened 2020-04-06
Removed the notion of "low level file watcher" from FileSystem. Added…

erictraut:
… support for vscode client-base file watcher for files within all workspaces. Chokidar file watcher is no longer used for VS Code extension — only for the command-line version of Pyright.

### DIFF
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -11,7 +11,7 @@
 
 import * as fs from 'fs';
 import * as path from 'path';
-import { commands, ExtensionContext, Position, Range, TextEditor, TextEditorEdit } from 'vscode';
+import { commands, ExtensionContext, Position, Range, TextEditor, TextEditorEdit, workspace } from 'vscode';
 import { LanguageClient, LanguageClientOptions, ServerOptions, TextEdit, TransportKind } from 'vscode-languageclient';
 
 import { Commands } from '../../server/src/commands/commands';
@@ -52,7 +52,14 @@ export function activate(context: ExtensionContext) {
         ],
         synchronize: {
             // Synchronize the setting section to the server.
-            configurationSection: ['python', 'pyright']
+            configurationSection: ['python', 'pyright'],
+
+            // Send file system change events within the workspace.
+            fileEvents: [
+                workspace.createFileSystemWatcher('**/pyrightconfig.json'),
+                workspace.createFileSystemWatcher('**/mspythonconfig.json'),
+                workspace.createFileSystemWatcher('**/*.{py,pyi,pyd}')
+            ]
         },
         connectionOptions: { cancellationStrategy: cancellationStrategy }
     };
--- a/server/src/languageServerBase.ts
+++ b/server/src/languageServerBase.ts
@@ -9,6 +9,7 @@
 
 import './common/extensions';
 
+import * as fs from 'fs';
 import {
     CancellationToken,
     CodeAction,
@@ -45,8 +46,14 @@ import { ConfigOptions } from './common/configOptions';
 import { ConsoleInterface } from './common/console';
 import { Diagnostic as AnalyzerDiagnostic, DiagnosticCategory } from './common/diagnostic';
 import { LanguageServiceExtension } from './common/extensibility';
-import { createFromRealFileSystem, FileSystem } from './common/fileSystem';
-import { convertPathToUri, convertUriToPath } from './common/pathUtils';
+import {
+    createFromRealFileSystem,
+    FileSystem,
+    FileWatcher,
+    FileWatcherEventHandler,
+    FileWatcherEventType
+} from './common/fileSystem';
+import { containsPath, convertPathToUri, convertUriToPath } from './common/pathUtils';
 import { Position } from './common/textRange';
 import { AnalyzerServiceExecutor } from './languageService/analyzerServiceExecutor';
 import { CompletionItemData } from './languageService/completionProvider';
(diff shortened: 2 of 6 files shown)


server/src/languageServerBase.ts
```diff
@@ -507,6 +569,18 @@ export abstract class LanguageServerBase implements LanguageServerInterface {
             service.setFileClosed(filePath);
         });
 
+        this._connection.onDidChangeWatchedFiles(params => {
+            params.changes.forEach(change => {
+                const filePath = convertUriToPath(change.uri);
+                const eventType: FileWatcherEventType = change.type === 1 ? 'add' : 'change';
```
jakebailey, line 575: Perhaps this is something to be left for later (as I'm noticing it's missing from previous enums like `FileWatcherEventType`), but do we not support full removals? I can see this coming up if a module gets deleted, which would probably have a `delete` event. Perhaps you're modelling that as a change and it works well enough.

client/src/extension.ts
```diff
@@ -52,7 +52,14 @@ export function activate(context: ExtensionContext) {
...
         synchronize: {
             // Synchronize the setting section to the server.
-            configurationSection: ['python', 'pyright']
+            configurationSection: ['python', 'pyright'],
+
+            // Send file system change events within the workspace.
+            fileEvents: [
+                workspace.createFileSystemWatcher('**/pyrightconfig.json'),
+                workspace.createFileSystemWatcher('**/mspythonconfig.json'),
+                workspace.createFileSystemWatcher('**/*.{py,pyi,pyd}')
```
jakebailey, line 61: I think these are the pre-defined ones being passed in the initialization unconditionally. Will this be overzealous? Or do we need to do this later once we know the search paths so we're able to exclude "bad" paths (other virtualenvs, for example) which won't be used?

I haven't tested this PR yet (I may do so tomorrow), but was wondering.

jakebailey: Oh, this is in the client... I'll have to go read into what this does on this side of the equation.

server/src/languageServerBase.ts
```diff
@@ -208,6 +226,50 @@ export abstract class LanguageServerBase implements LanguageServerInterface {
...
+            if (workspaces.some(workspace => containsPath(workspace.rootPath, path))) {
+                workspacePaths.push(path);
+            } else {
+                nonWorkspacePaths.push(path);
+            }
+        });
+
+        // For any non-workspace paths, use the node file watcher.
+        const nodeWatchers = nonWorkspacePaths.map(path => {
+            return fs.watch(path, { recursive: true }, listener);
```
jakebailey, line 250: This bypasses the VFS layer, which doesn't seem right from the abstraction point of view, as I assume we're going to want to test this.

server/src/languageServerBase.ts
erictraut: Chokidar and the node version don't differentiate between "change" and "delete" events. They're all reported as "change". I'm just normalizing everything here.

I'm not sure what you mean by "full removals"? Are you talking about when the root directory of a file watcher is deleted?

client/src/extension.ts
erictraut: I assume the client installs a file watcher for the entire workspace (which makes sense, since it needs to update the file hierarchy in the editor). 

There's no way to omit "bad" paths in any of the three file watchers. You need to install the file watcher and ignore any events that you aren't interested in.

server/src/languageServerBase.ts
erictraut: This isn't any different from the current implementation, which no-ops file watchers in the mocked file system implementation.

server/src/languageServerBase.ts
jakebailey: Sorry, by "full" I meant "distinct from changed". I just didn't want to end up in a scenario where someone deletes a module but it remains resolvable.

server/src/languageServerBase.ts
jakebailey: I don't agree that it isn't different; now this file imports `fs` directly whereas before it had to do so via the VFS interface. The only file that should know how to access the file system for real should be the "real" implementation of the VFS interface, but now both the VFS and the language server does it.

client/src/extension.ts
jakebailey: I would think that it's an implementation detail of the server to be requesting file watches, not of the client. The above reads to me like the client forcing the watching of specific files onto the server (which may know better). A client which is not the pyright extension will not have this code.

It's more typical for the server to do this itself on handling of the `initialized` notification (which is when the client says "it's okay for you to send requests to me now", as it's illegal to make requests during the `initialize` handshake). See https://microsoft.github.io/language-server-protocol/specifications/specification-current/#initialized ("The server can use the initialized notification for example to dynamically register capabilities" -- file watching is a dynamically registered capability)

erictraut: Do you know of any code samples that show how to use dynamically registered capabilities — preferably for file watcher events? I couldn't find any. 

BTW, I looked in the debugger at the client capabilities sent from VS Code to the server, and they don't include the capability for file watcher dynamic registrations.

jakebailey: Here is what the Go language server does: https://github.com/golang/tools/blob/8cb32c46614565a65efe53a937bd4144a4642627/internal/lsp/general.go#L143

Note that they only register a single watcher and give it a statically generated name; you'll want to create some unique identifier as you'll need more than one. The API is nearly identical to the client side version.

jakebailey: > BTW, I looked in the debugger at the client capabilities sent from VS Code to the server, and they don't include the capability for file watcher dynamic registrations.

You mean that VS Code claims to not support file watching at all? That's strange... and goes against my experience so far. Maybe there's some option being missed as lacking file watch support is a biggie.

PR #610, opened 2020-04-08
Upgrade to Prettier 2.0, reformat

jakebailey:
This uses the new Prettier 2.0 defaults, with the exception of retaining the old `endOfLine` behavior as the new version always uses LF (whereas we're all using git's auto mode to handle that).


server/src/parser/tokenizer.ts
```diff
@@ -924,7 +924,7 @@ export class Tokenizer {
         const comment = Comment.create(start, length, value);
 
         if (value.match(/^\s*type:\s*ignore(\s|$)/)) {
-            if (this._tokens.findIndex(t => t.type !== TokenType.NewLine && t && t.type !== TokenType.Indent) < 0) {
+            if (this._tokens.findIndex((t) => t.type !== TokenType.NewLine && t && t.type !== TokenType.Indent) < 0) {
```
Zzzen, line 927: `&& t` is redundant?

PR #662, opened 2020-05-07
Add extraPath support, accept minor versions in version_info, make completion extension async

jakebailey:
Also enables the `eqeqeq` lint rule. 🙂


jakebailey:
Kicking pipelines to get it to run (it will likely give an error message for this but it's a lie).

/azp run

PR #663, opened 2020-05-09
Introduced UnboundVariable and UndefinedVariable diagnostic rules and…

heejaechang:
… refactored "auto import" code.

Also, unit tests covering auto imports are added.


heejaechang:
will check in once test passes.

PR #708, opened 2020-06-04
Added pre-commit hook

nihaals:
* [Documentation](https://pre-commit.com/#new-hooks)
* [Examples of other projects that have pre-commit hooks](https://pre-commit.com/hooks.html)

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,6 @@
+- id: pyright
+  name: pyright
+  entry: pyright
+  language: node
+  types: [python]
+  pass_filenames: false


erictraut:
Thanks for the contribution.

nihaals:
@erictraut I’m unsure if this requires `require_serial`, after looking at other projects like isort, flake8 and mypy, they all set it to `true`, but I’m not sure what the affect of this would be. The documentation says:
> (optional: default `false`) if `true` this hook will execute using a single process instead of in parallel.

erictraut:
Yeah, I think this is needed only for hooks that write to the file (like tools that have reformatting modes). Pyright doesn't fall into this category.

PR #734, opened 2020-06-15
Don't show errors for other files by default

WhyNotHugo:
There's a setting to show errors for files _other_ than the files currently open. While I can't think of a scenario where this would make sense, I guess there are and some people find this desireable.

However, this is set to true by default as well. This results in pyright showing errors for files _other_ than the currently open file, which doesn't really make sense. Aside from being extremely confusing, it's quite orthogonal to what any other checker does too.

This change makes the default what any developer would expect: when opening a file, see errors for that file and not for other files.

It also documents what the alternative setting does, since it's not at all obvious.

(I was going to open a bug report thinking `pyright` was broken, but it seems its just an oversight in having the werid-case being the default, rather than the normal case).

Thanks for maintaining this, btw!

### DIFF
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -20,7 +20,7 @@ The Pyright VS Code extension honors the following settings.
 
 **python.analysis.stubPath** [path]: Path to directory containing custom type stub files.
 
-**python.analysis.diagnosticMode** ["openFilesOnly", "workspace"]: Determines whether pyright analyzes (and reports errors for) all files in the workspace, as indicated by the config file. If this option is set to "openFilesOnly", pyright analyzes only open files.
+**python.analysis.diagnosticMode** ["openFilesOnly", "workspace"]: Determines whether pyright analyzes (and reports errors for) all files in the workspace, as indicated by the config file. If this option is set to "openFilesOnly", pyright analyzes only open files. If this option is set to "workspace", pyright will also analyze and show errors for files that are _not_ open.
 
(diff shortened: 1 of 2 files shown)


erictraut:
I think the current code is correct. It defaults to showing errors only for files that are currently open in the editor. That is the intended default behavior, which is why openFilesOnly is set to true by default. Your code change breaks this behavior by setting this flag to false by default.

If you want Pyright to report errors for all files in your workspace (including open files and non-open files), you can set "python.analysis.diagnosticMode" to "workspace". This is the mode I typically use because I'm interested in seeing bugs reported across all of my source base, not just in files that happen to be open in the editor at the moment.

erictraut:
@WhyNotHugo - I'm going to close this PR for the reasons I mentioned above. If I'm missing something, let me know.

WhyNotHugo:
Yes, it seems that my setup is screwed up in a different way, since I'm getting errors for files that are **not** open when not setting this preference (e.g.: it should fall back to the default).

erictraut:
Let me know how I can help debug the problem. If you'd like to open a bug, we can try to track down the underlying cause. Or since you seem to have checked out the code for pyright, perhaps you'd be comfortable debugging it on your system? If you look in the file "program.ts", you'll find a method called getDiagnostics. It should include diagnostics only for files that are open in the editor if the checkOnlyOpenFiles is true.

PR #766, opened 2020-06-26
Add severity level enum to package.json

9at8:
I'm working on making object settings editable from the GUI in vscode https://github.com/microsoft/vscode/issues/99635. The change in this PR would make the `diagnosticSeverityOverrides`  setting editable from the GUI. We can make it stricter by adding schemas for all of these options mentioned here: https://github.com/microsoft/pyright/blob/master/docs/configuration.md#type-check-diagnostics-settings.


erictraut:

PR #797, opened 2020-07-06
Add reference to PEP-561 for background information

IanHopkinson:
PEP 561 - distributing and packaging type information for background provides useful background reading, including sample code for type annotations.

Addition is in order by PEP number, checked validity of Markdown by looking at "Preview changes" in GitHub.


erictraut:
Thanks for the contribution.

PR #807, opened 2020-07-08
Fixes from Pylance.

huguesv:
Code Action range fix
Display of multi-line literals
Remove self from signature in completion item when more appropriate


huguesv:
@erictraut thanks I don't have permissions to merge, so feel free to merge for me

Issue #99, opened 2019-04-29
Add standalone language server executable for other editors to consume


dsifford:
Hi there. Awesome project!

**Is your feature request related to a problem? Please describe.**

Perhaps this is already possible in the distributed package but I wasn't able to figure it out. What I'd like to do is consume the LSP server this project defines in vim (using ALE). 

As an example, here's how the pylint LSP server is able to be consumed in ALE... https://github.com/w0rp/ale/blob/master/ale_linters/python/pyls.vim

**Describe the solution you'd like**

Add an ability to invoke the language server from the command line. (If not there already).

erictraut:
This is outside the scope of this project. If you want to start another project to provide the necessary glue to host Pyright or other LSP servers in other editors, please feel free to do so.

w0rp:
Could you offer any advice on how to build a command line utility for starting the language server? Without a CLI utility, the language server cannot be used outside of VSCode, or in some other editors written in JavaScript.

erictraut:
Here are a couple of pointers...

The [pyright.ts](https://github.com/Microsoft/pyright/blob/master/server/src/pyright.ts) code file implements the command-line interface provided by Pyright today. It's a small wrapper around the [analyzer service](https://github.com/Microsoft/pyright/blob/master/server/src/analyzer/service.ts).

The language server is implemented in [server.ts](https://github.com/Microsoft/pyright/blob/master/server/src/server.ts) and is also a small wrapper around the analyzer service. The language server uses VS Code APIs to communicate with the VS Code language client.

If you want to support other editors, you can create a new project and specify pyright as a dependency through the package.json file. Then create a new top-level code file that imports the [analyzer service](https://github.com/Microsoft/pyright/blob/master/server/src/analyzer/service.ts). You'll need to instantiate it (it's a singleton class) and call it as appropriate (e.g. in response to various events from your language service client). You can use pyright.ts and server.ts as examples of how to call it.

Hope that helps!

w0rp:
Cool, thank you for the information.

lkgarrison:
@erictraut I'm unclear why a separate project needs to be created in order to use the Language Server instead of being able to run it from this project. Is it too tightly coupled with VSCode? Or would it need an additional CLI command to start up the language server that is undesirable to add? I was surprised to see you push for a separate project that would expose the language server for other IDEs. It seems a bit silly and against the spirit of LSP to develop a tool like this that is only compatible with VSCode?

erictraut:
[...]

PR #810, opened 2020-07-08
Add new entrypoint for pyright-langserver

zeronone:
- As discussed in #99. This PR adds a new executable `pyright-langserver` that is installed when `npm install -g pyright` is done. This would make it easy to be utilized by other editors.

### DIFF
--- a/langserver.index.js
+++ b/langserver.index.js
@@ -3,4 +3,4 @@
 // Stash the base directory into a global variable.
 global.__rootDirectory = __dirname + '/dist/';
 
-require('./dist/pyright');
+require('./dist/pyright-langserver');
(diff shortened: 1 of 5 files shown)


erictraut:

WhyNotHugo:
Thanks for addressing this!

w0rp:
Thanks to this patch, `pyright` is now supported by [ALE](https://github.com/dense-analysis/ale), and enabled by default. ALE will try to figure out where your virtualenv might be and which Python version to use, so it should work automatically for a lot of people without having to configure anything.

Feel free to mention ALE in [README.md](https://github.com/microsoft/pyright/blob/master/README.md).

PR #915, opened 2020-08-03
Use `sys.version_info` to query interpreter version

jakebailey:
Syncing from Pylance.

Fixes the case where `python` points to Python 2, as Python 2's `--version` flag outputs via stderr. Parsing `--version` is not recommended by the Python docs in favor of `sys.version_info`. Do that instead as we already do for `sys.path`.

### DIFF
--- a/server/src/common/configOptions.ts
+++ b/server/src/common/configOptions.ts
@@ -16,7 +16,13 @@ import { ConsoleInterface } from './console';
 import { DiagnosticRule } from './diagnosticRules';
 import { FileSystem } from './fileSystem';
 import { combinePaths, ensureTrailingDirectorySeparator, FileSpec, getFileSpec, normalizePath } from './pathUtils';
-import { latestStablePythonVersion, PythonVersion, versionFromString, versionToString } from './pythonVersion';
+import {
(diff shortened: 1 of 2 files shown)


server/src/common/configOptions.ts
```diff
@@ -1254,22 +1263,13 @@ export class ConfigOptions {
                 execOutput = child_process.execFileSync('python', commandLineArgs, { encoding: 'utf8' });
             }
 
-            // Parse the execOutput. It should be something like "Python 3.x.y".
-            const match = execOutput.match(/[0-9]+.[0-9]+.[0-9]+/);
-            if (!match) {
-                console.info(`Unable to get Python version from interpreter. Received "${execOutput}"`);
-                return undefined;
-            }
+            const versionJson: { major: number; minor: number } = JSON.parse(execOutput);
```
erictraut, line 1266: I don't see any error handling here. JSON.parse will throw an exception if it receives anything that's not JSON compliant, so you should be ready to handle that.

jakebailey: This whole function is a try-catch block, so handles those cases.

PR #934, opened 2020-08-11
Declare severity overrides for VS Code UI, fix up some configuration pulling code

jakebailey:
- Use the new VS Code object setting schema support to declare in the UI which settings are available, their options, and their descriptions. This raises the minimum VS Code version to 1.47, and ensures this list is always up to date and synced with the enum and pyrightconfigschema.

image

- Update configuration querying code to support querying top-level settings (the `undefined` scope) in Pylance.

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -11,9 +11,9 @@
             "dev": true
         },
         "@types/vscode": {
-            "version": "1.46.0",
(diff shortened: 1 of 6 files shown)


jakebailey:
Note that the above is an example of the UI, but in the submitted code the order of the severity levels is the more correct: none, information, warning, error

PR #940, opened 2020-08-14
Changed "find workspace symbols" to return only symbols from within u…

erictraut:
…ser code or opened files, not library files that are closed.

### DIFF
--- a/server/src/analyzer/program.ts
+++ b/server/src/analyzer/program.ts
@@ -1106,9 +1106,8 @@ export class Program {
             }
 
             for (const sourceFileInfo of this._sourceFileList) {
-                // "Find symbols" will only include references from user code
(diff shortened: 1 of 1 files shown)


server/src/analyzer/program.ts
```diff
@@ -1106,13 +1106,17 @@ export class Program {
             }
 
             for (const sourceFileInfo of this._sourceFileList) {
-                this._bindFile(sourceFileInfo);
+                // "Find symbols" will only include references from user code
+                // unless the file is explicitly opened in the editor.
+                if (this._isUserCode(sourceFileInfo) || sourceFileInfo.isOpenByClient) {
```
jakebailey, line 1111: I think the `isOpenByClient` could be dropped, but otherwise this seems correct to me.

PR #971, opened 2020-08-26
Update deps, pin LSP libraries together, use Node 12.x

jakebailey:
Updates the deps, but pins a set of known-compatible LSP libraries. I've also pinned the node types to 12, as VSC runs 12 (not 14).


client/package.json
```diff
@@ -614,14 +614,16 @@
...
+        "@types/node": "~12.12.54",
+        "@types/vscode": "^1.48.0",
+        "typescript": "^4.0.2",
+        "vsce": "^1.79.3"
     },
     "dependencies": {
         "leven": "^3.1.0",
-        "vscode-languageclient": "^7.0.0-next.8",
-        "vscode-languageserver": "^7.0.0-next.6",
+        "vscode-jsonrpc": "6.0.0-next.4",
```
jakebailey, line 624: This was a phantom dependency; the code imports this but it was never added to the deps list.

client/package.json
```diff
@@ -13,7 +13,7 @@
         "url": "https://github.com/Microsoft/pyright"
     },
     "engines": {
-        "vscode": "^1.47.0"
+        "vscode": "^1.48.0"
```
jakebailey, line 16: The engine and `@types/vscode` need to be kept in sync, that's why the other PR pipeline had failed.

PR #980, opened 2020-08-29
Add sublime text to the list of supported editors

AmjadHD:


erictraut:
Thanks for the contribution.

PR #984, opened 2020-08-31
Suppress the use of "Unnecessary" diagnostic hints (used to display v…

erictraut:
…ariables and code blocks in gray) if the LSP client claims not to support this tag.


erictraut:
@edricgarran - FYI

PR #991, opened 2020-09-02
fix: reportMissingModuleSource default none

fannheyward:
schema default is none https://github.com/microsoft/pyright/blob/58a349b50e5212d29ab1606f8f3cb8b6557c7aa5/packages/vscode-pyright/schemas/pyrightconfig.schema.json#L138-L142

### DIFF
--- a/packages/vscode-pyright/package.json
+++ b/packages/vscode-pyright/package.json
@@ -130,7 +130,7 @@
                         "reportMissingModuleSource": {
                             "type": "string",
                             "description": "Diagnostics for imports that have no corresponding source file. This happens when a type stub is found, but the module source file was not found, indicating that the code may fail at runtime when using this execution environment. Type checking will be done using the type stub.",
-                            "default": "warning",
+                            "default": "none",
                             "enum": [
                                 "none",
                                 "information",


erictraut:
This isn't correct. If you don't specify reportMissingModuleSource, it will default to "warning".

fannheyward:
Did I missed something? As schema defined, `reportMissingModuleSource`'s default is none, but in `python.analysis.diagnosticSeverityOverrides`, default is warning.

Others, for example `reportInvalidStubStatement`'s default are same in both schema and python.analysis.diagnosticSeverityOverrides properties.

erictraut:
You are correct that there's a bug here, but it's in the schema. Both of them should be "warning".

Here's the table in the source code that specifies the default values when "typeCheckingMode" is "off": https://github.com/microsoft/pyright/blob/7fd612ad703f04d0f6133e8820c705a91f3bb9ca/packages/pyright-internal/src/common/configOptions.ts#L277

And here's the table that specifies the default values when "typeCheckingMode" is "basic": https://github.com/microsoft/pyright/blob/7fd612ad703f04d0f6133e8820c705a91f3bb9ca/packages/pyright-internal/src/common/configOptions.ts#L332

It's not clear which of these tables should be reflected in the JSON schema and the package.json diagnosticSeverityOverrides properties. I opted to reflect the type checking mode "off" setting because that's the default type checking mode for Pylance, and I wanted the Pyright defaults to be consistent with Pylance.

erictraut:
I just pushed a fix to the schema. Thanks for pointing out the inconsistency.

PR #1007, opened 2020-09-04
Added support for py.typed files to distinguish between packages that…

erictraut:
… claim to have inlined types versus those that don't. Also added support for "partial" stub packages (a "-stub" package that has a py.typed file that includes the line "partial\n" as described in PEP 561).


jakebailey (approved):
Seems correct to me, though with the return type change of `resolveAbsoluteImport`, you might be able to remove a few truthiness checks elsewhere as a cleanup later.

PR #1137, opened 2020-11-04
recommend pylance over pyright for vs code

jli:


jakebailey:
I think some aspect of this is covered, if you hadn't seen it: https://github.com/microsoft/pyright#using-pyright-with-vs-code-python-extension

erictraut:
Thanks for the suggestion. I've updated the README to include the gist of your changes, but I reworded it to add more clarity.

jli:
Thanks for the update Eric!

Noticed a small typo:
> For most VS Code users, we recommend using the Pylance extension **rathern** than Pyright.

erictraut:
Good catch. Just fixed. I forgot that I had temporarily disabled the spell-checker extension in VS Code.

PR #1165, opened 2020-11-12
Support detecting multiple site-packages directories in venvs

truls:
On Red Hat derived systems, Python's `purelib` and `platlib` are set to `<prefix>/lib` and `<prefix>/lib64` respectively. This means that Python packages are installed to either of those directories depending on whether they contain native libraries or not (as I understand it). This distinction is also present in pipenv's virtualenvs. See:

```
$  . /home/truls/.local/share/virtualenvs/pyright-test-Fm_xQ9MX/bin/activate
(pyright-test) truls@dhcp-111-236 ~/foo/pyright-test $ python
Python 3.9.0 (default, Oct  6 2020, 00:00:00) 
[GCC 10.2.1 20200826 (Red Hat 10.2.1-3)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import sysconfig
>>> sysconfig.get_path('purelib')
'/home/truls/.local/share/virtualenvs/pyright-test-Fm_xQ9MX/lib/python3.9/site-packages'
>>> sysconfig.get_path('platlib')
'/home/truls/.local/share/virtualenvs/pyright-test-Fm_xQ9MX/lib64/python3.9/site-packages'
>>> 
```

The current virtualenv include path detection logic in pyright returns only the first extant `site-packages` directory it finds. This is insufficient for discovering the full include path for virtualenvs where packages may be installed in both the `lib` and `lib64` directories. This PR changes the virtualenv include path detection logic so that it returns all of the extant `site-packages` dirs it finds in the `lib`, `lib64` or `Lib` directories.

### DIFF
--- a/packages/pyright-internal/src/analyzer/pythonPathUtils.ts
+++ b/packages/pyright-internal/src/analyzer/pythonPathUtils.ts
@@ -59,6 +59,39 @@ export function getTypeshedSubdirectory(typeshedPath: string, isStdLib: boolean)
     return combinePaths(typeshedPath, isStdLib ? stdLibFolderName : thirdPartyFolderName);
 }
 
+function checkVenvPath(fs: FileSystem, libPath: string, importFailureInfo: string[]): string | undefined {
+    if (fs.existsSync(libPath)) {
+        importFailureInfo.push(`Found path '${libPath}'; looking for ${pathConsts.sitePackages}`);
+    } else {
+        importFailureInfo.push(`Did not find '${libPath}'`);
+        return undefined;
+    }
+
+    const sitePackagesPath = combinePaths(libPath, pathConsts.sitePackages);
+    if (fs.existsSync(sitePackagesPath)) {
+        importFailureInfo.push(`Found path '${sitePackagesPath}'`);
+        return sitePackagesPath;
+    } else {
+        importFailureInfo.push(`Did not find '${sitePackagesPath}', so looking for python subdirectory`);
+    }
+
+    // We didn't find a site-packages directory directly in the lib
+    // directory. Scan for a "python*" directory instead.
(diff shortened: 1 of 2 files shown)


packages/pyright-internal/src/analyzer/pythonPathUtils.ts
```diff
@@ -69,59 +102,30 @@ export function findPythonSearchPaths(
...
             }
+        });
+
+        if (foundPaths.length > 0) {
+            importFailureInfo.push(`Found the following '${pathConsts.sitePackages}' dirs`);
+            foundPaths.forEach((path) => {
+                importFailureInfo.push(`  ${path}`);
+            });
+            return foundPaths;
+        } else {
```
jakebailey, line 124: Nit: else block is redundant; the previous block returns and the old code would be equivalent.

packages/pyright-internal/src/common/pathConsts.ts
```diff
@@ -8,5 +8,7 @@
 
 export const typeshedFallback = 'typeshed-fallback';
 export const lib = 'lib';
+export const lib64 = 'lib64';
+export const Lib = 'Lib';
```
jakebailey, line 12: Technically I think we got this one for free; `Lib` in caps is only on Windows, which is case insensitive anyway. Doesn't hurt.

truls: Thanks for the review. Are you suggesting to remove it? I wondered which platform the capitalized Lib was for.

jakebailey: No, it's fine; I can see that the original code checked it (missed it on my first read).

erictraut:
Thanks for the contribution!

PR #1168, opened 2020-11-13
Improve suggestion results

cdce8p:
A first attempt to improve the suggestion list. See: https://github.com/microsoft/pylance-release/issues/608

Instead of the Levenshtein Distance I though that Longest Common Subsequence might work better. The only drawback being that that user must only enter chars that appear in the symbol he/she is looking for and they must be in the correct order.

An example result would look like this:
Screen Shot 2020-11-13 at 15 18 38

Please let me know if that's something worth exploring further.

--
Algorithm used: Similar to `isPatternInWord` from [VS Code](https://github.com/microsoft/vscode/blob/21dc66054203ab742d36be9f0ef6ecb774ae62f2/src/vs/base/common/filters.ts#L506-L514)

### DIFF
--- a/packages/pyright-internal/src/common/stringUtils.ts
+++ b/packages/pyright-internal/src/common/stringUtils.ts
@@ -11,86 +11,53 @@ import leven from 'leven';
 
 import { compareComparableValues, Comparison } from './core';
 
-// Determines how closely a typed string matches a symbol
-// name. An exact match returns 1. A match that differs
-// only in case returns a slightly lower number. A match
-// that involves a few missing or added characters returns
-// an even lower number.
+// Determines if typed string matches a symbol
+// name. Characters must appear in order.
+// Return 1 if all typed characters are in symbol
 export function computeCompletionSimilarity(typedValue: string, symbolName: string): number {
-    if (symbolName.startsWith(typedValue)) {
-        return 1;
-    }
-
-    const symbolLower = symbolName.toLocaleLowerCase();
     const typedLower = typedValue.toLocaleLowerCase();
-
-    if (symbolLower.startsWith(typedLower)) {
-        return 0.75;
-    }
-
-    // Check longest common subsequence of symbolLower and typedLower
-    // Store max number of continuous characters that match
-    let numCommonChars = 0;
-    let longestSequence = 0;
-    let currentLongestSequence = 0;
-    let currentSearchIndex = 0;
-    for (let i = 0; i < typedLower.length; i++) {
-        const index = symbolLower.indexOf(typedLower[i], currentSearchIndex);
-        if (index !== -1) {
-            numCommonChars++;
-            currentLongestSequence++;
-            currentSearchIndex = index + 1;
-            if (longestSequence < currentLongestSequence) {
-                longestSequence++;
-            }
-        } else {
-            currentLongestSequence = 0;
-            currentSearchIndex = i + 1;
+    const symbolLower = symbolName.toLocaleLowerCase();
+    const typedLength = typedLower.length;
+    const symbolLength = symbolLower.length;
+    let typedPos = 0;
+    let symbolPos = 0;
+    while (typedPos < typedLength && symbolPos < symbolLength) {
+        if (typedLower[typedPos] === symbolLower[symbolPos]) {
+            typedPos += 1;
         }
-        if (currentSearchIndex > symbolLower.length - 1) {
-            break;
-        }
+        symbolPos += 1;
     }
-
-    // Exclude symbol if the user has typed characters that didn't match
-    if (typedValue.length !== numCommonChars) {
-        return 0;
-    }
-
-    // If typed string matches continuously, it's probably a good fit
-    if (typedValue.length === longestSequence) {
-        return 0.7;
-    }
-
-    // Everything else that matched, probably not as good
-    // TODO some logic that evaluates longestSequence in comparison to typedValue.length
-    return 0.4;
-
-    // How far apart are the two strings? Find the smallest edit
-    // distance for each of the substrings taken from the start of
-    // symbolName.
-    // let symbolSubstrLength = symbolLower.length;
-    // let smallestEditDistance = Number.MAX_VALUE;
-    // while (symbolSubstrLength > 0) {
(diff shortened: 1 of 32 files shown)


erictraut:
We've received feedback from many other pyright/pylance users that they don't want to be restricted to only characters that appear within the symbol, so I think we'd receive a significant backlash if we went with this approach. The approach needs to accommodate small typos (different order, different capitalization, small deviations in characters).

cdce8p:
Capitalization is already taken care of, since it compares all lower case strings.
Regarding the other points: Maybe a combined approach would work. I'll think about it.

heejaechang:
returning result that doesn't contain given pattern will be filtered out by vscode so regardless whether LS returns them user will never be able to see those?

you can try that by enabling LSP logging and what shows  up in completion. (https://github.com/microsoft/pylance-release/issues/362#issuecomment-726565780)

by the way, I think we don't need to do anything more than what vscode recommended (https://github.com/microsoft/vscode/blob/21dc66054203ab742d36be9f0ef6ecb774ae62f2/src/vs/base/common/filters.ts#L506-L514) for vscode since it does its own filter over data returned.

not sure about other editors that support LSP such as atom, vim-lsp and etc. for those, we might use different algorithm?

cdce8p:
I tried a few things and basically observed what @heejaechang mentioned already. VSC does exact pattern matching for the typed string. Setting `"editor.suggest.filterGraceful": true` (the default) allows for probably one permutation but that's it. Other results are filtered although Pyright provided them (using Levenshtein Distance). Additionally if the user provided chars that don't appear in the symbol, it is also filtered.

Having this in mind, I don't see much value in sticking with the current approach. Then again it's not my decision to make.
We could also use a combined method which I suggested previously. However this would probably come with at least a small performance hit, depending on the project size.

erictraut:
Yeah, if that's the case, then I agree that sticking with the current approach doesn't provide much value.

cdce8p:
I've change the algorithm to the one @heejaechang suggested and VS Code uses: `isPatternInWord` ([Link](https://github.com/microsoft/vscode/blob/21dc66054203ab742d36be9f0ef6ecb774ae62f2/src/vs/base/common/filters.ts#L506-L514)). Before that I also did some tests with a recursive one to allow for permutations. I'll include it below, however I'm not sure if it's worth it.

```ts
export function computeCompletionSimilarity(typedValue: string, symbolName: string): number {
    if (recurseSimilarity(typedValue.toLocaleLowerCase(), symbolName.toLocaleLowerCase())) {
        return 1;
    }
    return 0;
}

function recurseSimilarity(typedLower: string, symbolLower: string): boolean {
    if (typedLower.length === 0) {
        return true;
    }
    const index = symbolLower.indexOf(typedLower[0]);
    if (index === -1) {
        return false;
    }
    if (typedLower.length === 1) {
        return true;
    } else if (typedLower.length === 2) {
        symbolLower = symbolLower.slice(index + 1);
        return recurseSimilarity(typedLower.slice(1), symbolLower);
    } else {
        symbolLower = symbolLower.slice(index + 1);
        return (
            recurseSimilarity(typedLower.slice(1), symbolLower) ||
            recurseSimilarity(typedLower[2] + typedLower[1] + typedLower.slice(3), symbolLower)
        );
    }
}
```

packages/pyright-internal/src/common/stringUtils.ts
```diff
@@ -7,51 +7,25 @@
...
 
-// Determines how closely a typed string matches a symbol
-// name. An exact match returns 1. A match that differs
-// only in case returns a slightly lower number. A match
-// that involves a few missing or added characters returns
-// an even lower number.
+// Determines if typed string matches a symbol
+// name. Characters must appear in order.
+// Return 1 if all typed characters are in symbol
 export function computeCompletionSimilarity(typedValue: string, symbolName: string): number {
```
heejaechang, line 15: can you leave this code here and add another method that uses isPatternWord? asking since I still need old one for code action.

heejaechang:
I agree that we don't do permutation for now. for me, code LGTM, but I will leave it to @erictraut to decide.

packages/pyright-internal/src/common/stringUtils.ts
jakebailey: Yes, the old one should stay (and its dependency left in package.json). We can remove the dead code later if we eliminate it.

cdce8p:
I've pushed a new commit with all requested changes, see commit description.
Should I remove all references to `similarityLimit`, now that it isn't used anymore?

jakebailey:
`similarityLimit` appears to be a local in a bunch of modules, I think, so would be safe to remove if truly not used. It's the exported code that we wouldn't want to remove (given we do reference things from pylance).

cdce8p:
@jakebailey Should it remain in function signatures of exported classes?

jakebailey:
Oh, my mistake. I see, it's a parameter. I would just leave it and we can perform the cleanup later as it's a part of the API.

packages/pyright-internal/src/tests/common.test.ts
```diff
@@ -6,7 +6,6 @@
 
 import * as assert from 'assert';
 
-import { computeCompletionSimilarity } from '../common/stringUtils';
```
heejaechang, line 9: can you leave this test in since this still exist in pyright?

heejaechang: oops. ignore, you have the test in other places.

heejaechang:
it looks good to me? if there is no objection, I can accept the PR.

packages/pyright-internal/src/tests/common.test.ts
cdce8p: I moved it to `tests/stringUtils.test.ts`, since that file already existed.

heejaechang:
@cdce8p we will take the PR soon. but probably for the next release (Pylance) due to some change we need to make for add import code actions.

cdce8p:
@heejaechang Thanks for letting me know. Looking forward to using it when it's released!

jakebailey:
Thanks for this; it's not likely to come out in Pylance next week (holiday week), but should make it afterward. 🙂

PR #1173, opened 2020-11-15
Add __future__ module as import suggestion

cdce8p:
See: https://github.com/microsoft/pylance-release/issues/539


jakebailey:
Please don't add "fixes" to the PR description for the linked issues; we only close issues once the code is actually released (to better indicate when it's actually available) and GitHub will auto-close them.

erictraut:
This change looks good to me, but a unit test should be added for it.

cdce8p:
I've added a unit test

erictraut:

PR #1174, opened 2020-11-15
Expand type alias for completion provider

cdce8p:
Match type alias from completion provider with hover provider. The type alias will only be expanded if the completion item is the alias itself to preserve readability.

See: https://github.com/microsoft/pylance-release/issues/562

 https://github.com/microsoft/pyright/blob/51ad4c20df30bae7acbdd23fdb04ab88cda7b3f5/packages/pyright-internal/src/languageService/hoverProvider.ts#L161-L174

Screen Shot 2020-11-16 at 00 06 32

Screen Shot 2020-11-16 at 00 08 14


erictraut:
This change looks good to me, but a unit test should be added for it.

cdce8p:
I've added a unit test

erictraut:

PR #1177, opened 2020-11-16
Don't add suggestions for 'with Y as [ ]'

cdce8p:
Pyright shouldn't attempt to provide suggestions after `with Y as [ ]`.

### Before
Screen Shot 2020-11-16 at 14 31 13

### After
Screen Shot 2020-11-16 at 14 27 48


erictraut:

PR #1178, opened 2020-11-16
Improve handling of public symbols for imports

cdce8p:
* Don't exclude CamelCase variables / type aliases
* Exclude protected and private constants

See: https://github.com/microsoft/pylance-release/issues/606


erictraut:
Nice! And thanks for providing the unit tests too.

erictraut:
@heejaechang, could you please review this change?

packages/pyright-internal/src/tests/analyser.test.ts
```diff
@@ -0,0 +1,77 @@
+/*
+ * symbolNameUtils.test.ts
```
jakebailey, line 2: Nit: Filename does not match, and by this repo's convention the spelling would be "analyzer" if anything. I'd prefer if it were named "symbolNameUtils.test.ts", however.

cdce8p: Good catch! `symbolNameUtils.test.ts` was what I originally though off. Will change it back

packages/pyright-internal/src/languageService/autoImporter.ts
```diff
@@ -279,11 +279,11 @@ export class AutoImporter {
 
             if (
                 !isStubOrHasInit.isStub &&
-                autoImportSymbol.kind === CompletionItemKind.Variable &&
-                /[a-z]/.test(name)
+                (autoImportSymbol.kind === undefined || autoImportSymbol.kind === CompletionItemKind.Variable) &&
```
heejaechang, line 282: which Symbol Kind is "autoImportSymbol.kind === undefined" for?

packages/pyright-internal/src/languageService/autoImporter.ts
```diff
@@ -279,11 +279,11 @@ export class AutoImporter {
...
+                !SymbolNameUtils.isPublicConstantOrTypeAlias(name)
```
heejaechang, line 283: previously, name check was negative check based on the fact that the lexer has used proper identifier check (https://github.com/microsoft/pyright/blob/master/packages/pyright-internal/src/parser/characters.ts#L52) which allows non enlgish unicode as variable name.

now, only English name can be used in auto import variable. probably okay? but want to make sure that everyone is okay with that.

ex) before one could have 검은색상값 = "BLACK" and it will show up in auto import, now it won't.

packages/pyright-internal/src/languageService/autoImporter.ts
heejaechang: I am asking since now it will check every symbol rather than only variable symbol kind.

packages/pyright-internal/src/languageService/autoImporter.ts
erictraut: @heejaechang, good question. Are you familiar with the naming conventions used in locales that don't use Roman (ASCII) characters? Would it be common to see someone use "검은색상값" as the name of a type alias, or would that be unexpected and atypical? If it's expected, then is there a common heuristic we could use here? How about "starts with a capital letter and has no underscores?"

packages/pyright-internal/src/languageService/autoImporter.ts
heejaechang: I think what you want is tweaking this (https://github.com/microsoft/pyright/blob/d5b7725a6f60248d515e1541a5674f9f487443e0/packages/pyright-internal/src/languageService/autoImporter.ts#L94), not kind === undefined.

heejaechang: exiting code basically says, if pyright already figured out that variable is "Final" or "Const", then we will always show it in auto-import regardless of how the name is written.

packages/pyright-internal/src/languageService/autoImporter.ts
heejaechang: I think using non-roman character as a variable name is not common (in Korea at least) since usually, tool support is so bad even if language spec allows it :) but that being said, I believe in some countries (one of them is Japan I believe), it is used more often.

but, this is only for auto-import, but not actual variable name, so might be okay?

erictraut: OK, sounds good to me. Let's stick with this for now. If it's too restrictive, I'm sure we'll get feedback.

packages/pyright-internal/src/languageService/autoImporter.ts
cdce8p: I see. Originally I noticed that protected constants where also showing up, which is why I added that check. However after you mentioning it, I now believe that it might be for the better the way constants are currently handled. Unlike with variables, there are at least some scenarios in which you might what to import them.

I'll change it.

heejaechang (approved):

PR #1179, opened 2020-11-16
Improve except suggestions

cdce8p:
Don't include `NotImplemented` in except suggestions. It should be `except NotImplementedError`.
This is a common beginner mistake.
See: https://github.com/microsoft/pylance-release/issues/563

--
The change works as intended, but it doesn't feel right. Maybe there is another position to insert this check. Ideally that would also allow us to also check every other suggestion if it inherits from `BaseException`.

### DIFF
--- a/packages/pyright-internal/src/languageService/completionProvider.ts
+++ b/packages/pyright-internal/src/languageService/completionProvider.ts
@@ -991,6 +991,16 @@ export class CompletionProvider {
             }
         }
 
+        if (parseNode.parent?.nodeType === ParseNodeType.Except) {
+            const completionItems: CompletionItem[] = [];
+            completionList.items.forEach((item) => {
+                if (item.label !== 'NotImplemented') {
+                    completionItems.push(item);
+                }
+            });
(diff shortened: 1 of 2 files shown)


jakebailey:
We haven't really made a decision on this yet (https://github.com/microsoft/pylance-release/issues/563#issuecomment-722541593), but I'd think the fix would not be to special case this one thing, but to not try and offer non-exception typed things in this block.

erictraut:
Why would we special-case this particular symbol? If there's a general fix to this problem, I'm open to it, but I don't want to start hacking in knowledge of specific classes like this.

cdce8p:
The only reason to add it, would be to prevent a common beginner mistake. As I've said above I don't really like this change either. However as far as my limited understanding goes there isn't an option to check for the general case, at least not without some bigger changes.

Maybe you know a better place that I've missed. If not, I would completely understand it if you closed this issue.

erictraut:
I don't think we should close the issue, but this isn't the right fix. We'll keep it open until we have time to look at it.

PR #1259, opened 2020-12-09
Changed encoding of diagnostics reported through the LSP interface. T…

erictraut:
…he diagnostic rule (if applicable) is now reported in the "code" field, and a URL points to general documentation for diagnostic rules.

### DIFF
--- a/packages/pyright-internal/src/languageServerBase.ts
+++ b/packages/pyright-internal/src/languageServerBase.ts
@@ -1229,9 +1229,9 @@ export abstract class LanguageServerBase implements LanguageServerInterface {
         });
     }
 
-    protected getDocumentationUrlForDiagnosticRule(rule: string) {
+    protected getDocumentationUrlForDiagnosticRule(rule: string): string | undefined {
         // For now, return the same URL for all rules. We can separate these
         // in the future.
-        return 'https://github.com/microsoft/pylance-release/blob/master/DIAGNOSTIC_SEVERITY_RULES.md';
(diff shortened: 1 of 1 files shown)


packages/pyright-internal/src/languageServerBase.ts
```diff
@@ -1225,4 +1228,10 @@ export abstract class LanguageServerBase implements LanguageServerInterface {
             workspace.serviceInstance.recordUserInteractionTime();
         });
     }
+
+    protected getDocumentationUrlForDiagnosticRule(rule: string) {
+        // For now, return the same URL for all rules. We can separate these
+        // in the future.
+        return 'https://github.com/microsoft/pylance-release/blob/master/DIAGNOSTIC_SEVERITY_RULES.md';
```
jakebailey, line 1235: Maybe the pyright equivalent here, and then we override for pylance?

packages/pyright-internal/src/languageServerBase.ts
```diff
@@ -1225,4 +1228,10 @@ export abstract class LanguageServerBase implements LanguageServerInterface {
             workspace.serviceInstance.recordUserInteractionTime();
         });
     }
+
+    protected getDocumentationUrlForDiagnosticRule(rule: string) {
```
jakebailey, line 1232: Should this always return a string, or would it be `string | undefined`? Looking at the usage above I'm not sure, but I can't see us ever returning nothing for a given rule.

rwols (approved):

jakebailey (approved):

PR #1267, opened 2020-12-11
Expand ${HOME} in settings

ashb:
It is useful to be able to have virtualenvironments in a shared folder
(`~/.virtualenvs` in my case) and have the settings work without hard-coding a
path.

This probably doesn't make any sense on Windows.

(I use coc-pyright, not VSCode if it makes any difference)


erictraut:
No, we don't plan to add support for shell variables of any sort in pyrightconfig.json, which is meant to be OS-agnostic and shell-agnostic.

ashb:
Is there any way I can achieve this? I want to use the same config on multiple machines where I have different user names.

If I extended this to instead allow _any_ environment variable would it be acceptable, as it is then not OS specific.

erictraut:
The config mechanism is already designed to be used with a virtual environment in a shared folder. You can specify the location of your venv shared folder via the setting `python.venvPath` (or if you're using the CLI version, with the `--venv-path` switch). This allows each user to specify a different path location. The pyrightconfig.json file (which is assumed to be the same for all users working on a project) can then refer to a desired venv by name (which will be common to all users).

ashb:
It's precisely this `python.venvPath` setting I want to have expand `~`/`${HOME}` variable in

erictraut:
You should use the expanded version in the setting.

Sorry, but we're not going to support shell variable expansion in pyrightconfig.json. As I said, it is meant to be OS and shell agnostic.

ashb:
I might be mis-understanding where this is coming from then. Is this code only for pyrightconfig.json? I thought it was also applied to  workspace/config message from the editor/language server

erictraut:
This was my misunderstanding. You are correct — this code path is not used for pyrightconfig.json. It's used for settings messages coming from the LSP client.

As you can see in this code, we do support expansion of the VS Code variable "workspaceFolder". I think we would entertain adding support for one or more additional [VS Code variables](https://code.visualstudio.com/docs/editor/variables-reference). That would be consistent with the existing code. I don't think it's a good idea to expand variables from other shells.

jakebailey:
This code is unfortunately not used super consistently (see microsoft/pylance-release#548). It'd be best to perform this processing early in the settings change code, but this isn't called for that path I don't think.

I think making this consistently used in settings would be key, as well as having the same replacements as the Python extension. In this case, they would have needed `${env:HOME}`, not `${HOME}`.

ashb:
Thanks for the explanations - this code base (the whole LSP ecosystem is new to me.)

Should I adjust this pr to use `env:` style prefix, or is bigger change needed to handle it consistently?

jakebailey:
I think this is fine if we only accept exactly `env:HOME` for now and no other variables.

To handle further replacements for Python extension consistency as well as to deal with changes/env files, that will take more effort, and shifting stuff around and defining the custom APIs to handle that is on my plate next.

erictraut:
OK, that sounds like a reasonable solution. "env:HOME" is effectively in the same namespace as what we support currently. @ashb, if that sounds good to you, you can push the change to your branch.

ashb:
Will do! Are there any tests I should update too?

jakebailey:
No, not for this code path at the moment. The three line change should be sufficient.

ashb:
Done @jakebailey @erictraut.

I didn't pull in @types/node or import anything for process, following existing patterns `packages/pyright-internal/src/pyright.ts` mostl.

ashb:
I think this fix didn't quite work as hoped. If I set this in my editor config (`coc-settings.json`):

```json
  "python.venvPath":"${env:HOME}/.virtualenvs",
```

I then see this in the language server logs.

```
[Error  - 10:55:37] venvPath /home/ash/code/airflow/airflow/home/ash/.virtualenvs is not a valid directory.
```

D'oh!

ashb:
Setting it to

```json
  "python.venvPath":"/${env:HOME}/.virtualenvs",
```

Does work --verified by setting a junk `venv` in pyrightconfig.json, where I see this error:

```
[Error  - 11:08:44] venv airflow23432 subdirectory not found in venv path /home/ash/.virtualenvs.
```

I'm trying to work out how to actually get a debug/test build of this used in my editor :)

ashb:
Oh I see the problem, https://github.com/microsoft/pyright/blob/6ff4201d89df2a32f4ce670551ff6f03a90cac4a/packages/pyright-internal/src/server.ts#L77-L80 makes the path "absolute", so by the time the analyzser service see it has been changed to `/home/ash/code/airflow/airflow/${env:HOME}`.

This would be the case even with the existing `${workspaceFolder}` expansion used in venvPath I think.

PR #1269, opened 2020-12-11
Replaced bespoke progress-reporting messages with LSP-defined server-…

erictraut:
…initiated progress reporting mechanism.


packages/pyright-internal/src/languageServerBase.ts
```diff
@@ -139,6 +139,7 @@ export interface LanguageServerInterface {
 // in the public interface.
 export interface ProgressReporterConnection {
     sendNotification: (method: string, params?: any) => void;
+    window: RemoteWindow;
```
jakebailey, line 142: This interface must not contain any types that come from the LSP, otherwise the use of this in Pylance breaks at compile time due to the LSP library unfortunately using concrete types in some of its methods. See the comment above.

packages/pyright-internal/src/server.ts
```diff
@@ -203,15 +206,27 @@ function reporterFactory(connection: ProgressReporterConnection): ProgressReport
         },
 
         begin(): void {
-            connection.sendNotification('pyright/beginProgress');
+            _workDoneProgress = connection.window.createWorkDoneProgress();
```
jakebailey, line 209: This variable would be better served as a variable defined inside of this function. It's not accessed anywhere else.

jakebailey: "This function" meaning reporterFactory, I mean.

packages/pyright-internal/src/languageServerBase.ts
erictraut: I don't fully understand the problem. The "window" variable isn't a generic type, so I would think this addition would work fine.

packages/pyright-internal/src/server.ts
erictraut: Good point. I've updated the code accordingly.

packages/pyright-internal/src/languageServerBase.ts
jakebailey: It's due to the fact that we have multiple `node_modules` folders, and two LSP server libs are treated as distinct. Many types (including `RemoteWindow` from what I am reading) are defined in terms of specific classes in the LSP server library, so when we try to implement this interface from Pylance, it fails because it believes that the types are different due to private class members breaking comparison.

The only way that we'll be able to fix this is by getting dependency hoisting to work to share dependencies between packages including within our nested monorepo setup for Pylance, something I've been working on in my free time (and am close to, but only for finalized builds).

jakebailey: If we want to add to this, we'll have to ensure that we use a type that's compatible. `WindowProgress` _might` work as a replacement, but I'd have to test it by porting this to test.

jakebailey: More likely, we can instead refactor this to just provide the factory that only returns the reporter, then both codebases can implement it as they need without having to expose the internals.

jakebailey: Did some checking now that I could look, and this would break things due to `RemoteWindow` depending on `Connection` depending on `Request0`, which is where the concrete classes with private properties begin setting in. Typing this as `WindowProgress` would work, though, as it is free of any bad types.

But the more I look at this, the more I see that the "reporter factory" could have just been an abstract method and be implemented per LS class, preventing more of this type oddity. I could make that change in Pylance first and merge it back to lower the merge annoyance.

erictraut: I'm not sure I fully follow the details, but it sounds like a reasonable plan to me. If it's helpful to do so, feel free to use this branch.

erictraut:
Abandoning this PR in favor of a better implementation that Jake is working on.

PR #1272, opened 2020-12-12
Added a new diagnostic rule called "reportInvalidTypeVarUse" that fla…

erictraut:
…gs errors when TypeVars are used incorrectly. In particular, it flags the use of a single instance of a TypeVar within a generic function signature.


erictraut:
@jakebailey, I'm interested in your thoughts on this change. It's one that I introduced a while back and then backed out. But we're continuing to get a stream of bug reports based on a misuse of TypeVars in generic function signatures. Most of this confusion could be avoided through the introduction of this diagnostic.

jakebailey:
In principle I'm for it, if this is a bad-behavior that type checkers are going to agree on disallowing. I think there is still some typing-sig threading happening here, though.

PR #1283, opened 2020-12-16
Expand variables in paths early

ashb:
Variables can be used in paths (`${workspaceFolder}` or `${env:HOME}`)
but they were already being resolved to absolute paths before expanding
the variables, which made the expansions not do the "right" thing.

This commit changes it to expand the variables early, in the server
constructor.

Relates to #1267

### DIFF
--- a/packages/pyright-internal/src/server.ts
+++ b/packages/pyright-internal/src/server.ts
@@ -52,6 +52,24 @@ class PyrightServer extends LanguageServerBase {
         this._controller = new CommandController(this);
     }
 
+    // Expands certain predefined variables supported within VS Code settings.
+    // Ideally, VS Code would provide an API for doing this expansion, but
+    // it doesn't. We'll handle the most common variables here as a convenience.
+    _expandPathVariables(rootPath: string, value: string): string {
+        const regexp = /\$\{(.*?)\}/g;
+        return value.replace(regexp, (match: string, name: string) => {
+            const trimmedName = name.trim();
+            if (trimmedName === 'workspaceFolder') {
+                return rootPath;
+            }
+            if (trimmedName === 'env:HOME' && process.env.HOME !== undefined) {
+                return process.env.HOME;
+            }
+
+            return match;
+        });
+    }
+
     async getSettings(workspace: WorkspaceServiceInstance): Promise<ServerSettings> {
         const serverSettings: ServerSettings = {
             watchForSourceChanges: true,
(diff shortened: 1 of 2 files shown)


erictraut:
Thanks for debugging this. I had assumed you tested it prior to submitting the previous PR.

The problem with the new change is that you've taken the variable expansion code out of a module that is common to Pyright and Pylance and moved it into a code module (`server.ts`) that is specific to Pyright. That means this change will break variable expansion for Pylance. We'll have to find some other way to address this.

ashb:
> Thanks for debugging this. I had assumed you tested it prior to submitting the previous PR.

I thought I did, but just badly. Sorry.

ashb:
>  That means this change will break variable expansion for Pylance.

Ahh right.

One possible change might be to not have server.ts normalize the path, and instead delay it until the previous location (which was doing it again anyway).

I haven't yet looked at where else those serverSettings might be used.

packages/pyright-internal/src/server.ts
```diff
@@ -52,6 +52,24 @@ class PyrightServer extends LanguageServerBase {
         this._controller = new CommandController(this);
     }
 
+    // Expands certain predefined variables supported within VS Code settings.
+    // Ideally, VS Code would provide an API for doing this expansion, but
+    // it doesn't. We'll handle the most common variables here as a convenience.
+    _expandPathVariables(rootPath: string, value: string): string {
```
jakebailey, line 58: This should probably be `protected`.

jakebailey:
I left one comment about adding `protected` to a function (that we'll have to call from Pylance), but other than that I think this PR looks fine (but maybe should be tested this time! 🙂). I'll just need to ensure that when we pull next, this code is updated so we don't end up in a bad state.

packages/pyright-internal/src/server.ts
ashb: Python developer -- not use to having an actual formal `protected` :)

jakebailey: No specifier would mean public, so it's not a huge problem for us, just a style nit. 🙂

ashb:
Editor config:

```
{
  "python.venvPath":"${env:HOME}/.virtualenvs"
}
```

pyrightconfig.json:
```
  "venv": "airflow23432"
```

Language server logs:

```
[Info  - 19:25:37] Pyright language server 1.1.95 starting
[Info  - 19:25:37] Server root directory: /home/ash/.config/coc/extensions/node_modules/coc-pyright/node_modules/pyright/dist/
[Info  - 19:25:37] Loading configuration file at /home/ash/code/airflow/airflow/pyrightconfig.json
[Info  - 19:25:37] Assuming Python version 3.7
[Info  - 19:25:37] Assuming Python platform Linux
[Info  - 19:25:37] No include entries specified; assuming /home/ash/code/airflow/airflow
[Error  - 19:25:37] venv airflow23432 subdirectory not found in venv path /home/ash/.virtualenvs.
[Error  - 19:25:37] stubPath /home/ash/code/airflow/airflow/typings is not a valid directory.
[Error  - 19:25:37] Exception received when installing recursive file system watcher
[Info  - 19:25:37] Searching for source files
[Info  - 19:25:37] Found 2599 source files
```

_Actually_ tested it this time (I sort of assumed the `${workspaceFolder}` expansion would work how I expected. Sorry), and the variables are expanded, and correctly/as expected, from the config in my editor (`coc-settings.json` in my case.)

packages/pyright-internal/src/server.ts
ashb: I've made this `protected expandPathVariables` now.

jakebailey (approved):
Thanks!

PR #1294, opened 2020-12-18
Added a heuristic to skip call-site return type inference if the numb…

erictraut:
…er of arguments is above a threshold (6). This avoids long analysis times for complex unannotated functions.

### DIFF
--- a/packages/pyright-internal/src/analyzer/typeEvaluator.ts
+++ b/packages/pyright-internal/src/analyzer/typeEvaluator.ts
@@ -590,6 +590,12 @@ interface ReturnTypeInferenceContext {
 // performance.
 const maxReturnTypeInferenceStackSize = 3;
 
+// What is the max number of input arguments we should allow
+// for call-site return type inference? We've found that large,
+// complex functions with many arguments can take too long to
+// analyze.
+const maxReturnTypeInferenceArgumentCount = 6;
+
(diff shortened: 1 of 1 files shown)


packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -14625,6 +14631,13 @@ export function createTypeEvaluator(importLookup: ImportLookup, evaluatorOptions
             return undefined;
         }
 
+        // Very complex functions with many arguments can take a long time to analyze,
+        // so we'll use a heuristic and avoiding this inference technique for any
+        // call site that involves too many arguments.
+        if (args.length > maxReturnTypeInferenceArgumentCount) {
+            return undefined;
+        }
```
bschnurr, line 14639: I was also thinking about counting the unknown parameters when calculating allArgTypesAreUnknown and if the number of unknown parameters was over a certain value dont evaluate inferFunctionReturnType

bschnurr: No your fix is better

bschnurr: testing this again.

bschnurr (approved):

Issue #862, opened 2020-07-20
pre-commit hook raising an error about missing `./dist/pyright`


brettcannon:
**Describe the bug**
```
pyright..................................................................Failed
- hook id: pyright
- exit code: 1

internal/modules/cjs/loader.js:1033
  throw err;
  ^

Error: Cannot find module './dist/pyright'
Require stack:
- /Users/brettcannon/.cache/pre-commit/repopxecb0be/index.js
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:1030:15)
    at Function.Module._load (internal/modules/cjs/loader.js:899:27)
    at Module.require (internal/modules/cjs/loader.js:1090:19)
    at require (internal/modules/cjs/helpers.js:75:18)
    at Object.<anonymous> (/Users/brettcannon/.cache/pre-commit/repopxecb0be/index.js:6:1)
    at Module._compile (internal/modules/cjs/loader.js:1201:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1221:10)
    at Module.load (internal/modules/cjs/loader.js:1050:32)
    at Function.Module._load (internal/modules/cjs/loader.js:938:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:71:12) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [ '/Users/brettcannon/.cache/pre-commit/repopxecb0be/index.js' ]
}
```

**To Reproduce**
Set up pyright in pre-commit:
```yaml
 - repo: https://github.com/microsoft/pyright
   rev: 1.1.53
   hooks:
   - id: pyright
```

bschnurr:
'./dist/pyright'  gets built by running "npm run package"

erictraut:
@brettcannon, the hook assumes that the Pyright CLI has been installed via npm. Have you done that? Alternatively, you could build the full cli using "npm run package", as Bill mentioned.

brettcannon:
@erictraut no as pre-commit is often used to isolate a project from having to install individual tools (e.g. just run `pre-commit` in CI and have everything just work), and specifically in pyright's case I was hoping I could avoid doing anything directly npm-related.

jakebailey:
The pre-commit hook appears to declare pyright as using the python language (technically true). Does the config need to include more steps / more dependencies to make it work? I'm unfamiliar with pre-commit.

[...]

PR #1318, opened 2020-12-27
feat: Integration with pre-commit

kumaraditya303:
Integration with [pre-commit](https://github.com/pre-commit/pre-commit) would make *Pyright* easier to run on CI environments.


erictraut:
We previously had a similar pre-commit hook in this repo, but it didn't work. See the discussion in [this issue](https://github.com/microsoft/pyright/issues/862). We opted instead to add [this documentation](https://github.com/microsoft/pyright/blob/master/docs/ci-integration.md) for users who were interested in using pyright for CI hooks. 

Please let us know how your proposed hook is different from the one that we ended up deleting and why it's preferable to the documentation that's already in place.

kumaraditya303:
> We previously had a similar pre-commit hook in this repo, but it didn't work. See the discussion in [this issue](https://github.com/microsoft/pyright/issues/862). We opted instead to add [this documentation](https://github.com/microsoft/pyright/blob/master/docs/ci-integration.md) for users who were interested in using pyright for CI hooks.
> 
> Please let us know how your proposed hook is different from the one that we ended up deleting and why it's preferable to the documentation that's already in place.

The pre-commit hooks proposed by this PR 🚀 is different as it installs pyright as a dependency hence it is installed normally using node, hence by this pyright is already built during the install process.

erictraut:
@jakebailey, @brettcannon, you have more experience than me with github hooks, so I'd appreciate your review here.

README.md
```diff
@@ -89,6 +89,18 @@ Pyright’s type-checking functionality and language features are now incorporat
 Installing both Pyright and Pylance at the same time is not recommended. If both are installed and enabled, you will see duplicate errors, hover text, and completion suggestions. 
 
 
+## Using pyright with pre-commit
```
erictraut, line 92: I don't think this documentation belongs in the core README.md file. It should replace the current documentation in the `ci-integration.md` file.

kumaraditya303: In the next commit, it would be updated. 🚀

.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,6 @@
+- id: pyright
+  name: pyright
+  entry: pyright .
+  language: node
+  types: [python]
+  additional_dependencies: ['pyright@1.1.97']
```
erictraut, line 6: This will require updating every time I release a new version of pyright. I'm not crazy about adding that additional maintenance work. What is the benefit of adding a ".pre-commit-hooks.yaml" to this repo versus documenting the contents of a ".pre-commit-hook.yaml" that someone else would need to check in to their repo to use pyright in their CI? That way, they can decide which version of pyright to use and when to upgrade it.

kumaraditya303:
I have created a automated repo for `pyright`  for pre-commit  with all the previous versions. You may mention in docs. https://github.com/kumaraditya303/mirrors-pyright

PR #1331, opened 2020-12-31
Improving error messages for expected TypeDicts

sransara:
When TypedDict is an expected value, but the Dict expression does not fulfill the requirements, the generated error message does not provide enough information to narrow down the error.

```python
from typing import TypedDict, Dict

Movie = TypedDict("Movie", name=str, year=int)
move: Movie = {
    "name": "M",
    "year": "5"
    # this should generate an error about year
}

class Inner(TypedDict, total=False):
    inner_key: Dict[str, 'Inner']
    something: str

a: Inner = {
    "inner_key": {
        "akey": {
            "inner_key": {
                "bkey": {
                    "something": 22
                    # this should generate an error about "something"
                }
            }
        }
    }
}

```

Before this change:
```
  4:15 - error: Expression of type "dict[str, str]" cannot be assigned to declared type "Movie"
    "dict[str, str]" is incompatible with "Movie" (reportGeneralTypeIssues)
  14:12 - error: Expression of type "dict[str, dict[str, dict[str, dict[str, dict[str, int]]]]]" cannot be assigned to declared type "Inner"
    "dict[str, dict[str, dict[str, dict[str, dict[str, int]]]]]" is incompatible with "Inner" (reportGeneralTypeIssues)
```

After this change:
```
  4:15 - error: Expression of type "dict[str, str]" cannot be assigned to declared type "Movie"
    Type "Literal['5']" is not assignable to field "year" (reportGeneralTypeIssues)
  14:12 - error: Expression of type "dict[str, dict[str, dict[str, dict[str, dict[str, int]]]]]" cannot be assigned to declared type "Inner"
    Type "dict[str, dict[str, dict[str, dict[str, int]]]]" is not assignable to field "inner_key"
      Type "dict[str, dict[str, int]]" is not assignable to field "inner_key"
        Type "Literal[22]" is not assignable to field "something" (reportGeneralTypeIssues)
```

Ground work had been done in the commit: https://github.com/microsoft/pyright/commit/dc033502ee7df5471e527397c5733e85f9b6fa43. But it may have regressed somewhere along the line.

All tests are passing. Feel free to let me know if you have suggestions. I would also like to work on to see if we can report the errors at a more a narrowed down levels than all at the outer expression.

PS: Thanks for this project. It has been an invaluable resource.

### DIFF
--- a/packages/pyright-internal/src/analyzer/typeEvaluator.ts
+++ b/packages/pyright-internal/src/analyzer/typeEvaluator.ts
@@ -8242,7 +8242,8 @@ export function createTypeEvaluator(importLookup: ImportLookup, evaluatorOptions
                 !!expectedType,
                 /* expectedKeyType */ undefined,
                 /* expectedValueType */ undefined,
-                expectedTypedDictEntries
+                expectedTypedDictEntries,
+                expectedDiagAddendum
             );
 
             if (
@@ -8290,7 +8291,9 @@ export function createTypeEvaluator(importLookup: ImportLookup, evaluatorOptions
             valueTypes,
             !!expectedType,
             expectedKeyType,
-            expectedValueType
+            expectedValueType,
+            undefined,
+            expectedDiagAddendum
         );
 
         const isExpectedTypeDict = isObject(expectedType) && ClassType.isBuiltIn(expectedType.classType, 'dict');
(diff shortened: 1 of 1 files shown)


erictraut:
Thanks for the contribution. Yeah, it looks like I broke this at some point and didn't notice the regression.

PR #1333, opened 2021-01-01
Go full typescript

kumaraditya303:


kumaraditya303:
@erictraut Not sure why the style check is failing https://github.com/microsoft/pyright/pull/1333/checks?check_run_id=1633688381

packages/pyright/webpack.config.ts
```diff
@@ -0,0 +1,84 @@
...
+        all: false,
+        errors: true,
+        warnings: true,
+    },
+    resolve: {
+        extensions: ['.ts', '.js'],
+        plugins: [
+            // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+            // @ts-ignore
+            new TsconfigPathsPlugin({
```
kumaraditya303, line 41: Upstream issue https://github.com/dividab/tsconfig-paths-webpack-plugin/issues/61

kumaraditya303:
@erictraut Not sure why the style check is failing

erictraut:
The existing build scripts are working fine for us and are designed to work with the build scripts for pylance, which is a private repo. It's not clear to me what benefit we get from these changes. It would have to be significant to offset the risk of regressions. In any case, I'm not in a position to review or approve this change because @jakebailey  is responsible for the existing scripts, so we'll need to wait for him to weigh in when he returns from the holidays.

kumaraditya303:
@erictraut only the style check is failing on `git diff --exit-code`  all the other checks have passed and also I have tested them locally. Only the style one is failing.

jakebailey:
Yeah, I don't know if this is a good idea quite yet; I'll have to ensure that this also happens for Pylance and update all of our scripts to be typescript aware, which will take some effort, plus dealing with the troubles of our nested project layout when it comes to TS configs.

This PR also adds in the old webpack 4 types, which aren't entirely accurate given we're on webpack 5. I'm not sure if that's really a regression, though; I think there are some oddities where TS was already picking up webpack types from the wrong place (even though webpack 5 should be shipping with webpack types).

There are also some package additions that I don't think are needed at all; the top-level workspace package definitely shouldn't need to know about webpack. pyright-internal is gaining more deps when no changes where made, etc.

erictraut:
I'm going to close this. If we decide to move forward with parts of it, we can refer back to it then.

PR #1344, opened 2021-01-08
Fix for missing docstrings in completions on a property where only the getter has documentation 

bschnurr:
Fix for missing docs in completion list due to only checking the setter for docs because its definition comes after the getter. This primary declaration logic comes from overloads were the doc strings are using on the last function. Now if the setter is missing a doc string we check the other declarations until we find a doc string.

During onHover we always look at the first declaration which would be the getter.

this._addResultsForDeclaration(format, sourceMapper, results.parts, declarations[0], node, evaluator);

in hoverProvider.ts

before 
image

after
image

better aligns with Hover
image

### DIFF
--- a/packages/pyright-internal/src/languageService/completionProvider.ts
+++ b/packages/pyright-internal/src/languageService/completionProvider.ts
@@ -31,6 +31,7 @@ import { getLastTypedDeclaredForSymbol } from '../analyzer/symbolUtils';
 import {
     getClassDocString,
     getFunctionDocStringFromDeclaration,
+    getFunctionDocStringFromDeclarations,
     getFunctionDocStringFromType,
     getModuleDocString,
     getOverloadedFunctionDocStrings,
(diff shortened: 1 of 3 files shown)


bschnurr:
One strange difference between completion and hover is that the hover shows Literal[3] being returned. Another is the ordering of "property". I see in Pycharm they mention the signature comes from the get property.

image

packages/pyright-internal/src/languageService/completionProvider.ts
```diff
@@ -1659,6 +1660,13 @@ export class CompletionProvider {
                             } else if (primaryDecl.type === DeclarationType.Function) {
                                 // @property functions
                                 documentation = getFunctionDocStringFromDeclaration(primaryDecl, this._sourceMapper);
```
bschnurr, line 1662: I could also just force this to be the first declaration

`getFunctionDocStringFromDeclaration(declarations[0], this._sourceMapper);`

erictraut:
Another approach (not sure if it's better) is to introspect the property object for the 'fget' method, which is the original getter method. The docstring will be intact there.

bschnurr:
Like you did for the typeDetail print ? with getGetterTypeFromProperty

bschnurr:
here is what PyCharm shows, their onHover "Accessor kind:" changes if the property is being used as a get or a set.
image

PR #1415, opened 2021-01-25
feat: Installation as a python package

kumaraditya303:
Platforms supported by this PR:

- [x]  Windows 64bit
- [x]  Windows 32bit
- [x] Linux 64bit
- [x] MacOS 64bit


kumaraditya303:
@erictraut I just saw that this Project is under MIT License which means I can distribute this so, do I have the permission to publish this package to PyPI as a python package?

erictraut:
I would really appreciate it if you didn't. We've already investigated this and concluded that it would not be a good idea.

kumaraditya303:
@erictraut Ok, I won't, thanks for the confirmation.

PR #1419, opened 2021-01-25
Fixed two issues in the import resolution logic. First, it was return…

erictraut:
…ing a namespace module if it found one in the workspace path or extraPaths even if a traditional (non-namespace) module satisfied the import from the PYTHONPATH. The interpreter searches all paths and always prefers a traditional module if it can find it. Second, it was resolving a namespace module if a traditional module only partially resolved the import path. The real interpreter always prefers a traditional module even if it partially resolves the path (in which case the full import fails).


jakebailey (approved):
Change looks good to me, though I'll be interested to see if more edge cases fall out of this handling.

Only nit is that the commit mentions PYTHONPATH, but meant `sys.path`.

PR #1496, opened 2021-02-16
Sync python.analysis.logLevel setting with the code

rchl:
According to the code the value of "python.analysis.logLevel" are
"error", "warn", "info" and "log", and not what was said in the
documentation and the schema.

### DIFF
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -20,7 +20,7 @@ The Pyright VS Code extension honors the following settings.
 
 **python.analysis.extraPaths** [array of paths]: Paths to add to the default execution environment extra paths if there are no execution environments defined in the config file.
 
-**python.analysis.logLevel** ["error", "warn", "info", or "log"]: Level of logging for Output panel. The default value for this option is "info".
+**python.analysis.logLevel** ["Error", "Warning", "Information", or "Trace"]: Level of logging for Output panel. The default value for this option is "Information".
 
 **python.analysis.stubPath** [path]: Path to directory containing custom type stub files.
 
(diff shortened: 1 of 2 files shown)


rchl:
I'm just wondering about a certain thing... Is "python.analysis.logLevel" setting specific to pyright or it's used more broadly in the VSCode ecosystem? If the latter then I guess it should be verified that all users of that option agree on the values.

jakebailey:
I'm pretty certain these log levels are read out from their setting and converted internally. They have a long history as the logLevel setting used in MPLS and they are used in Pylance as well. I don't think this should be changed, as it will just cause valid configs to annoy VS Code users.

Related: https://github.com/microsoft/pylance-release/issues/250

jakebailey:
What is the motivation behind the PR? Are you seeing your log level setting being disrespected where you give it something that follows the schema, but doesn't work at runtime?

jakebailey (changes requested):
Marking as do-not-merge to address the above.

rchl:
I've only noticed this problem because settings validation reported an issue when trying to set the value to the supposed default value `Info`.

I see now that the values provided by the user are converted here from the "long" to "short" variants:

https://github.com/microsoft/pyright/blob/f9a43214034936bdce64dadd347f751d8faac131/packages/pyright-internal/src/common/console.ts#L13-L18

where "short" ones are defined in:

https://github.com/microsoft/pyright/blob/f9a43214034936bdce64dadd347f751d8faac131/packages/pyright-internal/src/languageServerBase.ts#L1137-L1158

So yes, the code will work either way.

I've converted this PR to just fix the default value in the schema instead.

Feel free to update the PR's summary when merging.

erictraut:
Yeah, the default value in the schema should match the other values in the schema. I think this was an oversight on my part. Thanks for the fix.

PR #1524, opened 2021-02-23
Added support for ".pth" files (often used for editable installs) whe…

erictraut:
…n using the "venv" configuration option in pythonconfig.json.


erictraut:
This implements the protocol documented here: https://docs.python.org/3/library/site.html

Issue #1395, opened 2021-01-20
Separating JSON output to break up multiline messages into objects


Dr-Irv:
**Is your feature request related to a problem? Please describe.**
The JSON output from `--verifytypes` is just a list of messages.  A multiline message is one big string, which is hard to analyze.

**Describe the solution you'd like**
A more object-oriented format.

**Additional context**
Example.  If you run it on pandas (by creating a `py.typed` file and renaming `tests`), you get an item like this:
```json
 {
            "file": "",
            "severity": "error",
            "message": "Type partially unknown for class \"pandas.compat.numpy.function.CompatValidator\"\n  Type partially unknown for class \"CompatValidator\"\n    Type partially unknown for method \"__init__\"\n      Parameter \"defaults\" is missing a type annotation\n      Parameter \"fname\" is missing a type annotation\n      Parameter \"max_fname_arg_count\" is missing a type annotation\n    Type partially unknown for method \"__call__\"\n      Parameter \"args\" is missing a type annotation\n      Parameter \"kwargs\" is missing a type annotation\n      Parameter \"fname\" is missing a type annotation\n      Parameter \"max_fname_arg_count\" is missing a type annotation\n    Type unknown for variable \"fname\"\n    Type unknown for variable \"method\"\n    Type unknown for variable \"defaults\"\n    Type unknown for variable \"max_fname_arg_count\""
        },
```
Would be nicer if it was something like:
```json
{
error_type: "Type partially unknown",
class: "pandas.compat.numpy.function.CompatValidator",
alias: "CompatValidator",
methods: [
   {
   name: "__init__",
   parameters: [
       {
       name: "defaults",
       error:  "Missing type annotation"
       },
      {
      name:  "fname",
      error: "Missing type annotation"
       },
      {
      name: "max_fname_arg_count",
      error:  "Missing type annotation"
      }
  ],
  },
  {
  name: "__call__",
  parameters: [
     {
     name: "args",
     error:  "Missing type annotation"
     },
// etcera
```

Benefit would be that you could then analyze the JSON object and look for errors that are common across different classes and methods, meaning that if you fix one, you might get rid of a lot of problems.

erictraut:
I've started a new discussion about a rewrite of the verifytypes functionality here: https://github.com/microsoft/pyright/discussions/1568. I'd appreciate any thoughts you have.

erictraut:
Thanks for your patience on this one. It took me longer than expected and required a complete rewrite of the type verifier.

[...]

PR #1566, opened 2021-03-03
Show file path in verifytypes output

vemel:
## Notes

Issue: somehow related to https://github.com/microsoft/pyright/issues/1395

File paths are relative to be consistent with other reports.

## Changes

Enhancement: `pyright --verifytypes` output shows file path
Enhancement: `pyright --verifytypes` output shows node range
Bugfix: `pyright --verifytypes` JSON output shows correct `file` path
Bugfix: `pyright --verifytypes` JSON output shows correct node `range` in diagnostics

### DIFF
--- a/packages/pyright-internal/src/analyzer/packageTypeVerifier.ts
+++ b/packages/pyright-internal/src/analyzer/packageTypeVerifier.ts
@@ -10,6 +10,7 @@
 
 import { ConfigOptions, ExecutionEnvironment } from '../common/configOptions';
 import { Diagnostic, DiagnosticAddendum, DiagnosticCategory } from '../common/diagnostic';
+import { FileDiagnostics } from '../common/diagnosticSink';
 import { FileSystem } from '../common/fileSystem';
 import { combinePaths, getDirectoryPath, getFileExtension, stripFileExtension } from '../common/pathUtils';
 import { getEmptyRange } from '../common/textRange';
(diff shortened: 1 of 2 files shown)


erictraut:
I'm planning on a general rewrite of the verifytypes functionality soon. I'd prefer not to make this change now because it modifies the output schema, and other users may be relying on the current format. I'd like to make a single large breaking change rather than multiple small breaking change.

I appreciate your contributions, but for efficiency reasons, it would work better if we discussed behavior changes before you submit a PR.

vemel:
Changes to the schema should be fully compatible to existing schema:

- `file` field still remains a string, but now it is not empty if diagnostics message comes from a module
- `range` field has been added and it is consistent with messages generated by `pyright` main mode

Probably, we can keep text output as it was and make changes only to JSON output. I noticed that both text and JSON outputs already had support for `range`.

erictraut:
Ah, I see. You're merging the file-level diagnostics into the common diagnostics at the end. Makes sense. OK, I'll accept this PR. But as I mentioned, this is all likely to get replaced soon.

vemel:
Thank you a lot! I did not split messages to array as the original issue suggested, but I think more JSON-friendly was can be added lated alongside with current `message`.

erictraut:
I've started a new discussion about a rewrite of the verifytypes functionality here: https://github.com/microsoft/pyright/discussions/1568. I'd appreciate any thoughts you have.

PR #1590, opened 2021-03-07
Added file system caching to virtual file system interface and made use of it within import resolver

erictraut:
This change significantly improves import resolution performance for large projects, especially those that use namespace modules internally. Namespace modules require full searches of all import resolution paths.

My team's code base consists of ~250K lines of code in over 700 files. The "Resolve Imports" time for analysis consumed about 2 sec several months ago, but it has ballooned to over 6 sec as we added bug fixes for namespace import resolutions.

This change gains back all of the perf loss and then some, resulting in a 30% overall perf improvement for analysis times.

Prior to change:
----------------
    Searching for source files
    Found 712 source files
    0 errors, 0 warnings, 0 infos 
    Completed in 20.721sec

    Analysis stats
    Total files analyzed: 1426

    Timing stats
    Find Source Files:    0.17sec
    Read Source Files:    0.56sec
    Tokenize:             0.48sec
    Parse:                0.94sec
    Resolve Imports:      6.25sec
    Bind:                 1sec
    Check:                10.04sec
    Detect Cycles:        0.65sec

After change:
-------------
    Searching for source files
    Found 712 source files
    0 errors, 0 warnings, 0 infos 
    Completed in 15.645sec

    Analysis stats
    Total files analyzed: 1426

    Timing stats
    Find Source Files:    0.17sec
    Read Source Files:    0.64sec
    Tokenize:             0.47sec
    Parse:                0.95sec
    Resolve Imports:      1.28sec
    Bind:                 0.93sec
    Check:                9.85sec
    Detect Cycles:        0.65sec

### DIFF
--- a/packages/pyright-internal/src/analyzer/importResolver.ts
+++ b/packages/pyright-internal/src/analyzer/importResolver.ts
@@ -20,11 +20,9 @@ import {
     getDirectoryPath,
     getFileExtension,
     getFileName,
-    getFileSystemEntries,
+    getFileSystemEntriesFromDirEntries,
     getPathComponents,
     getRelativePathComponentsFromDirectory,
-    isDirectory,
-    isFile,
     resolvePaths,
     stripFileExtension,
     stripTrailingDirectorySeparator,
@@ -87,6 +85,7 @@ export class ImportResolver {
         if (this.fileSystem instanceof PyrightFileSystem) {
             this.fileSystem.clearPartialStubs();
         }
+        this.fileSystem.invalidateCache();
     }
 
     // Resolves the import and returns the path if it exists, otherwise
@@ -250,7 +249,7 @@ export class ImportResolver {
         if (sourceFilePaths.length === 0) {
             // Simple case where the stub and source files are next to each other.
             const sourceFilePath = changeAnyExtension(stubFilePath, '.py');
-            if (this.fileSystem.existsSync(sourceFilePath)) {
+            if (this.fileSystem.dirExistsSync(sourceFilePath, /* canCache */ true)) {
                 sourceFilePaths.push(sourceFilePath);
             }
         }
@@ -292,7 +291,7 @@ export class ImportResolver {
                 for (const importRootPath of importRootPaths) {
                     const absoluteStubPath = resolvePaths(importRootPath, relativeStubPath);
                     let absoluteSourcePath = changeAnyExtension(absoluteStubPath, '.py');
-                    if (this.fileSystem.existsSync(absoluteSourcePath)) {
+                    if (this.fileSystem.fileExistsSync(absoluteSourcePath, /* canCache */ true)) {
                         sourceFilePaths.push(absoluteSourcePath);
                     } else {
                         const filePathWithoutExtension = stripFileExtension(absoluteSourcePath);
@@ -302,14 +301,14 @@ export class ImportResolver {
                             // Try equivalent: <root>/package.py
                             absoluteSourcePath =
                                 filePathWithoutExtension.substr(0, filePathWithoutExtension.length - 9) + '.py';
-                            if (this.fileSystem.existsSync(absoluteSourcePath)) {
+                            if (this.fileSystem.fileExistsSync(absoluteSourcePath, /* canCache */ true)) {
                                 sourceFilePaths.push(absoluteSourcePath);
                             }
                         } else {
                             // Did not match: <root>/package.py
                             // Try equivalent: <root>/package/__init__.py
                             absoluteSourcePath = combinePaths(filePathWithoutExtension, '__init__.py');
-                            if (this.fileSystem.existsSync(absoluteSourcePath)) {
+                            if (this.fileSystem.fileExistsSync(absoluteSourcePath, /* canCache */ true)) {
                                 sourceFilePaths.push(absoluteSourcePath);
                             }
                         }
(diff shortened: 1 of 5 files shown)


jakebailey:
I'm thinking the file watching callbacks need to call the FS invalidate method and not the import resolver; conceptually, file watchers are a part of the FS and the import resolver is not. The import resolver may clear its caches for reasons other than file changes (e.g. settings changes), but it is true that the import resolver is eventually called from file watching callbacks so this PR ends up working.

Unfortunately, the file watching provider stuff makes it really difficult to do so; we have it solely so we can use different underlying providers for different paths (and to ignore paths that we know the client will notify us of). It'd have been nice if the FS were handling these so we could hook in, but feasibly we could manually add the call to the handful of change callbacks.

erictraut:
I don't think we can rely on the file watchers within the fs implementation. So while I agree with you in principle, I don't think that's a viable approach.

packages/pyright-internal/src/pyrightFileSystem.ts
```diff
@@ -44,6 +55,53 @@ export class PyrightFileSystem implements FileSystem {
         return this._realFS.existsSync(this._getPath(path));
     }
 
+    fileExistsSync(path: string, canCache = false): boolean {
+        const splitPath = this._splitPath(path);
+
+        if (!canCache || !splitPath[0] || !splitPath[1]) {
+            if (!this.existsSync(path, canCache)) {
```
heejaechang, line 62: since you have try/catch, probably this check is not needed?

packages/pyright-internal/src/pyrightFileSystem.ts
```diff
@@ -44,6 +55,53 @@ export class PyrightFileSystem implements FileSystem {
...
+            try {
+                const stats = this.statSync(path);
+                return stats.isFile();
+            } catch {
+                return false;
+            }
+        }
+
+        const entries = this.readdirEntriesSync(splitPath[0], /* canCache */ true);
+        const entry = entries.find((entry) => entry.name === splitPath[1]);
```
heejaechang, line 74: can this be a problem if the directory has a lot of files?

packages/pyright-internal/src/pyrightFileSystem.ts
```diff
@@ -44,6 +55,53 @@ export class PyrightFileSystem implements FileSystem {
...
+
+        if (this._partialStubPackagePaths.has(path)) {
+            // Pretend partial stub package directory doesn't exist. To be 100% correct,
+            // we need to check whether a file is under partial stub package path,
+            // but for now, this is enough to make import resolver to skip this folder.
+            return false;
+        }
+
+        const entries = this.readdirEntriesSync(splitPath[0], /* canCache */ true);
+        const entry = entries.find((entry) => entry.name === splitPath[1]);
```
heejaechang, line 101: same question here.

packages/pyright-internal/src/pyrightFileSystem.ts
erictraut: Possibly. That's one of the tradeoffs this change is trying to balance.

We're using the caching functionality primarily in contexts where we're already calling readdirEntriesSync within the import resolver (because we're scanning for native libraries whose names can't be generated statically), so I'm not that worried about it.

jakebailey:
I guess my feeling is that we only have two file watcher providers, where each can somewhat easily call the invalidateCache function, then we can later improve this to shift code around. The chokidar code is a simple wrapper, and the LS-specific one is all just classifying paths as either in-workspace or out-of-workspace; both do no fancy processing for their callbacks.

We can certainly try this out for now, though. I was worried about the perf of doing this, but it is just caching of dirents (versus contents), so should be cheap. Have you tested this with the BG thread enabled to make sure nothing wonky happens with file watching?

packages/pyright-internal/src/analyzer/importResolver.ts
```diff
@@ -87,6 +85,7 @@ export class ImportResolver {
         if (this.fileSystem instanceof PyrightFileSystem) {
             this.fileSystem.clearPartialStubs();
         }
+        this.fileSystem.invalidateCache();
```
heejaechang, line 88: I assume any change from file watcher (inside of lib folder + workspace root + extra path) will cause invalidate cache to be called?

heejaechang: assuming it will get invalidated when the file watcher said something is changed, when would we ever want cache == false?

erictraut: The caller is responsible for calling invalidate. In this case, the import resolver is the caller, and it has enough knowledge to invalidate at appropriate times. Most other consumers of the vfs interface do not. Plus, there are cases where it's not safe to use caching, like in our file-based cancellation.

erictraut: I initially thought about doing the caching entirely within the import resolver, but there were a few related modules (like pyTypedUtils) that also needed to participate. Maybe it would still be better to do the caching in the import resolver. That would address most of the concerns you and Jake are making.

jakebailey: Now that you describe this as "the caller calls invalidate", I understand your implementation a lot better. A user of the FS can audit all of its calls, then call invalidate when it thinks something's wrong. The caches may be getting invalidated more often than needed if everyone does it, but it at least leads to a slow move of things to be cache-aware and then we can make some other change to improve it.

jakebailey: But I do see the inconsistent cache system as a potential bump in the road, yeah...

erictraut: I think I'll code a second version that includes all the caching in the import resolver. We can then compare/contrast the two approaches. I think you're both making good points, and the alternative approach would address all of them.

jakebailey: Another motivator for making the import resolver cache something is that I think the import resolver is the only code that has to rescan the same things over and over and over again; the rest of the code either reads files of known paths, or is the indexer, which does a one-pass over a directory recursively until it's requested to do it again. The fact that we got this giant perf boost just by opting only the import resolver in at least matches that model.

(But, having the import resolver do a bunch of work may not be super fun either; we'll see, I guess...)

heejaechang: I did a similar thing in indexer (https://github.com/microsoft/pyrx/blob/main/packages/pylance-internal/src/packageScanner.ts#L312)

I got a big perf win from it, so I actually tried it on import resolver before (got similar perf win) but dropped it since I wasn't sure when to invalidate the cache (indexer was easy since it runs once and thrown away)

heejaechang: did a similar thing in indexer (packageScanner.ts#L312)

I got a big perf win from it, so I actually tried it on import resolver before (got similar perf win) but dropped it since I wasn't sure when to invalidate the cache (indexer was easy since it runs once and thrown away)

PR #1599, opened 2021-03-08
Added file system caching to import resolver for performance reasons.

erictraut:


erictraut:
I confirmed this gives us the same perf benefit. I think this is cleaner than the previous one. Definitely simpler and more contained.

packages/pyright-internal/src/analyzer/importResolver.ts
```diff
@@ -8,6 +8,8 @@
  * runtime rules of Python.
  */
 
+import * as fs from 'fs';
```
jakebailey, line 11: This is a nit, but just to keep it clear that this file doesn't actually access the filesystem directly, can you do a `import type { Dirent } from 'fs'` or similar?

jakebailey: `fs` is also a variable used in a few places, so this may get shadowed.

heejaechang (approved):

packages/pyright-internal/src/pyrightFileSystem.ts
```diff
@@ -70,7 +74,11 @@ export class PyrightFileSystem implements FileSystem {
     }
 
     readdirSync(path: string): string[] {
-        const entries = this._realFS.readdirSync(path);
+        const entries = this._realFS.readdirSync(path).filter((item) => {
```
heejaechang, line 77: thank you!

PR #1607, opened 2021-03-09
Added support for symbolic links in import resolver

erictraut:
Added support for symbolic links in import resolver both for resolution of ".pth" files and for imports themselves.

### DIFF
--- a/packages/pyright-internal/src/analyzer/importResolver.ts
+++ b/packages/pyright-internal/src/analyzer/importResolver.ts
@@ -26,6 +26,8 @@ import {
     getFileSystemEntriesFromDirEntries,
     getPathComponents,
     getRelativePathComponentsFromDirectory,
+    isDirectory,
+    isFile,
     resolvePaths,
     stripFileExtension,
     stripTrailingDirectorySeparator,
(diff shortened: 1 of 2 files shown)


packages/pyright-internal/src/analyzer/importResolver.ts
```diff
@@ -502,7 +504,18 @@ export class ImportResolver {
 
         const entries = this.readdirEntriesCached(splitPath[0]);
         const entry = entries.find((entry) => entry.name === splitPath[1]);
-        return entry !== undefined && entry.isFile();
+        if (entry?.isFile()) {
+            return true;
+        }
+
+        if (entry?.isSymbolicLink()) {
+            const realPath = this.fileSystem.realpathSync(path);
```
jakebailey, line 512: I am sort of wondering if realpath+exists+stat is slower than just doing a single stat to get the missing info, but I suppose isFile does the throw handling and I doubt this is a bottleneck.

jakebailey (approved):

packages/pyright-internal/src/analyzer/importResolver.ts
erictraut: I don't think this is on any perf-critical paths, especially with all of the caching we have in place now.

PR #1614, opened 2021-03-10
Fix more symlink cases, add tryStat helper to clean up many try/catch stat calls

jakebailey:
There were still a few places in the code that didn't handle symlinks (via other readdirEntries variants), clean those up.

Since the "try/catch stat and do something" is done so often, add a helper for it and replace all of the easy cases with it.

### DIFF
--- a/packages/pyright-internal/src/analyzer/packageTypeVerifier.ts
+++ b/packages/pyright-internal/src/analyzer/packageTypeVerifier.ts
@@ -12,7 +12,7 @@ import { ConfigOptions, ExecutionEnvironment } from '../common/configOptions';
 import { Diagnostic, DiagnosticAddendum, DiagnosticCategory } from '../common/diagnostic';
 import { FileDiagnostics } from '../common/diagnosticSink';
 import { FileSystem } from '../common/fileSystem';
-import { combinePaths, getDirectoryPath, getFileExtension, stripFileExtension } from '../common/pathUtils';
+import { combinePaths, getDirectoryPath, getFileExtension, stripFileExtension, tryStat } from '../common/pathUtils';
 import { getEmptyRange } from '../common/textRange';
 import { DeclarationType, FunctionDeclaration, VariableDeclaration } from './declaration';
 import { ImportedModuleDescriptor, ImportResolver } from './importResolver';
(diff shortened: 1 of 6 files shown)


packages/pyright-internal/src/analyzer/packageTypeVerifier.ts
```diff
@@ -430,7 +430,17 @@ export class PackageTypeVerifier {
         const dirEntries = this._fileSystem.readdirEntriesSync(dirPath);
 
         dirEntries.forEach((entry) => {
-            if (entry.isFile()) {
+            let isFile = entry.isFile();
```
erictraut, line 433: We could just call pathUtils.getFileSystemEntries here rather than duplicating this logic.

packages/pyright-internal/src/pyrightFileSystem.ts
```diff
@@ -165,11 +166,15 @@ export class PyrightFileSystem implements FileSystem {
             }
 
             for (const entry of this._realFS.readdirEntriesSync(path)) {
-                if (!entry.isDirectory() || !entry.name.endsWith(stubsSuffix)) {
+                const partialStubPackagePath = combinePaths(path, entry.name);
```
erictraut, line 169: Here again, we could use pathUtils.getFileSystemEntries to simplify this.

packages/pyright-internal/src/pyrightFileSystem.ts
```diff
@@ -249,13 +254,24 @@ export class PyrightFileSystem implements FileSystem {
         const partialStubPathLength = ensureTrailingDirectorySeparator(path).length;
         const searchAllStubs = (path: string) => {
             for (const entry of this._realFS.readdirEntriesSync(path)) {
-                if (entry.isDirectory()) {
-                    searchAllStubs(combinePaths(path, entry.name));
+                const filePath = combinePaths(path, entry.name);
```
erictraut, line 257: Here again, we could simplify this with pathUtils.getFileSystemEntries.

erictraut (commented):
Overall, looks good. There are some places where we could further simplify if you'd like.

jakebailey:
Yep, I plan to go back to clean this up further; these were the minimal obvious pre-release changes. 🙂

PR #1627, opened 2021-03-12
Remove global search path cache

jakebailey:
Calls to this function only happen either on settings update (we do a bunch of work scanning files for other reasons that are no slower than calling Python), or in the import resolver (which has a cache for this call explicitly). Remove this cache so that editable installs modified while the LS is running are properly handled.

Last thing to finish https://github.com/microsoft/pylance-release/issues/78.

### DIFF
--- a/packages/pyright-internal/src/analyzer/pythonPathUtils.ts
+++ b/packages/pyright-internal/src/analyzer/pythonPathUtils.ts
@@ -29,8 +29,6 @@ interface PythonPathResult {
     prefix: string;
 }
 
-const cachedSearchPaths = new Map<string, PythonPathResult>();
-
 const extractSys = [
     'import os, os.path, sys',
     'normalize = lambda p: os.path.normcase(os.path.normpath(p))',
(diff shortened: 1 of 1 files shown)


packages/pyright-internal/src/analyzer/pythonPathUtils.ts
```diff
@@ -29,8 +29,6 @@ interface PythonPathResult {
     prefix: string;
 }
 
-const cachedSearchPaths = new Map<string, PythonPathResult>();
```
erictraut, line 32: I think we're caching enough in the import resolver that the removal of this cache won't matter. Out of curiosity, did you do any benchmarking to confirm this?

jakebailey: I haven't; I can try and clone your team's repo to test.

Note that the import resolver is already caching this call explicitly (was before per venv though that code got modified), so I wouldn't expect it to have changed.

jakebailey: I can't find your repo anymore, but on typeshed there is no difference in the numbers I get when using `--stats`.

erictraut: I'm not that concerned about it. I can test it after you land this change.

erictraut (approved):

PR #1632, opened 2021-03-15
Changed the way extraPaths works to avoid confusion

erictraut:
Added new top-level "extraPaths" config option for pythonconfig.json that specifies the default extraPaths to use when no execution environments apply to a file. Changed settings logic to use the new default extraPaths mechanism for the "python.analysis.extraPaths" setting.


jakebailey (approved):

jakebailey:
Vendored into pylance to test; works as expected in my test project.

PR #1636, opened 2021-03-16
Improved completion provider's handling of method overrides so it pro…

erictraut:
…perly handles properties.

### DIFF
--- a/packages/pyright-internal/src/languageService/completionProvider.ts
+++ b/packages/pyright-internal/src/languageService/completionProvider.ts
@@ -765,7 +765,7 @@ export class CompletionProvider {
                 const symbolTable = new Map<string, Symbol>();
                 for (const mroClass of classResults.classType.details.mro) {
                     if (isClass(mroClass)) {
-                        getMembersForClass(mroClass, symbolTable, false);
+                        getMembersForClass(mroClass, symbolTable, /* includeInstanceVars */ false);
                     }
                 }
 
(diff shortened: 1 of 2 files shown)


packages/pyright-internal/src/languageService/completionProvider.ts
```diff
@@ -926,15 +959,21 @@ export class CompletionProvider {
...
-        const parameters = getParameters(declaredType, decl);
+        const parameters = getParameters();
         if (decl.node.name.value !== '__init__') {
             sb += 'return ';
         }
 
+        if (isProperty) {
+            // If it's a property, it's not clear what to provide in the
+            // implementation. Assume that the user will fill in the expression.
+            return sb;
```
jakebailey, line 970: It'd actually be `return super().prop`, if we did want to generate something.

jakebailey: Or did you see an ambiguity in this case?

erictraut: Yeah, that's reasonable.

jakebailey:
This seems good as is for getters. I suppose we don't have anything for setters or anything, given to do overrides of that you have to do:

```python
class A:
    @property
    def foo(self): return ...

    @foo.setter
    def foo(self, x): ...


class B(A):
    @A.foo.setter
    def foo(self, x): ...
```

So hopefully we offer the completion `A.foo.setter` alright.

jakebailey (approved):

PR #1653, opened 2021-03-18
Avoid recursing infinitely in _matchFiles

jakebailey:
For https://github.com/microsoft/pylance-release/issues/1070.

This `_matchFiles` function recurses on `getFileSystemEntries`. When we changed this to follow symlinks (plus the bugfix to actually handle directories), this function now has the potential to keep going searching for source files to match forever.

Before recursing, check the realpath of the directory and skip if we're already visiting it.

As far as I'm aware, this is the only use of `getFileSystemEntries` which actually walks down the results recursively; the uses in the import resolver only ever go as far as any one import requires (which is bounded). `_matchFiles` is what lists out the files to check in workspace mode, so is not bounded.

While here, change a `try stat catch` block to a `tryStat`; the catch is likely to mask actual exceptions and this was the only place I didn't change to `tryStat`.

In the future, it'd be good to add a specific walker which does this work for us, but this is the only place in pyright that we do this. (Pylance's indexer may need this as well.)

### DIFF
--- a/packages/pyright-internal/src/analyzer/service.ts
+++ b/packages/pyright-internal/src/analyzer/service.ts
@@ -977,7 +977,7 @@ export class AnalyzerService {
         const envMarkers = [['bin', 'activate'], ['Scripts', 'activate'], ['pyvenv.cfg']];
         const results: string[] = [];
 
-        const visitDirectory = (absolutePath: string, includeRegExp: RegExp) => {
+        const visitDirectoryUnchecked = (absolutePath: string, includeRegExp: RegExp) => {
             if (this._configOptions.autoExcludeVenv) {
                 if (envMarkers.some((f) => this._fs.existsSync(combinePaths(absolutePath, ...f)))) {
                     this._console.info(`Auto-excluding ${absolutePath}`);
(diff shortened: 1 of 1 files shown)


packages/pyright-internal/src/analyzer/service.ts
```diff
@@ -1007,23 +1007,35 @@ export class AnalyzerService {
             }
         };
 
+        const seenDirs = new Set<string>();
+        const visitDirectory = (absolutePath: string, includeRegExp: RegExp) => {
+            const realDirPath = this._fs.realpathSync(absolutePath);
+            if (seenDirs.has(realDirPath)) {
+                this._console.warn(`Skipping recursive symlink "${absolutePath}" -> "${realDirPath}"`);
```
jakebailey, line 1014: I don't have to keep this if we want to just silently ignore these problems. Won't show by default anyway, more useful for debugging.

erictraut (approved):
Good find. I was watching for situations like this when we added the symlink code, but I missed this one. I was thinking primarily about the import resolver, which naturally limits walking to the length of the import path.

PR #1661, opened 2021-03-20
Allow printing out "absolute" type names (with dotted module paths)

necaris:
See https://github.com/microsoft/pyright/discussions/1342#discussioncomment-502941 for more context. This enables me to see diagnostic output like:

``` sh
error: Expression of type "tornado.concurrent.Future[None]" cannot be assigned to declared type "asyncio.futures.Future[None]"
```
rather than the less helpful previous output:

``` sh
error: Expression of type "Future[None]" cannot be assigned to declared type "Future[None]"
```
Not sure where at this point to add tests or to add the parsing of the flag (as a command-line or `pyrightconfig.json` option?) but all feedback very much appreciated!


erictraut:
Thanks for the PR, but I don't think this is a good change. Most users do not want verbose type output, and you've enabled this feature by default for all strict-mode types. This also breaks a bunch of tests.

I am not convinced this feature is needed generally. You're welcome to fork the repo and make the change locally if you have a strong need for it.

necaris:
Would it work better if it were gated behind a configuration flag? I doubt I'm alone in wanting more than the elided values, especially when debugging adding types to an existing codebase.

erictraut:
We generally don't add new configuration flags until we get some signal that the capability is desired by many users. My intuition is that your needs are pretty unique here, but I remain open minded and will reconsider if other users speak up and request this capability.

PR #1662, opened 2021-03-20
Use `asyncio.Future` for `tornado.concurrent.Future`

necaris:
As of
https://github.com/tornadoweb/tornado/commit/354d8c0ddac3809b2ee5528c5e58285f1d90053b
in v6.1, no other `Future` implementation has been available, and as of
https://github.com/tornadoweb/tornado/commit/ac13ee5f64cd3bceb1628dbcbbfc77445358038f
`asyncio.Future` has been preferred. Updating the type stubs to match.


erictraut:
Thanks for the contribution, but this isn't the right way to fix this. Pyright pulls its stubs from the typeshed repo. If you believe this stub is in error, you should make the change there, and pyright will incorporate it the next time we merge changes from typeshed.

necaris:
Ah, gotcha, I'll point this upstream at type shed then. Thanks!

PR #1698, opened 2021-03-29
Make calls to realpath safe, like stat

jakebailey:
For https://github.com/microsoft/pylance-release/issues/1102.

`realpath` can fail similarly to `stat` (because it's largely a stat wrapper). Wrap it like we do `stat` to ensure we don't fail to `realpath` a broken link.


erictraut (approved):
LGTM

PR #1699, opened 2021-03-30
fix typo in settings documentation.

spazm:


erictraut:
Thanks for the contribution. I didn't realize that "langue" was a word, but my spell checker seems to think it is. :)

PR #1726, opened 2021-04-04
Changed the logic that searches for a config file

erictraut:
It currently searches from the current working directory all the way up the folder hierarchy. This makes sense only for a command-line tool, not for a language server. The latter already knows the project root, and we should look only in that directory for a config file.


erictraut:
This PR is related to this issue: https://github.com/microsoft/pylance-release/issues/1079#issuecomment-813090062

jakebailey (approved):

PR #1732, opened 2021-04-05
Fixed bug in signature help provider where its heuristics were causin…

erictraut:
…g it to return a bad response when the insertion point was immediately after a comma and a call expression preceded the comma.


jakebailey (approved):
LGTM and passes the existing tests, though a fourslash test to cover this would be welcome.

PR #1735, opened 2021-04-06
Added support for an import edge case where a module's `__init__.py` …

erictraut:
…file is apparently importing from itself but intends instead to import from one of its submodules.


jakebailey (approved):
Seems to work; I can only get this to trigger for `apache_beam` and `sqlparse` (though the latter case's code doesn't really feel like it should be needed...).

PR #1749, opened 2021-04-09
Update command line directions with new output format

steverice:
Since https://github.com/microsoft/pyright/commit/ba901d1449fceb0642934bfd1993e79ed0fb136d#diff-c3f0fa903325059f6e7dda1116458aafa72061ac3d1a51ddfd077a5dc0801030R394, the `--outputjson` result schema has changed.


erictraut:
Thanks for the update. I missed that change when I modified the format last month.

PR #1773, opened 2021-04-19
Implemented first cut at generalized support for dataclass transforms.

erictraut:


specs/dataclass_transforms.txt
```diff
@@ -0,0 +1,365 @@
+Motivation
```
gvanrossum, line 1: Interesting. Is this meant to become a PEP?

specs/dataclass_transforms.txt
```diff
@@ -0,0 +1,399 @@
...
```
jakebailey, line 1: This file appears to be markdown; can you make in an `md` so github formats it nicely?

erictraut: I've sent a file link to several people already. If I rename the file, those links will be invalidated.

specs/dataclass_transforms.txt
```diff
@@ -0,0 +1,399 @@
...
+        init: Optional[bool] = True,
+        **kwargs: Any
+    ) -> None: ...
+
+@typing.dataclass_transform(kw_only_default=True, field_descriptors=(ModelField, ))
+def create_model(cls: type) -> Callable[[_T], _T]: ...
+
+
+# Code that imports this library:
+@create_model(init=False)
```
heejaechang, line 261: so init:False has no meaning in dataclass_transform (init seems not part of dataclass_trasform's known parameters)? and ModelField has init == true as default, and init in field descriptor seems part of known parameters, so will it affect how type checker treat the type?

erictraut: This parameter instructs the transform function `create_model` to not synthesize an `__init__` method for the decorated class. The stdlib `dataclass` decorator supports the same semantics.

```python
@dataclass(init=False)
class Foo:
    a: int

Foo(a=3) # Error because Foo has no __init__ method
```

When the `init` parameter is used in a field descriptor class like `ModelField`, it has a slightly different meaning. In particular, it means that this field should not be included in the synthesized `__init__` method.

specs/dataclass_transforms.txt
```diff
@@ -0,0 +1,399 @@
...
+```python
+# Library code (within type stub or inline):
+class ModelField:
+    def __init__(
+        self,
+        *,
+        default: Optional[Any] = ...,
+        init: Optional[bool] = True,
+        **kwargs: Any
+    ) -> None: ...
```
heejaechang, line 254: > Libraries that support dataclass-like semantics and support field descriptor
classes typically use common parameter names to construct these field
descriptors

so, do that mean libraries that use names not defined in here for field descriptor like types, they need to make breaking changes to opt in? or can they provide decorator to map names?

specs/dataclass_transforms.txt
```diff
@@ -0,0 +1,399 @@
...
+
+Here is its complete implementation.
+
+```python
+def dataclass_transform(
+    *,
+    eq_default: bool = True,
+    order_default: bool = False,
+    kw_only_default: bool = False,
+    field_descriptors: Tuple[Union[type, Callable[..., Any]], ...] = (()),
```
heejaechang, line 283: can this decorator cause perf impact on runtime? like allocations and etc?

jakebailey: I can't imagine it would; if you look at the example code for "early access", it's just a function that returns what it's given, meaning the decorator would be a no-op for the function it's wrapping.

erictraut: No, it should not have any measurable impact. If you're calling this in a tight loop, you're doing something very wrong.

specs/dataclass_transforms.txt
erictraut: Libraries would need to standardize on these names, which is good for users because the same concepts will have the same names across libraries. This sort of standardization appears to have already happened for the most popular libraries. If a new library is developed that has dataclass-like semantics, they should be highly encouraged to adopt the existing names.

specs/dataclass_transforms.txt
```diff
@@ -0,0 +1,399 @@
...
+
+
+Dataclass Semantics
+-------------------
+
+The following dataclass semantics are implied when dataclass_transform is
+specified.
+
+Frozen classes cannot inherit from non-frozen classes. A class that
+directly specifies a metaclass that has been decorated with
```
heejaechang, line 296: frozen seems not part of known parameters for data_transform decorator?

erictraut: No, it's not part of the `data_transform` decorator. It's optionally supported by the decorator function that `data_transform` decorates. You need to think a metalevel deeper here.

specs/dataclass_transforms.txt
```diff
@@ -0,0 +1,399 @@
...
+`dataclass_transform` will not be considered non-frozen. In the example
+
+```python
+@typing.dataclass_transform()
+class ModelMeta(type): ...
+
+# ModelBase is not considered either "frozen" or "non-frozen"
+# because it directly specifies ModelMeta as its metaclass.
+class ModelBase(metaclass=ModelMeta): ...
```
heejaechang, line 305: so here type checker treats ModelMeta as dataclass ? and apply dataclass rules? and this, frozen, non-frozen behavior, is normal data class behavior?

specs/dataclass_transforms.txt
heejaechang: looks like this is by design.

specs/dataclass_transforms.txt
heejaechang: I see. so this just follow whatever dataclass existing rule.

specs/dataclass_transforms.txt
heejaechang: my confusion was why rule on frozen instance's inheritance rule was specified here. is using dataclass_tranform decorator changes behavior on this compared to regular dataclass case?

PR #1774, opened 2021-04-19
Optimize getUnescapedString and _skipToEndOfStringLiteral

jakebailey:
Through profiling, I found that these functions accounted for more than 75% of the memory allocations when running on a test application. Both build strings via `+=`. Rewrite them to push into an array instead and `join('')` at the end, which is much faster overall. Presumably v8 knows to preallocate the output to the correct size when it sees it's joining an array of strings.

This halves the time spent in GC from 11% to around 5% of total CPU time.

(There are further tweaks that could be made to reuse these arrays and allocate less, but this fixes the great bulk of the problem.)


erictraut (approved):
LGTM

PR #1776, opened 2021-04-19
Make fewer calls to python at config time, print actual import roots

jakebailey:
Through profiling, I found that we were calling python 4 times at startup. This was due to a couple of issues:

- We wanted to log the search paths (for helping debug users), but called out to python to get them directly (distinctly from the import resolver). Instead, don't do this, and print the actual set of import roots later (per execution environment), which is what we actually want to know.
- The file watcher setup functions were invalidating the import cache. I have no idea why; the setup functions are called directly after a new import resolver is constructed. The import resolver's caches are certainly empty.
- The ensure partial stub function was run on setConfig, but then all callers of setConfig immediately created a new import resolver, leading to an extra python call as the ensure function fetches search paths. (It's also dubious that the partial stubs are even ready in the old code, if the import resolver was created _after_ the ensure call; @heejaechang I'm curious if you think this is alright in the new code.)

Now, we only call twice, once to get the default interpreter's version, once to get the import roots. Those are in entirely different parts of the code, so is less simple to fix (but would be nice later).

### DIFF
--- a/packages/pyright-internal/src/analyzer/backgroundAnalysisProgram.ts
+++ b/packages/pyright-internal/src/analyzer/backgroundAnalysisProgram.ts
@@ -62,13 +62,13 @@ export class BackgroundAnalysisProgram {
         this._configOptions = configOptions;
         this._backgroundAnalysis?.setConfigOptions(configOptions);
         this._program.setConfigOptions(configOptions);
-
-        configOptions.getExecutionEnvironments().forEach((e) => this._ensurePartialStubPackages(e));
     }
 
     setImportResolver(importResolver: ImportResolver) {
         this._importResolver = importResolver;
         this._program.setImportResolver(importResolver);
+
+        this._configOptions.getExecutionEnvironments().forEach((e) => this._ensurePartialStubPackages(e));
 
         // Do nothing for background analysis.
         // Background analysis updates importer when configOptions is changed rather than
--- a/packages/pyright-internal/src/analyzer/backgroundAnalysisProgram.ts
+++ b/packages/pyright-internal/src/analyzer/backgroundAnalysisProgram.ts
@@ -212,13 +212,6 @@ export class BackgroundAnalysisProgram {
         this._program.markAllFilesDirty(true);
     }
 
-    invalidateCache() {
-        // Invalidate import resolver because it could have cached
-        // imports that are no longer valid because a source file has
-        // been deleted or added.
-        this._importResolver.invalidateCache();
-    }
-
     restart() {
         this._backgroundAnalysis?.restart();
     }
--- a/packages/pyright-internal/src/analyzer/service.ts
+++ b/packages/pyright-internal/src/analyzer/service.ts
@@ -59,7 +59,7 @@ import { AnalysisCompleteCallback } from './analysis';
 import { BackgroundAnalysisProgram, BackgroundAnalysisProgramFactory } from './backgroundAnalysisProgram';
 import { ImportedModuleDescriptor, ImportResolver, ImportResolverFactory } from './importResolver';
 import { MaxAnalysisTime } from './program';
-import { findPythonSearchPaths, getPythonPathFromPythonInterpreter } from './pythonPathUtils';
+import { findPythonSearchPaths } from './pythonPathUtils';
 import { TypeEvaluator } from './typeEvaluator';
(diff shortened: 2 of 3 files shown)


packages/pyright-internal/src/analyzer/service.ts
```diff
@@ -645,36 +645,6 @@ export class AnalyzerService {
                     }
                 }
             }
-        } else {
-            const importFailureInfo: string[] = [];
```
jakebailey, line 649: This block is only for logging, and has been moved down.

packages/pyright-internal/src/analyzer/service.ts
```diff
@@ -1069,8 +1039,6 @@ export class AnalyzerService {
     private _updateSourceFileWatchers() {
         this._removeSourceFileWatchers();
 
-        this._backgroundAnalysisProgram.invalidateCache();
```
jakebailey, line 1072: These invalidations don't do anything; this function is always run on a fresh import resolver.

packages/pyright-internal/src/analyzer/backgroundAnalysisProgram.ts
```diff
@@ -62,14 +62,14 @@ export class BackgroundAnalysisProgram {
...
         this._program.setConfigOptions(configOptions);
-
-        configOptions.getExecutionEnvironments().forEach((e) => this._ensurePartialStubPackages(e));
     }
 
     setImportResolver(importResolver: ImportResolver) {
         this._importResolver = importResolver;
         this._program.setImportResolver(importResolver);
 
+        this._configOptions.getExecutionEnvironments().forEach((e) => this._ensurePartialStubPackages(e));
```
heejaechang, line 71: seems fine to me. especially if it improve perf.

packages/pyright-internal/src/analyzer/service.ts
erictraut: IIRC, we previously cached the results of the call to Python, so this code path was effectively "free". But it appears that we no longer cache this value (at least not on this code path), so it results in a redundancy. I agree that it's fine to remove this.

packages/pyright-internal/src/analyzer/service.ts
erictraut: I'm fine with this change as long as HeeJae is. I'm pretty sure he added this originally, and I'm not sure what the reasoning was. I agree that there doesn't seem to be any reason why we'd need to explicitly invalidate the cache in this code path.

erictraut (approved):

packages/pyright-internal/src/analyzer/service.ts
```diff
@@ -1292,6 +1258,17 @@ export class AnalyzerService {
         const importResolver = this._importResolverFactory(this._fs, this._backgroundAnalysisProgram.configOptions);
         this._backgroundAnalysisProgram.setImportResolver(importResolver);
 
+        if (this._commandLineOptions?.fromVsCodeExtension || this._configOptions.verboseOutput) {
+            const logLevel = this._configOptions.verboseOutput ? LogLevel.Info : LogLevel.Log;
+            for (const execEnv of this._configOptions.getExecutionEnvironments()) {
+                log(this._console, logLevel, `Search paths for ${execEnv.root}`);
+                const roots = importResolver.getImportRoots(execEnv);
```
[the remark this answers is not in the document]
jakebailey: Here's what it looks like after:

```
Search paths for E:\work\demo-apps\ds-1
  E:\work\pyright\packages\pyright\dist\typeshed-fallback\stdlib
  E:\work\demo-apps\ds-1
  E:\work\demo-apps\ds-1\typings
  E:\work\pyright\packages\pyright\dist\typeshed-fallback\stubs\...
  C:\Users\jabaile\scoop\apps\python38\current\DLLs
  C:\Users\jabaile\scoop\apps\python38\current\lib
  C:\Users\jabaile\scoop\apps\python38\current
  E:\work\demo-apps\ds-1\.venv
  E:\work\demo-apps\ds-1\.venv\lib\site-packages
```

Perhaps `E:\work\pyright\packages\pyright\dist\typeshed-fallback\stubs\*` is more informative? This is for logging only, of course.

heejaechang (approved):

PR #1783, opened 2021-04-22
Use `null` instead of `-1` for `SignatureHelp.activeParameter`

pbzweihander:
LSP 3.16 (current) specification says that `SignatureHelp.activeParameter` should be `uinteger`. (https://microsoft.github.io/language-server-protocol/specifications/specification-3-16/#textDocument_signatureHelp) But `pyright-langserver` returns `-1` in certain situation. It makes Rust lsp-types break. (https://docs.rs/lsp-types/0.89.0/lsp_types/struct.SignatureHelp.html#structfield.active_parameter)

### DIFF
--- a/packages/pyright-internal/src/languageServerBase.ts
+++ b/packages/pyright-internal/src/languageServerBase.ts
@@ -694,11 +694,7 @@ export abstract class LanguageServerBase implements LanguageServerInterface {
             }
 
             if (this.client.hasActiveParameterCapability || activeSignature === null) {
-                // A value of -1 is out of bounds but is legal within the LSP (should be treated
-                // as undefined). It produces a better result in VS Code by preventing it from
-                // highlighting the first parameter when no parameter works, since the LSP client
-                // converts null into zero.
-                activeParameter = -1;
+                activeParameter = null;
             }
 
             return { signatures, activeSignature, activeParameter };


jakebailey (changes requested):
The comment I left there (that you're removing) still stands. I don't want to change this without verifying that VS Code (and maybe other editors that depend on this behavior) doesn't regress. The spec being changed to uinteger is a relatively new thing; it used to say "number" and the reference LSP client handles this in the way we want it to.

I'm surprised the client isn't being permissive. I've seen issues in the past where clients are too restrictive on their inputs (with regards to extra properties), and this seems similar (though more problematic).

jakebailey:
Yeah, I'm afraid just removing this hack causes a regression; in the below, we want none of them to be highlighted:

image

For your client, does it work if instead of `-1`, a large value like `9999` (or at least `max(parameters)+1` for all signatures) is used instead? This does appear to work for VS Code and is in the range of `uinteger` as defined in LSP.

Unfortunately, the spec instead says that clients should treat out-of-range values as `0` if there are parameters at all, which isn't what we want (since I want it to highlight nothing), not the zero-th parameter:

```ts
	/**
	 * The active parameter of the active signature. If omitted or the value
	 * lies outside the range of `signatures[activeSignature].parameters`
	 * defaults to 0 if the active signature has parameters. If
	 * the active signature has no parameters it is ignored.
	 * In future version of the protocol this property might become
	 * mandatory to better express the active parameter if the
	 * active signature does have any.
	 */
	activeParameter?: uinteger;
```

So I'll have to open an issue to clarify this with the LSP folks. But, if a positive number outside the range of the parameters works, then that could be used as the workaround instead.

erictraut:
@jakebaily, any update on this issue?

@pbzweihander, could you answer Jake's question above about using a large value like 9999?

jakebailey:
Agh, sorry, forgot about this over my pycon and Pylance GA prep.

@dbaeumer may be able to comment on finding a way in the LSP to handle this behavior, as we really do want the "nothing is highlighted" behavior, but the spec appears to imply that clients should treat all out of bounds indexes as zeros. VS Code doesn't behave that way to our benefit, and changing that will be a UI regression. We can likely move this line of thought to the LSP or client tracker.

I'd still like an answer from the PR author about the impact of using a large number as the out of bounds value; I don't really have the ability to check what they were using, but if it works, I can certainly see changing this to just pick a number that's above the range of signatures/parameters for the specific call.

dbaeumer:
@jakebailey actually setting this to `null` will not work due to this code: https://github.com/microsoft/vscode-languageserver-node/blob/main/client/src/common/protocolConverter.ts#L593

In addition it is not part of the specification.

I see the following way forward:

- allow `null` in the spec guarded by a capability and define it as none is highlighted.
- for VS Code only you could test what happens if you use a number > `signatures[activeSignature].parameters`. If would violate the spec but would for now give you the desired result I guess.

jakebailey:
I'll have to make an issue in the LSP tracker to request an explicit way to say "highlight nothing", then. I just don't want the VS Code LSP client to "fix" this behavior and leave us with no way to actually do what we want.

As far as I can tell from the wording in the spec, it's "legal" to have an out of bounds value, but the client should consider that to be equivalent to zero. The VS Code client just passes the value through, then VS Code does something different with it than treating it as zero, leading to the "good" behavior.

I could just go out of bounds the other way to fix clients which are being strict about `uinteger`; I tested it when the PR was sent by just hardcoding 9999, and it worked as expected. The question is whether or not other clients will break if the value is moved to out of bounds the other way around; presumably the existing clients that aren't the OPs haven't hit crashes like this when we were using `-1`.

jakebailey:
I just checked `gopls`, and they achieve this by using indexes past the end of the parameter list, so I think that might be fine to do for now. The only difference is that Python has overloads, so multiple signatures that can have their own active parameters, so we have to handle the `activeParameter` fallback and pick some value that's out of bounds for _all_ overloads (or maybe just always specify `parameters.length` for "no active" per-signature).

I'll try and send a PR to replace this one.

jakebailey:
I've sent #1876 to replace this PR; it should have the same effect and not break the behavior we want while we mull over how to really handle this in the LSP.

PR #1791, opened 2021-04-23
Changed binder logic for "from .a import b" statements in `__init__.p…

erictraut:
…y` files so implicit import of submodule "a" is performed only if it's not a multi-part module path.


jakebailey (approved):

jakebailey:
Can confirm that this works on the django stubs.

image

PR #1811, opened 2021-05-01
Add ``match`` and ``case`` keywords to completion provider

cdce8p:


erictraut:
Thanks for the contribution.

PR #1812, opened 2021-05-01
Improved performance of string token value unescape logic by handling…

erictraut:
… the common cases (no format string and no escape characters) using a fast path.

### DIFF
--- a/packages/pyright-internal/src/parser/stringTokenUtils.ts
+++ b/packages/pyright-internal/src/parser/stringTokenUtils.ts
@@ -83,8 +83,40 @@ function completeUnescapedString(incomplete: IncompleteUnescapedString): Unescap
 export function getUnescapedString(stringToken: StringToken): UnescapedString {
     const escapedString = stringToken.escapedValue;
     const isRaw = (stringToken.flags & StringTokenFlags.Raw) !== 0;
+
+    if (isRaw) {
+        return {
+            value: escapedString,
+            unescapeErrors: [],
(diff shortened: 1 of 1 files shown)


erictraut:
@jakebailey, let me know if you'd like me to merge this into the master branch or hold it on a side branch until after the next pylance release.

packages/pyright-internal/src/parser/stringTokenUtils.ts
```diff
@@ -83,8 +83,40 @@ function completeUnescapedString(incomplete: IncompleteUnescapedString): Unescap
...
+    if (isRaw) {
+        return {
+            value: escapedString,
+            unescapeErrors: [],
+            nonAsciiInBytes: false,
+            formatStringSegments: [],
+        };
+    }
+
+    const charCodes: number[] = [];
```
bschnurr, line 96: maybe there is a minor benefit to pre-allocate the array size. 

something like 
```
const charCodes: number[] = new Array(escapedString.length);
```

jakebailey: That won't do quite what you expect:

```
> const arr = new Array(5)
undefined
> arr.push("test")
6
> arr
[ <5 empty items>, 'test' ]
>
```

Array is weird and has "empty items" as a thing. Maybe preallocs are possible by doing the above and setting `arr.length = 0`, but that would require profiling to confirm.

jakebailey (approved):
LGTM, and seems safe to me.

PR #1816, opened 2021-05-02
Removed PEP 563 (deferred type annotation) behavior as default for Py…

erictraut:
…thon 3.10, reflecting a recent decision by the Python steering committee.

### DIFF
--- a/packages/pyright-internal/src/analyzer/typeEvaluator.ts
+++ b/packages/pyright-internal/src/analyzer/typeEvaluator.ts
@@ -1334,7 +1334,11 @@ export function createTypeEvaluator(
     }
 
     function isAnnotationEvaluationPostponed(fileInfo: AnalyzerFileInfo) {
-        return fileInfo.futureImports.get('annotations') !== undefined || fileInfo.isStubFile;
+        return (
+            fileInfo.futureImports.get('annotations') !== undefined ||
+            fileInfo.executionEnvironment.pythonVersion >= PythonVersion.V3_11 ||
+            fileInfo.isStubFile
+        );
     }
 
     function getTypeOfAnnotation(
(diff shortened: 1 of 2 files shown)


erictraut:
@jakebailey, let me know if you'd like me to merge this into the master branch or hold it on a side branch until after the next pylance release.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -1334,11 +1334,7 @@ export function createTypeEvaluator(
     }
 
     function isAnnotationEvaluationPostponed(fileInfo: AnalyzerFileInfo) {
-        return (
-            fileInfo.futureImports.get('annotations') !== undefined ||
-            fileInfo.executionEnvironment.pythonVersion >= PythonVersion.V3_10 ||
```
jakebailey, line 1339: Should this be bumped to 3.11 instead, in hopes it's there?

erictraut: Yeah, I'm fine with that.

jakebailey:
I'm honestly surprised this passes with only this change... Aren't there tests that specify 3.10 to verify this behavior?

erictraut:
There are tests for the `from __future__ import annotations`, which is the complicated part of PEP 563. There aren't any tests specifically for which version this becomes the default. This involves very simple logic in one line of code, and it's unlikely to regress.

Let me know if you want me to merge this change or hold it on a branch until after the code freeze is lifted.

jakebailey (approved):

jakebailey:
Probably want this so that anyone relying on this assuming 3.10 has it will realize it's not ready (given 3.10 is going into beta today?), so, LGTM.

PR #1819, opened 2021-05-03
Added support for typeshed stdlib VERSIONS file, which indicates the …

erictraut:
…minimum (and optionally the maximum) version supported by each module in stdlib.

### DIFF
--- a/packages/pyright-internal/src/analyzer/importResolver.ts
+++ b/packages/pyright-internal/src/analyzer/importResolver.ts
@@ -1357,7 +1357,7 @@ export class ImportResolver {
             this._cachedTypeshedStdLibModuleVersions = this._readTypeshedStdLibVersions(execEnv, importFailureInfo);
         }
 
-        const versionRange = this._cachedTypeshedStdLibModuleVersions!.get(moduleDescriptor.nameParts[0]);
+        const versionRange = this._cachedTypeshedStdLibModuleVersions.get(moduleDescriptor.nameParts[0]);
         if (versionRange) {
             if (execEnv.pythonVersion < versionRange.min) {
                 return false;
@@ -1387,7 +1387,8 @@ export class ImportResolver {
                 if (fileStats.size > 0 && fileStats.size < 256 * 1024) {
                     const fileContents = this.fileSystem.readFileSync(versionsFilePath, 'utf8');
                     fileContents.split(/\r?\n/).forEach((line) => {
-                        const colonSplit = line.split(':');
+                        const commentSplit = line.split('#');
+                        const colonSplit = commentSplit[0].split(':');
                         if (colonSplit.length !== 2) {
                             return;
                         }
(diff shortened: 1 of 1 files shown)


erictraut:
We cannot integrate this change until the following typeshed issue is resolved:
https://github.com/python/typeshed/issues/5328

packages/pyright-internal/src/analyzer/importResolver.ts
```diff
@@ -1310,14 +1324,112 @@ export class ImportResolver {
...
+    private _isStdlibTypeshedStubValidForVersion(
+        moduleDescriptor: ImportedModuleDescriptor,
+        execEnv: ExecutionEnvironment,
+        importFailureInfo: string[]
+    ) {
+        if (!this._cachedTypeshedStdLibModuleVersions) {
+            this._cachedTypeshedStdLibModuleVersions = this._readTypeshedStdLibVersions(execEnv, importFailureInfo);
+        }
+
+        const versionRange = this._cachedTypeshedStdLibModuleVersions!.get(moduleDescriptor.nameParts[0]);
```
srittau, line 1360: Nit: Is `!` necessary here? Typescript should have narrowed this due to the guard above.

packages/pyright-internal/src/analyzer/importResolver.ts
```diff
@@ -1310,14 +1324,112 @@ export class ImportResolver {
...
+        // Read the VERSIONS file from typeshed.
+        const typeshedStdLibPath = this._getTypeshedSubdirectory(/* isStdLib */ true, execEnv, importFailureInfo);
+
+        if (typeshedStdLibPath) {
+            const versionsFilePath = combinePaths(typeshedStdLibPath, 'VERSIONS');
+            try {
+                const fileStats = this.fileSystem.statSync(versionsFilePath);
+                if (fileStats.size > 0 && fileStats.size < 256 * 1024) {
+                    const fileContents = this.fileSystem.readFileSync(versionsFilePath, 'utf8');
+                    fileContents.split(/\r?\n/).forEach((line) => {
```
srittau, line 1389: Currently, the `VERSIONS` file doesn't support comments, and I think neither mypy nor pytype support them, but maybe it's a good idea to add something like this for future proofing:

```typescript
line = line.split('#')[0];
```

srittau (commented):
This is the first time I had a look at the pyright code base, but with that limited experience the PR looks good. I hope you don't me adding a few comments, though.

packages/pyright-internal/src/analyzer/importResolver.ts
JelleZijlstra: mypy does support it: https://github.com/python/mypy/blob/866ee316ded61e1f61207ee33b6f31450493ffa3/mypy/modulefinder.py#L761. I agree that it would be useful to support. Once my typeshed PR is done I'd like to write a slightly more precise specification for the VERSIONS file.

packages/pyright-internal/src/analyzer/importResolver.ts
erictraut: You're correct. Removed.

packages/pyright-internal/src/analyzer/importResolver.ts
erictraut: Good catch. I've added support for comments.

PR #1895, opened 2021-05-21
pylance-1301: make it work based on PWD in the worst case

smekkley:
My first attempt with typescript. Rather hacky but it's better than nothing. 
On Linux `.`, resolves to the pyright script directory, $PWD returns where editor(vim) is executed.
https://github.com/microsoft/pylance-release/issues/1312
https://github.com/microsoft/pylance-release/issues/1301

### DIFF
--- a/packages/pyright-internal/src/analyzer/service.ts
+++ b/packages/pyright-internal/src/analyzer/service.ts
@@ -917,9 +917,23 @@ export class AnalyzerService {
 
         timingStats.findFilesTime.timeOperation(() => {
             const matchedFiles = this._matchFiles(this._configOptions.include, this._configOptions.exclude);
+            if (matchedFiles.length == 0) {
+                const envPwd = process.env.PWD;
+                if (envPwd) {
+                    var filespec : FileSpec = {
+                        wildcardRoot: envPwd,
+                        regExp: RegExp("","")
+                    }
+                    let filespecs: FileSpec[] = [filespec];
+                    for (const file of this._matchFiles(filespecs, this._configOptions.exclude)) {
+                        fileMap.set(file, file);
+                    }
 
-            for (const file of matchedFiles) {
-                fileMap.set(file, file);
+                }
+            } else {
+                for (const file of matchedFiles) {
+                    fileMap.set(file, file);
+                }
             }
         });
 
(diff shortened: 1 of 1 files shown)


packages/pyright-internal/src/analyzer/service.ts
```diff
@@ -917,9 +917,23 @@ export class AnalyzerService {
 
         timingStats.findFilesTime.timeOperation(() => {
             const matchedFiles = this._matchFiles(this._configOptions.include, this._configOptions.exclude);
+            if (matchedFiles.length == 0) {
+                const envPwd: string = (process.env.PWD as string);
```
jakebailey, line 921: This is unsafe and will crash at runtime if `PWD` is unset. In TS, `as` is a cast with no type checks.

jakebailey (changes requested):
I don't think this is the right fix. This will end up loading files when the configuration matches nothing, which is a real possibility, and I don't think this the right location to do this work anyway if we want to better support the "rootless" LSP mode.

What'd be more valuable is an LSP trace of how your editor is initializing pyright. If it really is the open file mode with no workspace, then we should be looking at that scenario as a whole and how we want to support it. I don't think that hacking it to pick out pwd is going to be the right solution cross-editor.

packages/pyright-internal/src/analyzer/service.ts
smekkley: Ah, thanks. unset PWD couldn't make it crash with vim, but maybe it's respawning it immediately. Anyway, I pushed the fix, hope it's correct.

jakebailey: I'm afraid it's not; the correct fix would just be `const envPwd = process.env.PWD` and then `if (envPwd)` to check truthiness.

It's good if this works for you (and absolutely fork and use it), but I don't know that we're going to accept this PR or other changes to fix this quite yet until we spec out how we intend for the rootless scenario to work and compare it to our current behavior.

smekkley: Yes, sure. I'm not entirely sure it should be merged either. I was hoping I kick off something at least.
Thank you for the heads up and code explanation.

erictraut:
I agree this isn't the correct fix for this issue, so I'm going to go ahead and close this PR.

PR #1905, opened 2021-05-24
Don't use process.exit in CLI to prevent truncated stdout

jakebailey:
I'm working on the type completeness dashboard and hit https://github.com/nodejs/node/issues/6379 (and associated issues) when capturing pyright's output.

Avoid using `process.exit` to stop the process, instead exiting gracefully with the exit code set. `process.exit` synchronously exits the process, but doesn't flush `console`'s internal buffers or stdout. This can lead to a truncated output, especially when the output is large (which is the case with `--verifytypes` and many libraries). There isn't a good way to force a flush (besides calling undocumented functions to put stdout into blocking mode), so avoiding `process.exit` is a better method and fits with the NodeJS docs that say `process.exit` is not recommended.

After #1822 fixed the file watcher hang, this should be safe to do, but does rely on all promises resolving before the process can exit.

### DIFF
--- a/packages/pyright-internal/src/pyright.ts
+++ b/packages/pyright-internal/src/pyright.ts
@@ -16,7 +16,6 @@ import { timingStats } from './common/timing';
 import chalk from 'chalk';
 import commandLineArgs from 'command-line-args';
 import { CommandLineOptions, OptionDefinition } from 'command-line-args';
-import * as process from 'process';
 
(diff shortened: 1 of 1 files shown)


erictraut:
Ah, good find. I didn't realize that process.exit exited without flushing. That explains some problems I've experienced in other node-based command-line tools I've written in the past :).

The changes LGTM.

erictraut (approved):

PR #1918, opened 2021-05-28
Avoid reanalyzing a file when it is opened in the editor if we have a…

erictraut:
…lready analyzed it and the file contents are the same as before. This restores the behavior that existed prior to this change: https://github.com/microsoft/pyright/commit/2d90df325fd9ffb6bd72e727ec59039b50cdf537#diff-32dc01623d13bd967864d6e0b14c73f5e490534e7bb337a1d182169b8c670a79R426.

### DIFF
--- a/packages/pyright-internal/src/analyzer/sourceFile.ts
+++ b/packages/pyright-internal/src/analyzer/sourceFile.ts
@@ -421,7 +421,17 @@ export class SourceFile {
                 this._clientDocument = TextDocument.create(this._filePath, 'python', version, '');
             }
             this._clientDocument = TextDocument.update(this._clientDocument, contents, version);
-            this.markDirty();
+
+            const fileContents = this._clientDocument.getText();
+            const contentsHash = StringUtils.hashString(fileContents);
+
+            // Have the contents of the file changed?
+            if (fileContents.length !== this._lastFileContentLength || contentsHash !== this._lastFileContentHash) {
+                this.markDirty();
+            }
+
+            this._lastFileContentLength = fileContents.length;
+            this._lastFileContentHash = contentsHash;
         }
     }
 


packages/pyright-internal/src/analyzer/sourceFile.ts
```diff
@@ -421,7 +421,17 @@ export class SourceFile {
                 this._clientDocument = TextDocument.create(this._filePath, 'python', version, '');
             }
             this._clientDocument = TextDocument.update(this._clientDocument, contents, version);
-            this.markDirty();
+
+            const fileContents = this._clientDocument.getText();
+            const contentsHash = StringUtils.hashString(fileContents);
```
bschnurr, line 426: you could delay calculating the hash till after you check the file length

erictraut: How so? I need to calculate the hash so I can verify that it hasn't changed. If it has changed, I need to use it to update the value in `this._lastFileContentHash`.

jakebailey:
I think what we want to do is this:

```ts
setClientVersion(version: number | null, contents: TextDocumentContentChangeEvent[]): void {
    if (version === null) {
        this._clientDocument = undefined;
    } else {
        const wasClosed = !this._clientDocument;
        if (wasClosed) {
            this._clientDocument = TextDocument.create(this._filePath, 'python', version, '');
        }
        this._clientDocument = TextDocument.update(this._clientDocument, contents, version);

        if (wasClosed) {
            const fileContents = this._clientDocument.getText();
            const contentsHash = StringUtils.hashString(fileContents);

            // Have the contents of the file changed?
            if (fileContents.length !== this._lastFileContentLength || contentsHash !== this._lastFileContentHash) {
                this.markDirty();
            }

            this._lastFileContentLength = undefined;
            this._lastFileContentHash = undefined;
        }
    }
}
```

The "_last" properties are only set for unopened files. The above will optimize the case where we move from unopened to opened.

erictraut:
That change handles closed->opened transitions, but it doesn't handle opened->closed transitions properly.

I agree that it avoids some hash calculations, but I'm not that concerned about hash calculations. They're cheap, especially given that the contents will already be in the L2 cache.

jakebailey (approved):

packages/pyright-internal/src/analyzer/sourceFile.ts
bschnurr: i was thinking if the file length changed then you wouldn't need to hash it. but I see now it still needs to be stored.

PR #1973, opened 2021-06-10
Added error handling for very large source files. This situation is n…

erictraut:
…ow detected earlier so we don't attempt to load the file contents and crash the language server in the process.


packages/pyright-internal/src/analyzer/sourceFile.ts
```diff
@@ -520,6 +524,15 @@ export class SourceFile {
             let fileContents = this.getFileContents();
             if (fileContents === undefined) {
                 try {
+                    const fileStat = this.fileSystem.statSync(this._filePath);
```
jakebailey, line 527: Can you move this section down into the `timeOperation` below? That way, we'll count the time it takes to stat in addition to the read itself and not have unaccounted time.

jakebailey (approved):

PR #1980, opened 2021-06-12
Update command-line.md to emphazise flags with arguments

Daggy1234:
image

As per the usage, , the --ignoreexternal flag must be placed after the package and hence this PR adds a note. I struggled a lot with this and hope it helps other people!

### DIFF
--- a/docs/command-line.md
+++ b/docs/command-line.md
@@ -6,32 +6,30 @@ Pyright can be run as either a VS Code extension or as a node-based command-line
 
 | Flag                               | Description                                           |
 | :--------------------------------- | :---------------------------------------------------  |
-| --createstub IMPORT                 | Create type stub file(s) for import                  |
-| --dependencies                      | Emit import dependency information                   |
-| -h, --help                          | Show help message                                    |
-| --ignoreexternal                    | Ignore external imports for --verifytypes (2)        |
-| --lib                               | Use library code for types when stubs are missing    |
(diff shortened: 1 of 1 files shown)


erictraut:
I appreciate the contribution, but I don't think the change is correct. When using the "--verifytypes" option, the `[..files]` are not used. Instead, the library module or submodule specified after `--verifytypes` is analyzed. You must provide the library module or submodule as part of the `--verifytypes` option (e.g. `--verifytypes pandas`). You can optionally add `--ignoreexternal` either before or after the `--verifytypes` option — `--verifytypes pandas --ignoreexternal` or `--ignoreexternal --verifytypes pandas`.

Daggy1234:
Oh sorry my bad, any particular reason why the order of ` --verifytypes --ignoreexternal pandas` doesn't work?

erictraut:
Some of the command-line options require parameters. The `--verifytypes` option requires such a parameter, which is the library module name. That parameter must come directly after the `--verifytypes` option. That's why the documentation indicates `--verifytypes IMPORT`.

erictraut:
There's already documentation here: https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#verifying-type-completeness

Daggy1234:
Oh I'm extremely sorry for wasting your time. Thank you for clearing that up!

erictraut:
Thanks, this looks good. I'll also update the `--help` output to match this formatting change.

PR #1984, opened 2021-06-14
Return empty completionList for Ellipsis

cdce8p:
Currently, writing overloads can be quite cumbersome as Ellipsis don't work too well with completion results. Since `.` is treated as a separator, all possible results are suggested. When combined with `editor.acceptSuggestionOnEnter = on` (the default), it means you first need to press escape, before you can continue with the next line.

Adding `...` to the list of keywords also doesn't work, as it would just add three additional `.` The best solution I found was to check if the previous text is `...` and just return an empty completion list for it.

Screen Shot 2021-06-14 at 18 20 31


erictraut:
Is there a reason why we wouldn't want to apply this logic after two periods in a row? Why wait for three periods in a row?

jakebailey:
Probably not; two is probably a fine assumption for now.

I tried doing something similar to this, but also eliminate completions on a top-level `.` (e.g. eliminate all completions on `.` until `...`, where you're completing on the ellipsis itself), but it was a bit difficult to figure figure out with the current error node setup and how we try to find the right context. This partial fix is an improvement, in any case.

cdce8p:
I've updated the code to work with just two. In a short test, this seems be work even a bit better then before with three.

erictraut:

PR #2064, opened 2021-07-06
minor change to assert, execute command

bschnurr:
Rollup of:
 - ExecuteCommand now can return a result instead of just applying workspaceEdits
 - New CommandResult wrapper for returning custom data along with workspaceEdits
 - Changed Program API to expose tracked files
 - updated assert signature

### DIFF
--- a/packages/pyright-internal/src/commands/commandResult.ts
+++ b/packages/pyright-internal/src/commands/commandResult.ts
@@ -1,0 +1,12 @@
+import { WorkspaceEdit } from 'vscode-languageserver-types';
+
+export interface CommandResult {
+    data?: any;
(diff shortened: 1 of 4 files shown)


jakebailey (approved):

packages/pyright-internal/src/commands/commandResult.ts
```diff
@@ -0,0 +1,12 @@
+import { WorkspaceEdit } from 'vscode-languageserver-types';
```
erictraut, line 1: Please include a header comment and copyright notice.

bschnurr: thanks

PR #2095, opened 2021-07-16
Update dependencies, eliminate broken webpack plugin, begin browser refactorings

jakebailey:
Rollup of:

- Update dependencies, which fixes outstanding `npm audit` problems.
- Eliminate `tsconfig-paths-webpack-plugin`, which has a number of bugs, including one that silently breaks packaging when two files have the same name in two of our packages.
- Begin refactorings for de-nodifying codebase.

### DIFF
--- a/build/lib/webpack.js
+++ b/build/lib/webpack.js
@@ -1,6 +1,8 @@
 const fs = require('fs');
 const path = require('path');
 const glob = require('glob');
+const JSONC = require('jsonc-parser');
+const assert = require('assert');
 
 /**
  * Builds a faked resource path for production source maps in webpack.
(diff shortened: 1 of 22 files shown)


jakebailey:
Hm, that build failure is odd. I wonder if that plugin was masking yet another bug...

jakebailey:
There was a latent bug that my plugin removal revealed. We had:

https://github.com/microsoft/pyright/blob/1aec5188e071ce87012cfcc5d92d1462b17ff9df/packages/pyright-internal/src/server.ts#L38

But:

https://github.com/microsoft/pyright/blob/1aec5188e071ce87012cfcc5d92d1462b17ff9df/packages/pyright-internal/src/pyright.ts#L609

In the same package.

jakebailey:
I'll change the pylance build to ensure that pyright packages properly, as we don't seem to check that (oops).

PR #2155, opened 2021-08-05
Update issue template labels

nihaals:


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -2,7 +2,7 @@
 name: Bug report
 about: Create a report to help us improve pyright
 title: ''
-labels: ''
+labels: bug
```
nihaals, line 5: It seems not *all* issues created with this template are manually labelled with `bug` and instead with `enhancement request`. Maybe it should still be the default with the form and manually changed if needed?

jakebailey, line 5: I think that @erictraut typically avoids labelling issues as bugs until they've been looked at, hence why this is set to a blank. (The other change seems fine given the current state of the labels.)

nihaals: Related: https://github.com/microsoft/pyright/pull/2155#discussion_r683814633

PR #2238, opened 2021-08-28
Fix docs/configuration.md misleading case sensitive `Pyproject.toml` typo

omertuc:
Creating `Pyproject.toml` as it is currently shown in the docs title will not work,
because the file name is case sensitive. It should be exactly `pyproject.toml`


erictraut:
Thanks for the contribution.

PR #2263, opened 2021-09-03
Add missing `strictSetInference` option to table

nullromo:
The `strictSetInference` row was missing from the Diagnostic Rule Defaults table.


erictraut:
Thanks for catching that omission.

PR #2264, opened 2021-09-04
Adjust auto-import sorting to better match isort

cdce8p:
TypeScript and isort handle sorting of `_` differently. This can result in the auto-import insert position not matching the isort behavior.

This PR adds a workaround to better handle underscores when sorting import names.

#### Example
```py
from lib import (
    ...
    # auto-import
    MY_CONST_VAR,
    # isort
    ....
)

MYA_CONST_VAR  # import from lib
```

### DIFF
--- a/packages/pyright-internal/src/analyzer/importStatementUtils.ts
+++ b/packages/pyright-internal/src/analyzer/importStatementUtils.ts
@@ -164,12 +164,12 @@ export function getTextEditsForAutoImportSymbolAddition(
             // This can't be reproduced by a normal string compare in TypeScript, since '_' > 'A'.
             // Replace all '_' with '=' which guarantees '=' < 'A'.
             // Safe to do as '=' is an invalid char in Python names.
-            const symbolNameCompare = symbolName.replace('_', '=');
+            const symbolNameCompare = symbolName.replace(/_/g, '=');
             for (const curImport of importStatement.node.imports) {
                 const curImportNameType = _getImportSymbolNameType(curImport.name.value);
                 if (
                     (curImportNameType === symbolNameType &&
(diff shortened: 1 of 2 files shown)


erictraut:
Thanks for the contribution, @cdce8p. The change looks good to me.

@jakebailey, could you please give it a quick review as well?

packages/pyright-internal/src/analyzer/importStatementUtils.ts
```diff
@@ -160,10 +160,16 @@ export function getTextEditsForAutoImportSymbolAddition(
             // Insert new symbol by import symbol type and then alphabetical order.
             // Match isort default behavior.
             const symbolNameType = _getImportSymbolNameType(symbolName);
+            // isort will prefer '_' over alphanumerical chars
+            // This can't be reproduced by a normal string compare in TypeScript, since '_' > 'A'.
+            // Replace all '_' with '=' which guarantees '=' < 'A'.
+            // Safe to do as '=' is an invalid char in Python names.
+            const symbolNameCompare = symbolName.replace('_', '=');
```
jakebailey, line 167:
```suggestion
            const symbolNameCompare = symbolName.replace(/_/g, '=');
```

I'm thinking you intended this to replace all underscores? replace only replaces the first match unless you use a regex. (We can't use replaceAll yet, so this is the right way to replace all.)

cdce8p: 👍🏻 Updated the code and added an additional test case for it.

erictraut: Good catch @jakebailey!

PR #2279, opened 2021-09-08
Added missing check for access to a module member that is meant to be…

erictraut:
… private to that module — e.g. a private-name type alias, a private-named function in a ".py" file within a py.typed library, or a non-re-exported import within a type stub.

### DIFF
--- a/packages/pyright-internal/src/analyzer/binder.ts
+++ b/packages/pyright-internal/src/analyzer/binder.ts
@@ -199,13 +199,13 @@ export class Binder extends ParseTreeWalker {
         id: getUniqueFlowNodeId(),
     };
 
-    // Map of symbols at the module level that may be private depending
-    // on whether they are listed in the __all__ list.
-    private _potentialPrivateSymbols = new Map<string, Symbol>();
-
     // Map of symbols at the module level that may be externally
     // hidden depending on whether they are listed in the __all__ list.
     private _potentialHiddenSymbols = new Map<string, Symbol>();
+
+    // Map of symbols at the module level that may be private depending
+    // on whether they are listed in the __all__ list.
+    private _potentialPrivateSymbols = new Map<string, Symbol>();
 
     constructor(fileInfo: AnalyzerFileInfo, private _moduleSymbolOnly = false) {
         super();
@@ -255,7 +255,7 @@ export class Binder extends ParseTreeWalker {
         this._bindDeferred();
 
         // Use the __all__ list to determine whether any potential private
-        // symbols should be made externally invisible.
+        // symbols should be made externally hidden or private.
         this._potentialHiddenSymbols.forEach((symbol, name) => {
             if (!this._dunderAllNames?.some((sym) => sym === name)) {
                 symbol.setIsExternallyHidden();
--- a/packages/pyright-internal/src/analyzer/binder.ts
+++ b/packages/pyright-internal/src/analyzer/binder.ts
@@ -203,6 +203,10 @@ export class Binder extends ParseTreeWalker {
     // on whether they are listed in the __all__ list.
     private _potentialPrivateSymbols = new Map<string, Symbol>();
 
+    // Map of symbols at the module level that may be externally
+    // hidden depending on whether they are listed in the __all__ list.
+    private _potentialHiddenSymbols = new Map<string, Symbol>();
+
     constructor(fileInfo: AnalyzerFileInfo, private _moduleSymbolOnly = false) {
         super();
 
@@ -252,11 +256,15 @@ export class Binder extends ParseTreeWalker {
 
         // Use the __all__ list to determine whether any potential private
         // symbols should be made externally invisible.
-        this._potentialPrivateSymbols.forEach((symbol, name) => {
-            // If this symbol was found in the dunder all, don't mark it
-            // as externally hidden.
+        this._potentialHiddenSymbols.forEach((symbol, name) => {
             if (!this._dunderAllNames?.some((sym) => sym === name)) {
                 symbol.setIsExternallyHidden();
+            }
+        });
+
+        this._potentialPrivateSymbols.forEach((symbol, name) => {
+            if (!this._dunderAllNames?.some((sym) => sym === name)) {
+                symbol.setIsPrivateMember();
             }
         });
 
@@ -1438,18 +1446,17 @@ export class Binder extends ParseTreeWalker {
             const symbol = this._bindNameToScope(this._currentScope, symbolName);
             if (
                 symbol &&
+                (this._currentScope.type === ScopeType.Module || this._currentScope.type === ScopeType.Builtin) &&
                 (!node.alias ||
                     node.module.nameParts.length !== 1 ||
                     node.module.nameParts[0].value !== node.alias.value)
             ) {
-                if (this._fileInfo.isStubFile) {
+                if (this._fileInfo.isStubFile || this._fileInfo.isInPyTypedPackage) {
                     // PEP 484 indicates that imported symbols should not be
                     // considered "reexported" from a type stub file unless
                     // they are imported using the "as" form and the aliased
                     // name is entirely redundant.
-                    symbol.setIsExternallyHidden();
-                } else if (this._fileInfo.isInPyTypedPackage && this._currentScope.type === ScopeType.Module) {
-                    this._potentialPrivateSymbols.set(symbolName, symbol);
+                    this._potentialHiddenSymbols.set(symbolName, symbol);
                 }
             }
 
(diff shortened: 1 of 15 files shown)


packages/pyright-internal/src/analyzer/binder.ts
```diff
@@ -255,7 +255,9 @@ export class Binder extends ParseTreeWalker {
         this._potentialPrivateSymbols.forEach((symbol, name) => {
             // If this symbol was found in the dunder all, don't mark it
             // as externally hidden.
-            if (!this._dunderAllNames?.some((sym) => sym === name)) {
+            const isInDunderAll = this._dunderAllNames?.some((sym) => sym === name);
+            const isTypedStubSymbol = this._fileInfo.isStubFile && symbol.getTypedDeclarations().length > 0;
+            if (!isInDunderAll && !isTypedStubSymbol) {
                 symbol.setIsExternallyHidden();
```
jakebailey, line 261: I vaguely recall @heejaechang having to modify this code to ensure we could handle untyped libraries; I think this change should be safe given it explicitly applies to stubs only, but worth a second set of eyes.

erictraut: Yeah, I think this is safe for stubs.

Some of the typeshed stubs include symbols whose names would seem to indicate that they shouldn't be exported. For example, the `_exit` function in the `os.pyi` stub. That same stub also defines a symbol `_T`. I verified that mypy treats  `_exit` as an exported symbol but `_T` as a private (not exported) symbol. 

Note that this change results in a difference between a ".pyi" file and a pytyped ".py" file, at least by our current rules for pytype interpretation.

jakebailey: Would this change only add the diagnostic (to warn uses of these things), and not affect the visibility of the symbols? That'd be my worry, that code that already uses something like `_exit` now getting changed to `Unknown`, when it's explicitly declared in the stub for a reason.

erictraut: There are two parts to the change in this PR. The first part avoids marking symbols as "externally hidden" if they come from a stub file and have a declared type. The second part adds a missing diagnostic check when accessing an "externally hidden" symbol from a module using a member access expression.

For clarity, here's how the behavior changes when accessing `os._exit` from the `os.pyi` stub.

Previous to this PR:
A1. We include `_exit` in the symbol table but mark it as "externally hidden".
A2. If you access this symbol using `from os import _exit`, you will see an error and the type will be marked unknown.
A3. If you access it using `import os` and `os._exit`, you will not see an error and the type will not be unknown.

Clearly the behavior between A2 and A3 is inconsistent and wrong.

With this PR:
B1. We include `_exit` in the symbol table and do not mark it as externally hidden because it comes from a stub file and it has a declared type (whereas `_T` does not have a declared type)
B2. If you import it using `from os import _exit`, you will not see an error
B3. If you import it using `import os` and `os._exit`, you will not see an error

So this change makes B2 and B3 consistent when the symbol comes from a stub.

I was previously concerned that this PR would change the visible behavior with py.typed libraries, but I've confirmed that it doesn't. That's because we don't enter private symbols 

Let's look at a different example where a symbol `foo` comes from a `py.typed` library, but `foo` is not publicly exported (e.g. it's imported from another module but not re-exported and not included in the `__all__` list).

Previous to this PR:
C1. We include `foo` in the symbol table and mark it externally hidden
C2. If you access this symbol using `from os import _exit`, you do not see an error and the type is known (this is a bug)
C3. If you access it using `import os` and `os._exit`, you do not see an error and the type is known (this is a bug)

With this PR:
D1. We include `foo` in the symbol table and mark it externally hidden
D2. If you access this symbol using `from os import _exit`, you do not see an error and the type is known (this is still a bug)
D3. If you access it using `import os` and `os._exit`, you see an error and the type is known

The visible behavior change is between C3 and D3, since users will now see an error. That's where we're likely to get some negative reaction to this change.

There's still a bug in case C2/D2. (I just discovered this as I was manually verifying these various cases.) I think we should fix this as part of this change. It's also likely to get some negative reactions, but I'd rather get all these fixes in place at the same time and "rip the bandaid off".

erictraut:
I've fixed the bug for case C2, and that case now generates an error too. 

I think the changes in this PR are correct, but I worry that they're going to generate a lot of new errors for pyright/pylance users who have type checking enabled. I needed to make a few hundred changes to my team's code base to fix local type stubs that were not correctly re-exporting imported symbols. This is going to expose a lot of similar bugs in public type stubs. I wonder if we should move these new errors to a new diagnostic rule so it can be disabled separately from other general type errors. Or I could report them using the existing `reportPrivateUsage` diagnostic rule, which is off by default in basic type checking mode. @jakebailey, thoughts?

jakebailey:
I'd be locking them behind the private usage check (you can use it, but a stricter mode may yell at you); I was actually thinking it was that way already.

packages/pyright-internal/src/analyzer/binder.ts
```diff
@@ -252,14 +256,18 @@ export class Binder extends ParseTreeWalker {
 
         // Use the __all__ list to determine whether any potential private
         // symbols should be made externally invisible.
```
heejaechang, line 258: update comments?

packages/pyright-internal/src/analyzer/binder.ts
```diff
@@ -1438,18 +1446,17 @@ export class Binder extends ParseTreeWalker {
             const symbol = this._bindNameToScope(this._currentScope, symbolName);
             if (
                 symbol &&
+                (this._currentScope.type === ScopeType.Module || this._currentScope.type === ScopeType.Builtin) &&
                 (!node.alias ||
                     node.module.nameParts.length !== 1 ||
                     node.module.nameParts[0].value !== node.alias.value)
             ) {
-                if (this._fileInfo.isStubFile) {
+                if (this._fileInfo.isStubFile || this._fileInfo.isInPyTypedPackage) {
```
heejaechang, line 1454: so, now even if import is not written as "import x as x", if x is in __ all __ , it is not marked as hidden?

previously, stub didn't care about __ all __ to mark symbol as hidden or not. it only cared about 'import x as x" form.

erictraut: Yeah, the challenge I ran into is that there are symbols in some stubs that begin with `_` that are meant to be exported. The `__all__` mechanism was the only way I could think of supporting it. This came up frequently in the hand-rolled stubs that I maintain for my team's code base, but some of these `_` symbols show up even in typeshed stubs. For example, consider the function `os._exit`.

packages/pyright-internal/src/analyzer/binder.ts
```diff
@@ -1604,21 +1611,21 @@ export class Binder extends ParseTreeWalker {
...
-                                symbol.setIsExternallyHidden();
-                            } else if (
-                                this._fileInfo.isInPyTypedPackage &&
-                                this._currentScope.type === ScopeType.Module
+                        if (
+                            this._currentScope.type === ScopeType.Module ||
+                            this._currentScope.type === ScopeType.Builtin
+                        ) {
+                            if (
+                                !importSymbolNode.alias ||
```
heejaechang, line 1619: same here.

packages/pyright-internal/src/analyzer/binder.ts
```diff
@@ -2952,11 +2959,21 @@ export class Binder extends ParseTreeWalker {
...
+                    if (isPrivateOrProtectedName(name)) {
+                        if (isPrivateName(name)) {
+                            // Private names are obscured, so they are always externally hidden.
+                            symbol.setIsExternallyHidden();
+                        } else if (this._fileInfo.isStubFile || this._fileInfo.isInPyTypedPackage) {
+                            if (this._currentScope.type === ScopeType.Builtin) {
+                                // Don't include private-named symbols in the builtin scope.
+                                symbol.setIsExternallyHidden();
+                            } else {
+                                this._potentialPrivateSymbols.set(name, symbol);
```
heejaechang, line 2972: now symbol in stub can affected by __ all __ when it was always hidden before.

packages/pyright-internal/src/analyzer/binder.ts
heejaechang: this will impact auto-import behavior. but I guess that's fine. (since auto import decide what to include in auto import using the symbol. externallyHidden flag.

packages/pyright-internal/src/analyzer/binder.ts
erictraut: Yes, this is intended. See my comment above.

packages/pyright-internal/src/analyzer/binder.ts
erictraut: Yeah, I consider the current behavior for auto-import to be a bug. It suggests completions that will result in runtime exceptions. This change addresses that bug.

erictraut: I just posted a question in typeshed about symbols within a type stub that begin with an underscore. https://github.com/python/typeshed/issues/6018

PR #2290, opened 2021-09-10
Improve suite 'Expected ":"' knock-on errors

microbit-matt-hillsdon:
Before this change, simple scripts with missing colons such as:

```python
if True
    pass
```

result in quite a lot of diagnostic output:

- Expected ":"
- Expected expression
- Unindent not expected
- Expected expression
- Statements must be separated by newlines or semicolons

Some of this output is also seen when part way through writing (eventually) valid code.

The parser consumes the newline in error recovery which causes some onward confusion as it tries to parse the indent as a simple statement.

The PR avoids consuming the newline in error recovery, giving it a chance to parse the following indented block which first looks for that newline. I've also avoided parsing as a simple statement if the parser has consumed the line in the error recovery, as I can't see how that can succeed. After the change we just get the first error.

I wasn't sure how best to approach the testing. Most of the parser tests are indirect but it seemed worth having direct tests for error recovery.

I hope the error recovery doesn't add too much complexity. I think it's worthwhile here as the knock-on errors can be quite confusing and can be intimidating to new programmers.

### DIFF
--- a/packages/pyright-internal/src/parser/parser.ts
+++ b/packages/pyright-internal/src/parser/parser.ts
@@ -1180,10 +1180,16 @@ export class Parser {
         if (!this._consumeTokenIfType(TokenType.Colon)) {
             this._addError(Localizer.Diagnostic.expectedColon(), nextToken);
 
-            // Try to perform parse recovery by consuming tokens until
-            // we find the end of the line.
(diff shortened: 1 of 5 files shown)


erictraut:
This is a big improvement. Thanks for the contribution!

PR #2291, opened 2021-09-10
Ambiguous tuples become as specific as possible.

sasano8:
The rationale is as follows.

>Tuple Expressions
>When inferring the type of a tuple expression (in the absence of bidirectional inference hints), Pyright assumes that the tuple has a fixed length, and each tuple element is >typed as specifically as possible.

```python
# The inferred type is Tuple[Literal[1], Literal["a"], Literal[True]]
var1 = (1, "a", True)

def func1(a: int):
    # The inferred type is Tuple[int, int]
    var2 = (a, a)

    # If you want the type to be Tuple[int, ...]
    # (i.e. a homogenous tuple of indeterminate length),
    # use a type annotation.
    var3: Tuple[int, ...] = (a, a)
```


erictraut:
Good catch. Thanks for the contribution.

PR #2326, opened 2021-09-22
Added support for reporting the use of types deprecated in PEP 585, etc.

erictraut:
This PR adds the use of the "Deprecated" hint tag. In VS Code, these are styled with a strike-through effect. Deprecated symbols are not flagged at the time they are imported, but rather when they are used. 

Screen Shot 2021-09-21 at 11 48 03 PM

I found what appears to be a bug in VS Code. It is applying the strike-through style for text ranges that were not included in the hint's text range. We will need to work with the VS Code team to diagnose and fix this issue.

Screen Shot 2021-09-21 at 11 47 27 PM

### DIFF
--- a/packages/pyright-internal/src/analyzer/checker.ts
+++ b/packages/pyright-internal/src/analyzer/checker.ts
@@ -17,6 +17,7 @@ import { DiagnosticLevel } from '../common/configOptions';
 import { assert } from '../common/debug';
 import { Diagnostic, DiagnosticAddendum } from '../common/diagnostic';
 import { DiagnosticRule } from '../common/diagnosticRules';
+import { PythonVersion, versionToString } from '../common/pythonVersion';
 import { TextRange } from '../common/textRange';
 import { Localizer } from '../localization/localize';
 import {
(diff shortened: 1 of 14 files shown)


jakebailey:
I have yet to read the code, but just from the screenshots, only the type itself should be marked as deprecated, not things derived from it like instances.

(Unless this is the bug you're describing?)

erictraut:
That's the bug I'm describing. I confirmed that the language server is sending only one diagnostic range for the text "Deque". VS Code is stylizing "q" without any directive to do so.

erictraut:
Oh, never mind. I just figured out the problem. It was a bug in my implementation.

jakebailey:
Huh, looks like both are possible.

jakebailey:
Overall, this looks good, but the main thing I'm wondering about here is if this is going to end up being too distracting; I think it's pretty realistic that the bulk of typed code out there still uses these constructs and will for a good while, and then we strike things out for anyone who simply updates their interpreter (which implicitly sets their python version to one that can support things like `|` instead of `Union`). We'll probably need to ensure settings like `"editor.showDeprecated": false` are documented so people can revert back if they wish, if anything.

Given it's Wednesday, I'd probably want to hold off on merging this quite yet if possible to keep it out of the Pylance release later today.

erictraut:
Yeah, I wasn't planning to get this into this week's pylance release. I agree that it needs more discussion before we move forward with it. I consider this a proof of concept more than a final ready-to-go feature.

PR #2334, opened 2021-09-23
Add support for completing keys for TypedDict arguments

RobertCraigie:
This PR adds support for autocompleting keys for `TypedDict` arguments to functions.

Image showcasing completions

It should be noted that this does _not_ add support for https://github.com/microsoft/pylance-release/issues/654.

### Nested TypedDicts

Image of nested TypedDicts

### Unions of TypedDicts

Image of unions of TypedDicts

### Narrows unions of TypedDicts

If the keys that are already present in the dictionary are not valid for one or more TypedDicts then suggestions will be removed.

Image of narrowing unions of TypedDicts

### Removes redundant completions

Keys that are already present in the dictionary will not be suggested.

Image showcasing removing redundant completions

### Lists of TypedDicts

Image showcasing completions for lists of TypedDicts

### Erroneous parsing

Dictionary expressions that are not fully complete, e.g.

```py
{
    'hello': 'world',
    '.'
}
```

are handled naively, I simply check that the cursor is inside a string inside a dictionary and that the previous token is a comma. I do not know if there is a better method for achieving this.

### Completion bug

While writing tests for this feature I ran into a bug. Parameter arguments are suggested while inside string literals, see [this](https://github.com/microsoft/pyright/compare/main...RobertCraigie:typeddict-arg-complete?expand=1#diff-091452b93d749b8b9475a3aa506d5121c4eb0c67f37b24403fa8812d59ad699bR15) test case.

## Motivation

Adding key completions for TypedDict arguments will significantly improve the experience using https://github.com/RobertCraigie/prisma-client-py

### DIFF
--- a/packages/pyright-internal/src/languageService/completionProvider.ts
+++ b/packages/pyright-internal/src/languageService/completionProvider.ts
@@ -59,12 +59,10 @@ import {
     isNone,
     isOverloadedFunction,
     isUnbound,
-    isUnion,
     isUnknown,
     OverloadedFunctionType,
     Type,
     TypeBase,
-    UnionType,
     UnknownType,
 } from '../analyzer/types';
 import {
@@ -92,6 +90,7 @@ import {
     ArgumentCategory,
     DecoratorNode,
     DictionaryKeyEntryNode,
+    DictionaryNode,
     ErrorExpressionCategory,
     ErrorNode,
     ExpressionNode,
@@ -104,6 +103,7 @@ import {
     ParameterNode,
     ParseNode,
     ParseNodeType,
+    SetNode,
     StringNode,
 } from '../parser/parseNodes';
 import { ParseResults } from '../parser/parser';
@@ -298,11 +298,6 @@ interface CompletionDetail {
     edits?: Edits | undefined;
     sortText?: string | undefined;
     itemDetail?: string | undefined;
-}
-
-interface DictPathEntry {
-    key: string;
-    entry: DictionaryKeyEntryNode;
 }
 
 export const autoImportDetail = 'Auto-import';
@@ -1301,8 +1296,6 @@ export class CompletionProvider {
                             priorText,
                             priorWord,
                             postText,
-                            undefined /* TODO: base expression node */,
-                            parseNode,
                             completionList
                         );
                     }
@@ -1386,15 +1379,7 @@ export class CompletionProvider {
                 }
 
                 // Add literals that apply to this parameter.
-                this._addLiteralValuesForArgument(
-                    signatureInfo,
-                    priorText,
-                    priorWord,
-                    postText,
-                    signatureInfo.callNode.arguments[callInfo.activeIndex]?.valueExpression,
-                    parseNode,
-                    completionList
-                );
+                this._addLiteralValuesForArgument(signatureInfo, priorText, priorWord, postText, completionList);
             }
         }
     }
@@ -1404,8 +1389,6 @@ export class CompletionProvider {
         priorText: string,
         priorWord: string,
         postText: string,
-        baseExpressionNode: ExpressionNode,
-        parseNode: ParseNode,
         completionList: CompletionList
     ) {
         signatureInfo.signatures.forEach((signature) => {
@@ -1421,15 +1404,7 @@ export class CompletionProvider {
             }
 
             const paramType = type.details.parameters[paramIndex].type;
-            this._addLiteralValuesForTargetType(
-                paramType,
-                priorText,
-                priorWord,
-                postText,
-                baseExpressionNode,
-                parseNode,
-                completionList
-            );
+            this._addLiteralValuesForTargetType(paramType, priorText, priorWord, postText, completionList);
             return undefined;
         });
     }
@@ -1439,8 +1414,6 @@ export class CompletionProvider {
         priorText: string,
         priorWord: string,
         postText: string,
-        baseExpressionNode: ExpressionNode | undefined,
-        parseNode: ParseNode,
         completionList: CompletionList
     ) {
         const quoteValue = this._getQuoteValueFromPriorText(priorText);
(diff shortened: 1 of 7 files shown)


packages/pyright-internal/src/languageService/completionProvider.ts
```diff
@@ -1427,6 +1456,299 @@ export class CompletionProvider {
...
+            // this function is only designed return true for states that
+            // are discarded by the parse tree, for example:
+            // {'foo': 'bar', '|}
+            const offset = convertPositionToOffset(provider._position, provider._parseResults.tokenizerOutput.lines)!;
+            const tokens = provider._parseResults.tokenizerOutput.tokens;
+            const index = tokens.getItemAtPosition(offset);
+            const token = tokens.getItemAt(index);
+            const prevToken = tokens.getItemAt(index - 1);
+            return token.type === TokenType.String && prevToken.type === TokenType.Comma;
+        }
```
jakebailey, line 1542: I don't think you need to have `provider` as a parameter here; `this` should work just fine, or just have it be a method like most other helpers.

packages/pyright-internal/src/tests/harness/fourslash/testState.ts
```diff
@@ -274,6 +274,13 @@ export class TestState {
         return [...this.testData.markerPositions.keys()];
     }
 
+    getStringPositionRange(markerString: string, start = 1, end = 1) {
```
heejaechang, line 277: I dont think this has anything special about string. either merge this into getPositionRange. or get string node and get string range from it.

ex)
https://github.com/microsoft/pyright/blob/main/packages/pyright-internal/src/analyzer/parseTreeUtils.ts#L1832

packages/pyright-internal/src/languageService/completionProvider.ts
bschnurr: or maybe just pass in `ParseResults` and `position`

RobertCraigie: I got `this` implicitly has type `any` errors.

I was wary about making it a method as this is the part of the code that I am least confident with and thought it might be a misleading helper as it should only be used with surrounding context.

Passing in `ParseResults` and `position` seems like the best solution to me.

jakebailey: I see; could probably be a `() =>` function at that point (but Bill's suggestion makes more sense).

packages/pyright-internal/src/tests/harness/fourslash/testState.ts
RobertCraigie: You're right, I'd still like to keep this though, what about renaming to `getOffsetPositionRange`?

packages/pyright-internal/src/languageService/completionProvider.ts
```diff
@@ -1427,6 +1456,299 @@ export class CompletionProvider {
...
+    private _getTypedDicts(baseExpressionNode: ExpressionNode, parseNode: ParseNode, type: ClassType | UnionType) {
+        if (!(tokenIsAtKey(this) || this._isInDictionaryStringKey(parseNode))) {
+            return [];
+        }
+
+        // path to resolve nested dictionaries
+        let path: DictPathEntry[] = [];
+        let curNode: ParseNode | undefined = parseNode;
+
+        while (curNode && curNode.id !== baseExpressionNode.id) {
```
heejaechang, line 1506: rather than resolving containing types using parse tree, can you just bind expression of where the key is used and find one what type you are supposed to use?

heejaechang: I guess that didn't work? and that is why you have to find parameter type and dig in to find type of outter key type?

packages/pyright-internal/src/languageService/completionProvider.ts
```diff
@@ -1427,6 +1456,299 @@ export class CompletionProvider {
...
+
+        function tokenIsAtKey(provider: CompletionProvider): boolean {
+            // this function is only designed return true for states that
+            // are discarded by the parse tree, for example:
+            // {'foo': 'bar', '|}
+            const offset = convertPositionToOffset(provider._position, provider._parseResults.tokenizerOutput.lines)!;
+            const tokens = provider._parseResults.tokenizerOutput.tokens;
+            const index = tokens.getItemAtPosition(offset);
+            const token = tokens.getItemAt(index);
+            const prevToken = tokens.getItemAt(index - 1);
```
heejaechang, line 1540: what is index == 0, I think this might be good to be in like parsetreeutils.ts file so can be reused.

getTokenAt(parseResults, position) and getPreviousToken(parseResults, token) ?

packages/pyright-internal/src/languageService/completionProvider.ts
heejaechang: this seems good helper method to put in parse tree utils as well if just binding containing expression can't figure out what outer typeddict current expression is in. and we have to do this search up to find call and find parameter type and dig in to find type of key of nested typed dict.

erictraut: Yeah, a language service provider ideally shouldn't need to do so much parse tree walking to try to determine type information. That's the job of the type evaluator. Currently, the type evaluator doesn't expose a way to extract an "expected type" for a given node, but this is something that I could look at adding. This would allow us to remove all of this complex code from the completion provider. It would also allow us to support much more complex scenarios where expressions are deeply nested (e.g. a TypedDict within a TypedDict within a list within a tuple). Give me a few hours to investigate the feasibility of exposing this sort of functionality from the type evaluator.

erictraut: @RobertCraigie, here's a PR that adds a new entry point to the typeEvaluator called getExpectedType. It allows you to ask what the expected type is for a specified node. I think this would allow you to remove a bunch of code from this PR and make it work in the more general case (including nested expressions). https://github.com/microsoft/pyright/pull/2336

packages/pyright-internal/src/languageService/completionProvider.ts
RobertCraigie: index === 0 is the current position of the cursor. Yeah that's a good suggestion.

packages/pyright-internal/src/tests/harness/fourslash/testState.ts
heejaechang: any issue with merging this with getPositionRange ?

packages/pyright-internal/src/languageService/completionProvider.ts
heejaechang: sorry, typo, what I meant was what "if" index == 0 :)

RobertCraigie: Ah I see, that would be a bug, thanks for spotting it.

I don't think the index could be 0 in this context but it would probably be better to safeguard against it anyway.

packages/pyright-internal/src/tests/harness/fourslash/testState.ts
RobertCraigie: I didn't really want to have to duplicate `getPositionRange('marker', 1, 1)` all over the place but that's not a big deal, happy to merge.

packages/pyright-internal/src/languageService/completionProvider.ts
RobertCraigie: @erictraut Thank you! I've given it a shot and it works perfectly for non-nested TypedDicts but I can't figure out which node I should call `getExpectedType` on to resolve nested expressions.

For example, given `thing3({'wrapped': {'[|/*marker13*/|]'}})` and calling `getExpectedType` on the `Set` node representing `{'[|/*marker13*/|]'}` resolves to `Any`. I also tried the node at the marker and all of its parents but they also resolved to `Any`. Am I doing something wrong?

RobertCraigie: Think I've figured out why the issue mentioned above is happening, it's to do with how `evaluateTypesForExpressionInContext` works with call expressions as explicitly assigning a variable to a TypedDict correctly resolves the nested TypedDict type.

```py
user: User = {
    'meta': {
        '|'
    }
}
```

erictraut: I can take a look at it if you commit and push your updated code that uses `getExpectedType`. I would expect the case covered by `marker13` to work, so if it's not working, I'd like to debug it.

RobertCraigie: @erictraut thank you, I've pushed my WIP code that makes use of `getExpectedType`.

packages/pyright-internal/src/languageService/completionProvider.ts
```diff
@@ -1923,12 +1639,14 @@ export class CompletionProvider {
...
                     priorWord,
                     postText,
-                    undefined /* TODO: base expression node */,
-                    parseNode,
                     completionList
                 );
                 return { completionList };
             }
+
+            if (this._tryAddTypedDictKeys(parseNode, priorText, postText, completionList)) {
```
RobertCraigie, line 1647: I don't know if this is the best place to add the completions but keeping the code in `_addLiteralValuesForTargetType` resulted in incorrect completions as the function is called from `_addLiteralValuesForArgument` which passes the signature type instead of the expected type at the current parse node

erictraut:
@RobertCragie, I simplified your logic in the completionProvider, fixed a couple of test cases, and merged the changes into the main branch. Let me know if you see any issues with the changes I made.

Thanks for this contribution. Much appreciated! Pyright and pylance users will benefit from your work! We can close this PR.

PR #2336, opened 2021-09-24
Implemented getExpectedType function within typeEvaluator

erictraut:
... for the benefit of language service providers.

### DIFF
--- a/packages/pyright-internal/src/analyzer/typeEvaluator.ts
+++ b/packages/pyright-internal/src/analyzer/typeEvaluator.ts
@@ -147,6 +147,7 @@ import {
     EffectiveTypeResult,
     EvaluatorFlags,
     EvaluatorUsage,
+    ExpectedTypeResult,
     FunctionArgument,
     FunctionTypeResult,
     TypeEvaluator,
@@ -506,6 +507,7 @@ export function createTypeEvaluator(importLookup: ImportLookup, evaluatorOptions
     const isExceptionContextManagerCache = new Map<number, boolean>();
     const codeFlowAnalyzerCache = new Map<number, CodeFlowAnalyzer>();
     const typeCache: TypeCache = new Map<number, CachedType>();
+    const expectedTypeCache = new Map<number, Type>();
     const speculativeTypeTracker = new SpeculativeTypeTracker();
     const effectiveTypeCache = new Map<number, EffectiveTypeCacheEntry[]>();
     const suppressedNodeStack: ParseNode[] = [];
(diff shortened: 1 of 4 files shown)


jakebailey:
Very neat; this is something I was hoping to get eventually to use in places like the completion provider to rank completions based on their type similarity (this is what Go/TS do, I think?), at least once we're confident enough to do something like that.

erictraut:
Yeah, I think this approach is an improvement over what we have today, but it does have some limitations:
1. If there is no node (e.g. a missing argument to a call), you cannot call it and ask "what type should this non-existent, missing node be?".
2. It is limited to cases where we use bidirectional type inference. That excludes some important cases for completion suggestions such as index values within subscript expressions.
3. The current implementation in this PR will return only one of the possible expected types in some cases. In particular, if it's an argument provided to a call that is a union of callables or a single callable with overloads, the resulting "expected type" will be one of the argument types, not the union of all possible types that could possibly make sense in that context.

erictraut:
@jakebailey, does this approach sound worthwhile? If so, I'll merge this change so Robert can use it in his PR.

jakebailey:
Yes, I think so, though I hope we can come up with a way around those limitations at some point.

The first limitation is something I already hit in signature help; it'd be nice to ensure the parser outputs error nodes in all of the right places.

jakebailey:
One thing I do see in this PR is that it changes how we do some existing completion; is that likely to cause a performance loss, or are there new completions it's enabling that we didn't already support?

jakebailey:
Ah, no, I see now. It's just replacing "give me the type of the declaration on the left of the assignment" with "give me the expected type for the right side".

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -694,6 +696,28 @@ export function createTypeEvaluator(importLookup: ImportLookup, evaluatorOptions
...
     }
 
+    // Determines the expected type of a specified node based on surrounding
+    // context. For example, if it's a subexpression of an argument expression,
+    // the associated parameter type might inform the expected type.
+    function getExpectedType(node: ExpressionNode): ExpectedTypeResult | undefined {
+        evaluateTypesForExpressionInContext(node);
+
+        let curNode: ParseNode | undefined = node;
+        while (curNode !== undefined) {
```
heejaechang, line 706: so, I guess expectedType returns what code says on what type it expects and getType returns what type evaluator actually evaluated the type to?

basically, we have these functions that return type on expression node

getType: (node: ExpressionNode) => Type | undefined; 
getTypeOfExpression: (node: ExpressionNode, expectedType?: Type, flags?: EvaluatorFlags) => TypeResult;
getDeclaredTypeForExpression: (expression: ExpressionNode) => Type | undefined;

and wondering whether it can be merged to one of existing API? rather than consumer has to use right API at right location.

erictraut: getType and getExpectedType are the two main high-level APIs, and they return different things.

The others are lower-level APIs that I'd prefer not to expose outside of the type evaluator, but they're needed for some other tightly-bound modules like the checker.

erictraut:
@jakebailey said:
>  it'd be nice to ensure the parser outputs error nodes in all of the right places.

Theres nothing for the parser to emit in the case of missing arguments. We don't even know that they're missing until we are able to resolve the type of the call and examine its signature, and that happens way later than parsing.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
heejaechang: so what is difference between getType and getExpectedType? is my understanding correct?

erictraut: getType returns the evaluated type for that node. getExpectedType returns the type that is expected when using bidirectional inference. It is used in the process of evaluating the actual type.

For example:
```python
x: list[int] = ["a"]
```

In the case of the parse node for `"a"`, `getType` would return `Literal["a"]`, and `getExpectedType` would return `int`.

PR #2400, opened 2021-10-08
Implemented logic to retain type aliases in printed form of a union t…

erictraut:
…ype when aliased types are combined in a union.

### DIFF
--- a/packages/pyright-internal/src/analyzer/typeEvaluator.ts
+++ b/packages/pyright-internal/src/analyzer/typeEvaluator.ts
@@ -12690,9 +12690,10 @@ export function createTypeEvaluator(importLookup: ImportLookup, evaluatorOptions
             if (annotatedType) {
                 // PEP 484 indicates that if a parameter has a default value of 'None'
                 // the type checker should assume that the type is optional (i.e. a union
-                // of the specified type and 'None').
+                // of the specified type and 'None'). Skip this step if the type is already
+                // optional to avoid losing alias names when combining the types.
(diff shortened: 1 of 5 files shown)


packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -12680,10 +12690,9 @@ export function createTypeEvaluator(importLookup: ImportLookup, evaluatorOptions
             if (annotatedType) {
                 // PEP 484 indicates that if a parameter has a default value of 'None'
                 // the type checker should assume that the type is optional (i.e. a union
-                // of the specified type and 'None'). Skip this step if the type is already
-                // optional to avoid losing alias names when combining the types.
+                // of the specified type and 'None').
                 if (param.defaultValue && param.defaultValue.nodeType === ParseNodeType.Constant) {
-                    if (param.defaultValue.constType === KeywordType.None && !isOptionalType(annotatedType)) {
+                    if (param.defaultValue.constType === KeywordType.None) {
```
jakebailey, line 12695: I think this is still a valid change to keep around (even if the comment above is removed); it feels worthwhile to avoid creating a union that will end up being the same (and potentially defeat fast-path type comparisons by reference).

erictraut: I don't think this case is common enough to justify a special case, but I don't have a strong opinion. I've added back the special-case logic.

PR #2448, opened 2021-10-18
 Rik de Graaff Added broader support for context managers in with statements which swallow exceptions.

rik-degraaff:


erictraut:
Thinking about this a bit more... The limitations in your function call handling is consistent with the limitations in the existing supported cases (constructor calls and simple names), so I think that's defensible if someone complains about not handling more general cases.

I'm going to merge the PR. Thanks for the contribution.

erictraut:
I'll fix the style issues after I merge.

PR #2520, opened 2021-11-01
Added support in completion provide for enum members. They are now pr…

erictraut:
…operly identified as such and prioritized higher in the completion list than other non-member symbols.


erictraut:
This is related to this feature request: https://github.com/microsoft/pylance-release/issues/1977

I happened to fix a related bug in the completion provider earlier today (it wasn't properly handling members bound to classes). This was an easy fix addition on top of my bug fix.

jakebailey (approved):
Thanks!

Issue #2562, opened 2021-11-12
Pyright with venv still uses Python 2.7 on MacOS


jdtsmith:
**Describe the bug**
The (unsupported) python2.7 interpreter bundled with MacOS is still chosen by pyright (despite the fixes in #829), but only when a venv is specified.

**To Reproduce**
Specify a "venvPath" and "venv" in config.  Start `pyright --verbose" in that directory:

```
  Python version 2.7 from interpreter is unsupported
```

**Expected behavior**
pyright checks for python3 and finds it in the venv (even `python` in the venv points to python3).

**Additional context**
It appears the fallback python3 -> python is disabled in the presence of a venv, _and_ the venv is not first activated when determining the interpreter.

erictraut:
There should be additional details logged to the console. Can you provide those? Also, can you please post your config file?

Pyright never activates a venv. It uses the directory of the venv that you specify. The only time that pyright invokes the python interpreter is when no venv is specified and it needs to fall back to the import resolution paths provided by the default python3 interpreter (i.e. the interpreter that runs when you type "python3" from the shell).

jdtsmith:
Sure here's more detail:

```
% cat pyproject.toml
[tool.pyright]
venvPath = "./.venv"
venv = "default"

% python3 --version
Python 3.9.7

rain% ls .venv 
.venv@

% ls .venv/
default/

% pyright --verbose
No configuration file found.
pyproject.toml file found at /Users/XXX/code/python/scraps.
Loading pyproject.toml file at /Users/XXX/code/python/scraps/pyproject.toml
Python version 2.7 from interpreter is unsupported
Assuming Python platform Darwin
No include entries specified; assuming /Users/XXX/code/python/scraps
Auto-excluding **/node_modules
Auto-excluding **/__pycache__
Auto-excluding .git
stubPath /Users/XXX/code/python/scraps/typings is not a valid directory.
Search paths for /Users/XXX/code/python/scraps
  /usr/local/Cellar/pyright/1.1.186/libexec/lib/node_modules/pyright/dist/typeshed-fallback/stdlib
  /Users/XXX/code/python/scraps
  /Users/XXX/code/python/scraps/typings
  /usr/local/Cellar/pyright/1.1.186/libexec/lib/node_modules/pyright/dist/typeshed-fallback/stubs/...
  /Users/XXX/code/python/venv/default/lib/python3.9/site-packages
Searching for source files
```

erictraut:
[...]

PR #2581, opened 2021-11-18
Use shared helper method to call python interpreter to have consistent behavior between different calls.

heejaechang:
fixes https://github.com/microsoft/pyright/issues/2562


erictraut:
LGTM

heejaechang:
thank you @erictraut 

...

I just confirmed that it now calls through new code path for ensurePythonVersion when pyproject.toml exists.

PR #2641, opened 2021-12-01
Sped up "find reference" by performing a quick text search for the re…

erictraut:
…ference symbol and avoiding additional work for that file if there is no chance of finding a reference within it.

### DIFF
--- a/packages/pyright-internal/src/analyzer/program.ts
+++ b/packages/pyright-internal/src/analyzer/program.ts
@@ -1259,12 +1259,10 @@ export class Program {
                         !invokedFromUserFile ||
                         this._isUserCode(curSourceFileInfo)
                     ) {
-                        this._parseFile(curSourceFileInfo);
-
                         // See if the reference symbol's string is located somewhere within the file.
                         // If not, we can skip additional processing for the file.
-                        const parseResults = curSourceFileInfo.sourceFile.getParseResults();
-                        if (!parseResults || parseResults.text.search(referencesResult.symbolName) >= 0) {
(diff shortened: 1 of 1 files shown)


packages/pyright-internal/src/analyzer/program.ts
```diff
@@ -1259,19 +1259,26 @@ export class Program {
                         !invokedFromUserFile ||
                         this._isUserCode(curSourceFileInfo)
                     ) {
-                        this._bindFile(curSourceFileInfo);
+                        this._parseFile(curSourceFileInfo);
```
heejaechang, line 1262: its okay but you might even skip parsing if text doesn't have text you are looking for?

heejaechang: ex) https://github.com/microsoft/pyright/blob/fdd1e3b85feb8693a2e95ce37d762cb53ef3a310/packages/pyright-internal/src/analyzer/program.ts#L1933

erictraut: The cost of tokenizing and parsing is small compared to the cost of reading in the file contents the first time. And once it's parsed, we retain the file contents as part of the parse results, so it's faster on subsequent searches.

erictraut: I stand corrected. I implemented your suggested change. It didn't have that much of an effect on subsequent searches (3.3s vs 3.2s), but it greatly sped up the first search. From 16s to 6.8s. So I think that's a good optimization.

PR #2677, opened 2021-12-08
Fixed bug that resulted in the import resolution paths retrieved from…

erictraut:
… the currently-selected Python interpreter to omit the working directory if it happens to be in the "sys.path" list.


jakebailey (approved):
LGTM, though you probably don't need the empty string check since I think the code that uses this drops them anyway (and then you'd get exactly what python says is in sys.path).

PR #2712, opened 2021-12-15
Protocol type var variance fix

### LINKED ISSUE #2711: False positive about type-var variance in callable protocol when used in union type
This is a tracking issue.

Originated at microsoft/pylance-release#2172.

Azureblade3808:
Fixes #2711.


Azureblade3808:
Some explanation:

- What is an arbitary type?
An arbitary type used here is an imaginal type that subclasses nothing (maybe not even `object`), and is subclassed by nothing.

- Why using an arbitary type?
The most important feature of an arbitary type is that it doesn't interfere with other types in any unexpected ways, especially when `T | ArbitaryType` would always stay as `T | ArbitaryType` no matter what `T` should be. By contrast, `T` gets swallowed when `T | Any` is evaluated as `Any` in the original logic.

- Is there a better way to define `ArbitaryType` in the checker?
Probably. But I haven't been able to make it out by myself.

erictraut:
Thanks for the contribution. This is a clever fix.

PR #2824, opened 2022-01-06
Fix bytes mixin labled as str in urllib.parse.pyi stub

cccs-jh:
This was causing ParseResultBytes.geturl() to be labelled  with the return type str when it returns bytes.


erictraut:
You would need to fix this in the typeshed github repo, not in the pyright repo. We take the typeshed stubs unmodified from the typeshed repo.

erictraut:
Let me know once your PR is merged into the typeshed repo, and I will prioritize getting the change into pyright. Thanks!

cccs-jh:
@erictraut https://github.com/python/typeshed/pull/6846 has been merged.

erictraut:
Thanks, I've pulled the updated typeshed stubs into pyright, so this will be included in the next release.

Issue #2882, opened 2022-01-17
Support multiple packages with overlapping namespaces


SandyChapman:
In our projects, we use multiple libraries distributed by Google for interacting with Google Cloud. Two such libraries are `google-cloud-bigquery` and `google-cloud-storage`. One feature of these libraries, is that they share a common namespace when importing them. For these two examples, they share `google.cloud` (I.e. `google.cloud.storage` and `google.cloud.bigquery`). Google implements this pattern using [`pkgutil`'s `extend_path`](https://docs.python.org/3/library/pkgutil.html#pkgutil.extend_path) function (see [here](https://github.com/googleapis/python-storage/blob/main/google/cloud/__init__.py) and [here](https://github.com/googleapis/python-bigquery/blob/main/google/cloud/__init__.py)).

However, this pattern introduces problems for Pyright as it appears it only resolves the first instance of a namespace when searching for members.

Here's a simple example:

We have two packages, `dep_1` and `dep_2`, contained in a directory `deps`. Both of these deps have a package with the same name, and subpackages with differing names.

E.g.

```
.
┣ deps
┃  ┣ dep_1
┃  ┃ ┗ pkg
┃  ┃   ┣ __init__.py
┃  ┃   ┗ a
┃  ┃     ┗ __init__.py
┃  ┗ dep_2
┃    ┗ pkg
┃      ┣ __init__.py
┃      ┗ b
┃        ┗ __init__.py
┗ main.py
```

After adding the packages to my PYTHONPATH, I can execute main.py which imports from both pkg.a and pkg.b and works at runtime.

```python
from pkg import a
from pkg import b

print(a.A())
print(b.B())
```

However, this results in Pyright detecting `a` as a valid namespace for `pkg` but not `b`.

```
➤ pyright ./main.py
Loading configuration file at <snip>/pyrightconfig.json
Assuming Python version 3.8
Assuming Python platform Darwin
Auto-excluding **/node_modules
Auto-excluding **/__pycache__
Auto-excluding .git
stubPath <snip>/typings is not a valid directory.
Searching for source files
Found 1 source file
<snip>/main.py
  <snip>/main.py:2:17 - error: "b" is unknown import symbol (reportGeneralTypeIssues)
1 error, 0 warnings, 0 infos 
Completed in 0.737sec
```

For ease of reproducibility, you can find this example here: https://github.com/SandyChapman/pkgutil_import_example

**Describe the solution you'd like**

[...]

PR #2891, opened 2022-01-18
Support multiple packages with overlapping namespaces

SandyChapman:
This allows Pyright to properly infer namespace packages spread
across multiple directories using pkgutil or pkg_resources as specified
here:

https://packaging.python.org/en/latest/guides/packaging-namespace-packages/#pkgutil-style-namespace-packages

See https://github.com/microsoft/pyright/issues/2882 for more details.

### DIFF
--- a/packages/pyright-internal/src/analyzer/importResolver.ts
+++ b/packages/pyright-internal/src/analyzer/importResolver.ts
@@ -12,6 +12,7 @@ import type { Dirent } from 'fs';
 
 import { flatten, getMapValues, getOrAdd } from '../common/collectionUtils';
 import { ConfigOptions, ExecutionEnvironment } from '../common/configOptions';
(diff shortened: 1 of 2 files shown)


SandyChapman:
I'd like some feedback on this PR and if it's useful or if there's a better approach that would be acceptable. If it's a welcome addition, I'll get my company to sign the CLA and reopen the PR under their GitHub account.

erictraut:
Thanks for taking a crack at this. I'll look further into the problem, but I won't be using the code in this PR. It violates a bunch of architectural principles in pyright. In particular, the import resolver module shouldn't have any knowledge of parse nodes or language syntax, and it definitely shouldn't call the parser directly.

SandyChapman:
@erictraut : That makes sense. Thanks again!

PR #2996, opened 2022-02-06
Document reveal_type() and reveal_locals()

JelleZijlstra:
reveal_locals() was not documented anywhere, which may be part of why so few people know about it.

Happy to put this somewhere else if you prefer.


erictraut:

PR #3040, opened 2022-02-14
Fixed long-standing bug in logic that applies config file settings fo…

erictraut:
…r diagnostic rule severity levels. The bug caused all settings overrides to be ignored if a pyrightconfig.json file was present. The new logic applies the default values, then the settings overrides, then the pyrightconfig.json file values. The change also simplifies the code, which was getting a bit unmaintainable.

### DIFF
--- a/packages/pyright-internal/src/analyzer/typeEvaluator.ts
+++ b/packages/pyright-internal/src/analyzer/typeEvaluator.ts
@@ -6582,10 +6582,10 @@ export function createTypeEvaluator(importLookup: ImportLookup, evaluatorOptions
                 node.leftExpression.value === 'reveal_type'
             ) {
                 // Handle the implicit "reveal_type" call.
-                returnResult = getTypeFromRevealType(node);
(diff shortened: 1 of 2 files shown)


erictraut:
@jakebailey, could I get your eyes on this change? You wrote the original logic, and I think its behavior was incorrect. I want to make sure that I'm not missing something here.

jakebailey:
I don't remember writing this bit (maybe blame says otherwise), but I vaguely recall it being designed this way as to let the pyrightconfig have the final say as to the settings. But, I'm not sure on that front. It sure seems better at least to not have to duplicate code, not that that's the main point of the PR.

Half related but it looks like you pushed a commit to this branch rather than main, unrelated to this PR.

erictraut:
pyrightconfig is supposed to have final say, but it shouldn't overwrite settings with defaults. The implementation prior to this change always overwrote all settings values if there was a config file present, even if the config file was effectively empty. I don't think that was the intent, was it?

PR #3051, opened 2022-02-16
Psycopg analysis

erictraut:
Documented analysis of psycopg library and recommendations for changes in pyright. Posting as a PR to encourage feedback in the form of PR comments.

This PR won't be merged into the main branch. Its branch will be deleted after the discussion is complete and we settle on a final plan.

### DIFF
--- a/specs/psycopg_analysis.md
+++ b/specs/psycopg_analysis.md
@@ -1,0 +1,176 @@
+# Analysis of psycopg package
+
+## Overview
+
+The Python package `psycopg` is a "py.typed" package with inlined type annotations.
+
+Users of Pylance reported that they were not receiving completion suggestions for certain symbols imported from psycopg. Running "pyright --verifytypes" on the library revealed that 349 of 1472 public symbols exported from the library were lacking type annotations and were therefore treated by Pyright’s logic as an `Unknown` type.
+
+Daniele Varrazzo, the maintainer of `psycopg`, started a [discussion in the python/typing github project](https://github.com/python/typing/discussions/1058#discussioncomment-2153366) to discuss his concerns about the way Pyright was interpreting the unannotated variables in the package.
+
+## Analysis
+
+### Inheritance of Type Declarations
+
+I analyzed the 349 symbols in `psycopg` that were flagged by Pyright as having "unknown or partially unknown types".
+
+The vast majority of these cases were based on a common pattern where a base class provided a type annotation for a class or instance variable but a subclass simply assigned a value to that same class or instance variable without providing its own type declaration. Pyright's type evaluator already handles this case, using the declared type from the base class. The "--verifytypes" feature contained a bug where it incorrectly flagged these cases. I introduced a bug fix in Pyright 1.1.221 and updated the "Guidance for Library Authors" document to clarify the behavior. This eliminated 267 out of 349 symbols from the list.
+
+It is worth noting that Python type checkers do not handle inherited type declarations in a consistent manner. Consider the following case:
+
+```python
+class Parent:
+    foo: str | float
+
+class Child(Parent):
+    foo = 0
+
+c = Child()
+reveal_type(c.foo)
+c.foo = "Hi"
+```
+
+Pyright always honors a type declaration if it is present. It falls back on type inference only when the developer has omitted a type annotation. This rule is applied consistently in all cases, including types of inherited class and instance variables. In the example above, Pyright evaluates the type of `c.foo` as `str | float`, and it allows a `str` value to be assigned to it.
+
+By contrast, mypy treats some assignment statements as though they are implied type declarations. In the example above, mypy assumes that the value assignment `foo = 0` overrides the `foo: str | float` declaration. It therefore evaluates the type of `c.foo` as `int`, and it emits an error when attempting to assign a `str` value. I don't know if mypy authors intended this behavior, but it seems inconsistent and confusing. There is nothing in PEP 484 that specifies this behavior.
+
+While this particular difference in behavior doesn't affect `psycopg`, it's a good example of inconsistent results between type checkers that can result when a type annotation is omitted. These are the types of inconsistencies that we want to avoid with "py.typed" libraries.
+
+### Instance variable initialization from parameters
+
+Another source of "symbols with unknown types" in `psycopg` are instance variables that have no type annotation but are assigned a value that comes from an annotated input parameter within the `__init__` method.
+
+```python
+class Foo:
+    def __init__(self, name: str, age: int, weight: float) -> None:
+        self.name = name
+        self.age = age
+        self.weight = weight
+```
+
+This pattern is common in Python code. It's probably safe to assume that all Python type checkers will infer the same types for these instance variables if the following conditions are met:
+
+1. The value of the instance variable is assigned only once within the `__init__` method
+2. The assignment is unconditional (e.g. not within an `if` statement)
+3. There is no class variable declared with the same name within the class body
+4. There is no instance or class variable of the same name declared within any of the classes that this class derives from (see comments above about inheritance)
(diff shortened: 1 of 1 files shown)


specs/psycopg_analysis.md
```diff
@@ -0,0 +1,176 @@
...
+
+## Recommendations
+
+I'm recommending the following changes:
+
+### Pyright's treatment of unannotated symbols in "py.typed" libraries
+
+Today, Pyright treats unannotated symbols in a "py.typed" library as "Unknown" if type checking is enabled. When typeCheckingMode is "off", it falls back to its inference logic. This effectively "punishes" users who want to use Pyright and Pylance for static type checking. I recommend that we change this behavior and always fall back on type inference logic when annotations are not present. This change will unfortunately reduce the visibility of missing annotations, so I worry that it will slow efforts to improve type completeness and consistency across the Python ecosystem, but I think it represents a pragmatic tradeoff.
+
+As a mitigation for this lack of visibility, we might want to modify Pyright and Pylance to display "ambiguous types" in a way that differentiates them from "unambiguous types". For example, we could prepend a `~` character to indicate that the type originated from a "py.typed" library and was inferred in a way that might be ambiguous.
```
Akuli, line 154: I'm not convinced that most users will care about consistency so much that a prefix is useful. For example, mypy prepends a `*` prefix in front of inferred types, and nobody seems to find it useful: https://github.com/python/mypy/issues/10076

Akuli (commented):
I like this overall! Just one nit.

specs/psycopg_analysis.md
```diff
@@ -0,0 +1,176 @@
...
+
+It is perhaps time to consider an alternative approach, one that better balances the interests of users and library authors.
+
+## Recommendations
+
+I'm recommending the following changes:
+
+### Pyright's treatment of unannotated symbols in "py.typed" libraries
+
+Today, Pyright treats unannotated symbols in a "py.typed" library as "Unknown" if type checking is enabled. When typeCheckingMode is "off", it falls back to its inference logic. This effectively "punishes" users who want to use Pyright and Pylance for static type checking. I recommend that we change this behavior and always fall back on type inference logic when annotations are not present. This change will unfortunately reduce the visibility of missing annotations, so I worry that it will slow efforts to improve type completeness and consistency across the Python ecosystem, but I think it represents a pragmatic tradeoff.
```
hmc-cs-mdrissi, line 152: Will this continue to be controlled by, `useLibraryCodeForTypes`? I have that set to False as inferred types for tensorflow produce a good number of false positives that they've caused a fair amount of annoyance.

specs/psycopg_analysis.md
hmc-cs-mdrissi: I see one place where it can matter. Should type errors be reported for ambiguous types? A fair amount of false positives get produced if you let pyright infer types on tensorflow mostly because of it's dynamic magic.

It'd be ideal for me if ambiguous types had hover/suggestions but did not trigger errors. Although if it's possible to disable this inference with `useLibraryCodeForTypes ` I could also just continue to set that to False.

specs/psycopg_analysis.md
erictraut: This document is specifically talking about "py.typed" libraries. The `useLibraryCodeForTypes` setting does not apply to "py.typed" libraries. Since tensorflow is not currently "py.typed", `useLibraryCodeForTypes` would still apply.

specs/psycopg_analysis.md
hmc-cs-mdrissi: My concern goes away given this is restricted to py.typed. I would expect a library that claims to be typed to have minimal false positives from ambiguous errors or be willing to improve them at a reasonable pace. If a library has a lot dynamic magic in public api and doesn't type hint/stub it appropriately then I think it's misclaiming to be "py.typed".

specs/psycopg_analysis.md
```diff
@@ -0,0 +1,176 @@
+# Analysis of psycopg package
+
+## Overview
+
+The Python package `psycopg` is a "py.typed" package with inlined type annotations.
+
+Users of Pylance reported that they were not receiving completion suggestions for certain symbols imported from psycopg. Running "pyright --verifytypes" on the library revealed that 349 of 1472 public symbols exported from the library were lacking type annotations and were therefore treated by Pyright’s logic as an `Unknown` type.
```
dlax, line 7: Many things are not specific to psycopg; e.g., while trying to run `pyright --verifytypes` on [click](https://github.com/pallets/click/) (tried quickly, because this raises a *lot* or "errors"), I found a lot of similar issues.

Which makes me wonder if it'd be worth to have a "primer" similar to https://github.com/hauntsaninja/mypy_primer for this verifytypes?

(Note: I'm not a user of pyright myself, mostly coming from psycopg perspective.)

Bibo-Joshi (commented):
Thank you very much @erictraut for taking the time to write this up!

I agree with the main points of your analysis except for a few questions that I left below. However, I'm not sure what your intended goal of this discussion is. In the `Discussion` and `Recommendations` paragraphs, you mainly address Pyright and Pylance and changes to be made in these. But this thread was mainly triggered by https://github.com/python/typing/discussions/1058#discussioncomment-2140526, where the punchline is

>  […] I would prefer that your tools [Pyright] "features" don't become a spec.

So what I had hoped and expected after your comment https://github.com/python/typing/discussions/1058#discussioncomment-2152785 was a proposal on how to adjust the guide at https://typing.readthedocs.io/en/latest/source/libraries.html.

IMO the major points of both your analysis and the thread at https://github.com/python/typing/discussions/1058 are

* there are some situations where types can be inferred with little risk of ambiguity
* ambiguous type hints are often more helpful than no type hints at all
* rules for type inference are not standardized

Putting these all together, I guide on what a `py.typed` lib should do could IMO look somewhat like this

1. Definition of what the public interface is (as is currently)
2. Definition on how to unambiguously annotate the public interface (as currently)
3. Explanation of ambiguously types and their limitations. This should include the punshlines
    1. A type checker should aim to infer a reasonable type hint even for not annotated symbols. So if the inference is "reasonably unambiguous" (to be judged by the libs author), one can omit annotations
    2. examples of cases where type inference is likely to be unambiguous (i.e. the cases analysed in the thread)
    3. Explanation that type inference is not standardized and hence this has some culprits, including inconsistencey across different checkers as well as speed

---

Additionally, a thought that I had while reading your analyis was: __Why__ is type inference not standardized? TBH, I'm little involved in the development of Python itself and I can only assume that this just hasn't developed yet. If there has been discussion on this and standardizing type inteference was rejected for specific reasons, I'd be happy to hear those.

In case standardizing type inference is something one could aim for in the future (would probably require a PEP?), I had the idea of a "benchmark"/set of test cases. This would be a collection of code snippets accompanied by the types that a type checker should infer from those snippets. A type checker could then include this benchmark in the unit tests.
Such a benchmark would surely have to evolve over time, given that the typing system is still evolving. It also wouldn't have to be complete from the beginning, but special cases could be added step by step. A type checker could e.g. say "compatible with type inference standard X.Y".

What are your thoughts on this idea?

erictraut:
@Bibo-Joshi, I don't think it's feasible to standardize type inference rules, nor do I think it's wise to attempt. As I mention in the doc, significant innovation has occurred with respect to type inference, and we're likely to see additional innovation over time. Also, type inference logic reflects the philosophies and design priorities of individual type checkers. Some type checkers value "looseness" (fewer false positives) over "strictness" (catching all potential type violations at the expense of some false positives). You can see from my document how difficult it would be to agree on the type inference rules for even the most basic of circumstances. Specifying the "correct" inference behavior in every case would take years of effort.

If the intent of such an effort is to eliminate type ambiguity, I would argue that it would be a wasted effort because we already have a well-defined way to eliminate such ambiguity: type annotations.

PR #3063, opened 2022-02-16
Update lerna.json and package.jsons with version 1.1.222

debonte:
You published 1.1.222 last night, but when we pulled pyright this morning I noticed that the versions in package.json and lerna.json are still 1.1.221.

Not sure if taking this PR is the right fix, but wanted to save you some time if it is.

Does Pyright 1.1.222 need to be republished with this change?

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -2,7 +2,7 @@
     "packages": [
         "packages/*"
     ],
-    "version": "1.1.221",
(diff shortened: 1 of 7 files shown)


erictraut:
Oops, I forgot to push the commit that updated the package.json. The published versions were built correctly, so no need to republish. I just pushed the commit.

PR #3070, opened 2022-02-18
Handle __set__ overloads with transform_descriptor_types

debonte:
I tested your transform_descriptor_types prototype and it wasn't working for me because my __set__ method had overloads. To be honest, it's not clear to me why overloading `__set__` would be interesting, but [SQLAlchemy's `Mapped` class](https://github.com/sqlalchemy/sqlalchemy/blob/f24a34140f6007cada900a8ae5ed03fe40ce2631/lib/sqlalchemy/orm/base.py#L656) has overloads for `__set__`, so I think we should handle this more gracefully.

Currently I'm getting the value type from the implementation function. However SQLAlchemy's current `Mapped` class has no annotation on the `__set__` impl's value parameter. We could use a Union of the types found on the overloads, but perhaps we should require library authors to annotate the `__set__` impl's value parameter?

What do you think?

### DIFF
--- a/packages/pyright-internal/src/analyzer/dataClasses.ts
+++ b/packages/pyright-internal/src/analyzer/dataClasses.ts
@@ -547,7 +547,17 @@ function transformDescriptorType(evaluator: TypeEvaluator, type: Type): Type {
         return type;
     }
 
-    const setMethodType = evaluator.getTypeOfMember(setMethodInfo);
+    let setMethodType = evaluator.getTypeOfMember(setMethodInfo);
+
+    if (isOverloadedFunction(setMethodType)) {
+        // Find the implementation, not the overloaded signatures.
+        const setMethodImplType = setMethodType.overloads.find((func) => !FunctionType.isOverloaded(func));
+        if (!setMethodImplType) {
+            return type;
+        }
+        setMethodType = setMethodImplType;
+    }
+
     if (!isFunction(setMethodType)) {
         return type;
     }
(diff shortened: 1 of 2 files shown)


erictraut:
I don't think this is the right solution. If the class happens to be in a stub file, there will be no implementation. Plus, implementations with overloads sometimes don't have annotations; only the overloads do. It's perfectly legitimate for a "py.typed" library to do this, since the overloads represent the "public interface" of the method or function. You can see this is the case in the current sqlalchemy implementation.

It's very odd for a `__set__` to have overloads because all such methods return `None` and take only two parameters. Why is there an overload in this case? It looks to me like there's no justification for this. Why isn't there a single `__set__` method that takes an input type of `_T | SQLCoreOperations`?

debonte:
> Why isn't there a single `__set__` method that takes an input type of `_T | SQLCoreOperations`?

Totally agree. That's why I was saying I didn't understand the point of overloading `__set__`. Mike apparently also agrees since he removed the overloads today :) -- https://github.com/sqlalchemy/sqlalchemy/commit/cb35da20d7b93618814a9582a122cc1b24f806b2#

Anyhow, say a Pyright user has an overloaded `__set__` for some weird reason (likely not a good reason), how would they figure out that that's why `transform_descriptor_types` wasn't working for them? Should we flag it as a warning/error?

erictraut:
Maybe we should document it as unsupported in the PEP. I can't think of a good reason to support it, and I also can't think of a good way to implement it. A union of all types from the individual overloads might make sense, but maybe not.

PR #3072, opened 2022-02-18
Changed heap overflow detection to use a dynamic value based on avail…

erictraut:
…able memory rather than a hard-coded size. For the pyright VS Code extension, changed the default "max heap size" from 1.7GB to 3.0GB on 32-bit systems. On 64-bit systems, this value appears to already be a higher value (4GB), and it doesn't appear to get overridden by the lower number.

### DIFF
--- a/packages/pyright-internal/src/analyzer/program.ts
+++ b/packages/pyright-internal/src/analyzer/program.ts
@@ -2089,12 +2089,17 @@ export class Program {
         // Don't bother doing this until we hit this point because the heap usage may not
         // drop immediately after we empty the cache due to garbage collection timing.
         if (typeCacheSize > 750000 || this._parsedFileCount > 1000) {
-            const heapSizeInMb = Math.round(process.memoryUsage().heapUsed / (1024 * 1024));
+            const memoryUsage = process.memoryUsage();
 
-            // Don't allow the heap to get close to the 2GB limit imposed by
-            // the OS when running Node in a 32-bit process.
-            if (heapSizeInMb > 1536) {
-                this._console.info(`Emptying type cache to avoid heap overflow. Heap size used: ${heapSizeInMb}MB`);
+            // If we use more than 90% of the available heap size, avoid a crash
+            // by emptying the type cache.
+            if (memoryUsage.heapUsed > memoryUsage.rss * 0.9) {
+                const heapSizeInMb = Math.round(memoryUsage.rss / (1024 * 1024));
+                const heapUsageInMb = Math.round(memoryUsage.heapUsed / (1024 * 1024));
+
+                this._console.info(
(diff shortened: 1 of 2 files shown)


erictraut:
I'm not extremely confident in this change. A code review is appreciated. How can we test this on a wide range of platforms and architectures (32-bit and 64-bit) while stressing it with a large code base?

packages/pyright-internal/src/analyzer/program.ts
```diff
@@ -2089,12 +2089,17 @@ export class Program {
...
-            const heapSizeInMb = Math.round(process.memoryUsage().heapUsed / (1024 * 1024));
+            const memoryUsage = process.memoryUsage();
 
-            // Don't allow the heap to get close to the 2GB limit imposed by
-            // the OS when running Node in a 32-bit process.
-            if (heapSizeInMb > 1536) {
-                this._console.info(`Emptying type cache to avoid heap overflow. Heap size used: ${heapSizeInMb}MB`);
+            // If we use more than 90% of the available heap size, avoid a crash
+            // by emptying the type cache.
+            if (memoryUsage.heapUsed > memoryUsage.rss * 0.9) {
```
debonte, line 2096: Should you be using memoryUsage.heapTotal rather than rss? My impression is that rss is the amount of heap in physical memory, whereas heapTotal/heapUsed include heap in both physical and virtual memory.

debonte:
> How can we test this on a wide range of platforms and architectures (32-bit and 64-bit) while stressing it with a large code base?

Azure VMs perhaps? That would let us test Windows and Linux at least. Mac testing might have to be manual.

packages/pyright-internal/src/analyzer/program.ts
erictraut: That's what I thought initially, but it turns out that `heapTotal` is the total amount allowed in the current heap. When the current heap is exhausted, it is extended up to the point where it hits the `rss` value.

JelleZijlstra:
At work I run pyright on a very large codebase (several million lines) on a Mac. Anecdotally it feels slow sometimes, and I'm happy to help out testing changes to pyright if that's helpful to you.

jakebailey:
LGTM, but I personally have only managed to hit this cache dropping code when I was working on making things not crash while running on the `sympy` codebase.

bschnurr (approved):

erictraut:
@JelleZijlstra, thanks for the offer. Several million lines? Wow! I presume this is the source base for Quora? 

When you run pyright on this source base, do you use the command-line version? If so, do you see evidence that it is hitting the heap limit and clearing the in-memory type cache? It will output a log messages when it does this.

JelleZijlstra:
Yes, it's Quora. I normally just have pyright running in VSCode, but I am trying the command-line version now. I'll email you with some details.

PR #3101, opened 2022-02-22
Make pyright command fail if option value is missing

not-my-profile:
For options like -v,--venv-path <DIRECTORY> omitting the <DIRECTORY>
should result in the command failing. command-line-args apparently
doesn't take care of that and just sets venvpath to null.

This is especially unfortunate because users might assume that -v
enables verbose output, so running `pyright -v` should really fail
instead of happily running as if no option was passed.


erictraut:
Thanks for the contribution.

PR #3195, opened 2022-03-15
Fix tsc type generation error

nchen63:
When building pyright-internal with `tsc --emitDeclarationOnly --declaration`, it gives an error:

```
src/analyzer/dataClasses.ts:68:16 - error TS4078: Parameter 'evaluator' of exported function has or is using private name 'TypeEvaluator'.

68     evaluator: TypeEvaluator,
                  ~~~~~~~~~~~~~
```

This is caused by a duplicate definition of `TypeEvaluator`. The definition on L119 is not exported and tsc seems to confuse it with the exported parameter on line 68, which is exported and references `TypeEvalulator` defined in typeEvaluatorTypes.ts L279.

### DIFF
--- a/packages/pyright-internal/src/analyzer/dataClasses.ts
+++ b/packages/pyright-internal/src/analyzer/dataClasses.ts
@@ -116,8 +116,8 @@ export function synthesizeDataClassMethods(
     }
 
     // Maintain a list of "type evaluators".
-    type TypeEvaluator = () => Type;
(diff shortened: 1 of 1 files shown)


erictraut:
The proposed change isn't consistent with the naming conventions used throughout pyright. Names of interfaces and classes should be nouns, not verbs. I've checked in a change that addresses this issue while retaining consistency with the naming conventions. Closing this PR.

PR #3266, opened 2022-03-29
Fix cache cleanup threshold calculation

m-novikov:


m-novikov:
Also doesn't work and crashes, on average better runtime as it triggers when threshold is close. Probably multiplier should be lower 0.8 or 0.7

erictraut:
The logic introduced in this PR doesn't look correct to me. It's mixing stats from v8 and from node. I think you're going to need to do more analysis to determine why this is crashing on your system.

Perhaps I should back out the other change since it doesn't address your crash.

Surprisingly, no one else has reported this crash, and of the ~4M pylance users, we haven't seen this crash in any of our crash telemetry.

erictraut:
I'm going to close this PR because I don't think this is the right fix.

PR #3271, opened 2022-03-29
Add support for type narrowing of NamedTuple fields with "is None"

coffeyk:
Unions of NamedTuples were not being narrowed when comparing fields against `None` with the `is` operator. This change allows NamedTuples to narrow Unions with both numerical indexing and attribute access. There are a few examples of the new behavior in the two test samples.

Attribute narrowing was added through a new `narrowTypeForDiscriminatedFieldNoneComparison` function. This code is based on both the `narrowTypeForDiscriminatedFieldComparison` and `narrowTupleTypeForIsNone` functions. There was some commonality with the narrowing for `X.Y is <literal>` but I kept the functions separate because the `None` specific logic cluttered the function. Tests for this addition are in [typeNarrowingIsNoneNamedTuple1.py](https://github.com/microsoft/pyright/pull/3271/files#diff-2a1199dba8a8bf5992864c66bcd22cab47e502c5204e2e2ba57be13dc36736db)

Index narrowing was added by modifying `narrowTupleTypeForIsNone` to support not only `Tuples` but also subclasses. Tests for this addition are in [typeNarrowingIsNoneNamedTuple2.py](https://github.com/microsoft/pyright/pull/3271/files#diff-0b982abad7e4d103fd0776f7daba84d2e879999f526a648ba5d4ff226c93ab7b)

### DIFF
--- a/packages/pyright-internal/src/tests/typeEvaluator1.test.ts
+++ b/packages/pyright-internal/src/tests/typeEvaluator1.test.ts
@@ -309,7 +309,7 @@ test('TypeNarrowingIsNoneNamedTuple1', () => {
 });
 
 test('TypeNarrowingIsNoneNamedTuple2', () => {
-    const analysisResults = TestUtils.typeAnalyzeSampleFiles(['TypeNarrowingIsNoneNamedTuple2.py']);
+    const analysisResults = TestUtils.typeAnalyzeSampleFiles(['typeNarrowingIsNoneNamedTuple2.py']);
 
     TestUtils.validateResults(analysisResults, 0);
 });
--- a/packages/pyright-internal/src/tests/typeEvaluator1.test.ts
+++ b/packages/pyright-internal/src/tests/typeEvaluator1.test.ts
@@ -302,6 +302,18 @@ test('TypeNarrowingIsNoneTuple1', () => {
     TestUtils.validateResults(analysisResults, 0);
 });
 
+test('TypeNarrowingIsNoneNamedTuple1', () => {
+    const analysisResults = TestUtils.typeAnalyzeSampleFiles(['typeNarrowingIsNoneNamedTuple1.py']);
+
+    TestUtils.validateResults(analysisResults, 0);
+});
+
+test('TypeNarrowingIsNoneNamedTuple2', () => {
+    const analysisResults = TestUtils.typeAnalyzeSampleFiles(['TypeNarrowingIsNoneNamedTuple2.py']);
+
+    TestUtils.validateResults(analysisResults, 0);
+});
+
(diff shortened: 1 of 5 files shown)


erictraut:
Thanks for the contribution.

We generally add new type narrowing patterns only if there is signal for sufficient demand and/or the pattern is already supported by mypy or other Python type checkers. I'm somewhat surprised that this is supported because it seems extremely specific and not likely to have broad applicability. But it appears that this form is supported by mypy, so someone evidently has seen a need for this in the past.

In the future, I recommend filing an enhancement request prior to submitting a PR so we can discuss whether we would accept the enhancement.

packages/pyright-internal/src/tests/typeEvaluator1.test.ts
```diff
@@ -302,6 +302,18 @@ test('TypeNarrowingIsNoneTuple1', () => {
...
 });
 
+test('TypeNarrowingIsNoneNamedTuple1', () => {
+    const analysisResults = TestUtils.typeAnalyzeSampleFiles(['typeNarrowingIsNoneNamedTuple1.py']);
+
+    TestUtils.validateResults(analysisResults, 0);
+});
+
+test('TypeNarrowingIsNoneNamedTuple2', () => {
+    const analysisResults = TestUtils.typeAnalyzeSampleFiles(['TypeNarrowingIsNoneNamedTuple2.py']);
```
erictraut, line 312: The file name here must be lowercase. Otherwise the test will fail on OSes (like Linux) that have case-sensitive file systems.

erictraut (changes requested):
The documentation in `type-concepts.md` in the section "Type Guards" should be updated to include the new type narrowing pattern.

erictraut:
I needed to revert the change because it introduced a bunch of false positives in internal Microsoft code bases. I'll need to debug it to determine the root cause.

erictraut:
I've merged the portion of the PR that supports `a[I] is None` for subtypes of tuples.

I reworked the other portion of the change (to support discriminated member accesses based on an `is None` pattern) in [this branch](https://github.com/microsoft/pyright/pull/3273). The code in this PR wasn't properly handling properties, descriptors, or type variables. I think I've handled all of the cases, but I still consider this change highly risky in terms of regressions. I'm therefore going to leave it parked on a branch until after this week's pylance release is published. I'll include it in the next pyright release after that.

Type narrowing logic is very difficult to get correct. It's also tends to be computationally expensive. That's why we don't add support for every pattern that is suggested by a user. I'm still not fully convinced that this pattern is worth supporting, but as I said above, since mypy supports it, I'm willing to use that as a signal that it is generally useful.

PR #3289, opened 2022-04-02
Add support for going to definition for TypedDict keys

RobertCraigie:
I added support for this to both "Go to Definition" and "Go to Type Definition" as in my opinion they are interchangeable in this case and supporting both cases provides the best UX.

## Demo

https://user-images.githubusercontent.com/23125036/161362858-782fa9e5-59db-4800-9dc5-339f84fa567f.mp4

### DIFF
--- a/packages/pyright-internal/src/languageService/definitionProvider.ts
+++ b/packages/pyright-internal/src/languageService/definitionProvider.ts
@@ -17,14 +17,14 @@ import { Declaration, DeclarationType, isFunctionDeclaration } from '../analyzer
 import * as ParseTreeUtils from '../analyzer/parseTreeUtils';
 import { isStubFile, SourceMapper } from '../analyzer/sourceMapper';
 import { TypeEvaluator } from '../analyzer/typeEvaluatorTypes';
-import { ClassType, isClassInstance, isOverloadedFunction, TypeCategory } from '../analyzer/types';
-import { doForEachSubtype, lookUpObjectMember } from '../analyzer/typeUtils';
+import { isOverloadedFunction, TypeCategory } from '../analyzer/types';
+import { doForEachSubtype } from '../analyzer/typeUtils';
 import { throwIfCancellationRequested } from '../common/cancellationUtils';
 import { appendArray } from '../common/collectionUtils';
 import { isDefined } from '../common/core';
 import { convertPositionToOffset } from '../common/positionUtils';
 import { DocumentRange, Position, rangesAreEqual } from '../common/textRange';
-import { ParseNodeType, StringNode } from '../parser/parseNodes';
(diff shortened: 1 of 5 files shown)


packages/pyright-internal/src/languageService/definitionProvider.ts
```diff
@@ -207,6 +219,65 @@ export class DefinitionProvider {
...
+            }
+
+            field = DefinitionProvider._resolveField(base, name, checkedTypes);
+            if (field !== undefined) {
+                return field;
+            }
+        }
+
+        return undefined;
+    }
```
RobertCraigie, line 279: I'm not sure if this is the correct place for this function. I was also overly cautious with a recursion guard that may not be necessary.

I also would be surprised if a utility function like this didn't already exist somewhere in the codebase but I couldn't find it. If such a function does exist I'd be more than happy to refactor my code to use that instead.

packages/pyright-internal/src/languageService/definitionProvider.ts
```diff
@@ -207,6 +219,65 @@ export class DefinitionProvider {
...
+                        definitions,
+                        sourceMapper
+                    );
+                }
+            }
+        });
+    }
+
+    private static _resolveField(type: ClassType, name: string, checkedTypes?: Set<TypeSourceId>): Symbol | undefined {
+        // resolve a field from the symbol table for the given class and all it's base classes
```
erictraut, line 249: Minor nits: Comments should generally be complete sentences starting with capital letter and ending in period. Also, "it's" should be "its".

packages/pyright-internal/src/languageService/definitionProvider.ts
erictraut: This function should probably be in the type evaluator, not in the definition provider. It's analogous to `getDeclarationsForNameNode` but for a StringNode, so maybe call it `getDeclarationsForStringNode`.

Within the type evaluator, you can use `lookUpObjectMember` to find the declaration. This will use proper MRO ordering and avoid returning overridden declarations. Refer to the memberAccess section of `getDeclarationsForNameNode`.

erictraut (changes requested):

RobertCraigie:
@erictraut I've refactored it to make use of `lookUpObjectMember`, thanks for pointing that out to me.

One issue I've encountered is when an entry overrides a synthesized symbol:

```py
class User(TypedDict, total=False):
    get: str

user: User = {|'get'|: 'foo'}
```

Clicking go to definition on 'get' will not resolve any definitions as the symbol lookup resolves to the synthesized `dict.get` method instead of the symbol for the entry. I don't know if anyone would ever actually name an entry like that and It looks like supporting this edge case would complicate the code a lot more.

Is this an acceptable edge case or should we support it as well?

PR #3298, opened 2022-04-06
[PylanceBot] Pull Pylance with Pyright 1.1.235

PylanceBot:
rollup of:

- rely on vscode's file watcher for all folders. remove non-workspace paths node file watcher
- fix normalize path and slash for LSP
- fix missing docstrings in Pillow due to improper typeshed typeshedFallback stub folder path parsing

### DIFF
--- a/packages/pyright-internal/src/common/realFileSystem.ts
+++ b/packages/pyright-internal/src/common/realFileSystem.ts
@@ -10,9 +10,7 @@ import * as fs from 'fs';
 import * as tmp from 'tmp';
 import { URI } from 'vscode-uri';
 
-import { WorkspaceMap } from '../workspaceMap';
 import { ConsoleInterface, NullConsole } from './console';
-import { isDefined } from './core';
 import {
     FileSystem,
     FileWatcher,
@@ -23,7 +21,6 @@ import {
     nullFileWatcherProvider,
     TmpfileOptions,
 } from './fileSystem';
-import { containsPath } from './pathUtils';
 import { getRootLength } from './pathUtils';
 
 // Automatically remove files created by tmp at process exit.
@@ -300,7 +297,10 @@ class RealFileSystem implements FileSystem {
     }
 
     createFileSystemWatcher(paths: string[], listener: FileWatcherEventHandler): FileWatcher {
-        return this._fileWatcherProvider.createFileWatcher(paths, listener);
+        return this._fileWatcherProvider.createFileWatcher(
+            paths.map((p) => this.realCasePath(p)),
+            listener
+        );
     }
 
     createReadStream(path: string): fs.ReadStream {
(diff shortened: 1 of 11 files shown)


packages/pyright-internal/src/common/realFileSystem.ts
```diff
@@ -400,54 +400,12 @@ interface WorkspaceFileWatcher extends FileWatcher {
...
-                try {
-                    return fs.watch(path, { recursive: true }, (event, filename) =>
-                        listener(event as FileWatcherEventType, filename)
-                    );
-                } catch (e: any) {
-                    this._console.warn(`Exception received when installing file system watcher: ${e}`);
-                    return undefined;
-                }
-            })
-            .filter(isDefined);
```
bschnurr, line 439: fyi @erictraut. I'm not familiar with non-vscode clients but removing this might affect them.

packages/pyright-internal/src/languageServerBase.ts
```diff
@@ -598,25 +603,47 @@ export abstract class LanguageServerBase implements LanguageServerInterface {
                     await this.updateSettingsForWorkspace(newWorkspace);
                 });
             });
+
+            this._setupFileWatcher();
+        }
+    }
+
+    private _setupFileWatcher() {
+        if (!this.client.hasWatchFileCapability) {
```
bschnurr, line 612: Now we rely on the client's filewatching

erictraut: But clients don't provide file watching outside of the workspace. What about site-packages?

packages/pyright-internal/src/common/realFileSystem.ts
bschnurr: VSCode added support for filewatching outside the workspace

bschnurr: We were under the assumption that outside the workspace filewatching has also been broken.  But maybe it was only windows.

heejaechang: unfortunately, it never worked before. 

basically, this code (https://github.com/microsoft/pyright/blob/main/packages/pyright-internal/src/analyzer/service.ts#L1414) ignored any events from nodejs since our code expects full path for the changed file (the behavior of vscode file watcher), but nodejs returns just filename. 

I confirm the behavior on both windows, and linux. pretty sure it will be same on mac.

basically, nodejs fs watcher never worked.

and on how to fix it, we decided to just rely on new vscode ability (https://code.visualstudio.com/updates/v1_64#_vscodeworkspacecreatefilesystemwatcher-now-supports-any-path) which will be supported by LSP 3.17 (which we are planning to move to https://github.com/microsoft/pyrx/pull/2289)

if we worry about non vscode client and keep old behavior on pyright and fix nodejs fs watcher issue rather than moving to vscode client (lsp 3.17) 's new fs ability, we can do that. but I dont think we need to that in this PR since outside of workspace nodejs file watcher was never worked before.  (at least last 2 years)

packages/pyright-internal/src/languageServerBase.ts
heejaechang: @erictraut see my other comment (https://github.com/microsoft/pyright/pull/3298#discussion_r844260706) unfortunately, I don't believe it has worked for long time.

erictraut: If it never worked, then I'm OK removing it. I haven't received any bug reports from non-VSCode client maintainers, so I'll conclude that this wasn't a problem.

PR #3348, opened 2022-04-15
Add support for showing key on hover for unions of TypedDicts

RobertCraigie:
## Screenshot

union-typeddict

### DIFF
--- a/packages/pyright-internal/src/languageService/hoverProvider.ts
+++ b/packages/pyright-internal/src/languageService/hoverProvider.ts
@@ -28,7 +28,12 @@ import {
     Type,
     UnknownType,
 } from '../analyzer/types';
-import { ClassMemberLookupFlags, isMaybeDescriptorInstance, lookUpClassMember } from '../analyzer/typeUtils';
+import {
+    ClassMemberLookupFlags,
+    doForEachSubtype,
+    isMaybeDescriptorInstance,
+    lookUpClassMember,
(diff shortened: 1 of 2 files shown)


packages/pyright-internal/src/languageService/hoverProvider.ts
```diff
@@ -255,37 +260,49 @@ export class HoverProvider {
...
         // with the type of the TypedDict key and its docstring, if available.
+        doForEachSubtype(type, (subtype) => {
+            if (isClassInstance(subtype) && ClassType.isTypedDictClass(subtype)) {
+                const entry = subtype.details.typedDictEntries?.get(node.value);
+                if (entry) {
+                    // If we have already added parts for another declaration (e.g. for a union of TypedDicts that share the same key)
+                    // then we need to add a separator to prevent a visual bug.
+                    if (parts.length > 0) {
+                        parts.push({ text: '\n\n---\n' });
+                    }
```
RobertCraigie, line 276: Visual bug encountered:
bug

RobertCraigie: This fix is probably in the wrong place but I didn't want to make a potentially mass breaking change.

erictraut:
Thanks for the contribution. This strikes me as a rare edge case, but since it doesn't increase the complexity in a significant way, I think it's a reasonable addition.

PR #3349, opened 2022-04-15
Fixed file watcher regression.

heejaechang:
Previous change around file watcher added isTracked check so that we don't refresh service just because py file under project root is changed. isTracked will make sure the py file changed is actually part of the source files.

But the fix found a bug where some of the paths that are used under file watcher are not normalized. because of that, isTracked returned some files as not part of the source files breaking file watcher in some cases.

Now, we make sure all paths involved in the file watcher are all normalized.

### DIFF
--- a/packages/pyright-internal/src/common/pathUtils.ts
+++ b/packages/pyright-internal/src/common/pathUtils.ts
@@ -710,12 +710,12 @@ export function hasPythonExtension(path: string) {
     return path.endsWith('.py') || path.endsWith('.pyi');
 }
 
-export function getFileSpec(rootPath: string, fileSpec: string): FileSpec {
+export function getFileSpec(fs: FileSystem, rootPath: string, fileSpec: string): FileSpec {
     let regExPattern = getWildcardRegexPattern(rootPath, fileSpec);
     const escapedSeparator = getRegexEscapedSeparator();
     regExPattern = `^(${regExPattern})($|${escapedSeparator})`;
 
-    const regExp = new RegExp(regExPattern);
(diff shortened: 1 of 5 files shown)


packages/pyright-internal/src/common/pathUtils.ts
```diff
@@ -710,13 +710,13 @@ export function hasPythonExtension(path: string) {
     return path.endsWith('.py') || path.endsWith('.pyi');
 }
 
-export function getFileSpec(rootPath: string, fileSpec: string): FileSpec {
-    let regExPattern = getWildcardRegexPattern(rootPath, fileSpec);
+export function getFileSpec(fs: FileSystem, rootPath: string, fileSpec: string): FileSpec {
+    let regExPattern = normalizePathCase(fs, getWildcardRegexPattern(rootPath, fileSpec));
```
erictraut, line 714: This looks a little suspect. Does "normalizePathCase" handle regular expressions correctly?

erictraut (approved):
Looks good overall. One question for you to consider...

packages/pyright-internal/src/common/pathUtils.ts
heejaechang: I looked what is added as regex pattern and none of them seems to be affected by lower casing, but I can take another approach where regex include /i flag rather than making whole thing lower case.

erictraut: OK, as long as you've considered all of the possible escaping behaviors and the impact that lower-casing may have on those cases.

heejaechang:
I changed approach to just adding /i flag to regex if it is case insensitive file system. this should be much safer approach I think since It won't actually change data, but only change compare behavior.

erictraut (approved):

PR #3360, opened 2022-04-20
push Pylance into pyright

PylanceBot:
    improved source mapper to handle builtin modules such as _io or _collections and updated LSP version.


debonte (approved):

debonte:
@msfterictraut, FYI, the LSP upgrade is in preparation for our LSP notebooks experiment. We shipped it in Pylance's 2022.4.1-pre.1 and 2022.4.2-pre.1 releases and didn't see any issues, so we're shipping it in the stable release this week.

erictraut:
@debonte, OK sounds good. Do we need to update the "engines" version in the pyright VS Code extension? I presume that the new LSP support has a dependency on newer versions of VS Code?

Here's the current version in package.json.

```python
    "engines": {
        "vscode": "^1.63.1"
    },
```

debonte:
The new vscode-languageclient's package.json has the engines version set to `^1.66.0`. So we should update vscode-pyright and vscode-pylance to match? Is there no transitive thing going on here where depending on this new vscode-languageclient forces the "effective" engine version of our packages to be higher?

erictraut:
I don't know where there's a hard dependency between the languageclient library and 1.66.0. It's possible that it's a soft dependency and only certain (new) capabilities depend on the new engine version.

It's unlikely that pylance or pyright users are on versions of VS Code prior to 1.66.0, so I think it's safe (and prudent) to update the engines version for both the pyright and pylance extensions.

PR #3403, opened 2022-04-29
Add "eglot" to the Emacs section

buhtz:
Eglot is an alternative to lsp-mode that is lightweight and uses more Emacs inbuild functionatlity than lsp-mode.


Codeberg-AsGithubAlternative-buhtz:
I don't talk to bots.
And I won't sign anything coming from Microsoft.

If you don't accept: Close the PR.

PR #3410, opened 2022-04-30
Recognize "__" as unused name

carlbordum:


erictraut:
Unlikely a single underscore, I don't think that a double underscore is a standard way to represent an unused name in Python, so I don't think this change is appropriate.

carlbordum:
`__` is useful when you use [gettext](https://docs.python.org/3/library/gettext.html) as `_` is used in a common `gettext` idiom for a real variable.

It also makes sense to use `__` because `_` has special meaning in the standard REPL.

erictraut:
I've never seen `__` used before, nor have I seen any documentation that indicates it should be treated this way.

My recommendation is that you open a discussion topic and ask the community for input. If there is some consensus, then I think it makes sense to make this behavioral change. If this is a convention that is particular to your code base, then I don't think it makes sense.

carlbordum:
Completely fair. No need for a discussion. I just wanted to share a small fix, as I had `__` all over my codebase. 

Back in the day, I learned it from [The Hitchhiker’s Guide to Python!](https://docs.python-guide.org/writing/style/#create-an-ignored-variable), but I don't think thats up-to-date anymore.

carlbordum:
Related issue; sometimes you need implement a protocol with multiple parameters you don't need in your specific use case.

An example would be `__exit__`, that is typically implemented as `__exit__(self, exc_type, exc_val, exc_tb)`, or [pydantic validators](https://docs.pydantic.dev/usage/validators/) that looks like `f(cls, v, values, **kwargs)`.

Let's say I don't need `values` and `kwargs` for my particular validator. Normally, I would implement this as `(cls, v, _, **__)` to make it clear to the reader that they are unused. This would generalise to "ignore names consisting of only underscores".

What do you think would be most idiomatic here, @erictraut ?

erictraut:
Typically, protocols define methods whose parameter names are significant — i.e. they can be used by callers with keyword arguments. In this case, your implementation must provide the correct names for these parameters if you want to match the protocol.

In the less-common case where a protocol defines a method with positional-only parameters, you can choose whatever names you like for those parameters. It's common to use a double underscore followed by a meaningful name. That doesn't imply that they are "unused", but it means they are positional-only and the names are insignificant.

The `__exit__` method used in context managers is an example of a method whose parameters are all positional-only. You can choose whatever name you would like in this case. The typeshed stubs, for example, tend to use:

```python
    def __exit__(
        self, __exc_type: type[BaseException] | None, __exc_val: BaseException | None, __exc_tb: TracebackType | None
    ) -> None: ...
```

PR #3549, opened 2022-06-06
when writing stubs replace callable return types with a valid python…

bschnurr:
… annotation using the 'Callable' keyword.

ex:
(....) -> int
becomes
Callable[..., int]


packages/pyright-internal/src/analyzer/typePrinter.ts
```diff
@@ -76,7 +79,12 @@ export function printType(
 ): string {
     const parenthesizeUnion = (printTypeFlags & PrintTypeFlags.ParenthesizeUnion) !== 0;
     const parenthesizeCallable = (printTypeFlags & PrintTypeFlags.ParenthesizeCallable) !== 0;
-    printTypeFlags &= ~(PrintTypeFlags.ParenthesizeUnion | PrintTypeFlags.ParenthesizeCallable);
+    const annotationFormat = (printTypeFlags & PrintTypeFlags.AnnotationFormat) !== 0;
+    printTypeFlags &= ~(
+        PrintTypeFlags.ParenthesizeUnion |
+        PrintTypeFlags.ParenthesizeCallable |
+        PrintTypeFlags.AnnotationFormat
```
erictraut, line 86: I don't think you want to clear the AnnotationFormat flag here. It should be included recursively for other types that are printed within the outer type.

bschnurr:
Note I didn't add a new diagnosticRuleSet or update `_getPrintTypeFlags` to parse any new rule related to 'AnnotationFormat'

packages/pyright-internal/src/analyzer/typePrinter.ts
```diff
@@ -245,11 +253,14 @@ export function printType(
                 // If it's a Callable with a ParamSpec, use the
                 // Callable notation.
                 const parts = printFunctionParts(type, printTypeFlags, returnTypeCallback, recursionTypes);
-                const paramSignature = `(${parts[0].join(', ')})`;
+                const paramSignature = annotationFormat ? `${parts[0].join(', ')}` : `(${parts[0].join(', ')})`; // no brackets if it's an annotation
```
erictraut, line 256: Please don't use end-of-line comments. In the pyright code base, comments should be block comments prior to the line, and they should be complete sentences starting with a capital letter engine with a period. I'm not sure this comment is needed since the intent is pretty obvious.

packages/pyright-internal/src/analyzer/typePrinter.ts
```diff
@@ -245,11 +253,14 @@ export function printType(
...
+
                 if (FunctionType.isParamSpecValue(type)) {
                     return paramSignature;
                 }
-                const fullSignature = `${paramSignature} -> ${parts[1]}`;
+                const fullSignature = annotationFormat
+                    ? `Callable[${paramSignature}, ${parts[1]}]`
```
erictraut, line 262: There are some signatures that cannot be represented by `Callable`. For example, signatures that include position-only separators, keyword-only arguments, *args, **kwargs, ParamSpecs, TypeVarTuples, and parameters with default arguments. It's not clear we should do in those cases. Perhaps fall back to `Callable[..., <return>]`?

PR #3578, opened 2022-06-15
adding back function getConflictedFile for pylance

bschnurr:


bschnurr:
i'll take a closer look. it wasn't just simply removed

PR #3618, opened 2022-06-26
Treat the type of anonymous classes and functions as not declared

Azureblade3808:
... so that the anonymous variable can still be re-assigned with any values even if it's declared as a class or function in the same scope.


erictraut:
I'm not completely sure what is motivating this change, but I don't think this is the right way to implement this. Please open a discussion topic or feature request that explains the motivation behind the change, and we can decide whether such a change makes sense. I'm going to close the current PR.

Issue #3962, opened 2022-09-18
pyright-langserver v. 1.1.271 doesn't detect source files 


yuval-piiano:
**Describe the bug**
When starting  pyright-langserver (from python) , version 1.1.271 doesn't detect source files while versions before 1.1.270 do .

(I install pyright using pip in python and use this command to start the langserver:

```
pyls_cmd = ["pyright-langserver", "--stdio"]
process = subprocess.Popen(
    pyls_cmd, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE
)
```

**To Reproduce**
 When setting PYRIGHT_PYTHON_FORCE_VERSION=1.1.271 , 
I am getting: 'No source files found.'

When setting  PYRIGHT_PYTHON_FORCE_VERSION=1.1.2710, 
I am getting 'Found 1638 source files'
(for a local clone of https://github.com/shuup/shuup ) 

(Attaching logs of LSP initialization for each run) 
[
[lsp-log1.1.271.txt](https://github.com/microsoft/pyright/files/9593939/lsp-log1.1.271.txt)
](url)

[lsp-log1.1.270.txt](https://github.com/microsoft/pyright/files/9593944/lsp-log1.1.270.txt)

**Expected behavior**

The same 1638 source files should be found also in version 1.1.271

**VS Code extension or command-line**

command line

erictraut:
@yuval-piiano, I'm not aware of any changes between 1.1.270 and 1.1.271 that would impact this.

Are you using the pyright LSP in conjunction with one of the major client plugins like [coc-pyright for vim](https://github.com/fannheyward/coc-pyright), [lsp-pyright for emacs](https://github.com/emacs-lsp/lsp-pyright), or [lsp-pyright for sublime](https://github.com/sublimelsp/LSP-pyright)? If so, I recommend that you work with the maintainer of the respective plugin. 

If you are using the lsp in some other context, you'll need to debug this yourself. As I said, I'm not aware of anything that would have changed the behavior, and it works fine for our primary supported scenarios including the command-line version and the LSP for VS Code.

yuval-piiano:
It's definitely a change in the versions, located the change in 
pyright/packages/pyright-internal/src/analyzer/service.ts, the function _getConfigOptions
Prior to 271, the code is: 

` if (commandLineOptions.fileSpecs.length > 0) {
            commandLineOptions.fileSpecs.forEach((fileSpec) => {
[...]

PR #3965, opened 2022-09-19
Use default includes/excludes when user has pyproject.toml

debonte:
Fix https://github.com/microsoft/pyright/issues/3962

Revert `service.ts` change that caused a regression where users with `pyproject.toml` files didn't get the default include and exclude file specs.


skilkis:
Thank you for this fix!

PR #3973, opened 2022-09-20
When logging is active, log the state of the TypeEvaluator

rchiodo:
This builds on my previous PR: https://github.com/microsoft/pyright/pull/3967

After a bunch of discussion with the team, one of the things that's hard to reason about is the state of the TypeEvaluator as the checker runs. 

This change logs the state of the TypeEvaluator before and after every call. 

This is the first step in a plan to write a tool that parses the log and shows diffs so you can track down where state went wrong.

The idea is based on [redux logging](https://github.com/LogRocket/redux-logger#readme) where tracking down changes to state is done one 'action' at a time.


packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -595,6 +595,42 @@ export function createTypeEvaluator(importLookup: ImportLookup, evaluatorOptions
     const returnTypeInferenceContextStack: ReturnTypeInferenceContext[] = [];
     let returnTypeInferenceTypeCache: TypeCache | undefined;
 
+    function getState(): object {
+        // Return the current state of the TypeEvaluator.
+        // Note this is not type constrained so that it might change in the future.
+        // It's intended to be used for logging.
+        return {
```
erictraut, line 602: Hmm. I'm not crazy about this. These are all internal to the type evaluator. About half of them (e.g. noneType through incompleteTypeCache) are effectively late-bound constants. Most of the others are caches which can grow very large. Almost none of these are interesting from a logging perspective.

rchiodo: What do you think would be useful to log? I started with all the local state in the `createTypeEvaluator`. A subset would work too.

erictraut: I'm not sure what your objective is with this change, but personally, I don't think any of this state is useful to log. As I said, it's mostly static (and therefore uninteresting to log) or represents caches that grow very large (and therefore would be completely inappropriate to log).

Is your intent to log information locally on your machine to aid in debugging? If so, I don't think any of this information will be useful. There are other logging facilities that are already present and designed for this purpose. They include the "printExpressionTypes" flag in typeEvaluator.ts and "isPrintControlFlowGraphEnabled" and "isPrintCallNoReturnEnabled" flags in codeFlowEngine.ts.

rchiodo: When discussing this yesterday with the rest of the team, we weren't sure what state would be useful in tracking down bugs (or understanding the typeevaluator), so I added it all.

rchiodo: We've been discussing how to better understand the TypeEvaluator as it's mostly voodoo magic to all of us. There was a discussion around how does the internal state change over time and when debugging it, it's hard to understand why or how it changed. That discussion reminded me of debugging redux enabled apps, and how logging state changes made debugging redux apps a lot simpler. 

Of course the 'state' logged has to make sense. Logging constants seems not useful, but maybe there's some aspect of the caches that should be logged. Like maybe if it's using the wrong cache, or the cache in use determines the logic for a subsequent call. 

The logging is already logging all function calls, this is adding the internal state each function call relies on.

rchiodo: I think we'd start with using this locally but could potentially ask customers to turn this flag on. And you're right, it logs a lot of data. Like 5000 lines for a hover.

erictraut: Yeah, I don't think this approach will accomplish your goal. I recommend enabling "printExpressionTypes" locally. Most of the complexity in the type evaluator is related to expressions, and this will give you a good idea of how that works.

rchiodo:
Closing in lieu of discussion around custom commands for logging

Issue #3974, opened 2022-09-21
1.1.172: False negative for calling positional-only argument by name 


rsokl:
```python
def f(x, /, *args, **kw): ...

f(x=1)   # pyright 1.1.271: error: Expected 1 more positional argument (reportGeneralTypeIssues)
         # pyright 1.1.272: no errors
```
<details><summary>pyright 1.1.271</summary>
<p>

```
Loading configuration file at C:\Users\rsokl\hydra-zen\pyrightconfig.json
Assuming Python version 3.9
Assuming Python platform Windows
stubPath C:\Users\rsokl\hydra-zen\typings is not a valid directory.
Searching for source files
Found 1 source file
pyright 1.1.271
C:\Users\rsokl\hydra-zen\scratch\scratch.py
  C:\Users\rsokl\hydra-zen\scratch\scratch.py:3:5 - error: Expected 1 more positional argument (reportGeneralTypeIssues)
1 error, 0 warnings, 0 informations
Completed in 0.799sec
```

</p>
</details>

<details><summary>pyright 1.1.272</summary>
<p>

```console
$ pyright scratch.py
Loading configuration file at C:\Users\rsokl\hydra-zen\pyrightconfig.json
Assuming Python version 3.9
Assuming Python platform Windows
stubPath C:\Users\rsokl\hydra-zen\typings is not a valid directory.
Searching for source files
Found 1 source file
pyright 1.1.272
0 errors, 0 warnings, 0 informations 
Completed in 0.796sec
```

</p>
</details>

I also noticed that the following:

```python
def f(x, /, *args): ...

f(x=1)
```
now raises two, contradictory errors:
```
\scratch.py:3:3 - error: No parameter named "x" (reportGeneralTypeIssues)
\scratch.py:3:1 - error: Argument missing for parameter "x" (reportGeneralTypeIssues)
```

erictraut:
@rsokl, out of curiosity, how did you detect a false negative so quickly after the 1.1.272 release? False negatives generally go undetected a long time because they don't show up as errors, so it takes a keen eye to spot them.

rsokl:
> @rsokl, out of curiosity, how did you detect a false negative so quickly after the 1.1.272 release? False negatives generally go undetected a long time because they don't show up as errors, so it takes a keen eye to spot them.

[...]

PR #3975, opened 2022-09-21
Fix argsRemainingCount calculation

debonte:
Possible fix for #3974. When calculating the number of remaining positional args (`argsRemainingCount`) I think we shouldn't subtract one for a `VarArgList` that is after `positionalOnlyLimitIndex`.

There's a problem with verifying the new tests. Each of the two scenarios should produce one error, but the bug caused them to produce zero and two errors respectively. So both with and without my fix, the total number of errors is 2. Since the validation logic checks the total number of errors in the sample file, it can't tell if the bug has occurred. Should I remove one of the test cases?


erictraut:
The change looks good to me. Thanks for the fix. As for the test, I think it's fine. It's unlikely that a regression would go uncaught.

erictraut (approved):

PR #4008, opened 2022-10-05
Added a global "cache manager"

erictraut:
Added a global "cache manager" to track heap usage for all type analyzer service instances. This prevents out-of-memory conditions when many roots are used in a multi-root workspace. This addresses bug https://github.com/microsoft/pylance-release/issues/3408#issuecomment-1266613951.


packages/pyright-internal/src/analyzer/cacheManager.ts
```diff
@@ -0,0 +1,88 @@
+/*
+ * cacheManager.ts
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ * Author: Eric Traut
+ *
+ * A singleton that track the size of caches and empty them
```
debonte, line 7: Nit: Grammar
```suggestion
 * A singleton that tracks the size of caches and empties them
```

packages/pyright-internal/src/languageServerBase.ts
```diff
@@ -632,20 +632,22 @@ export abstract class LanguageServerBase implements LanguageServerInterface {
...
             return;
         }
 
         this._connection.workspace.onDidChangeWorkspaceFolders((event) => {
-            event.removed.forEach((workspace) => {
-                const rootPath = this._uriParser.decodeTextDocumentUri(workspace.uri);
+            event.removed.forEach((workspaceInfo) => {
+                const rootPath = this._uriParser.decodeTextDocumentUri(workspaceInfo.uri);
+                const workspace = this._workspaceMap.getWorkspaceForFile(this, rootPath);
+                workspace.serviceInstance.dispose();
```
debonte, line 644: Can we call `workspace.serviceInstance.dispose()` in `WorkspaceMap.delete` instead?

If not, there are two places where `_workspaceMap.delete` is called in Pylance for notebook scenarios, and they'll also need to get updated to dispose the `serviceInstance`.

erictraut: Yes, good call. I made the change.

debonte (approved):

PR #4013, opened 2022-10-05
Don't treat classes decorated with `dataclass_transform` as dataclasses

debonte:
For https://github.com/microsoft/pylance-release/issues/3304

Attributes on classes decorated with `dataclass_transform` and their base classes should not be treated as dataclasses. That is, attributes on those classes are not fields.

Fixed by not calling `applyDataClassDefaultBehaviors` for these classes.

### DIFF
--- a/packages/pyright-internal/src/analyzer/typeEvaluator.ts
+++ b/packages/pyright-internal/src/analyzer/typeEvaluator.ts
@@ -15094,7 +15094,7 @@ export function createTypeEvaluator(importLookup: ImportLookup, evaluatorOptions
                 return (
                     isClass(mroClass) &&
                     mroClass.details.classDataClassTransform !== undefined &&
-                    mroClass !== classType
+                    !ClassType.isSameGenericClass(mroClass, classType)
                 );
             });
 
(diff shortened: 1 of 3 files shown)


packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -15091,7 +15091,11 @@ export function createTypeEvaluator(importLookup: ImportLookup, evaluatorOptions
             dataClassBehaviors = effectiveMetaclass.details.classDataClassTransform;
         } else {
             const baseClassDataTransform = classType.details.mro.find((mroClass) => {
-                return isClass(mroClass) && mroClass.details.classDataClassTransform !== undefined;
+                return (
+                    isClass(mroClass) &&
+                    mroClass.details.classDataClassTransform !== undefined &&
+                    mroClass !== classType
```
erictraut, line 15097: A direct comparison may not work here for generic classes. That's classes in the `mro` array are specialized (i.e. `ClassType.cloneForSpecialization` is called). For example, if `Foo[A]` derives from `list[B]` which derives from `MutableSequence[C]`, then the mro list will include `Foo[A]`, `list[A]`, and `MutableSequence[A]`. 

Rather than using a simple `!==` check here, it is better to use `!ClassType.isSameGenericClass(mroClass, classType)`. This will ignore any differences in specialization.

packages/pyright-internal/src/tests/samples/dataclassTransform2.py
```diff
@@ -23,6 +23,8 @@ def model_field(
     field_specifiers=(ModelField, model_field),
 )
 class ModelMeta(type):
+    not_a_field: str
```
erictraut, line 26: You might want to add a generic base class to the test to make sure it is properly handled.

erictraut (approved):

PR #4029, opened 2022-10-10
Update type-concepts.md

kwanghoon:
A fix to add a missing parameter to isinstance as:
 - isinstance(val, Bar)


erictraut:
Thanks for the fix.

PR #4041, opened 2022-10-12
Improved source mapper and made cache manager non static

heejaechang:
- improvement on source mapper to handle multiple pyi files for 1 py file case.
- made cache manager not static object but instance object which lifetime is tied to LS.


rchiodo:
sourceMapperTest.ts seems to be missing? Did I forget to do a squash merge so it will come later?

rchiodo:
Oh no nevermind. That's in pylance, not pyright.

rchiodo (approved):

bschnurr (approved):

PR #4141, opened 2022-11-05
Support __new__ for class constructor hover and signature help

bschnurr:
- fixes django stubs `CharField()`
- support overloads on dunder new/init
- changed function/method type prints that were preventing function names from being colored blue in vscode python textmate  due to an ":" being inserted after the function name. now function names also become blue.
 
image

ie.  `foo(..)` was being printed as `foo: (`
the code for adding ":" was ment for handling the property case only. no longer using `_getTypeText()` but instead using the underlining `printType()` call

https://github.com/microsoft/pylance-release/issues/3431

previous fix was a hack to just edit the django stubs to remove
`def __init__(self, *args: Any, **kwargs: Any) -> None: ...`
from the base class.

example of django-stubs's CharField's `__new__` signature 
image


packages/pyright-internal/src/languageService/hoverProvider.ts
```diff
@@ -348,7 +360,8 @@ export class HoverProvider {
         });
     }
 
-    private static _addInitMethodInsteadIfCallNode(
+    // logic similar to typeEvaluator.ts getCallSignatureInfo(...)
```
erictraut, line 363: Comments in the pyright code base should be complete sentences.

packages/pyright-internal/src/languageService/hoverProvider.ts
```diff
@@ -386,41 +399,81 @@ export class HoverProvider {
...
         }
 
-        const instanceType = evaluator.getType(callLeftNode.parent);
-        const functionType = evaluator.getTypeOfMember(initMethodMember);
+        let methodType: FunctionType | OverloadedFunctionType | undefined;
 
-        if (!instanceType || !functionType || !isClassInstance(instanceType) || !isFunction(functionType)) {
-            return false;
-        }
+        // Try to get the __init__ method first because it typically has
```
erictraut, line 409: What's up with the formatting of this comment?

packages/pyright-internal/src/languageService/hoverProvider.ts
```diff
@@ -386,41 +399,81 @@ export class HoverProvider {
...
+            const functionType = evaluator.getTypeOfMember(initMember);
 
-        if (!initMethodType || !isFunction(initMethodType)) {
-            return false;
+            if (isFunction(functionType) || isOverloadedFunction(functionType)) {
+                methodType = evaluator.bindFunctionToClassOrObject(instanceType, functionType);
+            }
+        }
+
+        // If there was no `__init__`, exluding `object` class `__init__`,
```
erictraut, line 421: Please format comment so its appearance isn't so sloppy.

packages/pyright-internal/src/languageService/hoverProvider.ts
```diff
@@ -386,41 +399,81 @@ export class HoverProvider {
...
+                isFunction(methodType) &&
+                (FunctionType.hasDefaultParameters(methodType) || methodType.details.parameters.length === 0))
+        ) {
+            const newMember = lookUpClassMember(classType, '__new__', ClassMemberLookupFlags.SkipInstanceVariables);
+
+            if (newMember) {
+                const functionType = evaluator.getTypeOfMember(newMember);
+
+                if (isFunction(functionType) || isOverloadedFunction(functionType)) {
+                    // set treatConstructorAsClassMember to so that the `__new__` methodType wont include `cls` as a parameter
```
erictraut, line 436: Comments in pyright code base should be complete sentences.

packages/pyright-internal/src/languageService/hoverProvider.ts
```diff
@@ -386,41 +399,81 @@ export class HoverProvider {
...
+                    methodType = evaluator.bindFunctionToClassOrObject(
+                        instanceType,
+                        functionType,
+                        /*memberClass*/ undefined,
```
erictraut, line 440: Include spaces around parameter names for consistency with reset of code base.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -2135,7 +2135,12 @@ export function createTypeEvaluator(importLookup: ImportLookup, evaluatorOptions
                         // If there was no `__init__` or the only `__init__` that was found
                         // was form the `object` class, see if we can find a better `__new__`
                         // method.
-                        if (!methodType || isObjectInit || isSkipConstructor) {
+                        if (
+                            !methodType ||
+                            isObjectInit ||
+                            isSkipConstructor ||
+                            (methodType && isFunction(methodType) && (FunctionType.hasDefaultParameters(methodType) || methodType.details.parameters.length === 0))
```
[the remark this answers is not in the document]
bschnurr: it seems this function `function getCallSignatureInfo()` is only called from completionProvider and signatureHelpProvider.  I was wondering if we should take it out of the typeEvaluator code?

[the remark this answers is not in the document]
erictraut: Ah, I didn't notice that this was in `getCallSignatureInfo`. Then it shouldn't affect any type evaluations, and I'm less concerned about the change.

When I originally wrote `getCallSignatureInfo`, it needed to be part of the typeEvaluator because it relied on functions that were not exported from the typeEvaluator. We've since exported a lot more functions, so it might be possible to implement it outside of the typeEvaluator. Not sure if it's worth moving though.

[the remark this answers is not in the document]
bschnurr: the logic is saying if __init__ is the following:

`def __init__(*args, **kwargs)`
`def __init__()`

then check for __new_  for a better constructor signature.  this was the case for django stubs, where they had a detailed `__new__` in a child class, but defaultParams for `__init__` in a parent base class.

there is also a way to skip the `isObjectInit` check. currently we call 

`methodType = getBoundMethod(subtype, '__init__');` above
and inside that it doesn't exclude methods from Object, when it makes a call to `lookUpClassMember`

erictraut:
@bschnurr, is this PR still relevant? Looks like it hasn't received any attention for more than two weeks.

bschnurr:
> @bschnurr, is this PR still relevant? Looks like it hasn't received any attention for more than two weeks.

I still want to check it in.. but I want to revert the part where I moved the word 'class' to its own line in the hover to give the signature more horizontal spacing. In the future I'm going to add line breaks to functions/constructors with long lists of parameters instead.

bschnurr:
done

heejaechang:
this fixes https://github.com/microsoft/pylance-release/issues/3549

PR #4158, opened 2022-11-08
[WIP] splitting checker to multiple visitor + walker

heejaechang:
Still working in progress. Just to show the approach I am taking.

The purpose of the refactoring is to make each concern to be isolated to each appropriate component. For example, complexity check should be done by only the walker, or each diagnostic check shouldn't be affected by other checks.

my goal (hope) is splitting out each check to each file that should be independent to other checks.

things need to be done

- [x] make sure the walker respects return value of the visitor (that controls whether its sub nodes should be visited or not) or find a way things to work without it.
- [ ] make sure the refactoring didn't break existing behavior
- [ ] perf test to make sure refactored checker doesn't cause perf issue

### DIFF
--- a/packages/pyright-internal/src/analyzer/checker.ts
+++ b/packages/pyright-internal/src/analyzer/checker.ts
@@ -41,7 +41,6 @@ import {
     FormatStringNode,
     ForNode,
     FunctionNode,
-    GlobalNode,
     IfNode,
     ImportAsNode,
     ImportFromAsNode,
@@ -57,11 +56,9 @@ import {
     ModuleNameNode,
     ModuleNode,
     NameNode,
-    NonlocalNode,
     ParameterCategory,
     ParameterNode,
     ParseNode,
-    ParseNodeArray,
     ParseNodeType,
     PatternClassNode,
     RaiseNode,
@@ -78,7 +75,6 @@ import {
     TupleNode,
     TypeAnnotationNode,
     TypeParameterListNode,
-    TypeParameterNode,
     UnaryOperationNode,
     UnpackNode,
     WhileNode,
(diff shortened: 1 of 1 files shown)


packages/pyright-internal/src/analyzer/checker.ts
```diff
@@ -233,223 +233,441 @@ const deprecatedSpecialForms = new Map<string, DeprecatedForm>([
...
+    }
+}
+
+class MainCheckerVisitor extends BaseCheckVisitor {
+    constructor(private _evaluator: TypeEvaluator, private _fileInfo: AnalyzerFileInfo) {
+        super();
+    }
+
+    override visitStatementList(node: StatementListNode) {
+        node.statements.forEach((statement) => {
```
heejaechang, line 530: this actually does many different checks but I left them together since it doesn't require any states (member field)

erictraut: I'm not liking this change.

IMO, it makes it more difficult to determine what's going on. I'm worried about introducing bugs where the code assumes that a "false" result terminates further walking; this code continues walking if one of the other walkers happens to return true.

This approach also requires multiple dispatches through a large case statement for every node, which will have performance impact.

I'd prefer not to continue down this path and stick with the current design with a unified walker.

heejaechang: > t makes it more difficult to determine what's going on. I'm worried about introducing bugs where the code assumes that a "false" result terminates further walking; this code continues walking if one of the other walkers happens to return true.

that's because I didn't finish that part as the comment says "put proper code here". and I wrote in the remaining tasks to  "make sure the walker respects return value of the visitor"

and I just put code to respect the return value of the visitor.

> This approach also requires multiple dispatches through a large case statement for every node, which will have performance impact.

the perf impact might be neglectable compared to the heavy work of checker actually doing for each node.

> I'd prefer not to continue down this path and stick with the current design with a unified walker.

that's fine but while I am doing this refactoring one thing I felt was it is very hard to refactor because this one walker is doing a lot of stuff and it controls what sub tree must be visited or not for some checks and it is hard to know what those checks are. making it hard to maintain.

erictraut: I don't find the current code hard to maintain. I would find it difficult to maintain if you were to split it up as you have done in this draft PR, so please don't do that.

PR #4164, opened 2022-11-08
"Expression value is unused" should not be reported for last expression in notebook cell

debonte:
Address https://github.com/microsoft/pylance-release/issues/3282

Don't report an error for an unused expression at the end of a notebook cell, as it is treated as the cell's value.

The tests are checking the diagnostic count which we would normally do by testing a sample `.py` file, but I don't believe we have a way to enable `ipythonMode` with that approach. If you'd prefer the tests to work that way, I can look into making that work.

### DIFF
--- a/packages/pyright-internal/src/analyzer/checker.ts
+++ b/packages/pyright-internal/src/analyzer/checker.ts
@@ -1538,8 +1538,9 @@ export class Checker extends ParseTreeWalker {
             reportAsUnused &&
             this._fileInfo.ipythonMode === IPythonMode.CellDocs &&
             node.parent?.nodeType === ParseNodeType.StatementList &&
+            node.parent.statements[node.parent.statements.length - 1] === node &&
             node.parent.parent?.nodeType === ParseNodeType.Module &&
-            node.parent.statements[node.parent.statements.length - 1] === node
+            node.parent.parent.statements[node.parent.parent.statements.length - 1] === node.parent
         ) {
             // Exclude an expression at the end of a notebook cell, as that is treated as
             // the cell's value.
--- a/packages/pyright-internal/src/analyzer/checker.ts
+++ b/packages/pyright-internal/src/analyzer/checker.ts
@@ -1534,6 +1534,18 @@ export class Checker extends ParseTreeWalker {
             }
         }
 
+        if (
+            reportAsUnused &&
+            this._fileInfo.ipythonMode === IPythonMode.CellDocs &&
+            node.parent?.nodeType === ParseNodeType.StatementList &&
+            node.parent.parent?.nodeType === ParseNodeType.Module &&
+            node.parent.statements[node.parent.statements.length - 1] === node
+        ) {
+            // Exclude an expression at the end of a notebook cell, as that is treated as
+            // the cell's value.
+            reportAsUnused = false;
+        }
+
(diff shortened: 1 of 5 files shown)


packages/pyright-internal/src/analyzer/checker.ts
```diff
@@ -1534,6 +1534,18 @@ export class Checker extends ParseTreeWalker {
...
+        if (
+            reportAsUnused &&
+            this._fileInfo.ipythonMode === IPythonMode.CellDocs &&
+            node.parent?.nodeType === ParseNodeType.StatementList &&
+            node.parent.parent?.nodeType === ParseNodeType.Module &&
+            node.parent.statements[node.parent.statements.length - 1] === node
+        ) {
+            // Exclude an expression at the end of a notebook cell, as that is treated as
+            // the cell's value.
+            reportAsUnused = false;
```
rchiodo, line 1546: Does this work for sub expressions too? Like

```
df['MyColumn']
```

packages/pyright-internal/src/analyzer/checker.ts
```diff
@@ -1534,6 +1534,18 @@ export class Checker extends ParseTreeWalker {
             }
         }
 
+        if (
+            reportAsUnused &&
+            this._fileInfo.ipythonMode === IPythonMode.CellDocs &&
+            node.parent?.nodeType === ParseNodeType.StatementList &&
+            node.parent.parent?.nodeType === ParseNodeType.Module &&
+            node.parent.statements[node.parent.statements.length - 1] === node
```
debonte, line 1542: Looks like there's something wrong with this check, since the following generates zero errors. I would expect the first line to still get an error.

```python
4
4
```

packages/pyright-internal/src/analyzer/checker.ts
debonte: It seems that this isn't considered an unused expression, so my change has no impact on it. I'm not sure why it isn't considered unused though.

packages/pyright-internal/src/analyzer/checker.ts
debonte: Needed to check that the expression's parent `StatementList` is the last statement in the module.

rchiodo (approved):

packages/pyright-internal/src/analyzer/checker.ts
debonte: `_reportUnusedExpression` only creates diagnostics for what it considers to be `simpleExpressionTypes` and non-comprehension collections. Index expressions are not currently considered "simple". However, if they were, this change would work for your example.

packages/pyright-internal/src/tests/ipythonMode.test.ts
```diff
@@ -560,3 +560,48 @@ function findCommentByOffset(tokens: TextRangeCollection<Token>, offset: number)
...
+// @filename: test.py
+// @ipythonMode: true
+//// if True:
+////     4[|/*marker*/|]
+    `;
+
+    verifyAnalysisDiagnosticCount(code, 1);
+});
+
+function verifyAnalysisDiagnosticCount(code: string, expectedCount: number) {
```
heejaechang, line 597: would it be too much if we actually check the error kind rather than just count?

PR #4165, opened 2022-11-08
Merged missing source import check to the existing walker.

heejaechang:


packages/pyright-internal/src/analyzer/checker.ts
```diff
@@ -256,6 +255,21 @@ export class Checker extends ParseTreeWalker {
...
+        // type evaluation to happen and generate diagnostics if there are issues.
+        // Since we are lazy evaluator, unless someone calls type eval, no diagnostic
+        // will be generated. This walker basically walks entire tree and force evaluation.
+        // Second is using the result of the evaluation, look for issues that require larger
+        // context such as whether overloads have consistent signature. or diagnostics that
+        // doesn't require internal state of type eval to generate.
+
+        // This walker bypasses visiting ModuleNode, so checking module level complexity
+        // directly here.
+        if (!this._checkModuleComplexity()) {
```
heejaechang, line 269: I am not sure why the walker bypass visitModule and have check method that directly calls statements and etc. this all seems could be done in visitModule method.

packages/pyright-internal/src/analyzer/checker.ts
```diff
@@ -745,6 +762,8 @@ export class Checker extends ParseTreeWalker {
 
         this._scopedNodes.push(node);
 
+        // It visits all child nodes, but it changes process ordering from pre-order to post-order
+        // by doing its own walking.
         return false;
```
heejaechang, line 767: there are a few visit methods that return false and does its own walking. I know some of them do that to avoid visit name but some of them, not sure why they are doing it. I tried change a few to true and see what tests are failing, but didn't try every cases. 

at some point, it would be nice if we can write it down the reason why it doesn't walk some of its child nodes. 

even if I find tests failing, and open the test files (since test itself just check error count but doesn't say what error), it just show the error without much explanation on why it fails with the walking child nodes.

erictraut:
@heejaechang, I think you're significantly overthinking (and overcomplicating) this. I've checked in a change that trivially merges the two tree walkers, so this PR is no longer needed.

PR #4266, opened 2022-12-02
Put back checking for type for module subnames

rchiodo:
Addresses https://github.com/microsoft/pylance-release/issues/3636

This shortcut added in https://github.com/microsoft/pyright/commit/b02711fd941beb5b85bcf1c754cf2d0ffd645927 to improve perf treats module subnames as not resolvable. 

This makes hover over different parts of an import statement inconsistent.


rchiodo:
Example:

- Before this change:
image


- After:
image

rchiodo:
Okay maybe I need something more nuanced. 678 failures after this small change.

rchiodo:
The 678 failures were the 'reportUnusedVariable' that was being incorrectly detected for the module subname case.

rchiodo:
Different idea - use the hover provider to find doc strings when hovering over import nodes. These nodes are not symbols so they shouldn't be parsed that way.

Issue #4269, opened 2022-12-02
reportShadowedImports is warning about a relative package import


joerick:
**Describe the bug**
reportShadowedImports is warning about a relative package import. For example- 

**To Reproduce**
Steps to reproduce the behavior.

**Expected behavior**

**Screenshots or Code**
If applicable, add screenshots or the text of the code (surrounded by triple back ticks) to help explain your problem.
```python
def foo(self) -> str:
    return 3
```

If your code relies on symbols that are imported from a third-party library, include the associated import statements and specify which versions of those libraries you have installed.

**VS Code extension or command-line**
Are you running pyright as a VS Code extension or a command-line tool? Which version? You can find the version of the VS Code extension by clicking on the Pyright icon in the extensions panel.

**Additional context**

rchiodo:
This should be relatively easy to fix for the import line (check if it's relative), but it also shows up in the typing.py file too:

![image](https://user-images.githubusercontent.com/19672699/205369303-0f3c0c2c-865d-4ea6-8a7d-5ac3d808a809.png)

The only way to turn that warning off (at the moment) is to disable all errors in that file or add this setting:

```jsonc
"python.analysis.diagnosticSeverityOverrides": {
    "reportShadowedImports": "none"
  },
```

joerick:
Thanks for looking at this. FWIW, I don't see the error in the typing.py file on my project-

<img width="1595" alt="image" src="https://user-images.githubusercontent.com/1244307/205494686-e7266386-f060-4a99-8a6a-6dab0d74134f.png">

erictraut:
This will be fixed in the next release.

joerick:
Thanks!

erictraut:
This is addressed in pyright 1.1.283, which I just published. It will also be included in a future version of pylance.

PR #4270, opened 2022-12-02
Allow relative imports for stdlib modules

rchiodo:
Addresses https://github.com/microsoft/pyright/issues/4269

Allow relative imports for stdlib modules without showing a warning

### DIFF
--- a/packages/pyright-internal/src/analyzer/checker.ts
+++ b/packages/pyright-internal/src/analyzer/checker.ts
@@ -3666,6 +3666,14 @@ export class Checker extends ParseTreeWalker {
         if (this._fileInfo.diagnosticRuleSet.reportShadowedImports === 'none') {
             return;
         }
+
+        // Skip this check for relative imports.
+        const nodeModule = node.nodeType === ParseNodeType.ImportFromAs ? (node.parent as any)?.module : node.module;
+        if (nodeModule?.importInfo?.isRelative) {
+            return;
+        }
+
+        // Otherwise use the name to determine if a match for a stdlib module.
         const namePartNodes =
             node.nodeType === ParseNodeType.ImportAs
                 ? node.module.nameParts
--- a/packages/pyright-internal/src/analyzer/checker.ts
+++ b/packages/pyright-internal/src/analyzer/checker.ts
@@ -3668,8 +3668,13 @@ export class Checker extends ParseTreeWalker {
         }
 
         // Skip this check for relative imports.
-        const nodeModule = node.nodeType === ParseNodeType.ImportFromAs ? (node.parent as any)?.module : node.module;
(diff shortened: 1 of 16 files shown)


packages/pyright-internal/src/analyzer/program.ts
```diff
@@ -240,6 +240,9 @@ export class Program {
                 const filePath = normalizePathCase(this._fs, oldFile.sourceFile.getFilePath());
                 if (!newFileMap.has(filePath)) {
                     oldFile.isTracked = false;
+                } else {
+                    // Otherwise make sure this file is marked as tracked.
```
rchiodo, line 244: Noticed this potential problem when trying to track down why none of the files in my workspace were tracked.

The root cause of my problem was that 'pyvenv.cfg' was in my root. Added this issue to maybe add a warning for this case:

https://github.com/microsoft/pylance-release/issues/3706

packages/pyright-internal/src/analyzer/checker.ts
```diff
@@ -3666,6 +3666,14 @@ export class Checker extends ParseTreeWalker {
         if (this._fileInfo.diagnosticRuleSet.reportShadowedImports === 'none') {
             return;
         }
+
+        // Skip this check for relative imports.
+        const nodeModule = node.nodeType === ParseNodeType.ImportFromAs ? (node.parent as any)?.module : node.module;
```
erictraut, line 3671: This isn't the right way to test for this. You should see if the parent of the ImportFromAsNode is an ImportFromNode. If so, then you can check whether the `module.leadingDots.length > 0` to see if it's a relative import. 

What you're doing here is directly accessing the analyzerNodeInfo without going through the proper accessor function (`getImportInfo`). There's no need to use that information here since the parser output contains the necessary information.

packages/pyright-internal/src/analyzer/checker.ts
```diff
@@ -3666,6 +3666,14 @@ export class Checker extends ParseTreeWalker {
...
+        if (nodeModule?.importInfo?.isRelative) {
```
erictraut, line 3672: See above. Please don't access importInfo directly without going through the accessor function. And there's no need to use this here; you should use the syntax information (`module.leadingDots.length > 0`) instead.

packages/pyright-internal/src/analyzer/program.ts
erictraut: I don't understand why this is necessary. This should be handled by the subsequent call to `addTrackedFiles`. In any case, this change is unrelated to the bug at hand, so it would be better to make as a separate commit.

erictraut (changes requested):

erictraut (approved):

PR #4271, opened 2022-12-03
[WIP] add a new 'compactSignatureDisplay' option for hover and completions.…

bschnurr:
…  Currently its defaulted to false so that we insert newline between function parameters, but in tests we enable compact mode so that we don't need to re-write every test when we want to change the display format.

This time i dont modify `printType` and restrict the formatting to a higher level in `hoverProvider` and `completionProvider`.

todo:

- [x]  add some actual tests around the new non-compact hover/completion layouts
- [ ]  HeeJae also suggested we re-rewrite `printType` to use a `DisplayStruct` under the hood, with a list of string and token kind, so that we can do proper formatting outside of printType.  similar to `getFuncitonParts` but something like `getTypeParts`

One important note.
this doesn't currently handle the case of a function or overloaded function being assigned to a variable, and then having the variables hover show nicely formatted overloads with parameters separated on an individual line.  Variables use 'printType()' and not `getFunctionParts()` so I dont have access to the separate parts.  I probably could get the variables alias type and see if its a function/overload. Will investigate

example
class constructors
image

functions
image

currently if there is only 1 parameter will wont separate out the parens to new lines.  Also note I considered 2 options for the closing parens and decided to place the closing paren on a new line, in order to give return type annotations more horizontal space.
image


packages/pyright-internal/src/tests/harness/fourslash/testLanguageService.ts
```diff
@@ -88,6 +88,8 @@ export class TestLanguageService implements LanguageServerInterface {
     constructor(workspace: WorkspaceServiceInstance, readonly console: ConsoleInterface, readonly fs: FileSystem) {
         this._workspace = workspace;
         this._uriParser = new UriParser(this.fs);
+        const defaultTestConfig = new ConfigOptions('.');
+        defaultTestConfig.compactSignatureDisplay = true;
```
bschnurr, line 92: todo: will remove. logic is in `_applyTestConfigOptions` to set `compactSignatureDisplay ` to `True`

packages/pyright-internal/src/languageService/completionProvider.ts
```diff
@@ -2669,10 +2674,20 @@ export class CompletionProvider {
...
+                                            typeDetail = getFunctionTooltip(
+                                                '',
+                                                name,
+                                                functionType,
+                                                this._evaluator,
+                                                this._configOptions.compactSignatureDisplay
+                                            );
                                         } else {
                                             typeDetail = name + this._evaluator.printType(functionType);
                                         }
```
bschnurr, line 2693: below this code we handle class constructor completions.....

its interesting that we dont actually try to show the true constructor and just assume no params `foo()`

```typescript
		 case DeclarationType.Class:
	         case DeclarationType.SpecialBuiltInClass: {
	             typeDetail = 'class ' + name + '()';
	             break;
           }
```

packages/pyright-internal/src/languageService/completionProvider.ts
```diff
@@ -2669,10 +2674,20 @@ export class CompletionProvider {
                                         } else if (isOverloadedFunction(functionType)) {
                                             // 35 is completion tooltip's default width size
                                             typeDetail = getOverloadedFunctionTooltip(
```
bschnurr, line 2676: we should probably truncate all the additional overloads when asking for a completion. lots of extra work being done.

rchiodo (approved):

heejaechang:
I think our current approach of printType actually returning hard coded format of string representation of the type is okay but more flexible way to do it would be it returning some structured data that other formatter can run over to create string format whatever way they want. like how printFunctionParts or how roslyn does - https://sourceroslyn.io/#Microsoft.CodeAnalysis.CSharp/SymbolDisplay/SymbolDisplay.cs,109

this kind of infrastructure will provide PTVS (pylance in VS) to use VS LSP extension which let one does much richer user experience such as colorization in `find all reference` window or `workspace symbol` window.

which works by returning encoded string of each part of type signature (ex, markdown between parts) rather than plain text such as one returned by printType.

erictraut:
@bschnurr, is this PR still valid, or should it be closed?

bschnurr:
I need to update it to fix the latest conflicts, sorry.

bschnurr:
going to create a new PR.

PR #4283, opened 2022-12-05
Show doc strings on module part hover

rchiodo:
Addresses https://github.com/microsoft/pylance-release/issues/3636

Use parsed source file to determine module doc strings on hover even when the module is not used.

### DIFF
--- a/packages/pyright-internal/src/analyzer/typeDocStringUtils.ts
+++ b/packages/pyright-internal/src/analyzer/typeDocStringUtils.ts
@@ -188,7 +188,7 @@ export function getModuleDocString(
     if (!docString) {
         if (resolvedDecl && isStubFile(resolvedDecl.path)) {
             const modules = sourceMapper.findModules(resolvedDecl.path);
-            docString = _getModuleNodeDocString(modules);
+            docString = getModuleNodeDocString(modules);
         }
     }
 
@@ -366,7 +366,7 @@ function _getFunctionOrClassDeclsDocString(decls: FunctionDeclaration[] | ClassD
     return undefined;
 }
 
-function _getModuleNodeDocString(modules: ModuleNode[]): string | undefined {
+export function getModuleNodeDocString(modules: ModuleNode[]): string | undefined {
     for (const module of modules) {
         if (module.statements) {
             const docString = ParseTreeUtils.getDocString(module.statements);
--- a/packages/pyright-internal/src/analyzer/typeDocStringUtils.ts
+++ b/packages/pyright-internal/src/analyzer/typeDocStringUtils.ts
@@ -179,6 +179,19 @@ export function getVariableInStubFileDocStrings(decl: VariableDeclaration, sourc
     return docStrings;
 }
 
+export function getModuleNodeDocString(modules: ModuleNode[]): string | undefined {
+    for (const module of modules) {
+        if (module.statements) {
+            const docString = ParseTreeUtils.getDocString(module.statements);
+            if (docString) {
+                return docString;
+            }
+        }
+    }
+
+    return undefined;
+}
+
 export function getModuleDocString(
     type: ModuleType,
     resolvedDecl: DeclarationBase | undefined,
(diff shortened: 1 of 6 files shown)


packages/pyright-internal/src/tests/fourslash/hover.docFromSrc.fourslash.ts
```diff
@@ -55,19 +55,19 @@
 //// def func2() -> bool: ...
```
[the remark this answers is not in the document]
heejaechang: ah, ignore, you already have one.

packages/pyright-internal/src/languageService/hoverProvider.ts
```diff
@@ -294,7 +302,7 @@ export class HoverProvider {
 
             case DeclarationType.Alias: {
                 this._addResultsPart(parts, '(module) ' + node.value, /* python */ true);
```
heejaechang, line 304: change _addMouduleDocString to _addModuleParts ? and share this._addResultsparts part as well?

rchiodo: Yeah good idea. No need to have dupe code.

packages/pyright-internal/src/languageService/hoverProvider.ts
```diff
@@ -142,6 +144,12 @@ export class HoverProvider {
             if (type !== undefined) {
                 this._tryAddPartsForTypedDictKey(format, sourceMapper, evaluator, node, type, results.parts);
             }
+        } else if (node.nodeType === ParseNodeType.ModuleName || node.nodeType === ParseNodeType.Module) {
```
heejaechang, line 147: hmmm.. shouldn't be inside of node.nodeType === ParseNodeType.Name ?

import A.B

B is Name node under moduleName?

packages/pyright-internal/src/languageService/hoverProvider.ts
```diff
@@ -142,6 +144,12 @@ export class HoverProvider {
...
+            const importInfo = getImportInfo(node);
+            if (importInfo) {
+                this._addResultsPart(results.parts, '(module) ' + importInfo.importName, /* python */ true);
```
heejaechang: also, we have been trying to match tooltip behavior between completion provider and hover, so it would be nice if we can share these with completion provider as well.

rchiodo: The completion provider should provide doc strings as I type?

Like this would show them on the right?

image

rchiodo: Can I enter another issue to do that? Seems like a bigger issue.

packages/pyright-internal/src/languageService/hoverProvider.ts
heejaechang: actually, I would put this inside of _addDocumentationPart

if getType return nothing, then, check whether it is name node under moduleName, and do what you are doing here.

rchiodo: Sorry I'm missing what you mean. `_addDocumentationPart` isn't called for this scenario. It is for the A.B scenario you specified above. 

Are you suggesting if the nodeType is ModuleName or Module, it should call _addDocumentationPart which would then call this code if it couldn't find a type?

rchiodo: I can't repro this anymore so not sure why I had to add this case. I'll add the changes you suggested into _addDocumentationPart.

packages/pyright-internal/src/analyzer/typeDocStringUtils.ts
```diff
@@ -366,7 +366,7 @@ function _getFunctionOrClassDeclsDocString(decls: FunctionDeclaration[] | ClassD
     return undefined;
 }
 
-function _getModuleNodeDocString(modules: ModuleNode[]): string | undefined {
+export function getModuleNodeDocString(modules: ModuleNode[]): string | undefined {
```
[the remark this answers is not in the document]
rchiodo: Will do

packages/pyright-internal/src/languageService/hoverProvider.ts
heejaechang: ya, let me open an issue. we do provide doc string for other cases, it looks like somehow we don't do it for completion.

PR #4284, opened 2022-12-05
fix type for multiprocessing.log_to_stderr(): allow int level

mikepii:
Fixes this complaint:

```
/foobar.py
  /foobar.py:44:23 - error: Argument of type "int" cannot be assigned to parameter "level" of type "str | None" in function "log_to_stderr"
    Type "int" cannot be assigned to type "str | None"
      "int" is incompatible with "str"
      Type cannot be assigned to type "None" (reportGeneralTypeIssues)
```

For this code:

```python
import logging
from multiprocessing import log_to_stderr

log_to_stderr(logging.DEBUG)
```

See also typeshed stubs MR: https://github.com/python/typeshed/pull/9334


rchiodo (approved):

erictraut:
Changes to typeshed stubs should not be made in the pyright repo. They need to be made in the [typeshed project](https://github.com/python/typeshed). Pyright pulls stubs directly from typeshed periodically. If we were to accept this PR, it would be overwritten the next time we sync. Please submit an issue and/or a PR to typeshed.

JelleZijlstra:
The typeshed PR has already been merged, so presumably this should get fixed in pyright the next time you update pyright's bundled typeshed.

erictraut:
Thanks. I just pulled the latest version of typeshed stubs, and this fix will be included in the next weekly release of pyright (tomorrow).

PR #4316, opened 2022-12-11
Ensure reportImportCycles diagnostic has rule set

parched:
I noticed this in a tool I wrote that does some comparison between the errors generated and the settings in `pyrightconfig.json`. Note, I haven't actually tested this, I made the change in the browser.


erictraut:
Thanks for the contribution. The code wasn't quite right, but the idea was sound. I made the check directly rather than fixing it on your PR branch.

Issue #4322, opened 2022-12-12
Exclude recursive glob fails to exclude files in hidden directories


smackesey:
**Describe the bug**

Given this filesystem layout:

```
.buildkite
|--- .tox
|--- src
```

And this pyright config:

```
[tool.pyright]

include = [
  ".buildkite"
]

exclude = [
  "**/.tox"
]
```

`.buildkite/.tox` still gets included.

**Expected behavior**

If the semantics of `**/.tox` are "exclude a `.tox` directory anywhere in the included source tree, then this should also exclude `.buildkite/.tox`.

smackesey:
I noticed this because file discovery started taking interminably long in a large repo as soon as I tried to write my own `exclude` (which included `**/.tox`), and I could see that pyright was finding errors from libs inside `.buildkite/.tox`.

Here's a simple repro that works on MacOS:

```
#!/usr/bin/env zsh

local projroot=$(mktemp -d /tmp/temp.XXXX)
pushd $projroot >/dev/null

cat > pyproject.toml <<EOF
[tool.pyright]

include = [
  ".buildkite"
]

exclude = [
  "**/.tox"
]
EOF

mkdir .buildkite
mkdir .buildkite/src
mkdir .buildkite/.tox
touch .buildkite/.tox/foo.py

pyright .

popd >/dev/null
rm -rf $projroot
```

Output:

```
No configuration file found.
pyproject.toml file found at /private/tmp/temp.RwUR.
Loading pyproject.toml file at /private/tmp/temp.RwUR/pyproject.toml
Assuming Python version 3.10
Assuming Python platform Darwin
stubPath /private/tmp/temp.RwUR/typings is not a valid directory.
Searching for source files
Found 1 source file
pyright 1.1.283
0 errors, 0 warnings, 0 informations
Completed in 0.403sec
```

Notice `Found 1 source file`.

erictraut:
@smackeysey, this turned out to be a bug in a reg-ex that failed to properly handle globs that have path elements that begin with a "." like in your example. @rchiodo diagnosed and fixed the problem. This will be addressed in the next release.

erictraut:
This is included in pyright 1.1.284, which I just published. It will also be included in a future release of pylance.

PR #4323, opened 2022-12-12
Fix exclusions/inclusions to work with '.' prefixed directories

rchiodo:
Addresses https://github.com/microsoft/pyright/issues/4322

Generated regex wasn't handling '.foo' as a directory.


erictraut (approved):
LGTM

PR #4354, opened 2022-12-18
fix typos in mypy-comparison

JelleZijlstra:


docs/mypy-comparison.md
```diff
@@ -397,7 +397,7 @@ Pyright honors class decorators. Mypy largely ignores them. See [this issue](htt
 
 Versions of Python prior to 3.5 did not have a dedicated syntax for supplying type annotations. Annotations therefore needed to be supplied using “type comments” of the form `# type: <annotation>`. Python 3.6 added the ability to supply type annotations for variables. 
 
-Mypy has full support for type comments. Pyright supports type comments only in locations where there is a way to provide an annotation using modern syntax. Pyright was written to assume Python 3.5 and newer, so support for older versions was not a priority.
+Mypy has full support for type comments. Pyright supports type comments only in locations where there is a way to provide an annotation using modern syntax. Pyright was written to assume Python 3.6 and newer, so support for older versions was not a priority.
```
JelleZijlstra, line 400: "3.5 and newer" includes 3.5 itself, which required use of type comments, so I think you meant "3.6 and newer" here.

erictraut: Pyright was written for 3.5 and newer. Python 3.5 included support for function type annotations. It was lacking support for variable type annotations.

JelleZijlstra: 3.5 didn't add any annotation syntax though, that was in 3.0. (3.5 added the typing module, but no syntactic support for it; there was a backport for 3.4 and newer.) So I think I should revert the change in this paragraph and change "3.5" to "3.0" in the previous paragraph instead.

erictraut: Ah yes. I thought the basic annotation syntax was added in 3.5 along with PEP 484. I didn't realize it was in 3.0 already.

JelleZijlstra: Updated the PR.

PR #4492, opened 2023-01-19
Changed hover provider to limit its display of overloaded functions w…

erictraut:
…hen hovering over a name that is part of a call expression. In such cases, the context of the call's arguments often provides sufficient context to select one (or in some cases, several) overloads.

### DIFF
--- a/packages/pyright-internal/src/languageService/hoverProvider.ts
+++ b/packages/pyright-internal/src/languageService/hoverProvider.ts
@@ -271,9 +271,11 @@ export class HoverProvider {
                     label = isProperty ? 'property' : 'method';
                 }
 
-                const type = evaluator.getType(node);
+                let type = evaluator.getType(node);
                 const sep = isProperty ? ': ' : '';
                 if (type) {
+                    type = this._limitOverloadBasedOnCall(node, evaluator, type);
+
                     if (isOverloadedFunction(type)) {
                         this._addResultsPart(
                             parts,
@@ -482,8 +484,27 @@ export class HoverProvider {
     }
 
     private static _getTypeText(node: NameNode, evaluator: TypeEvaluator, expandTypeAlias = false): string {
-        const type = evaluator.getType(node) || UnknownType.create();
+        let type = evaluator.getType(node) || UnknownType.create();
+        type = this._limitOverloadBasedOnCall(node, evaluator, type);
         return ': ' + evaluator.printType(type, { expandTypeAlias });
+    }
+
+    private static _limitOverloadBasedOnCall(node: NameNode, evaluator: TypeEvaluator, type: Type) {
+        // If it's an overloaded function, see if it's part of a call expression.
+        // If so, we may be able to eliminate some of the overloads based on
+        // the overload resolution.
+        if (isOverloadedFunction(type)) {
+            const callNode = ParseTreeUtils.getCallForName(node);
+            if (callNode) {
+                const callTypeResult = evaluator.getTypeResult(callNode);
+
+                if (callTypeResult?.overloadsUsedForCall && callTypeResult.overloadsUsedForCall.length > 0) {
+                    type = OverloadedFunctionType.create(callTypeResult.overloadsUsedForCall);
+                }
+            }
+        }
+
+        return type;
     }
 
     private static _addDocumentationPart(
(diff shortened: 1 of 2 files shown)


erictraut:
Without the change:
Screenshot 2023-01-18 at 7 18 51 PM

With the change:
Screenshot 2023-01-18 at 7 18 18 PM

packages/pyright-internal/src/languageService/hoverProvider.ts
```diff
@@ -482,10 +484,29 @@ export class HoverProvider {
     }
 
     private static _getTypeText(node: NameNode, evaluator: TypeEvaluator, expandTypeAlias = false): string {
-        const type = evaluator.getType(node) || UnknownType.create();
+        let type = evaluator.getType(node) || UnknownType.create();
+        type = this._limitOverloadBasedOnCall(node, evaluator, type);
         return ': ' + evaluator.printType(type, { expandTypeAlias });
     }
 
+    private static _limitOverloadBasedOnCall(node: NameNode, evaluator: TypeEvaluator, type: Type) {
```
heejaechang, line 492: one refactoring idea is adding getBestMatchType or something in TypeUtils, that basically does

```
function getBestMatchType(evaluator: TypeEvaluator, node: NameNode): string {
        let type = evaluator.getType(node) || UnknownType.create();
        if (if (isOverloadedFunction(type)) {
            const callNode = ParseTreeUtils.getCallForName(node);
            if (callNode) {
                const callTypeResult = evaluator.getTypeResult(callNode);

                if (callTypeResult?.overloadsUsedForCall && callTypeResult.overloadsUsedForCall.length > 0) {
                    type = OverloadedFunctionType.create(callTypeResult.overloadsUsedForCall);
                }
            }
        }
        
        return type;
    }
```
and replace evaluator.getType(node) to TypeUtils.getBestMatchType(evaluator, node) for all places we want this behavior.

erictraut: TypeUtils is a utility module that is consumed by the TypeEvaluator. TypeUtils therefore cannot consume the TypeEvaluator without creating an inversion. If we were to factor this out, it would need to go into the TypeEvaluator itself. I don't think there's a good enough reason to do that right now.

heejaechang: I didn't mean literally the TypeUtils file we have. it could be any other (or new file) util file. since I believe all others will basically need to do the same work if they do.

debonte (approved):

packages/pyright-internal/src/languageService/hoverProvider.ts
```diff
@@ -482,10 +484,29 @@ export class HoverProvider {
     }
 
     private static _getTypeText(node: NameNode, evaluator: TypeEvaluator, expandTypeAlias = false): string {
-        const type = evaluator.getType(node) || UnknownType.create();
+        let type = evaluator.getType(node) || UnknownType.create();
+        type = this._limitOverloadBasedOnCall(node, evaluator, type);
```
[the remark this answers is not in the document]
erictraut: It would apply to variables and type aliases that evaluate to an overloaded callable. It would never apply to parameters or type parameters, but it doesn't hurt to have it in this common routine.

packages/pyright-internal/src/languageService/hoverProvider.ts
```diff
@@ -271,9 +271,11 @@ export class HoverProvider {
                     label = isProperty ? 'property' : 'method';
                 }
 
-                const type = evaluator.getType(node);
+                let type = evaluator.getType(node);
                 const sep = isProperty ? ': ' : '';
                 if (type) {
+                    type = this._limitOverloadBasedOnCall(node, evaluator, type);
```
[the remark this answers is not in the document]
erictraut: Yeah, I was surprised too. Yes, it's probably something we should include in the test suite.

rchiodo (approved):

packages/pyright-internal/src/languageService/hoverProvider.ts
bschnurr: It would help with inlay hints on call function parameters if I could match the overloads

packages/pyright-internal/src/languageService/hoverProvider.ts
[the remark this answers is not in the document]
erictraut: I've updated the test suite to include a test for this case.

PR #4498, opened 2023-01-20
fix typo in indentationUtils.ts

eltociear:
accomodate -> accommodate


erictraut:
Thanks for pointing this out. There were a number of other misspellings in the same file, so I'm going to fix them all at the same time rather than incorporating your PR.

Issue #3088, opened 2022-02-20
Cannot bind Generic class with TypeVar'ed Protocol


twoertwein:
**Describe the bug**
I'm not sure whether the following is supposed to work (but it works in mypy): Pyright doesn't seem to bind a generic class that uses a TypeVar'ed Protocol.

**To Reproduce**
```py
from typing import (
    Generic,
    Protocol,
    TypeVar,
)

T = TypeVar("T")

class Orderable(Protocol):
    def __sub__(self, __other):
        ...

    def __add__(self, __other):
        ...

    def __le__(self: T, __other: T) -> bool:
        ...

    def __lt__(self: T, __other: T) -> bool:
        ...

    def __ge__(self: T, __other: T) -> bool:
        ...

    def __gt__(self: T, __other: T) -> bool:
        ...

OrderableT = TypeVar("OrderableT", bound=Orderable)

class Interval(Generic[OrderableT]):
    def __init__(self, left: OrderableT, right: OrderableT):
        ...

reveal_type(Interval(1, 2))
# mypy: Interval[builtins.int*]
# pyright: Interval[Unknown]
```

**Additional context**
pyright 1.1.222
xref https://github.com/pandas-dev/pandas/pull/44922

erictraut:
You cannot use a regular TypeVar to annotate `self` in this manner. You would need to bind it to `Orderable`.

```python
T = TypeVar("T", bound="Orderable")
```

Issue #3084, opened 2022-02-20
Autocomplete `Literal` in `==` expressions


not-my-profile:
Within arguments pyright provides autocompletion:

![2022-02-20-085343_510x198_scrot](https://user-images.githubusercontent.com/73739153/154833643-79ba75ca-cce7-4cf0-86fb-6e0ee6a9b0fa.png)

It would be nice if pyright could provide the same autocompletion within `==` expressions:

![2022-02-20-085503_334x301_scrot](https://user-images.githubusercontent.com/73739153/154833689-026ec651-39fa-4362-a03f-541e0f0303cf.png)

erictraut:
This isn't a core type checking feature, so I'm transferring the enhancement request to the pylance-release project for consideration.

Issue #3039, opened 2022-02-14
diagnosticSeverityOverrides is not overriding properly when there is also pyrightconfig.json


tonyzhang9090:
Note: if you are reporting a wrong signature of a function or a class in the standard library, then the typeshed tracker is better suited for this report: https://github.com/python/typeshed/issues.

**Describe the bug**
When the VS Code project root has a `pyrightconfig.json` (for configs like “include”), setting `python.analysis.diagnosticSeverityOverrides` in settings.json doesn’t work as expected.

I think it’s caused by diagnosticSeverityOverrides logic in [configOptions.ts](https://github.com/microsoft/pyright/blob/b740ddfae5e6e74fcc819a1c8a78f83bc2571309/packages/pyright-internal/src/common/configOptions.ts#L899-L901) is incorrectly override by the the pyrightconfig config file, even when the config file doesn’t contain any diagnostic settings (like “reportMissingImports”).

**To Reproduce**
Any pyrightconfig.json, just an example below:
```json
{
  "include": ["**/*.py"]
}
```

VS code settings.json
```json
{
    "python.analysis.diagnosticSeverityOverrides": { 
        "reportMissingImports" : "none",
    }
}
```

`reportMissingImports` lints are still reported in VS Code.

**Expected behavior**
`python.analysis.diagnosticSeverityOverrides` should be working properly even when there is pyrightconfig.json in project root.

**VS Code extension or command-line**
VS Code version 1.64

erictraut:
Thanks for the bug report. I agree that the current behavior is wrong. This is a long-standing bug, a couple of years old. I'm surprised no one has reported it previously.

erictraut:
This is included in pyright 1.1.222, which I just published. It will also be included in the next release of pylance.

PR #4500, opened 2023-01-20
[PylanceBot] Pull Pylance with Pyright 1.1.290

PylanceBot:
    1. Officepy (#3039)
    2. improve when to show signature help inside of format string (#3088)
    3. now, we add extra commit chars to completion item only if there is matching pretext exists. (#3084)


heejaechang:
this PR is to test some automation. don't check in.

PR #4522, opened 2023-01-24
fix: empty range for submodule fallback

### LINKED ISSUE #4521: Invalid range when adding declaration for modules imported using `from ..a import b` syntax
**Describe the bug**

When adding declaration to symbols, there's at least one case where the resolved declaration (which is obtained via the submodule fallback) is of type `Alias` and it refers to a module, still the range is populated with the symbol data, causing clients to try to navigate to inexistent locations.

**To Reproduce**
Use the following project structure:
```
.
├── pyright_bug
│   ├── __init__.py
│   └── main.py
└── pyright_bug_2
    ├── __init__.py
    └── bug.py
```

```python
#  pyright_bug/main.py
from ..pyright_bug_2 import bug #  <- this symbol's declaration resolves to the right path but an inexistent range (the symbol's own range)
```

```python
# pyright_bug_2/bug.py
BUG="HERE"
```
**Expected behavior**
Declarations of modules should resolve with an empty range.

andrestone:
When adding declaration to symbols, there's at least one case where the resolved declaration (which is obtained via the submodule fallback) is of type `Alias` and it refers to a module (not a node), hence the empty range should be used.

Without this fix, LSP clients can fail to jump to the location because although the file exists the ranges might not.

Fixes #4521

### DIFF
--- a/packages/pyright-internal/src/analyzer/binder.ts
+++ b/packages/pyright-internal/src/analyzer/binder.ts
@@ -1883,7 +1883,7 @@ export class Binder extends ParseTreeWalker {
                             node: importSymbolNode,
                             path: implicitImport.path,
                             loadSymbolsFromPath: true,
-                            range: convertTextRangeToRange(importSymbolNode, this._fileInfo.lines),
+                            range: getEmptyRange(),
                             usesLocalName: false,
                             moduleName: this._fileInfo.moduleName,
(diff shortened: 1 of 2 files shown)


andrestone:

erictraut:
Thanks for the contribution! Looks good to me.

PR #4529, opened 2023-01-25
Adding a new setting ("formatFunctionSignature" in pyrightconfig and …

bschnurr:
…"python.analysis.formatFunctionSignature" in vscode) which will insert a newline between parameters to give more display room to right and also moving the return type annotation to its own line.  Formatting is defaulted to off.

also affects signatures displayed in completions.

image

Note: indentation is different compared to my last PR. this time i only indent a constant 4 spaces on each newline. Previously I tried to align the open and close braces `()`, but found that sacrificed a lot of room on the righthand side of the display

related to issue https://github.com/microsoft/pylance-release/issues/3425

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ Pyright supports [configuration files](/docs/configuration.md) that provide gran
 * [PEP 673](https://www.python.org/dev/peps/pep-0673/) Self type
 * [PEP 675](https://www.python.org/dev/peps/pep-0675/) arbitrary literal strings
 * [PEP 681](https://www.python.org/dev/peps/pep-0681/) dataclass transform
-* [PEP 692](https://www.python.org/dev/peps/pep-0692/) (draft) TypedDict for kwargs typing
+* [PEP 692](https://www.python.org/dev/peps/pep-0692/) TypedDict for kwargs typing
 * [PEP 695](https://www.python.org/dev/peps/pep-0695/) (draft) Type parameter syntax
 * [PEP 696](https://www.python.org/dev/peps/pep-0696/) (draft) Type defaults for TypeVarLikes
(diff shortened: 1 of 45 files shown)


erictraut:
I like the new formatting look. I think this will be appreciated by pylance users based on the feedback we've received. I don't understand why we'd add a configuration option though. Why not just change the behavior to always use the better formatting?

If we find that a configuration option is needed, I recommend that we don't expose it as a boolean. We're likely to add more options in the future, so an enumerated option would allow for that extensibility. 

I'd still vote for not exposing a configuration option at this time. Or if you want to expose a private option just for experimentation, that's fine, but then let's not document it or expose it in the settings schema.

bschnurr:
Ya. I think you're right. I'm not sure why someone would prefer the previous view.

Also I fyi I found that adding the `def` keyword in front of function hovers really helps the coloring of signature, but I didnt' want to include that change because it affects all the tests.

erictraut:
One small change that I recommend to the formatting ... the first parameter should also be moved to its own line. That's the standard that has emerged and the format that the "black" auto-formatter enforces. It looks weird to have all of the parameters on their own line except for the first one.

erictraut:
Oh, I also recommend dedenting the final line (the one that includes the closed paren and the return type). That's also standard for the "black" auto-formatter. And I like your idea of adding `def` to improve textmate coloring, but that could be done as a separate PR.

rchiodo (approved):
+1 on just making this the default. Not sure anybody would want the old way of doing it.

Maybe at some point we might want different formatting options (like flake/black/autopep8) but this is way better than what we had.

bschnurr:
> Oh, I also recommend dedenting the final line (the one that includes the closed paren and the return type). That's also standard for the "black" auto-formatter. And I like your idea of adding `def` to improve textmate coloring, but that could be done as a separate PR.


final line like this?
image

PR #4599, opened 2023-02-08
Completions and signature help for PEP 692

rchiodo:
Addresses https://github.com/microsoft/pylance-release/issues/3868

Special case completions and signature help for kwargs that come from TypedDicts.


packages/pyright-internal/src/tests/fourslash/completions.params.fourslash.ts
```diff
@@ -14,11 +30,29 @@ await helper.verifyCompletion('excluded', 'markdown', {
     marker2: {
         completions: [{ label: 'param1', kind: undefined }],
     },
+    marker4: {
```
heejaechang, line 33: can we test `include` rather than exclude?

heejaechang: oops. there is one below. ignore me :)

packages/pyright-internal/src/languageService/completionProvider.ts
```diff
@@ -2497,13 +2497,22 @@ export class CompletionProvider {
         });
     }
 
-    private _addNamedParametersToMap(type: FunctionType, paramMap: Map<string, string>) {
+    private _addNamedParametersToMap(type: FunctionType, names: Set<string>) {
         type.details.parameters.forEach((param) => {
-            if (param.name && !param.isNameSynthesized) {
+            if (
+                param.category === ParameterCategory.VarArgDictionary &&
```
heejaechang, line 2503: this condition seems duplicated here and in printer. can we share?

packages/pyright-internal/src/languageService/signatureHelpProvider.ts
```diff
@@ -141,13 +143,28 @@ export class SignatureHelpProvider {
                 paramName = params[params.length - 1].name || '';
             }
 
+            if (paramName && !paramString.includes(paramName) && paramName === 'kwargs') {
+                // Param string may have been expanded from kwargs.
+                // Pull out the param name from the string.
+                const match = /^(\w+):/.exec(paramString);
```
heejaechang, line 149: if type printer can return some kind of structured data rather than plain string to avoid doing this kind of thing would be nice. but I guess that's something we would do in different PR.

heejaechang (approved):

packages/pyright-internal/src/languageService/completionProvider.ts
rchiodo: Yeah I think so. I'll do that.

packages/pyright-internal/src/analyzer/parameterUtils.ts
```diff
@@ -0,0 +1,20 @@
...
+ * Utility functions for parameters.
+ */
+
+import { ParameterCategory } from '../parser/parseNodes';
+import { ClassType, FunctionParameter, TypeCategory } from './types';
+
+export function isTypedKwargs(param: FunctionParameter): boolean {
+    return (
+        param.category === ParameterCategory.VarArgDictionary &&
+        param.type.category === TypeCategory.Class &&
```
erictraut, line 15: Use `isClassInstance(param.type)` instead. Otherwise this might be a `type[class]` (an instantiable class).

packages/pyright-internal/src/analyzer/parameterUtils.ts
```diff
@@ -0,0 +1,20 @@
...
+        !!param.type.isUnpacked &&
```
erictraut, line 16: Use `isUnpackedClass(param.type)` for consistency.

packages/pyright-internal/src/languageService/signatureHelpProvider.ts
```diff
@@ -141,13 +143,28 @@ export class SignatureHelpProvider {
                 paramName = params[params.length - 1].name || '';
             }
 
+            if (paramName && !paramString.includes(paramName) && paramName === 'kwargs') {
```
erictraut, line 146: Don't look for the name `kwargs`. That's a convention, but a `**` parameter can have any legal identifier name. Instead, use `functionType.details.parameters[paramIndex].category === ParameterCategory.VarArgDictionary`. Also, `paramString.includes(paramName)` is not a safe test because the name of the parameter could be found in one of the TypedDict field names.

packages/pyright-internal/src/languageService/signatureHelpProvider.ts
erictraut: Yeah, this is pretty hack. I recommend finding a cleaner way to do this.

erictraut (changes requested):

packages/pyright-internal/src/languageService/completionProvider.ts
```diff
@@ -2497,13 +2498,16 @@ export class CompletionProvider {
         });
     }
 
-    private _addNamedParametersToMap(type: FunctionType, paramMap: Map<string, string>) {
+    private _addNamedParametersToMap(type: FunctionType, names: Set<string>) {
         type.details.parameters.forEach((param) => {
-            if (param.name && !param.isNameSynthesized) {
+            if (isTypedKwargs(param) && param.type.category === TypeCategory.Class) {
```
debonte, line 2503: Isn't `param.type.category === TypeCategory.Class` guaranteed if `isTypedKwargs(param)` since `isTypedKwargs` checks `isClassInstance`?

packages/pyright-internal/src/languageService/signatureHelpProvider.ts
```diff
@@ -141,13 +145,32 @@ export class SignatureHelpProvider {
                 paramName = params[params.length - 1].name || '';
             }
 
+            // If we have a typedKwargs, the param name will be wrong.
+            const kwargsIndex = paramIndex >= params.length ? params.length - 1 : paramIndex;
+            const kwargsParam = params[kwargsIndex];
+            if (
+                isTypedKwargs(kwargsParam) &&
+                isClassInstance(kwargsParam.type) &&
```
debonte, line 153: Similarly, I think this `isClassInstance` check is unnecessary?

packages/pyright-internal/src/languageService/signatureHelpProvider.ts
```diff
@@ -141,13 +145,32 @@ export class SignatureHelpProvider {
...
+
             parameters.push({
                 startOffset: label.length,
                 endOffset: label.length + paramString.length,
                 text: paramString,
                 documentation: extractParameterDocumentation(functionDocString || '', paramName),
             });
 
+            // Name match for active parameter as the set of parameters from the function
+            // may not match the actual string output from the typeEvaluator (kwargs for TypedDict is an example).
```
debonte, line 169: I'm not sure what this comment means. Maybe there's a typo / wrong word in there somewhere?

packages/pyright-internal/src/tests/fourslash/completions.params.fourslash.ts
```diff
@@ -1,10 +1,26 @@
...
+////
+//// def method2(param1: int, **kwargs: Unpack[Movie]):
+////     pass
+////
+//// method2(p/*marker4*/, k/*marker5*/)
+////
+//// def method3(param1: int, **kwargs: Any):
+////     pass
+////
+//// method3(p/*marker6*/, k/*marker7*/)
```
debonte, line 23: Looks like `marker6` isn't verified anywhere. Is that intentional?

packages/pyright-internal/src/languageService/completionProvider.ts
rchiodo: yes it is but typescript isn't smart enough to know that. Otherwise it shows a compiler error for using the details.

packages/pyright-internal/src/languageService/signatureHelpProvider.ts
rchiodo: Same reason for this one. Although maybe I can add a 'type is Class' to isTypedKwargs

packages/pyright-internal/src/languageService/signatureHelpProvider.ts
rchiodo: The old code would just compare the 'param' to the signature.activeParam. This doesn't work with the TypedDict stuff because there is no param that matches. 

So it has to do a match by 'name' instead.

rchiodo: Any suggestion for what the comment would be to explain that is appreciated. That's what the intent of the comment was.

packages/pyright-internal/src/languageService/signatureHelpProvider.ts
rchiodo: No that doesn't work because it's not passing the type into that function. So yeah this is necessary to get typescript to understand that the `.type.details` member actually exists.

packages/pyright-internal/src/tests/fourslash/completions.params.fourslash.ts
```diff
@@ -1,10 +1,26 @@
...
 ////     pass
 ////
 //// met/*marker1*/hod   /*marker2*/ (    /*marker3*/      param2 = 'test')
+////
+//// def method2(param1: int, **kwargs: Unpack[Movie]):
+////     pass
+////
+//// method2(p/*marker4*/, k/*marker5*/)
+////
+//// def method3(param1: int, **kwargs: Any):
```
debonte, line 20: Need to import `Any`?

packages/pyright-internal/src/tests/fourslash/completions.params.fourslash.ts
rchiodo: I should add a verification for this one too.

packages/pyright-internal/src/tests/fourslash/completions.params.fourslash.ts
rchiodo: It's not necessary, but wouldn't hurt I suppose. It's just ignoring the diagnostics produced.

debonte: It's fine either way. I only noticed when I copied this code to play with it. Then I was confused about why some things were imported but others were not. I guess `TypedDict` and `Unpack` are required for the `TypeEvaluator` to understand what's going on?

packages/pyright-internal/src/languageService/signatureHelpProvider.ts
rchiodo: Will change. Thanks for the suggestion.

packages/pyright-internal/src/tests/fourslash/completions.params.fourslash.ts
rchiodo: I think so. I was just copying other tests and forgot to include 'Any' in the import list. But the test passed without that part.

rchiodo:
@erictraut are you still requesting any changes?

Issue #4368, opened 2022-12-22
`reportUndefinedVariable` on multiple assignment


tushar-deepsource:
**To Reproduce**

The code: `asd.py`

```python3
z = [10]
a = z[a] = 0
```

```console
$ pyright asd.py
pyright asd.py
No configuration file found.
No pyproject.toml file found.
stubPath /Users/tusharsadhwani/code/marvin-python/typings is not a valid directory.
Assuming Python platform Darwin
Searching for source files
Found 1 source file
pyright 1.1.285
/Users/tusharsadhwani/code/marvin-python/asd.py
  /Users/tusharsadhwani/code/marvin-python/asd.py:2:7 - error: "a" is not defined (reportUndefinedVariable)
1 error, 0 warnings, 0 informations 
Completed in 0.559sec
```

erictraut:
Thanks for the bug report. This will be addressed in the next release.

erictraut:
This is included in pyright 1.1.286, which I just published. It will also be included in a future version of pylance.

PR #4607, opened 2023-02-10
Use binary search in findNodeByOffset if we can.

heejaechang:
I made the `findNodeByOffset` function to use `binarySearch` if possible, reducing the cost of `findNodeByOffset` to depend on the tree depth rather than both width + depth.

previously, if we want to find all `identifier` from a file, we will need to use parse tree walker since using token stream to find all `identifier` and use `findNodeByOffset` to find `name node` would be too costly since `findNodeByOffset` will always visit all siblings left to right for all matching sub tree until it finds the leaf node that match. and that can lead to n^2 cost compared to walker whose cost is linear to number of nodes.

Since parse tree doesn't guarantee increasing order between children's nodes (textual order), we might want to use tokenizer when we want the textual order to be intact. and this will let us use another approach when needed.

now, in practice, cost of repeatedly searching node using `findNodeByOffset` will remain almost constant. 

...

I wrote simple code to compare perf. what it does is, among 2k statements, finding the last statement 1000 times.

old `findNodeByOffset` took 20ms, new one took 1ms. making the size of code or repetition larger made old code number go up but new one stayed flat. for example, making 10K statements, made old code to take over 100ms, but new code stay at 2 ms.


packages/pyright-internal/src/analyzer/parseTreeUtils.ts
```diff
@@ -96,25 +96,46 @@ export function findNodeByOffset(node: ParseNode, offset: number): ParseNode | u
 
     // The range is found within this node. See if we can localize it
     // further by checking its children.
-    const children = getChildNodes(node);
+    let children = getChildNodes(node);
+    if (isCompliantWithNodeRangeRules(node) && children.length > 20) {
```
heejaechang, line 100: I asked chatGPT generally when binary search would be faster than linear search and answer was generally 20 elements, so choose. 20 :)

packages/pyright-internal/src/analyzer/parseTreeUtils.ts
```diff
@@ -96,25 +96,46 @@ export function findNodeByOffset(node: ParseNode, offset: number): ParseNode | u
...
+        // use Binary search to find child to visit. This should help
```
erictraut, line 101: Nit: "Use" and "Otherwise" should be capitalized because they start a sentence. "Binary" should not. "sibling" should be singular.

packages/pyright-internal/src/common/textRangeCollection.ts
```diff
@@ -96,27 +96,72 @@ export class TextRangeCollection<T extends TextRange> {
...
+    position: number,
+    min: number,
+    max: number
+): T | undefined {
+    const item = arr[position];
+    if (item) {
+        return item;
+    }
+
+    // Search forward and backward until it finds non null value
```
erictraut, line 151: Nit: Add period at end of sentence. "non-null" should have a hyphen.

packages/pyright-internal/src/analyzer/parseTreeUtils.ts
```diff
@@ -96,25 +96,46 @@ export function findNodeByOffset(node: ParseNode, offset: number): ParseNode | u
 
     // The range is found within this node. See if we can localize it
     // further by checking its children.
-    const children = getChildNodes(node);
+    let children = getChildNodes(node);
```
erictraut, line 99: Rather than jumping through all of these hoops to decide whether the range rules should be followed, would it make sense to push this to the layer below and have getChildNodes do the sorting — and perhaps even filtering out of null children? We could create a variant called getSorderChildNodes or something like that. It would simplify this logic significantly.

erictraut (commented):
I like the idea of using a binary search, but I think we might be able to simplify this a bit by pushing the sorting of child nodes down to a lower level. That way, this code could always use a binary search.

packages/pyright-internal/src/analyzer/parseTreeUtils.ts
heejaechang: I can make getChildNode not to return null value, but having node in sorted (and not overlapping) is not possible. for example, this change - https://github.com/microsoft/pyright/issues/4368 - made parse tree to look like this

```python
left = middle = right = 1

[112419] 'c:\pyTemp\junk\test.py:1:1' => Module (0,0)-(4,17) =>
  [112420] 'c:\pyTemp\junk\test.py:1:1' => StatementList (0,0)-(0,25) =>
    [112427] 'c:\pyTemp\junk\test.py:1:1' => Assignment (0,0)-(0,25) =>
      [112423] 'c:\pyTemp\junk\test.py:1:17' => Name (0,16)-(0,21) => 'c:\pyTemp\junk\test.py:1:17' (Identifier, (0,16)-(0,21)) {"start":16,"length":5,"type":7,"value":"right"} right
      [112426] 'c:\pyTemp\junk\test.py:1:1' => Assignment (0,0)-(0,25) =>
        [112422] 'c:\pyTemp\junk\test.py:1:8' => Name (0,7)-(0,13) => 'c:\pyTemp\junk\test.py:1:8' (Identifier, (0,7)-(0,13)) {"start":7,"length":6,"type":7,"value":"middle"} middle
        [112425] 'c:\pyTemp\junk\test.py:1:1' => Assignment (0,0)-(0,25) =>
          [112421] 'c:\pyTemp\junk\test.py:1:1' => Name (0,0)-(0,4) => 'c:\pyTemp\junk\test.py:1:1' (Identifier, (0,0)-(0,4)) {"start":0,"length":4,"type":7,"value":"left"} left
          [112424] 'c:\pyTemp\junk\test.py:1:25' => Number (0,24)-(0,25) => 1 int:(true) imaginary:(false)
```

basically, it made tree to be evaluation order not textual order and child's span is overlapping (start is same).

we could sort by both start and end location and iterate nodes until next sibling no longer contain the given position. but it might need to create more temp arrays and more if checks.

current code is to minimize those cases.

erictraut: OK, then I think your approach is fine.

heejaechang:
@erictraut are you okay me checking in?

erictraut:
Yes, I'm good. Thanks.

PR #4648, opened 2023-02-16
Idea to add unique identifiers to all Pyright diagnostics

rchiodo:
This is just an idea we discussed today:
https://github.com/microsoft/pylance-release/issues/3892#issuecomment-1433613560

With the goal being able to have docs for each specific type of message. Similar to how C# has docs like this example:
https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs0029

Having custom docs for each message allows:
- Searching on the message id. Like how searching for CS0029 brings up that example above.
- Allow for more information about each message that we can update independent from shipping
- Allow for translations (at least as much as a browser can)

This draft PR is just to get the idea in front of you @erictraut. This would be the first stage, just making sure we have a unique identifier for every message.

The second stage might be to add a new folder under [docs](https://github.com/microsoft/pyright/tree/main/docs) that would contain an `.md` for every identifier. 

3rd stage would be to start flushing out each error's description and give them examples. Might even use ChatGPT to generate the messages.


erictraut:
I think you're biting off a lot of work (both upfront and ongoing) for relatively little user benefit. I've maintained per-diagnostic error numbers in previous projects I've worked on, and it's a major pain to maintain. Also, don't underestimate the work involved in documenting all of these codes. My sense is that this isn't a good route to go down. Let's discuss more before you go further.

rchiodo:
Okay thanks. Hopefully we can still meet after today to discuss. 

I'm curious what the ongoing maintenance cost is. Well and if you have an alternative method that helps users.

rchiodo:
In this [file](https://github.com/microsoft/pyright/blob/7fc77ec1492a953ae980217f6b596416b7a70a36/packages/pyright-internal/src/localization/package.nls.en-us.json#L3) there are 672 different possible diagnostic messages.

Writing all of those might take a while but we could automate a script that generates the md files and just has the message in them. Filling in those that happen more often as we go.

erictraut:
If a diagnostic message is sufficiently unclear to be understood on its own, then we should consider ways to make them clearer. This would be more convenient to users than forcing them to go to another source for documentation and explanations.

rchiodo:
I wonder if we could contact the C# team and ask them their thoughts about the same thing.

rchiodo:
> If a diagnostic message is sufficiently unclear to be understood on its own, then we should consider ways to make them clearer. This would be more convenient to users than forcing them to go to another source for documentation and explanations.

I don't feel like this is always possible. At least in the UI that we have now. Perhaps we could show the extra information inside of VS code though? 

If you look at the C# error, it has a wall of text for a single error. Did they do that because they didn't find a way to explain things in the product? How come? What other things did they try? Maybe they would do things differently now.

Typescript also has error numbers but they don't document them like C# does. How come? Did they just let stack overflow document them? The internet has done a pretty good job of explaining them all. Could we do something like that? Just give the errors numbers and let the rest of the world document what to do about them?

I mean if we had error numbers, your responses to those errors would generate implicit documentation.

erictraut:
One idea to consider (not sure if I like it, but I figured I'd throw it out there) is to use the internal localization key (e.g. "overloadWithImplementation") rather than some randomly-assigned numeric value. We already maintain localization keys, guarantee that they're unique (one per diagnostic message), and they are semi-human-readable.

rchiodo:
I like that idea. Better than numbers anyway. More like C++ error messages.

erictraut:
I assume that we're not moving forward with this idea as prototyped, so I'm going to close this PR.

PR #4662, opened 2023-02-19
Support documentation hover for function parameters

ilanashapiro:
Closes https://github.com/microsoft/pylance-release/issues/2143

With docstring:
image

image

Without docstring:
image

image


ilanashapiro:

erictraut:
The approach you're using is a bit different than the one used in the other pending PR. I think the other approach (of caching the docstring during binding) is preferable and is more consistent with other docstring processing. I'm therefore going to close this PR.

Thanks for taking the time to submit the change.

Issue #2690, opened 2021-12-10
Pyright Attempts Triggering registerCapability handler despite dynamicRegistration set to false


holmanb:
**Describe the bug**
Pyright attempts [dynamic registration](https://microsoft.github.io/language-server-protocol/specification) despite the the fact that the client does not support it.

**To Reproduce**
[A minimal pyright configuration with neovim](https://github.com/neovim/nvim-lspconfig#quickstart) will reproduce. Requires (nvim > 0.5). Enable logging and check the logs.

**Expected behavior**
I expected to not see this warning.

**Screenshots or Code**
```
  4 [DEBUG][2021-12-09 21:30:06] .../vim/lsp/rpc.lua:496    "rpc.receive"   {  id = 0,  jsonrpc = "2.0",  method = "client/registerCapability",  params = {    registrations = { {        id = "b984e4cb-c66    3-453c-a46e-7eda50013708",        method = "workspace/didChangeWorkspaceFolders",        registerOptions = vim.empty_dict()      } }  }}
  3 [WARN][2021-12-09 21:30:06] ...lsp/handlers.lua:108     "The language server pyright triggers a registerCapability handler despite dynamicRegistration set to false. Report upstream, this warning is ha    rmless"
```

**VS Code extension or command-line**
Neither. Neovim 0.5.1 using native lsp.

**Additional context**

[Here is a minimal log containing all of the JSON RPC related to this issue.](https://github.com/microsoft/pyright/files/7690138/lsp.log) I am also using esm with flake8 as an LSP server, so there are traces of that in the log. Note `dynamicRegistration = false` on line 2.

erictraut:
I think pyright is doing the right thing here. It's registering for the "workspaceFolders" capability. Unlike most capabilities, this one doesn't have a "dynamicRegistration" indication. It's simply a boolean field indicating whether "workspaceFolders" is supported. NeoVim is indicating that it supports this capability ("workspaceFolders = true"). So I think the error message in the attached log is incorrect.

holmanb:
@erictraut Thanks for the response! It sounds like there is nothing to do here, so I will close this issue.

nschloe:
I'm debugging a similar issue and also found this error message. @holmanb Did you report this anywhere else?
In the message
```
[WARN][2022-04-06 09:38:58] ...lsp/handlers.lua:109     "The language server pyright triggers a registerCapability handler despite dynamicRegistration set to false. Report upstream, this warning is harmless"
```
[...]

Issue #1302, opened 2020-12-20
Pyright sending client/registerCapability even if dynamicRegistration set to false during initialization


mjlbach:
**Describe the bug**

We have support for pyright as one of the language servers for neovim's built in language client. One issue seems to be cropping up on node 15 where failed callbacks result in node terminating the process. The main callback that is failing is client/registerCapability, (our client doesn't support dynamic registration). When we report dynamicRegistration as false, pyright is still sending client/registerCapability requests

```
[ DEBUG ] 2020-12-20T13:55:08-0800 ] /home/michael/neovim/share/nvim/runtime/lua/vim/lsp/rpc.lua:489 ]	
"decoded"	{  id = 0,  jsonrpc = "2.0",  method = "client/registerCapability",  params = {    registrations = { {        
id = "807eedda-04cf-4637-9cb7-81d5f5d0e225",        method = "workspace/didChangeWorkspaceFolders",       
 registerOptions = vim.empty_dict()      } }  }}
```
We're initializing the server with all dynamicRegistration capabilities registered as false

<details open>
<summary>json of initialization params</summary>

```json
{
  "initialize_params": {
    "capabilities": {
      "callHierarchy": {
        "dynamicRegistration": false,
        "textDocument": {
          "codeAction": {
            "codeActionLiteralSupport": {
              "codeActionKind": {
                "valueSet": [
                  "",
                  "Empty",
                  "QuickFix",
                  "Refactor",
                  "RefactorExtract",
                  "RefactorInline",
                  "RefactorRewrite",
                  "Source",
                  "SourceOrganizeImports",
                  "quickfix",
                  "refactor",
                  "refactor.extract",
                  "refactor.inline",
                  "refactor.rewrite",
                  "source",
                  "source.organizeImports"
                ]
              }
            },
            "dynamicRegistration": false
          },
          "completion": {
            "completionItem": {
              "commitCharactersSupport": false,
              "deprecatedSupport": false,
              "documentationFormat": [
                "markdown",
                "plaintext"
              ],
              "preselectSupport": false,
              "snippetSupport": false
            },
            "completionItemKind": {
              "valueSet": [
                1,
                2,
                3,
                4,
                5,
                6,
                7,
                8,
                9,
                10,
                11,
                12,
                13,
                14,
                15,
                16,
                17,
                18,
                19,
                20,
                21,
                22,
                23,
                24,
                25
              ],
              "contextSupport": false,
              "dynamicRegistration": false
            },
[...]

PR #4666, opened 2023-02-20
Declare workspaceFolders support in server capabilities

the-mikedavis:
The `vscode-languageserver-node` dependency will attempt dynamic capability registration for workspace folder change notifications if this `workspace.workspaceFolders.changeNotifications` is not set in the `InitializeResult`. That is problematic for clients that don't support dynamic capability registration: if we send a JSONRPC error, the server will exit because of the rejected promise.

The server has support for workspace folders and requests and uses change notifications as far as I can see, so I believe it's more correct to declare support for these up-front anyways.

This is the recommended way to avoid dynamic capability registration for this case with `vscode-languageserver-node`: [(comment)](https://github.com/microsoft/vscode-languageserver-node/issues/713#issuecomment-754551117)

Fixes https://github.com/microsoft/pyright/issues/2690
Fixes https://github.com/microsoft/pyright/issues/1302
This and a change within Helix (https://github.com/helix-editor/helix/pull/6058) will resolve issues we started seeing in 1.1.293 (https://github.com/microsoft/pyright/discussions/4602)
See also https://github.com/microsoft/vscode-languageserver-node/commit/5bc82ac3c1c656aa70615c9e3e065fa04e41506e
Revelant part of the spec: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_workspaceFolders
[Here](https://github.com/microsoft/pyright/blob/d22850b041edbc747ec043553a23bdbff55696d6/packages/pyright-internal/src/languageServerBase.ts#L772-L786) is where the change notification handler is set up in pyright and what triggers dynamic capability registration within vscode-languageserver-node.


erictraut:

@heejaechang, I'm adding you as a reviewer also.

erictraut (approved):

erictraut:
@heejaechang, any thoughts on this PR? If I don't hear from you in the next day, I'll assume it's OK and merge it.

heejaechang (approved):

heejaechang:
I think it is right thing to do. we already listen for workspace folder change event and support workspace folders.

heejaechang:
It looks like when we register for workspace folder change event, LSP library automatically send registerCapabilities for workspace folder change support. and that's why it worked fine for vscode and we didn't notice.

PR #4676, opened 2023-02-23
Actions to publish to VS Code Marketplace

### LINKED ISSUE #4680: Automate publishing Pyright to VS Code Marketplace
Enable admins to publish to VS Code Marketplace without relying on their own personal credentials.

debonte:
Fixes #4680 

Adds two new GitHub actions based on Pylance's build and release actions.

I've tested both of them in my fork except for the final step of actually publishing. We can try that with the next release. The publishing PAT is already [in place](https://github.com/microsoft/pyright/settings/secrets/actions).

### build.yml
Triggered when a new `1.1.###` tag is pushed. Does the following:
1. Checks out that tag.
2. Installs node and runs `npm run install:all`
3. Builds a new VSIX with `npm run package`.
4. Saves that VSIX as a build artifact.
5. Creates a new release named "Published `<tag name>`". That naming matches the [existing releases](https://github.com/microsoft/pyright/releases).
6. Saves the VSIX as a release asset.

### release.yml
Triggered when someone edits the release and pushes the "Publish Release" button. Does the following:
1. Downloads the VSIX asset from the release.
2. Publishes it to the VS Code Marketplace via `vsce`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -50,7 +50,7 @@ jobs:
           path: packages/vscode-pyright/${{ env.VSIX_NAME }}
 
   create_release:
-    if: github.repository == 'microsoft/pyright' && startsWith(github.ref, 'refs/tags/')
+    if: github.repository == 'microsoft/pyright'
     runs-on: ubuntu-latest
     name: Create release
     needs: [build]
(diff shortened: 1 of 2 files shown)


erictraut (approved):
LGTM. I don't have much experience with this automation, so it would be good for others to review this as well.

heejaechang:
tagging @rchiodo and @bschnurr 

it looks good to me as well. 

one thing, some of github action shows warning to move to newer version of actions?

rchiodo (approved):

PR #4681, opened 2023-02-23
Change logic for import sorting

rchiodo:
For https://github.com/microsoft/pylance-release/issues/3429

Use new field priority instead of importtype for sorting

### DIFF
--- a/packages/pyright-internal/src/analyzer/importResolver.ts
+++ b/packages/pyright-internal/src/analyzer/importResolver.ts
@@ -43,7 +43,7 @@ import { equateStringsCaseInsensitive } from '../common/stringUtils';
 import * as StringUtils from '../common/stringUtils';
 import { isIdentifierChar, isIdentifierStartChar } from '../parser/characters';
 import { SupportPartialStubs } from '../pyrightFileSystem';
-import { ImplicitImport, ImportResult, ImportType } from './importResult';
+import { DefaultPriority, ImplicitImport, ImportResult, ImportType } from './importResult';
 import { getDirectoryLeadingDotsPointsTo } from './importStatementUtils';
 import { ImportPath, ParentDirectoryCache } from './parentDirectoryCache';
 import * as PythonPathUtils from './pythonPathUtils';
@@ -256,6 +256,7 @@ export class ImportResolver {
             importFailureInfo,
             resolvedPaths: [],
             importType: ImportType.Local,
+            priority: DefaultPriority,
             isStubFile: false,
             isNativeLib: false,
             implicitImports: [],
@@ -1161,6 +1162,18 @@ export class ImportResolver {
             importFound = resolvedPaths.length >= moduleDescriptor.nameParts.length;
         }
 
+        // Set priority based on if import is in the excluded list or not
+        let priority = DefaultPriority;
+        if (importFound && sourceFilePath && !isNamespacePackage && resolvedPaths.length > 0) {
+            // Check the resolved path. If it's for an included file or the importing
+            // file is also excluded, treat as lower priority.
+            priority =
+                !matchFileSpecs(this._configOptions, sourceFilePath, true) ||
+                matchFileSpecs(this._configOptions, resolvedPaths[resolvedPaths.length - 1], true)
+                    ? DefaultPriority
+                    : DefaultPriority - 10;
+        }
+
         return {
             importName,
             isRelative: false,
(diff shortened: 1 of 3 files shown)


packages/pyright-internal/src/analyzer/importResolver.ts
```diff
@@ -1161,6 +1162,18 @@ export class ImportResolver {
             importFound = resolvedPaths.length >= moduleDescriptor.nameParts.length;
         }
 
+        // Set priority based on if import is in the excluded list or not.
+        let priority = DefaultPriority;
+        if (importFound && sourceFilePath && !isNamespacePackage && resolvedPaths.length > 0) {
+            // Check the resolved path. If it's for an included file or the importing
+            // file is also excluded, treat as lower priority.
```
debonte, line 1169: Is the logic described in the comment backwards compared to the code? I'm assuming that a lower priority number means "lower priority" meaning that we'd push it further down the completion list.

It looks to me like the code is checking if the resolved path is included or the importing path is not, give it **higher** priority.

debonte (commented):
Do we need to hotfix this? Eric seemed to think the regression in Pyright was pretty bad. Is it less impactful in Pylance?

packages/pyright-internal/src/analyzer/importResolver.ts
rchiodo: Comment is backwards

rchiodo:
> Do we need to hotfix Pylance with this? Eric seemed to think the regression in Pyright was pretty bad. Is it less impactful in Pylance?

I don't see why not

packages/pyright-internal/src/analyzer/importResolver.ts
```diff
@@ -1534,10 +1548,20 @@ export class ImportResolver {
             }
 
             // Prefer local packages.
-            if (bestImportSoFar.importType === ImportType.Local && !bestImportSoFar.isNamespacePackage) {
+            if (
+                bestImportSoFar.importType === ImportType.Local &&
+                !bestImportSoFar.isNamespacePackage &&
+                newImport.importType !== ImportType.Local
```
erictraut, line 1554: I don't think this is correct. The import resolution ordering shouldn't be affected by the import type.

bschnurr (approved):

packages/pyright-internal/src/analyzer/importResolver.ts
```diff
@@ -1534,10 +1548,20 @@ export class ImportResolver {
...
+            ) {
                 return bestImportSoFar;
             }
-            if (newImport.importType === ImportType.Local && !newImport.isNamespacePackage) {
+
+            // Prefer higher priority.
+            if (
+                newImport.importType === ImportType.Local &&
+                !newImport.isNamespacePackage &&
+                newImport.priority > bestImportSoFar.priority
```
erictraut, line 1563: I don't think this is correct. The import resolution ordering shouldn't be affected by the import type or priority. It's fine to change the completion ordering based on the priority, but the import resolution ordering is dictated by the way the Python import loader works.

erictraut (changes requested):

packages/pyright-internal/src/analyzer/importResolver.ts
rchiodo: The user is asking to control just that. Perhaps there is another way?
https://github.com/microsoft/pylance-release/issues/3429

Like maybe we support .env files?

packages/pyright-internal/src/analyzer/importResolver.ts
rchiodo: I technically didnt change this one, it's just a side effect of the check below.

packages/pyright-internal/src/analyzer/importResolver.ts
erictraut: I think there's some confusion here about the meaning of "exclude". It indicates that the type checker shouldn't run on those files. It has no bearing on import resolution. If the user is asking to control import resolution using "exclude", then you should correct their misunderstanding.

rchiodo: They understand what is used for, but they want something new. This was one way to accomplish their goal, I'm wondering if there are others. 

Extra paths is used for controlling import resolution, perhaps it should behave like insert instead of append or have some other way to insert

erictraut: It wouldn't be appropriate to overload the use of "exclude" for this purpose. These are very different concepts.

Is the use case important enough to bother with? It was reported by one user and has no upvotes. Maybe just say "won't fix" or open for discussion.

rchiodo: Fair point. Ill close this then

PR #4691, opened 2023-02-25
added module name to moduleUnknownMember diagnostic to help inform wh…

bschnurr:
…ich module has the unknown member in the case of a union of modules being asked.

issue found in the tensorNet codebase here. https://github.com/taehoonlee/tensornets/blob/c9b1d78f806892193efdebee2789a47fd148b984/tensornets/utils.py#L118

previous diagnostic hover
image

where `tf` is a union of modules
image

new hover now mentions the name of the module
image


debonte:
In this particular scenario it's not clear to me that this is an improvement since at first glance I thought that the `"tf"` in the hover text was referring to the `tf` in `tf.GraphKeys` rather than helping me differentiate between `tf` and `tf.v1`.

Also, if you had a union of 3+ modules and 2+ of them were missing `GraphKeys`, would it give you a list of the modules that didn't have a `GraphKeys` member?

When there's a union of types and some of them are missing a member, we list out the types that are missing that member. For example, in this `Union[Good, Bad1, Bad2]` where only `Good` has the member `maybe_missing`:

image

erictraut (approved):

bschnurr:
i'll double check the cause of union with 3

bschnurr:
going to stop using printType inorder to remove the redundant Module("..") string 
image

PR #4714, opened 2023-03-01
[PylanceBot] Pull Pylance with Pyright 1.1.296

PylanceBot:

- made auto parenthese to check snippet support 
- node in a variable decl just points to the variable name. it should be expanded to the assignment statement to get proper end position. 
- Fix error node parsing for completions 
- Fix type ignore ending up on the wrong line
- Fix slots doc strings 
- Fix action warnings about deprecated features
- Remove unused imports after moving symbol.
- Fix offset when checking if current indent should be retained
- Fixed move symbol bug

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -32,7 +32,7 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v2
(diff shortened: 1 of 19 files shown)


heejaechang (approved):

bschnurr:
going to retrigger, due to updating branch with main

PR #4717, opened 2023-03-01
Fix indented code blocks in docstrings

gracepetryk:
This PR fixes an issue where markdown code blocks do not parse correctly when they started on an indented line in a docstring.

```python
def test():
    """
    A test function

    Example:
        ```python
        print(test())
        ```
    """
    pass
```

Currently this docstring looks like this when hovering `test`: 
![image](https://user-images.githubusercontent.com/5047809/222246442-6db923e7-8d59-4236-84e0-d42ada48295d.png)

After the fix it renders correctly:
![image](https://user-images.githubusercontent.com/5047809/222276577-57ad01d1-5be2-41b3-92bd-4e5a71bdd8ef.png)

### DIFF
--- a/packages/pyright-internal/src/analyzer/docStringConversion.ts
+++ b/packages/pyright-internal/src/analyzer/docStringConversion.ts
@@ -427,7 +427,10 @@ class DocStringConverter {
         const match = this._currentLine().match(CodeBlockStartRegExp);
         if (match !== null) {
             this._blockIndent = this._currentIndent()
-            this._appendLine('```' + match[1]) // remove indentation, preserve language tag
+
+            // Remove indentation and preserve language tag.
+            this._appendLine('```' + match[1])
+
             this._pushAndSetState(this._parseBacktickBlock);
             this._eatLine();
             return true;
@@ -436,7 +439,7 @@ class DocStringConverter {
     }
 
     private _parseBacktickBlock(): void {
-        // only match closing ``` at same indent level of opening
+        // Only match closing ``` at same indent level of opening.
         if (CodeBlockEndRegExp.test(this._currentLine()) && this._currentIndent() === this._blockIndent) {
             this._appendLine('```');
             this._appendLine();
--- a/packages/pyright-internal/src/analyzer/docStringConversion.ts
+++ b/packages/pyright-internal/src/analyzer/docStringConversion.ts
@@ -80,6 +80,8 @@ const PlusRegExp = /\+/g;
 const UnescapedMarkdownCharsRegExp = /(?<!\\)([_*~[\]])/g;
 const linkRegExp = /(\[.*\]\(.*\))/g;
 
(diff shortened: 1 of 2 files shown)


packages/pyright-internal/src/analyzer/docStringConversion.ts
```diff
@@ -431,7 +436,8 @@ class DocStringConverter {
     }
 
     private _parseBacktickBlock(): void {
-        if (this._currentLine().startsWith('```')) {
+        // only match closing ``` at same indent level of opening
```
erictraut, line 439: Nit: Comments within the pyright code base begin with capital letters and end in a period if they are complete sentences.

packages/pyright-internal/src/analyzer/docStringConversion.ts
```diff
@@ -421,8 +424,10 @@ class DocStringConverter {
     }
 
     private _beginBacktickBlock(): boolean {
-        if (this._currentLine().startsWith('```')) {
-            this._appendLine(this._currentLine());
+        const match = this._currentLine().match(CodeBlockStartRegExp);
+        if (match !== null) {
+            this._blockIndent = this._currentIndent()
+            this._appendLine('```' + match[1]) // remove indentation, preserve language tag
```
erictraut, line 430: Nit: Don't use end-of-line comments. Move to previous line. Comments should begine with capital letter and end in period if they are complete sentence.

erictraut (changes requested):
There should be a bug report associated with the change.

Change should also include a test to valid the change and prevent regressions.

gracepetryk:
https://github.com/microsoft/pyright/issues/4718

gracepetryk:

packages/pyright-internal/src/analyzer/docStringConversion.ts
```diff
@@ -421,8 +424,13 @@ class DocStringConverter {
...
 
     private _beginBacktickBlock(): boolean {
-        if (this._currentLine().startsWith('```')) {
-            this._appendLine(this._currentLine());
+        const match = this._currentLine().match(CodeBlockStartRegExp);
+        if (match !== null) {
+            this._blockIndent = this._currentIndent()
+
+            // Remove indentation and preserve language tag.
+            this._appendLine('```' + match[1])
```
debonte, line 432: Match group would be null if the language tag is missing?

debonte (changes requested):

packages/pyright-internal/src/analyzer/docStringConversion.ts
gracepetryk: match returns an empty string if a capture group is empty
```
> '    ```'.match(/\s*```(\w*)/)[1]
''
```

erictraut (approved):
Looks good to me. Thanks for the contribution!

PR #4747, opened 2023-03-08
Add pretty-printing to inline type-dicts

decorator-factory:
Before:
```
def origin() -> dict[{"x": int, "y": int}]:
    return {"x": 0, "y": 0}

p = origin()
reveal_type(p)  # Type of "p" is "<TypedDict>"
```

After:
```
def origin() -> dict[{"x": int, "y": int}]:
    return {"x": 0, "y": 0}

p = origin()
reveal_type(p)  # Type of "p" is "dict[{"x": int, "y": int}]"
```

I think it's a much more helpful description of the type


decorator-factory:
@microsoft-github-policy-service agree [company="myself"]

erictraut:
I don't think we'd want to do it this way. This logic should probably go into the typePrinter and take into account the various flags accepted by the typePrinter. It's also not clear whether it's a good idea to print out the entire definition. The pylance team should make that call, since they need to ensure that this works with hover text, completion suggestions, inlined types, etc.

Let's leave this out for now. We can refine it later if/when the PEP is filed.

PR #4790, opened 2023-03-17
Performance: Changed the logic in the type evaluator's code flow engi…

erictraut:
…ne to assume that any function that lacks a return type annotation is _not_ a NoReturn function. This greatly speeds up type analysis in certain unannotated code bases, but it results in less accurate type analysis.

### DIFF
--- a/packages/pyright-internal/src/analyzer/codeFlowEngine.ts
+++ b/packages/pyright-internal/src/analyzer/codeFlowEngine.ts
@@ -1583,7 +1583,15 @@ export function getCodeFlowEngine(
             return true;
         }
 
-        return evaluator.isAfterNodeReachable(functionType.details.declaration.node);
+        // This call is very expensive because it requires a full code flow analysis
+        // of this function and all (unannotated) functions that it calls, recursively.
+        // Rather than incurring this expense, we assume here that all unannotated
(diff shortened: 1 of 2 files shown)


packages/pyright-internal/src/analyzer/codeFlowEngine.ts
```diff
@@ -1583,7 +1583,15 @@ export function getCodeFlowEngine(
...
 
-        return evaluator.isAfterNodeReachable(functionType.details.declaration.node);
+        // This call is very expensive because it requires a full code flow analysis
+        // of this function and all (unannotated) functions that it calls, recursively.
+        // Rather than incurring this expense, we assume here that all unannotated
+        // functions are not NoReturn functions. To reverse this decision, uncomment
+        // the following line.
+
+        // return evaluator.isAfterNodeReachable(functionType.details.declaration.node);
+
```
bschnurr, line 1593: should we keep it for user code?

bschnurr: or maybe a setting

erictraut: No, type evaluation should be consistent, not depend on whether it's user code or not and should not depend on a setting.

PR #4796, opened 2023-03-17
fixed missing imports inside try/catch blocks when createTypeStub comm…

bschnurr:
…and is used.

example of lightgbm imports that weren't included in the .pyi

https://github.com/microsoft/LightGBM/blob/2fe2bf0675ef127a29f9bc49511fc0099cf3e140/python-package/lightgbm/__init__.py#L13


packages/pyright-internal/src/tests/fourslash/missingTypeStub.command.trycatchImport.fourslash.ts
```diff
@@ -0,0 +1,43 @@
...
+//// # This is a library file
+//// class MyLibrary:
+////     def DoEveryThing(self, code: str):
+////         ...
+
+// @filename: testLib/__init__.py
+// @library: true
+//// try:
+////     from .mylibrary import MyLibrary
+//// except ImportError:
```
bschnurr, line 19: previous `__init__.pyi` was missing `from .mylibrary import MyLibrary`

rchiodo (approved):

packages/pyright-internal/src/tests/fourslash/missingTypeStub.command.trycatchImport.fourslash.ts
erictraut: Hmm, I'm not sure this is what you want. If an `import` statement occurs within a `try` block, there's a good chance that there will also be an `except` that includes an import from another library or a local redefinition of the symbol imported within the `try` block.

This is a situation where manual intervention is often required to "fix up" the type stub.

If you are going to allow statements within a `try`/`except` block to be emitted by the stub generator, then I recommend that you only walk the `try` block and ignore the `except` and `else` blocks (if present). To do this, you would need to modify `visitTry` to call `this.walk(node.trySuite);` and then return `false`.

bschnurr: the issue was the pyi stub generated had `LGBModel` in its `__all__` but was missing the import.
image

erictraut: Yes, I understand. Keep in mind that the stub generator is not designed to create perfect stubs. It is meant to generate a "first draft" that you should then go and clean up. Needing to add a missing import statement is pretty typical.

bschnurr: for this case its not apparent to me how the import could fall to load a sibling file.

erictraut: It's not uncommon for a library to include something like this:
```python
try:
    from typing import TypeAlias
except:
    from typing_extensions import TypeAlias
```

With your proposed change, this would result in the type stub generating emitting both import statements, which is worse (IMO) than not emitting any. If there are no import statements emitted, then it's obvious to the person who is cleaning up the stub that an import is missing, and they can easily add it (using a few keystrokes if auto-import is enabled). If you emit two import statements, then no error is generated, yet the code is probably not what was intended. It's easy for the user to miss a problem like this when they're cleaning up the file.

Like I said, I think it's reasonable to follow the `try` block, but your proposed change also follows the `except` and `else` blocks, which I think will lead to problems.

bschnurr: Sorry.. I like your suggestion of walking only one.

erictraut (approved):
LGTM

PR #4858, opened 2023-03-28
Don't modify `Comment` range data

debonte:
Addresses https://github.com/microsoft/pylance-release/issues/4157

I haven't been able to reproduce this crash. Also I don't see a way for a `Comment` to be created with an out-of-range `start` value.

However, I noticed that `_addDiagnosticForRegionComment` was modifying `start` and `length` on `Comment` ranges to include the `#` sign. This was modifying the `Comment` data in `ParseResults`, so I could imagine it causing downstream problems. I verified that if `Checker` called `_checkRegions()` twice in a row we hit the `convertOffsetToPosition` assertion in the bug.

I considered [returning](https://github.com/microsoft/pyright/commit/9af9c78f32d27b1f77ec868549b9791da2e6fd8d) `undefined` from `getRegionCommentType` when `hashOffset` was out of range. But I think it's best to make this change and see if the crash disappears in telemetry.

### DIFF
--- a/packages/pyright-internal/src/analyzer/checker.ts
+++ b/packages/pyright-internal/src/analyzer/checker.ts
@@ -6218,11 +6218,10 @@ export class Checker extends ParseTreeWalker {
     }
 
     private _addDiagnosticForRegionComment(regionComment: RegionComment, message: string): Diagnostic | undefined {
-        // extend range to include # character
-        const range: TextRange = regionComment.comment;
-        range.start -= 1;
-        range.length += 1;
-
-        return this._evaluator.addDiagnosticForTextRange(this._fileInfo, 'error', '', message, range);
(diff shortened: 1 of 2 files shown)


rchiodo (approved):

erictraut (approved):

debonte:
Btw, @erictraut, this is the second time recently I've heard of a problem like this. Stella had a similar problem (in unsubmitted code) where she modified the path on a `Declaration` object not knowing that it would be reused. I wonder if there's a way that these objects could be made read-only?

erictraut:
Yeah, I think that's worth considering. I haven't personally hit this problem because I know that these data structures are meant to be immutable, so I treat them as such. But as others contribute to the code base, it would be nice if the TS compiler could enforce immutability.

For ranges in particular, I think it would require a change in the way the ranges are built in the parser. Descriptors (which are mostly built in the binder) might be an easier place to start.

PR #4882, opened 2023-03-31
Revert "when indexing is off, include import alias from user files"

bschnurr:
related to https://github.com/microsoft/pylance-release/issues/4179


erictraut:
@heejaechang, are you OK with this revert?

erictraut (approved):

erictraut:
I'm going to merge this change because this regression makes pyright very hard to use. 

@heejaechang, I'm going to assume that you're OK with this.

PR #4903, opened 2023-04-05
Changed the default of the `useLibraryCodeForTypes` from false to tru…

erictraut:
…e to bring pylance and pyright into alignment. Also deprecated the "--lib" command-line option, which was previously used to enable `useLibraryCodeForTypes` from the command line.


debonte (approved):

erictraut:
@debonte, I decided not to include this change in this week's pyright release. I'll merge it now, and it will be in next week's (1.1.303). This gives pyright users one week to respond to the [open issue](https://github.com/microsoft/pyright/issues/4904) that's tracking it.

PR #4923, opened 2023-04-08
Define python.analysis.extraPaths for stubs

costis-t:
python.analysis.extraPaths needs to be defined in settings.json to include the "typings" folder path. Full path was required for me.

### DIFF
--- a/docs/type-stubs.md
+++ b/docs/type-stubs.md
@@ -30,6 +30,11 @@ Click on the menu item to create the type stub. Depending on the size of the lib
 
 ![Pyright](img/CreateTypeStub2.png)
 
+Add the full path of the `typings` folder in `settings.json`:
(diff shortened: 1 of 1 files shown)


erictraut:
This is incorrect. It isn't recommended that you use `extraPaths` to define the stub locations. You should use `stubPath` instead. And `stubPath` should already default to a project-local directory called `typings`. If you're finding that this doesn't work as you expect, please file a bug with repro steps.

costis-t:
Correct. Setting it on stubPath was not working (local or absolute path). However, I could not reproduce it in an Ubuntu testing VM. I purged all the relevant config folders in my system and now it works as expected. I will keep an eye on it and file a relevant bug if it ever reappears. Thanks!

PR #4981, opened 2023-04-20
[WIP] Refactoring to move hover out of the program

heejaechang:
open to gather feedbacks.

### DIFF
--- a/packages/pyright-internal/src/analyzer/program.ts
+++ b/packages/pyright-internal/src/analyzer/program.ts
@@ -29,7 +29,7 @@ import { assert, assertNever } from '../common/debug';
 import { Diagnostic, DiagnosticCategory } from '../common/diagnostic';
 import { FileDiagnostics } from '../common/diagnosticSink';
 import { FileEditAction, FileEditActions, FileOperations, TextEditAction } from '../common/editAction';
-import { Extensions } from '../common/extensibility';
+import { Extensions, ProgramView } from '../common/extensibility';
 import { LogTracker } from '../common/logTracker';
 import {
     combinePaths,
@@ -76,15 +76,13 @@ import {
 import { DefinitionFilter } from '../languageService/definitionProvider';
 import { DocumentSymbolCollector, DocumentSymbolCollectorUseCase } from '../languageService/documentSymbolCollector';
 import { IndexOptions, IndexResults, WorkspaceSymbolCallback } from '../languageService/documentSymbolProvider';
-import { HoverResults } from '../languageService/hoverProvider';
 import { ImportAdder, ImportData } from '../languageService/importAdder';
 import { getModuleStatementIndentation, reindentSpan } from '../languageService/indentationUtils';
 import { getInsertionPointForSymbolUnderModule } from '../languageService/insertionPointUtils';
 import { ReferenceCallback, ReferencesResult } from '../languageService/referencesProvider';
 import { RenameModuleProvider } from '../languageService/renameModuleProvider';
 import { SignatureHelpResults } from '../languageService/signatureHelpProvider';
-import { ParseNodeType } from '../parser/parseNodes';
-import { StatementNode } from '../parser/parseNodes';
+import { ParseNodeType, StatementNode } from '../parser/parseNodes';
 import { ParseResults } from '../parser/parser';
 import { AbsoluteModuleDescriptor, ImportLookupResult } from './analyzerFileInfo';
 import * as AnalyzerNodeInfo from './analyzerNodeInfo';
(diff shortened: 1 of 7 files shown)


packages/pyright-internal/src/analyzer/program.ts
```diff
@@ -233,16 +231,28 @@ export class Program {
...
+
+    get evaluator(): TypeEvaluator | undefined {
+        return this._evaluator;
+    }
+
+    get configOptions(): ConfigOptions {
+        return this._configOptions;
+    }
+
+    get importResolver(): ImportResolver {
```
heejaechang, line 254: changed to property since it doesn't do any work to return the object.

packages/pyright-internal/src/analyzer/program.ts
```diff
@@ -725,6 +723,28 @@ export class Program {
...
     }
 
+    // This will allow the callback to execute a type evaluator with an associated
+    // cancellation token and provide a read-only program.
+    run<T>(callback: (p: ProgramView) => T, token: CancellationToken): T {
+        const evaluator = this._evaluator ?? this._createNewEvaluator();
+        return evaluator.runWithCancellationToken(token, () => callback(this));
+    }
+
+    createSourceMapper(
```
heejaechang, line 733: we could change name to `getSourceMapper` rather than `create` since that seems implementation detail.

I looked to see whether I can actually take this one out of program, but since we want to load some shadow files in program, it has to be in program. but it is not actually mutating program in anyway that can impact `semantic result`, so it is part of `programView`

packages/pyright-internal/src/common/extensibility.ts
```diff
@@ -39,12 +39,22 @@ export interface ProgramExtension {
 // Readonly wrapper around a Program. Makes sure it doesn't mutate the program.
 export interface ProgramView {
     readonly id: number;
-    rootPath: string;
-    getImportResolver(): ImportResolver;
-    console: ConsoleInterface;
-    getConfigOptions(): ConfigOptions;
+    readonly rootPath: string;
```
heejaechang, line 42: we should go through everything we return here and make sure we can't mutate them. for example, `configOptions` should have all field field marked as `readonly`. also, we might want to create `ImportResolverView` or something that only expose `readonly` APIs that LSP feature actually uses. so, blocking IDE features from calling things like `invalidateCache` and etc. We probably want to remove `Completion` code in `ImportResolver` as well.

currently we expose `SourceFileInfo` through view, we should either not expose it or expose only `readonly` fields.

packages/pyright-internal/src/analyzer/service.ts
```diff
@@ -29,16 +29,16 @@ import { BackgroundAnalysisBase, IndexOptions, RefreshOptions } from '../backgro
 import { CancellationProvider, DefaultCancellationProvider } from '../common/cancellationUtils';
 import { CommandLineOptions } from '../common/commandLineOptions';
 import { ConfigOptions, matchFileSpecs } from '../common/configOptions';
-import { ConsoleInterface, log, LogLevel, StandardConsole } from '../common/console';
+import { ConsoleInterface, LogLevel, StandardConsole, log } from '../common/console';
```
heejaechang, line 32: I am not sure why these ordering is changed. pyright and pylance might have different prettier setting?

packages/pyright-internal/src/languageService/hoverProvider.ts
```diff
@@ -648,98 +590,97 @@ export class HoverProvider {
...
-    if (hoverResults.lastKnownModule && format === MarkupKind.Markdown && includeHash) {
-        markupString += `\n<!--moduleHash:${hashString(hoverResults.lastKnownModule)}-->`;
+                return part.text;
+            })
+            .join('')
+            .trimEnd();
+
+        // If we have a lastKnownModule in the hover results, stick in a comment with
+        // the hashed module name. This is used by the other side to send telemetry.
+        if (hoverResults.lastKnownModule && this._format === MarkupKind.Markdown && this._supportTelemetry) {
```
heejaechang, line 663: since `hoverprovider` is instance class now, we can do this only in `pylance` and not leak this into `pyright`

heejaechang:
tagging @erictraut to get feedback

erictraut (commented):
This PR combines two significant changes:
1. It demonstrates a new pattern where a provider is passed the program rather than being called by the program.
2. It changes the implementation of the hover provider to be an instantiable class that holds state rather than a singleton.

I think both of these are fine changes, although it would perhaps have been preferable to separate them into two different PRs since each one involves quite a few modifications and they are separable.

That said, I'm fine with moving forward with this change and then using this as a pattern for the other providers. Please do the changes in bite-sized pieces so they can be reviewed separately and any regressions that are introduced can be pinpointed to a particular commit.

packages/pyright-internal/src/languageService/hoverProvider.ts
```diff
@@ -56,44 +54,52 @@ import {
...
+        this._parseResults = this._program.getParseResults(this._filePath);
+        this._sourceMapper = this._program.createSourceMapper(this._filePath, this._token, /* mapCompiled */ true);
+    }
 
-export interface HoverResults {
-    parts: HoverTextPart[];
-    lastKnownModule?: string;
-    range: Range;
-}
+    private get _evaluator(): TypeEvaluator {
```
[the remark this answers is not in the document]
heejaechang: will do.

heejaechang:
closed. I will split the PR into 2 PRs.

packages/pyright-internal/src/languageService/hoverProvider.ts
rchiodo: You mean we can subclass it, right? Or are we thinking of just forking the whole thing?

heejaechang: subclass.

PR #4989, opened 2023-04-21
Move HoverProvider out of Program

heejaechang:
This takes out HoverProvider out of program without any changes on HoverProvider.


erictraut (approved):
LGTM

PR #4990, opened 2023-04-22
Convert hover provider to regular class rather than static type.

heejaechang:
I will start sending PR for each IDE features in the program.


erictraut (approved):
LGTM

PR #5011, opened 2023-04-26
Refactored find all references out of program

heejaechang:
I think the rest of IDE features will be refactored out from `pylance` side and pushed up to `pyright`

IDE features left are

1. completion provider - it uses indices so it needs to be refactored with related indexing code in pylance.
2. document symbol, workspace symbol provider - they use indices as well. so, we need to refactor indices portion out first in pylance.
3. rename, move symbol, rename module provider - since all of them, at the end, rename references of a name, these are all in reference each other internally. but move symbol and rename module provider are split between in pylance and pyright. so doing it from pylance is much easier.

after those changes, we will probably take out any APIs in the program that doesn't require any internal state of program to its own type sucn as `getAutoImport` and etc.

### DIFF
--- a/packages/pyright-internal/src/analyzer/program.ts
+++ b/packages/pyright-internal/src/analyzer/program.ts
@@ -1856,7 +1856,7 @@ export class Program {
                 case 'singleFileMode':
                     referenceProvider.addReferencesToResult(
                         sourceFileInfo.sourceFile.getFilePath(),
-                        /*includeDeclaration*/ true,
+                        /* includeDeclaration */ true,
                         referencesResult
                     );
                     break;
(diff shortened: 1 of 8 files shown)


packages/pyright-internal/src/analyzer/program.ts
```diff
@@ -1947,10 +1850,15 @@ export class Program {
...
             }
 
+            const referenceProvider = new ReferencesProvider(this, token);
             const renameMode = this._getRenameSymbolMode(sourceFileInfo, referencesResult, isDefaultWorkspace);
             switch (renameMode) {
                 case 'singleFileMode':
-                    sourceFileInfo.sourceFile.addReferences(referencesResult, true, this._evaluator!, token);
+                    referenceProvider.addReferencesToResult(
+                        sourceFileInfo.sourceFile.getFilePath(),
+                        /*includeDeclaration*/ true,
```
bschnurr, line 1859: Eric wants these with space. /* includeDeclaration */

heejaechang: thank you

rchiodo (approved):

Issue #4994, opened 2023-04-23
pyright-lsp may enter a strange state reporting library not found when there are sub-project inside a folder


zhou13:
**Describe the bug**
Let's say we have a project structure like this:
```
$ tree
[ 288]  .
├── [   0]  example.py
├── [ 192]  projects
│   ├── [  49]  example.py
│   ├── [ 226]  requirements.txt
│   └── [ 160]  sac
│       ├── [   0]  __init__.py
│       └── [  21]  datasets.py
├── [  31]  pyproject.toml
```

I attached the example files in [Archive.zip](https://github.com/microsoft/pyright/files/11303247/Archive.zip)

With certain operation, pyright-lsp may enter a strange state reporting system library like os not found.

**To Reproduce**
The following sequence has 100% chance to trigger the bug on my Mac:
1. enable pyright in nvim
2. `nvim projects/example.py example.py`
3. switch to next buffer with `:bnext` or `:BufferPrev`
4. switch back with `bprev` or `:BufferPrev`
5. perform a format operation using isort inside vim
6. pyright-lsp will enter a strange state reporting `os` not found.

**Expected behavior**

pyright-lsp should not enter a strange state reporting `os` not found.

**Screenshots or Code**

![render1682238804660](https://user-images.githubusercontent.com/576382/233829166-3e581669-98dd-4dc5-a769-17a2b6e4c5f5.gif)

**VS Code extension or command-line**
I am using vim-lsp with pyright. I attached the lsp log for analysis, which contains all the communication in a json-like format:
[lsp.log](https://github.com/microsoft/pyright/files/11303255/lsp.log)

I am not sure whether it is possible to write a reproducible script by communicating with pyright-lsp in command-line tools. I will be happy to do that if there is a pointer.

**Additional context**
* If I delete `projects/requirements.txt`, then this bug is not reproducible. 
* step 5 is needed to trigger `textDocument/didChange`

erictraut:
I don't know anything about vim, so I'm trying to diagnose this based on the provided logs. Based on the symptoms, I suspect that the delta changes to the document are not being properly applied to the in-memory contents, so pyright is analyzing a corrupted source file.

Here's why I think this is the case.

The starting text in the document "./projects/example.py" is:
```python
import os
from sac.datasets import Data

os
Data
```

Here's the RPC message sent to the language server:
```json
{
  jsonrpc = "2.0",
  method = "textDocument/didChange",
  params = {
    contentChanges = { {
        range = {
[...]

PR #5019, opened 2023-04-27
Added a map of open documents to the languageServerBase class. It is …

erictraut:
…responsible for applying any delta changes and updating the contents for each of the workspaces associated with that file. This allows source code contents to be shared between workspaces, which reduces memory usage. It also eliminates a correctness issue that can occur if new workspaces are added after files are open. It addresses https://github.com/microsoft/pyright/issues/4994.

### DIFF
--- a/packages/pyright-internal/src/analyzer/service.ts
+++ b/packages/pyright-internal/src/analyzer/service.ts
@@ -16,7 +16,6 @@ import {
     CompletionItem,
     DocumentSymbol,
 } from 'vscode-languageserver';
-import { TextDocumentContentChangeEvent } from 'vscode-languageserver-textdocument';
 
 import { BackgroundAnalysisBase, IndexOptions, RefreshOptions } from '../backgroundAnalysisBase';
 import { CancellationProvider, DefaultCancellationProvider } from '../common/cancellationUtils';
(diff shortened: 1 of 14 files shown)


rchiodo (approved):

packages/pyright-internal/src/languageServerBase.ts
```diff
@@ -1210,6 +1233,8 @@ export abstract class LanguageServerBase implements LanguageServerInterface {
         workspaces.forEach((w) => {
             w.service.setFileClosed(filePath);
         });
+
+        this.openFileMap.delete(filePath);
```
bschnurr, line 1237: should we clear 'openFileMap' in 'onShutdown' ?

packages/pyright-internal/src/analyzer/service.ts
```diff
@@ -318,7 +317,7 @@ export class AnalyzerService {
     updateOpenFileContents(
         path: string,
         version: number | null,
-        contents: TextDocumentContentChangeEvent[],
```
rchiodo, line 321: @heejaechang I believe this would mean we'd have to do updates for say fixall differently. Maybe other spots too.

erictraut: I think this works fine for the current fixall strategy. These changes are applied only to source files associated with a cloned program.

rchiodo: Oh it looks like fixall uses something you already fixed. the workspaceEditUtils.applyDocumentChanges.

Issue #5025, opened 2023-04-27
Provide a way to `resolveAliasDeclaration` even if aliasDecl.path points to `unneeded` file.


heejaechang:
**Is your feature request related to a problem? Please describe.**
currently, when [resolveAliasDeclaration](https://github.com/microsoft/pyright/blob/main/packages/pyright-internal/src/analyzer/typeEvaluatorTypes.ts#L396) is called, it internally calls `program._lookUpImport` to follow up `alias` symbol. Since the system is lazy, if it sees the file to follow up is not bound yet (https://github.com/microsoft/pyright/blob/main/packages/pyright-internal/src/analyzer/program.ts#L2887), it will attempt to bind it. but it has this `perf optimization` where it will check whether the file is actually needed (rooted by either user file or opened in editor transitively) and if not, it will just bail out at that point.

this works fine for regular case where all library files are rooted to one of those, but it doesn't work if library symbol we want is not rooted by any of those.

some of examples would be https://github.com/microsoft/pylance-release/issues/3700, or our `go to definition` on pytest builtin fixture which doesn't require `explicit` import statement or if we ever support `find all reference` on closed library files as some user wanted.

This one can also be mixed with [this issue]( https://github.com/microsoft/pyright/pull/5023 ) and cause even more headaches.

erictraut:
If I understand you correctly, you're talking about a case where the specified file is not explicitly imported anywhere. If this file is resolved, we'd need to create a `SourceFile` object for it. Would you expect that the file would be added to the program? I think it would need to be, but I'm not sure what its lifetime would be.
* It can't be considered a "tracked" file. Tracked files are explicitly imported, and if their "importedBy" count drops to zero, they are automatically removed from the program.
* It can't be considered a "shadowed file" because their lifetime is defined by the file they are shadowing. 
* If it is opened in the editor, it could be marked as "isOpenByClient", and its lifetime would then be controlled by when the user closes the file, but it sounds like you want to follow entire alias chains, even in cases where you don't intend to open the intermediate files in the alias chain. So I'm not sure how that would work.
 
[...]

PR #5028, opened 2023-04-28
Added a flag in `resolveAliasDeclaration` so that we can skip early bail out (perf optimization) when needed.

heejaechang:
fix https://github.com/microsoft/pyright/issues/5025


erictraut (approved):
LGTM

PR #5031, opened 2023-04-28
Idea for copy on write

rchiodo:
Wrote a simple test, but also used this in pyrx to change how fix-all works. 

I put back the change events for the program/sourceFile. It was the easiest way to capture the events (otherwise I believe all the changes would be a full document replace).


packages/pyright-internal/src/analyzer/backgroundAnalysisProgram.ts
```diff
@@ -98,8 +100,8 @@ export class BackgroundAnalysisProgram {
     }
 
     setFileOpened(filePath: string, version: number | null, contents: string, options: OpenFileOptions) {
-        this._backgroundAnalysis?.setFileOpened(filePath, version, contents, options);
-        this._program.setFileOpened(filePath, version, contents, options);
+        this._backgroundAnalysis?.setFileOpened(filePath, version, [{ text: contents }], options);
```
rchiodo, line 103: Putting the events back made it easier to track them in the program setFileOpened. We could do something else too if we wanted to keep the setFileOpened simpler.

I was thinking potentially:
- We'd just return full edits for the entire file. This can be slow though.
- We track the edits at a higher level as they come into the service.

packages/pyright-internal/src/analyzer/program.ts
```diff
@@ -84,15 +85,16 @@ import { Scope } from './scope';
 import { getScopeForNode } from './scopeUtils';
 import { IPythonMode, SourceFile } from './sourceFile';
 import { collectImportedByFiles, isUserCode } from './sourceFileInfoUtils';
-import { isStubFile, SourceMapper } from './sourceMapper';
+import { SourceMapper, isStubFile } from './sourceMapper';
```
rchiodo, line 88: Oh I think the fixall is messing up imports. I probably have to turn it off before saving these.

packages/pyright-internal/src/analyzer/program.ts
```diff
@@ -351,8 +380,44 @@ export class Program {
...
+        version: number | null,
+        contents: TextDocumentContentChangeEvent[],
+        options?: OpenFileOptions
+    ) {
+        let chainedFilePath: string | undefined = options?.chainedFilePath;
+        const filePathKey = normalizePathCase(this.fileSystem, filePath);
         let sourceFileInfo = this.getSourceFileInfo(filePath);
+
+        // If in edit mode, we need to start over if this is the first edit.
+        if (this._editMode && sourceFileInfo && !this._copiedOnWriteFiles.has(filePathKey)) {
```
rchiodo, line 394: Need to change this to handle case where a new file is opened during edit mode.

packages/pyright-internal/src/analyzer/program.ts
```diff
@@ -240,6 +245,37 @@ export class Program {
         this._cacheManager.unregisterCacheOwner(this);
     }
 
+    enterEditMode() {
```
erictraut, line 248: It would probably be better to not expose `enterEditMode` and `exitEditMode` directly. These should be private. The public interface should be something like `useEditMode` with a try/catch handler, like this:

```typescript
useEditMode(callback: () => void) {
    this._enterEditMode();
    try {
        callback();
   } finally {
        this._leaveEditMode();
    }
}
```

This will properly handle cancellation exceptions.

packages/pyright-internal/src/analyzer/program.ts
```diff
@@ -240,6 +245,37 @@ export class Program {
...
+        this._editMode = true;
+    }
+
+    leaveEditMode() {
+        this._editMode = false;
+        const edits: FileEditAction[] = [];
+        this._copiedOnWriteFiles.forEach((fileInfo, path) => {
+            const changes = this._copiedOnWriteChanges.get(path);
```
erictraut, line 256: Hmm, this is way more complicated than what I had in mind. I proposed that each SourceFile have an "isEditMode" flag and maintain its own copy-on-write file contents. That way, the program doesn't need to maintain any additional state.

packages/pyright-internal/src/analyzer/program.ts
```diff
@@ -351,8 +387,50 @@ export class Program {
...
     }
 
-    setFileOpened(filePath: string, version: number | null, contents: string, options?: OpenFileOptions) {
+    setFileOpened(
+        filePath: string,
+        version: number | null,
+        contents: TextDocumentContentChangeEvent[],
+        options?: OpenFileOptions
+    ) {
+        let chainedFilePath: string | undefined = options?.chainedFilePath;
```
erictraut, line 396: See comment above. This is way more complicated than necessary. Please consider moving CoW logic to SourceFile.

packages/pyright-internal/src/analyzer/service.ts
```diff
@@ -222,6 +223,14 @@ export class AnalyzerService {
         return service;
     }
 
+    enterEditMode() {
```
erictraut, line 226: See note above about not exposing `enterEditMode` and `leaveEditMode`.

packages/pyright-internal/src/analyzer/sourceFile.ts
```diff
@@ -8,6 +8,7 @@
  */
 
 import { CancellationToken, CompletionItem, DocumentSymbol } from 'vscode-languageserver';
+import { TextDocument, TextDocumentContentChangeEvent } from 'vscode-languageserver-textdocument';
```
erictraut, line 11: Hmm, I was happy to have removed the language server notion of a TextDocument and TextDocumentContentChangeEvent from this file recently, but you've added them back. I don't think this is needed. The SourceFile can simply retain a string that holds its contents (and a second string that represents the "previous contents" if a modification is made while in edit mode).

erictraut (changes requested):

packages/pyright-internal/src/analyzer/sourceFile.ts
rchiodo: The problem with the 'previousContents' idea is you can't generate a very concise edit from it. It ends up being a full replace every time. This isn't so bad when the files are small, but sending full file contents back over LSP can be slow.

packages/pyright-internal/src/analyzer/program.ts
rchiodo: The SourceFileInfo needs to be recreated too I believe because of its list of imports. Since the SourceFileInfo doesn't have any code associated with it, I put it at this level.

packages/pyright-internal/src/analyzer/sourceFile.ts
erictraut: Isn't there a way to generate a minimal set of text edits from a "before" and "after" file contents? I thought that's what Heejae was doing previously, as opposed to recording all of the individual edit actions.

packages/pyright-internal/src/analyzer/program.ts
erictraut: I don't see why we'd need to create a new SourceFileInfo. The list of imports is dictated by the contents of the file. If you change the contents of the file, it will be marked dirty, reparsed, and its imports will be recomputed as needed. It leverages all of the existing machinery in the "program" and "sourceFile" modules.

packages/pyright-internal/src/analyzer/sourceFile.ts
rchiodo: Heejae's code was doing the full thing:
https://github.com/microsoft/pyright/blob/3be44c3242aac5644931c243c5f7a0dc4cbe55cf/packages/pyright-internal/src/common/workspaceEditUtils.ts#L189

He suggested it could use the vscode differ, but I've used the vscode diff before when working on notebook concat stuff. It's very messy and doesn't always guess right (and ends up turning stuff into a full edit). 

The easier way is to just remember the edits.

packages/pyright-internal/src/analyzer/program.ts
rchiodo: Ah okay. Sounds good. Didn't know what updated the import list. I'll try moving it to the SourceFile then.

rchiodo: Wait, I think this has to be at the program level because the source file may not exist. I could push the change tracking down to the SourceFile, but the list of changed source files has to be maintained in the program because it has to remove newly created files?

erictraut: Yes, you should be able to handle this in the SourceFile itself and leverage all of the other machinery in the program for tracking SourceFileInfos and their lifetimes.

packages/pyright-internal/src/analyzer/program.ts
heejaechang: if possible, it would be safer if we define something like ProgramView, like ClonedProgram or something like that and allow only what ProgramView allow + updating only text but doesn't allow updating like configOption (such as setConfiguration and etc)

by the way, will it allow modifying files? asking since one of common task on advanced refactoring is adding/removing/renaming a file.

packages/pyright-internal/src/analyzer/program.ts
rchiodo: I moved the logic to the SourceFile. Not sure if it's simpler or not. Copying the state out of the SourceFile was more selective so it changed a lot more. Handling it is certainly less complicated though.

packages/pyright-internal/src/analyzer/sourceFile.ts
```diff
@@ -80,7 +83,74 @@ export enum IPythonMode {
     CellDocs,
 }
 
+class WriteableData {
```
rchiodo, line 86: Made a class to hold all changeable data. Maybe I should call it MutableData. 

Anyway, this was a lot simpler than copying each item out by hand and felt like it would be more obvious in the future what needs to go here. 

However, it made a lot of changes in this file because of it.

packages/pyright-internal/src/analyzer/program.ts
rchiodo: Hmm, I get the same results. Maybe this file wasn't updated after a recent eslint change?

packages/pyright-internal/src/analyzer/sourceFile.ts
heejaechang: when you have nested edits, remembering edits won't work though if you were thinking sending those edits instead of sending whole documents.

if user change same file multiple time, you will need to adjust all spans since range would be wrong, and that's when there is no nested changes, if there are nested change, you need to merge those.

heejaechang: I would suggest just sending the last full content replacing whole content rather than trying to be micro-optimize. if we really want to have smaller edits, we should just diff content and combine that data with tokens or parse nodes to find chunk we want to replace together.

rchiodo: I'm not sure why the edits won't work? The same edits that were applied to change the file from one step to the next are returned.

heejaechang: if you are thinking about sending each edits with different `applyEdit` so multiple changes including nested edits work, then, it will become multiple undo.

rchiodo: So I think you're saying is that VS code doesn't handle the edits as one after another. The edits are actually all applied at once?

packages/pyright-internal/src/tests/workspaceEditUtils.test.ts
```diff
@@ -46,6 +48,99 @@ test('test applyWorkspaceEdits changes', async () => {
     assert.strictEqual(cloned.test_program.getSourceFile(range.fileName)?.getFileContent(), 'Text Changed');
 });
 
+test('test edit mode for workspace', async () => {
```
heejaechang, line 51: I think we need to have tests that changes same file multiple times and has nested changes such as

First add new import statement `from test import a, c` and then add new symbol to the new import you added in `from test import a[, b], c` and then remove import statement or reorder import statement in the file (assuming it had existing import statement)

or you can think about other multiple edits, nested edit case

let's say we provide refactoring to merge `import statements` and `remove unused imports`. so user doesn't have duplicated import statements or same symbol with different name such as `import os as o` and `import os as os` and etc.

packages/pyright-internal/src/analyzer/sourceFile.ts
heejaechang: edits and range are based on 1 version of document. not multiple version of document with changes.

packages/pyright-internal/src/analyzer/service.ts
```diff
@@ -222,6 +222,22 @@ export class AnalyzerService {
         return service;
     }
 
+    async useEditMode(callback: () => Promise<void>, token: CancellationToken) {
```
erictraut, line 225: Why is this async? I think it's important for this to be synchronous.

packages/pyright-internal/src/analyzer/backgroundAnalysisProgram.ts
```diff
@@ -242,6 +243,20 @@ export class BackgroundAnalysisProgram {
...
+
+        // Forward this request to the program.
+        this._program.enterEditMode();
+    }
+
+    exitEditMode() {
+        this._backgroundAnalysis = this._preEditAnalysis;
+        this._preEditAnalysis = undefined;
+        return this._program.exitEditMode();
+    }
```
erictraut, line 259: Nit: add extra line after this method.

packages/pyright-internal/src/analyzer/program.ts
```diff
@@ -187,6 +187,7 @@ export class Program {
     private _cacheManager: CacheManager;
     private _id: number;
     private static _nextId = 0;
+    private _editMode = false;
```
erictraut, line 190: Nit: rename to `_isEditMode` to make it obvious this is a boolean and not an enumeration or something like that.

packages/pyright-internal/src/analyzer/textEditUtils.ts
```diff
@@ -0,0 +1,24 @@
+import { Range, TextDocumentContentChangeEvent } from 'vscode-languageserver';
```
erictraut, line 1: This file is missing a header comment.

erictraut (commented):
Looking good.

packages/pyright-internal/src/analyzer/service.ts
rchiodo: The code in fix all (inside of pylance) is sort of async. It calls common functions that wait for code on the background thread if it exists. Meaning there could be sync versions of those. 

 I agree, we should just enforce synchronous code here. Pylance will just have to change to not make async calls when in edit mode.

packages/pyright-internal/src/analyzer/textEditUtils.ts
rchiodo: I realized I'm not using this anymore. I just deleted the file.

packages/pyright-internal/src/tests/workspaceEditUtils.test.ts
rchiodo: I changed the test to do a nested change. I didn't think it would matter if there was more than one, so I just added one extra nested change and verified we ended up with a single change.

packages/pyright-internal/src/analyzer/sourceFile.ts
```diff
@@ -536,23 +552,59 @@ export class SourceFile {
...
+    }
+
+    enterEditMode() {
+        this._isEditMode = true;
+    }
+
+    exitEditMode(): FileEditAction | undefined {
+        this._isEditMode = false;
+
+        // If we had an edit, return it
```
erictraut, line 581: Nit: include period at end of sentence.

packages/pyright-internal/src/analyzer/sourceFile.ts
```diff
@@ -536,23 +552,59 @@ export class SourceFile {
...
     getCheckTime() {
-        return this._checkTime;
+        return this._writableData.checkTime;
+    }
+
+    enterEditMode() {
+        this._isEditMode = true;
+    }
+
+    exitEditMode(): FileEditAction | undefined {
```
erictraut, line 578: I was hoping to remove knowledge of VSCode's FileEditAction and TextDocument from this file. Could we simply return the new contents as a string and let the caller construct the FileEditAction from it?

erictraut (commented):
Looking good. One minor consideration. See comment.

packages/pyright-internal/src/analyzer/sourceFile.ts
rchiodo: Sure that works. You mean at the program level?

erictraut: Actually, I should have said "languageServerBase", not "service".

rchiodo: Something needs the file path though? You mean create another interface that just stores the content and the file path?

erictraut: Ah, I see what you mean. Yeah, then keeping it in the service (in the `useEditMode` method) makes sense.

erictraut (approved):

PR #5034, opened 2023-04-28
Add support for PEP 712

debonte:
Addresses https://github.com/microsoft/pylance-release/issues/4290

Adds support for a `converter` parameter on `dataclasses.field` as described in https://peps.python.org/pep-0712/.

PEP 712 is still a draft. We're adding this support so users can play with the typing behavior. If the PEP is approved, this support will be updated to match the final spec. If the PEP is rejected, this support will be removed.

Users would also need to modify their `dataclasses.pyi` stub file to add support for `converter`. While testing I created and used the stubs below. Also note that the PEP author (@thejcannon) has implemented CPython support in a [personal branch](https://github.com/thejcannon/cpython/tree/converter).

<details>
<summary>Prototype stub change</summary>

```python
    from typing import ParamSpec, Concatenate

    _P = ParamSpec("_P")
    _InputT = TypeVar("_InputT")
    _ReturnT = TypeVar("_ReturnT")

    @overload
    def field(
        *,
        default: _InputT,
        init: bool = True,
        repr: bool = True,
        hash: bool | None = None,
        compare: bool = True,
        metadata: Mapping[Any, Any] | None = None,
        kw_only: bool = ...,
        converter: Callable[Concatenate[_InputT, _P], _ReturnT]
    ) -> _ReturnT: ...
    @overload
    def field(
        *,
        default: _T,
        init: bool = True,
        repr: bool = True,
        hash: bool | None = None,
        compare: bool = True,
        metadata: Mapping[Any, Any] | None = None,
        kw_only: bool = ...,
    ) -> _T: ...
    @overload
    def field(
        *,
        default_factory: Callable[[], _InputT],
        init: bool = True,
        repr: bool = True,
        hash: bool | None = None,
        compare: bool = True,
        metadata: Mapping[Any, Any] | None = None,
        kw_only: bool = ...,
        converter: Callable[Concatenate[_InputT, _P], _ReturnT]
    ) -> _ReturnT: ...
    @overload
    def field(
        *,
        default_factory: Callable[[], _T],
        init: bool = True,
        repr: bool = True,
        hash: bool | None = None,
        compare: bool = True,
        metadata: Mapping[Any, Any] | None = None,
        kw_only: bool = ...,
    ) -> _T: ...
    @overload
    def field(
        *,
        init: bool = True,
        repr: bool = True,
        hash: bool | None = None,
        compare: bool = True,
        metadata: Mapping[Any, Any] | None = None,
        kw_only: bool = ...,
        converter: Callable | None
    ) -> Any: ...
```
</details>


packages/pyright-internal/src/analyzer/dataClasses.ts
```diff
@@ -654,6 +673,93 @@ export function synthesizeDataClassMethods(
...
+function getConverterInputType(
+    evaluator: TypeEvaluator,
+    converterNode: ArgumentNode,
+    fieldType: Type,
+    fieldName: string
+): Type {
+    const converterType = evaluator.getTypeOfExpression(converterNode.valueExpression).type;
+
+    if (!isFunction(converterType) && !isOverloadedFunction(converterType)) {
+        return fieldType;
```
erictraut, line 687: Do you want to emit a diagnostic in this case rather than simply bailing? I could argue both sides of that.

I mentioned previous that it's probably OK to ignore unions for now. One other case that might be more common is a callable object. That would also be the same case as a object whose type conforms to a callback protocol. It may be fine to ignore for now and see if it ever arises.

packages/pyright-internal/src/analyzer/dataClasses.ts
```diff
@@ -654,6 +673,93 @@ export function synthesizeDataClassMethods(
...
+    }
+
+    // Create synthesized function of the form Callable[[T], fieldType] which
+    // will be used to check compatibility of the provided converter.
+    const typeVar = TypeVarType.createInstance('__converterInput');
+    typeVar.scopeId = evaluator.getScopeIdForNode(converterNode);
+    const targetFunction = FunctionType.createSynthesizedInstance('');
+    targetFunction.details.declaredReturnType = fieldType;
```
erictraut, line 695: You should set the details.typeVarScopeId for targetFunction as well. It should share the scope ID with its TypeVar.

packages/pyright-internal/src/analyzer/dataClasses.ts
```diff
@@ -654,6 +673,93 @@ export function synthesizeDataClassMethods(
...
+                argType: evaluator.printType(converterType),
+                fieldType: evaluator.printType(fieldType),
+                fieldName: fieldName,
+            }) + diagAddendum.getString(),
+            converterNode,
+            diagAddendum.getEffectiveTextRange() ?? converterNode
+        );
+    } else {
+        const acceptedTypes: Type[] = [];
+        const diagAddendums: DiagnosticAddendum[] = [];
```
erictraut, line 730: The DiagnosticAddendum is designed to handle multiple messages. Rather than accumulating an array of them like this, you should allocate one DiagnosticAddendum before the loop and use it repeatedly. The `assignType` will add all of the messages to it, and getString() will perform any necessary indentation and joining.

packages/pyright-internal/src/analyzer/dataClasses.ts
```diff
@@ -654,6 +673,93 @@ export function synthesizeDataClassMethods(
...
+
+        OverloadedFunctionType.getOverloads(converterType).forEach((overload) => {
+            const typeVarContext = new TypeVarContext(typeVar.scopeId);
+            const diagAddendum = new DiagnosticAddendum();
+
+            if (evaluator.assignType(targetFunction, overload, diagAddendum, typeVarContext)) {
+                const overloadSolution = applySolvedTypeVars(typeVar, typeVarContext, { unknownIfNotFound: true });
+                acceptedTypes.push(overloadSolution);
+            } else if (!diagAddendum.isEmpty()) {
+                diagAddendums.push(diagAddendum);
```
erictraut, line 740: This "else if" clause can be deleted if you don't accumulate diagAddendums manually.

packages/pyright-internal/src/localization/package.nls.en-us.json
```diff
@@ -77,6 +77,8 @@
         "constructorNoArgs": "Expected no arguments to \"{type}\" constructor",
         "dataClassBaseClassFrozen": "A non-frozen class cannot inherit from a class that is frozen",
         "dataClassBaseClassNotFrozen": "A frozen class cannot inherit from a class that is not frozen",
+        "dataClassConverterFunction": "Argument of type \"{argType}\" is not an acceptable converter for field \"{fieldName}\" of type \"{fieldType}\"",
+        "dataClassConverterOverloads": "None of the overloads of \"{funcName}\" are acceptable converters for field \"{fieldName}\" of type \"{fieldType}\"",
```
erictraut, line 81: Grammar nit: Either change to "None ... _is_ acceptable" or (perhaps better) change to "No overloads ... are acceptable".

Also, the word "acceptable" strikes me as a bit weird in these errors. We don't use that term anywhere else. It sounds a bit condescending to my ears. Perhaps switch to "valid" or "invalid".

packages/pyright-internal/src/tests/samples/dataclass25.py
```diff
@@ -0,0 +1,84 @@
...
+    return s
+
+def wrong_number_of_params(x: str, x2: str, /) -> int:
+    return 1
+
+@overload
+def wrong_converter_overload(s: float) -> str: ...
+
+@overload
+def wrong_converter_overload(s: str) -> str: ...
```
erictraut, line 75: In the test cases I write, I include a comment before each line where I expect an error to appear and the reason why. That way, it's easy to tell at a glance when opening the test whether the reported diagnostics match expectations.

erictraut (commented):
Looks really good. Just a few minor suggestions.

thejcannon:
Wow I'm flattered! I'll play around with this early next week.

🙌

packages/pyright-internal/src/analyzer/dataClasses.ts
debonte: Ah, I had naively thought that the `argAssignmentParam` would be reported in scenarios that I wasn't supporting.

Can you give me a code sample showing a union of functions/overloads? I wasn't able to figure out what syntax you were thinking of.

Here's what I found while playing with `Callable` and `Protocol` objects:

### Callable class instances
Instances of objects with `__call__` methods, as in the example below, don't work currently. Accepting `converterType` values that are class instances and validating them the same way I'm validating functions would help, but I'd also need to check if `__call__` is overloaded and special case that in the `OverloadedFunction` code path.

Is there some more general way to handle this? I saw `isCallableType` and thought that might help, but I'd still need a way to determine if there were any overloads in play and obtain them. Maybe some additional helper functions would be useful -- `isOverloadedCallableType` / `getOverloadsOfCallableType`?

Btw, I also noticed that if you remove the assignment of `callback`, the `TypeCategory` of the `converter` arg expression becomes `Unbound` and we fall back to returning `fieldType`. However, we also show the `reportUnboundVariable` diagnostic in that case, so maybe that's sufficient.

```python
class Callback(Protocol):
    def __call__(self, arg1: str) -> int: ...

class CallableObject(Callback):
    def __call__(self, arg1: str) -> int:
        return 1

callback: Callback = CallableObject()

@dataclass
class Foo:
    field5: int = field(converter=callback)
```

### Function as a Protocol
If I assign a function to the `Protocol` variable instead, validation of params, return value, etc work because we know that it is a function.

```python
class Callback(Protocol):
    def __call__(self, arg1: str) -> int: ...

def function1(arg1: str) -> int:
    return 1

callback: Callback = function1 

@dataclass
class Foo:
    field5: int = field(converter=callback)
```

### Callable
Similarly, validation works in the following `Callable` scenario, although if `callable` is unassigned we don't show the `reportUnboundVariable` diagnostic as we do for the `Protocol` case above, but instead get `object` from the `applySolvedTypeVars` call and use that as the `effectiveType` of the field. 

Is the difference in behavior with these two "unassigned" scenarios expected? Should an unassigned `Callable` also have `TypeCategory.Unbound`?

```python
def function1(arg1: str) -> int:
    return 1

callable: Callable[[str], int] = function1

@dataclass
class Foo:
    field5: int = field(converter=callable)
```

erictraut: Here's an (admittedly contrived) example of a union type:

```python
def get_class(converter: Callable[[str], str] | Callable[[int], str]):
    @dataclass
    class Foo:
        field5: int = field(converter=converter)

    return Foo
```

Here's another variant:

```python
support_str = load_config()
def fn1(val: int) -> int: ...
def fn2(val: str) -> int: ...

@dataclass
class Foo:
    field5: int = field(converter=fn2 if support_str else fn1)
```

Like I said, it's probably fine to ignore unions for now.

A an object with a `__call__` method internally is a `ClassType`, so your current logic will reject it. In particular, it's an instantiated class (i.e. `isClassInstance` will return true). You can call `getBoundMethod(type, '__call__')` to determine whether an instantiated class has a `__call__` method, and if so what its signature is. This will return a `FunctionType`, `OverloadedFunctionType` or `undefined` depending on the type and presence of a `__call__` method.

919o:
wow. that look like a good one!

erictraut (approved):
LGTM

919o:
what do your a blender addon?

PR #5046, opened 2023-05-01
Fix broken link in `mypy-comparison.md`

debonte:
Fixed broken link to conditional types docs in `mypy-comparison.md`.

### DIFF
--- a/docs/mypy-comparison.md
+++ b/docs/mypy-comparison.md
@@ -352,7 +352,7 @@ def func(a: AnyStr, b: T):
     return a + b # Mypy reports 4 errors
 ```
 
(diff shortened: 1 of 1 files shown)


docs/mypy-comparison.md
```diff
@@ -430,7 +430,7 @@ Pyright honors class decorators. Mypy largely ignores them. See [this issue](htt
 
 Versions of Python prior to 3.0 did not have a dedicated syntax for supplying type annotations. Annotations therefore needed to be supplied using “type comments” of the form `# type: <annotation>`. Python 3.6 added the ability to supply type annotations for variables. 
 
-Mypy has full support for type comments. Pyright supports type comments only in locations where there is a way to provide an annotation using modern syntax. Pyright was written to assume Python 3.5 and newer, so support for older versions was not a priority.
+Mypy has full support for type comments. Pyright supports type comments only in locations where there is no way to provide an annotation using modern syntax. Pyright was written to assume Python 3.5 and newer, so support for older versions was not a priority.
```
erictraut, line 433: This change is incorrect. It should be left as is. The only place where type comments are allowed is in locations where a modern type annotation could be used. For example, pyright doesn't support the use of type comments to annotate the types of variables defined in a tuple, as in `(a, b) = (1, 2) # type (int, int)`.

erictraut (changes requested):

erictraut (approved):

PR #5055, opened 2023-05-03
Fix extra references being returned for duplicate symbols

rchiodo:
Addresses https://github.com/microsoft/pylance-release/issues/4285

In the bug the same symbol is defined in two different modules and because both modules are referenced as implicit imports, both symbols get pulled in.

The fix here is to skip declarations that resolve to a different module than the original position.

All of the changes were in pyright, so I submitted the PR here.


heejaechang:
I think actual bug is `collectImportedByFiles` being used for non notebook files. I think we need to make `collectImportedByFiles` to check whether given file's `ipythonMode` is `Cell` otherwise, return nothing.

that should block it from gathering all `import alias` this file is `imported into`

PR #5142, opened 2023-05-19
`TextRange.combine` should not modify input ranges

debonte:
Addresses https://github.com/microsoft/pylance-release/issues/4393

`SmartSelectionProvider` calls `TextRange.combine` on `ParseNode` arrays. Currently that causes the `start` and/or `length` of the first node in the array to be changed to the combined range's length.

This would be an easy mistake to make in other places. In fact, I believe these other two usages are also resulting in unintended `ParseNode` changes:
- https://github.com/microsoft/pyright/blob/fedcafb706c335b17ca2617d3d2d1ab07030f51f/packages/pyright-internal/src/analyzer/typeEvaluator.ts#LL15171
- https://github.com/microsoft/pyrx/blob/main/packages/pylance-internal/src/languageService/completionProvider.ts#L579

So I've modified `combine` to leave the input ranges unchanged.

I looked through the other calls to `TextRange.combine` and none of them are relying on the current behavior.

### DIFF
--- a/packages/pyright-internal/src/common/textRange.ts
+++ b/packages/pyright-internal/src/common/textRange.ts
@@ -77,7 +77,7 @@ export namespace TextRange {
             return undefined;
         }
 
-        const combinedRange = ranges[0];
+        const combinedRange = { start: ranges[0].start, length: ranges[0].length };
         for (let i = 1; i < ranges.length; i++) {
             extend(combinedRange, ranges[i]);
         }
--- a/packages/pyright-internal/src/tests/textRange.test.ts
+++ b/packages/pyright-internal/src/tests/textRange.test.ts
@@ -1,0 +1,26 @@
+/*
+ * textRange.test.ts
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ */
+
+import * as assert from 'assert';
+
+import { TextRange } from '../common/textRange';
+
+test('textRange combine', () => {
+    const range1 = TextRange.create(10, 2);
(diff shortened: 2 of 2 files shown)


packages/pyright-internal/src/tests/textRange.test.ts
```diff
@@ -0,0 +1,29 @@
+/*
```
erictraut, line 1: I'm not sure we need a test for this. If this functionality doesn't work, it will be pretty obvious immediately. Each new test adds time and compute for every subsequent change, so I try to add tests only for things that have the potential for regressions that would otherwise go undetected. It's fine to leave this test here, but something to keep in mind for the future.

packages/pyright-internal/src/common/textRange.ts
```diff
@@ -77,7 +77,7 @@ export namespace TextRange {
             return undefined;
         }
 
-        const combinedRange = ranges[0];
+        const combinedRange = { start: ranges[0].start, length: ranges[0].length };
```
erictraut, line 80: Yes, good call.

erictraut (approved):

heejaechang:
it would be nice if all these parse node is marked as `readonly` outside of `parser.ts`

debonte:
> it would be nice if all these parse node is marked as `readonly` outside of `parser.ts`

We're tracking that work in general with https://github.com/microsoft/pyrx/issues/3333. It would be awesome to improve this. We've had several bugs of this type over the past few months.

PR #5152, opened 2023-05-22
Preserve parentheses around unary operators for printing

LyricLy:
The current parser does not keep track of whether or not there are parentheses around unary operators. This causes ambiguity when it comes to converting the parse tree back into a string, as there are some operators tighter than unary operators; for example, `reveal_type((-2) ** 2)` currently incorrectly reports the expression to be `-2 ** 2`, which is actually equivalent to `-(2 ** 2)`.

This simple change makes it so that the nodes for unary operators and `await` expressions store whether or not they were parenthesized in the same manner that binary operators already did, fixing the printing of expressions such as `(await f())[0]`.


erictraut:
Thanks for the contribution. The change looks good to me. I'll merge as soon as the CI tests pass.

LyricLy:

PR #5184, opened 2023-05-26
flatten loc files

bschnurr:
to better align with vscode/l10n-dev localization toollchain:

https://github.com/microsoft/vscode-l10n/tree/main/l10n-dev


rchiodo (approved):

erictraut:
What is the reason for this change? Are you preparing to add localized support for more languages? If there's not a good reason to make the change, I'd prefer to keep it as is. The new technique takes up a bunch of extra memory because it includes a bunch of redundant strings.

bschnurr:
Yes this is to add localization..

I think I could add a transformation to the localization pipeline to convert between our version and their expected format instead then.

PR #5238, opened 2023-06-05
Update typed-libraries.md

LiamSwayne:
The first paragraph of this page of documentation has been modified to be less wordy and more informative. It now provides a more detailed overview of this page of documentation with the same number of words as before.


erictraut:
I appreciate your interest in pyright, but I think the current wording in the documentation is preferable to your changes. I'm therefore going to leave it unmodified.

PR #5259, opened 2023-06-09
Update ci-integration.md

last-partizan:
When both scripts are executed in `scripts:` section, and `pyright` reports error, second script is not executed.

Best way to ensure report are always converted, is to use `after_script:` block.


erictraut:
Thanks for the contribution.

PR #5291, opened 2023-06-14
Changes for loc

bschnurr:
- adding a publisher name to pyright-internal to match vscode-pyright.  Loc pipeline uses this publisher name.
 - adding dev dependency on vscode's loc tools
 - current plan is to have pylance run the localization update for both repos


erictraut (approved):
LGTM

bschnurr:
i've reworked the loc pipeline so these changes are no longer needed. benefits is one less package to npm install

Issue #5307, opened 2023-06-15
Support PEP 712's new attribute assignment conversion


thejcannon:
**Describe the bug**
In the super cool experimental [PEP 712](https://peps.python.org/pep-0712/) support, we'll need to support attribute assignment as well: https://github.com/python/peps/pull/3152

**To Reproduce**
Just try and assign to a dataclass field with a converter and not use the field's type (but a supported conversion type)

**Expected behavior**
No error

**Code or Screenshots**

```python
from dataclasses import dataclass, field


@dataclass
class Foo:
    x: str = field(default=1, converter=str)


f = Foo(1)
f.x = 1
#  /I'm/a/path/tofoo.py:10:7 - error: Cannot assign member "x" for type "Foo"
#   "Literal[1]" is incompatible with "str" (reportGeneralTypeIssues)
```

**VS Code extension or command-line**
`pyright 1.1.314` :pie: 

**Additional context**
<3 y'all

erictraut:
My understanding is that the converter runs only on the values passed at initialization time. Is that not the case? The current PEP says nothing about converting values after initialization.

It's not immediately clear from the `attrs` documentation either, but I just verified through experimentation that it indeed does support conversion when assigning to an attribute with an associated field. I guess that means we need to support it here. I recommend adding more details to the PEP to make this clear.

I think the most straightforward (and accurate) way to handle this is to synthesize descriptor objects for each of the defined attributes. The `__set__` method for these descriptor objects will accept any type accepted by the converter. The `__get__` method would return the transformed type (the output of the converter).

@debonte, we can talk about the details later today, assuming that you want to implement this.

thejcannon:
> My understanding is that the converter runs only on the values passed at initialization time. Is that not the case? The current PEP says nothing about converting values after initialization.

You got that right. The new behavior is in [the linked PR](https://github.com/python/peps/pull/3152) (here's the PEP link if you want fewer clicks: https://pep-previews--3152.org.readthedocs.build/pep-0712/)

And your modeling this as a descriptor is likely a good way to model the world. However, wouldn't the `__get__` method be better modeled as returning the annotated attribute type since that is a superset of the converter return type?

erictraut:
[...]

PR #5343, opened 2023-06-19
Support PEP 712's new attribute assignment conversion

debonte:
Addresses #5307 

@erictraut, I'm not sure if this is working property for generic dataclasses when the descriptor's get and/or set types are generic. Currently I'm calling `makeTopLevelTypeVarsConcrete` on the get/set type when synthesizing the descriptor. Is this the right thing to do?

I initially tried to get this to work by passing the dataclass's `typeVarScopeId` to the descriptor. But then the revealed type of `g.field1` and `g.field2` at the end of the new test file was `Unknown` instead of `float`. It looked like this was because in `applyDescriptorAccessMethod` the return type of the synthesized descriptor's `__get__` method was not specialized (`_T` instead of `float`).


packages/pyright-internal/src/analyzer/dataClasses.ts
```diff
@@ -799,6 +812,86 @@ function getConverterAsFunction(
...
+        type: AnyType.create(),
+        hasDeclaredType: true,
+    });
+    FunctionType.addParameter(getFunction, {
+        category: ParameterCategory.Simple,
+        name: 'objtype',
+        type: AnyType.create(),
+        hasDeclaredType: true,
+    });
+    getFunction.details.declaredReturnType = evaluator.makeTopLevelTypeVarsConcrete(getType);
```
erictraut, line 888: See comment above about `makeTopLevelTypeVarsConcrete`.

packages/pyright-internal/src/tests/typeEvaluator4.test.ts
```diff
@@ -679,6 +679,12 @@ test('DataClass25', () => {
     TestUtils.validateResults(analysisResults, 17);
 });
 
+test('DataClass26', () => {
```
erictraut, line 682: I've been working on cleaning up the test cases, and things have changed a bit here. Please rebase before checking in this change. This new test case should be named `dataclassConverter2.py` to match the new test structure.

packages/pyright-internal/src/tests/samples/dataclass26.py
```diff
@@ -0,0 +1,54 @@
...
+foo.field0 = "2"
+
+# This should generate an error because only strs can be assigned to field0.
+foo.field0 = 2
+
+
+_T = TypeVar("_T", bound=float)
+
+@dataclass
+class GenericFoo(Generic[_T]):
```
erictraut, line 28: I'm thinking that we should not support generic converters for now. Your current code happens to work for this simple case, but it won't work for anything slightly more complicated like a converter that accepts or generates `list[_T]`.

packages/pyright-internal/src/tests/samples/dataclass26.py
```diff
@@ -0,0 +1,54 @@
...
+@dataclass
+class Foo:
+    # This should generate an error because "converter" is not an official property yet.
+    field0: int = field(converter=converter_simple)
+
+foo = Foo("1")
+reveal_type(foo.field0, expected_text="int")
+foo.field0 = "2"
+
+# This should generate an error because only strs can be assigned to field0.
```
erictraut, line 21: You might want to add a test case that verifies that type narrowing doesn't occur when assigning to a field with a converter. To do so, just copy the above `reveal_type` line after the `foo.field0 = "2"` assignment statement. It should still be considered an `int`, whereas normal narrowing rules would cause it to become a `str`.

erictraut: Another interesting test is the case where the converter doesn't actually do any conversion (i.e. it accepts the same type as it produces). That shouldn't be considered an asymmetric descriptor, so normal type narrowing rules would apply. For example, if the converter accepts a `str | int` and produces a `str | int` and you assign a value of "1" to the field, it should be narrowed to `Literal[1]`.

packages/pyright-internal/src/tests/samples/dataclass26.py
```diff
@@ -0,0 +1,54 @@
...
+foo.field0 = 2
+
+
```
erictraut, line 24: Another interesting test is to exercise the `__get__` and `__set__` methods by accessing the field from the class rather than a class instance. For example, what happens when you exec `Foo.field0 = ""`?

debonte: I added tests for this, but it's not clear to me if the behavior is as expected.

```python
# Asymmetric
reveal_type(Foo.asymmetric, expected_text="int")
Foo.asymmetric = "2"
reveal_type(Foo.asymmetric, expected_text="int")
Foo.asymmetric = 2
reveal_type(Foo.asymmetric, expected_text="Literal[2]")

# Symmetric
reveal_type(Foo.symmetric, expected_text="str | int")
Foo.symmetric = "1"
reveal_type(Foo.symmetric, expected_text="Literal['1']")
```

packages/pyright-internal/src/tests/samples/dataclassConverter2.py
```diff
@@ -0,0 +1,39 @@
...
+foo.symmetric = "1"
+reveal_type(foo.symmetric, expected_text="Literal['1']") # Symmetric -- type narrowing should occur
+
+
+reveal_type(Foo.asymmetric, expected_text="int")
+Foo.asymmetric = "2"
+reveal_type(Foo.asymmetric, expected_text="int")
+# This should generate an error because only strs can be assigned to field0.
+Foo.asymmetric = 2
+reveal_type(Foo.asymmetric, expected_text="Literal[2]")
```
erictraut, line 35: If you assign a value that generates a type violation (as in this case), pyright will narrow the type to the assigned type. This was a request/idea from Graham a while back. If the user has type checking disabled, it produces better (more intuitive) results for completion suggestions. I recommend deleting the `Literal[2]` reveal_type call because the narrowed type isn't really relevant once you assign an illegal value.

erictraut (approved):
LGTM

PR #5389, opened 2023-06-26
Create Traffic Singal Detection.py

iam100rav:
PS C:\Python Programing> & C:/Users/95sau/AppData/Local/Microsoft/WindowsApps/python3.11.exe "c:/Python Programing/Traffic singal detection.py" Traceback (most recent call last):
  File "c:\Python Programing\Traffic singal detection.py", line 2, in <module>
    from PIL import Image, ImageTk
ModuleNotFoundError: No module named 'PIL'
PS C:\Python Programing>

### DIFF
--- a/Traffic Singal Detection.py
+++ b/Traffic Singal Detection.py
@@ -1,0 +1,53 @@
+from tkinter import *
+from PIL import Image, ImageTk
+from tkinter import filedialog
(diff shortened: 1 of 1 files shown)


debonte:
Looks like you created a PR against pyright accidentally.

Issue #5415, opened 2023-06-30
Watch mode repeatedly prints `Error returned from file system watcher.` when virtual env is inside the directory being analyzed


adampauls:
**Describe the bug**
When running `pyright -w` on the command line, it correctly reports some errors then infinitely loops with `Error returned from file system watcher.` I'm on MacOS 13.4.1 (22F82). Happens both in iTerm2 and in the terminal within PyCharm. Perhaps it's at least worth exposing the error [here](https://github.com/microsoft/pyright/blob/c767228a82535b0bdd404480d6ffd5a1d24c8884/packages/pyright-internal/src/common/chokidarFileWatcherProvider.ts#L60)? I did so manually locally and see `Error: EMFILE: too many open files, watch`. It also only happens on larger projects, so [this comment](https://github.com/microsoft/pyright/blob/c767228a82535b0bdd404480d6ffd5a1d24c8884/packages/pyright-internal/src/common/chokidarFileWatcherProvider.ts#L42-L43) might be relevant. 

UPDATE: I verified that if I manually change [this line](https://github.com/microsoft/pyright/blob/c767228a82535b0bdd404480d6ffd5a1d24c8884/packages/pyright-internal/src/common/chokidarFileWatcherProvider.ts#L47) to 
```typescript
const excludes: string[] = ["**/node_modules/**", "**/__pycache__/**", "**/.*/**"];
```

then the problem is resolved. I see that plumbing through `configOptions.exclude` to the file system watcher constructor is not trivial, but hopefully someone more familiar with the code can do it quickly. 

**VS Code extension or command-line**
pyright 1.1.316 from the command line. 

**Additional context**
Only instance of the error message I could find on a Google search was [this bug](https://github.com/microsoft/pylance-release/issues/936).

adampauls:
Is there any chance that the pyright excludes mentioned [here](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#main-configuration-options) are not piped through to the file watcher [here](https://github.com/microsoft/pyright/blob/c767228a82535b0bdd404480d6ffd5a1d24c8884/packages/pyright-internal/src/common/chokidarFileWatcherProvider.ts#L47)? That would make sense, since we have a `.venv` folder inside the folder that we are running `pyright` on.

erictraut:
[...]

PR #5420, opened 2023-07-01
Exclude some common large directories in ChokidarFileWatchProvider

adampauls:
Temporary and hopefully inoffensive fix for https://github.com/microsoft/pyright/issues/5415. It would be better to pipe through `configOptions.exclude` but that's al arger change.


packages/pyright-internal/src/common/chokidarFileWatcherProvider.ts
```diff
@@ -44,7 +44,7 @@ export class ChokidarFileWatcherProvider implements FileWatcherProvider {
             watcherOptions.usePolling = false;
         }
 
-        const excludes: string[] = ['**/__pycache__/**'];
+        const excludes: string[] = ['**/node_modules/**', '**/__pycache__/**', '**/.*/**'];
```
rchiodo, line 47: Doesn't this exclude everything?

'**/.*/**'

rchiodo: oh no, it excludes any `.` folders, like `.git`. That seems okay?

adampauls: Yeah, I just copied from [here](https://github.com/microsoft/pyright/blob/0a67c65bef4ce95d4de8d2218a0bd875d2fb9230/packages/pyright-internal/src/analyzer/service.ts#L580)

adampauls:
Better to fix https://github.com/RobertCraigie/pyright-python/issues/191

PR #5503, opened 2023-07-14
Add fsevents as optional dependency to public package

adampauls:
I believe this should resolve https://github.com/microsoft/pyright/issues/5415, but I'm not an expert here. My understanding is that the published pyright package does the npm equivalent of statically linking the chokidar dependency, which means that fsevents is not pulled in if that linking is done on a machine that is not run on MacOS. 

This setup is not ideal since the version could drift from the one expected the chokidar embedded in `pyright`, but I think it's probably okay.

### DIFF
--- a/packages/pyright/package.json
+++ b/packages/pyright/package.json
@@ -22,6 +22,9 @@
         "prepack": "npm run clean && shx cp ../../README.md . && shx cp ../../LICENSE.txt . && npm run build",
         "webpack": "webpack --mode development --progress"
     },
+    "optionalDependencies": {
+        "fsevents": "~2.3.2"
+    },
     "devDependencies": {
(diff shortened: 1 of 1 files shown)


erictraut:
Please don't submit PRs if you're unsure of their effects or don't fully understand the problem. Issue #5415 requires deeper investigation. I'll get to that when I have a chance.

adampauls:
Yes. I tested by running `npm run install:all && npm run build:cli:dev`, then doing a local `npm install /path/to/local/pyright` in the `.cache` directory. I verified that repeating those steps without the change caused the error.

erictraut:
OK, that's pretty good evidence that this is a proper diagnosis and an effective fix. I'll merge your PR. Thanks!

Issue #5586, opened 2023-07-27
constructor types not recognized if meta class used


matt-the-midnight-hacker:
**Describe the bug**
I have a class with a constructor where a argument of certain type is expected e.g. `bool`. Pyright usually detects if I instantiate the class with a constructor argument of a wrong type. e.g. call it with a `string` instead of a `bool` type. If my class uses a metaclass which overwrites `__new__` and `__call__` then pyright can no longer detect the type of my class and suspects all arguments are from type `Any`.

**Code or Screenshots**

I hope this boiled down reproduction sample does dos not obscure the  issue I see in my actual code. I implemented here `__call__` and `__new__` minimal to trigger the behavior, My actual code does a little more, I believe type detection worked here once but not sure at which pyright version it stopped.

```python
class InitWithDefaults(type):
    def __call__(cls, *args: Any, **kwargs: Any):
        pass


class BaseMeta(InitWithDefaults):
    def __new__(mcs: Any, clsname: str):
        return type.__new__(mcs, clsname)


class Check(metaclass=BaseMeta):

    def __init__(self, success: bool) -> None:
        super().__init__()
        self.success: bool = success


c = Check(success="hallo")  # Here pyright does not complain about the wrong type

```

**VS Code extension or command-line**
I run pyright 1.1.318 in vscode.

erictraut:
This was an intentional change. I recently rewrote the code in pyright that validates constructor calls. It now more closely reflects the runtime behavior. For details about how this works, refer to [this documentation](https://microsoft.github.io/pyright/#/mypy-comparison?id=constructor-calls).

If a metaclass overrides `type.__call__`, a type checker cannot make assumptions about the behavior of the metaclass. The normal behavior of `type.__call__` is to invoke the `__new__` and `__init__` method of the class, but metaclasses that override `__call__` may do something different. For example, sometimes `__call__` overrides are used to prevent explicit instantiation of a class, in which case they raise an exception.

Why is your metaclass overriding `type.__call__`? Does it call through to `type.__call__`? If so, does it return the value returned by `type.__call__`?

matt-the-midnight-hacker:
Thank you for the fast answer and the information. The purpose in the code I work on is to "hook" the `__init__` function of each subclass to set default values for properties of specific types.

> Does it call through to `type.__call__`?

yes it does.

> does it return the value returned by `type.__call__`?

yes, it does after modifying the returned object.

If this is an intentional change, I guess I have to refactor this code then to have better type support.

[...]

PR #5605, opened 2023-07-29
Updated heuristics for how to handle custom metaclass `__call__` meth…

erictraut:
…ods based on feedback. Previously, if a metaclass `__call__` method was present, pyright assumed that `__new__` and `__init__` may not be called, so it ignored them. The new heuristic assumes that if the metaclass `__call__` returns a type that is consistent with the expected return type of `type.__call__` that it is probably mirroring the behavior of `type.__call__` and calling the class' `__new__` and `__init__` methods. This addresses https://github.com/microsoft/pyright/issues/5586.

### DIFF
--- a/packages/pyright-internal/src/analyzer/constructors.ts
+++ b/packages/pyright-internal/src/analyzer/constructors.ts
@@ -74,9 +74,6 @@ export function validateConstructorArguments(
     skipUnknownArgCheck: boolean,
     inferenceContext: InferenceContext | undefined
 ): CallResult {
-    // If there a custom `__call__` method on the metaclass, assume that it
-    // overrides the normal `type.__call__` logic and don't perform the usual
-    // __new__ and __init__ validation.
     const metaclassResult = validateMetaclassCall(
         evaluator,
         errorNode,
@@ -85,8 +82,17 @@ export function validateConstructorArguments(
         skipUnknownArgCheck,
         inferenceContext
     );
+
     if (metaclassResult) {
-        return metaclassResult;
+        const metaclassReturnType = metaclassResult.returnType ?? UnknownType.create();
+
+        // If there a custom `__call__` method on the metaclass that returns
+        // something other than Any or an instance of the class, assume that it
+        // overrides the normal `type.__call__` logic and don't perform the usual
+        // __new__ and __init__ validation.
+        if (metaclassResult.argumentErrors || !evaluator.assignType(convertToInstance(type), metaclassReturnType)) {
+            return metaclassResult;
+        }
     }
 
     // Determine whether the class overrides the object.__new__ method.
@@ -107,7 +113,7 @@ export function validateConstructorArguments(
 
     // If there is a constructor transform, evaluate all arguments speculatively
     // so we can later re-evaluate them in the context of the transform.
-    const returnResult = evaluator.useSpeculativeMode(useConstructorTransform ? errorNode : undefined, () => {
+    let returnResult = evaluator.useSpeculativeMode(useConstructorTransform ? errorNode : undefined, () => {
         return validateNewAndInitMethods(
             evaluator,
             errorNode,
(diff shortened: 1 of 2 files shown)


erictraut:
@Gobot1234, this change affects the `steam.py` code base. You'll need to modify your `EnumType` metaclass to prevent these type checking errors with the next release of pyright (1.1.320). 

Currently, you're lying to pyright about the presence of a custom `__call__` method. If you move this outside of the `not TYPE_CHECKING` block, the errors go away.

```python
class EnumType(_EnumMeta if TYPE_CHECKING else type):
    ...
    if not TYPE_CHECKING:
        def __call__(cls, value: Any) -> Enum:
            ...
```

Gobot1234:
Would there be a way to keep the special enum casing for it though where it returns the instance of the enum? Cause as is this is done so that calling Result (an enum defined in that file) returns an instance of Result rather than an arbitrary instance of Enum

erictraut:
Yes, that's possible with pyright's updated heuristics.

Gobot1234:
What should I change the code to look like to get that behaviour?

Gobot1234:
@erictraut `_EnumMeta` here should have a custom `__call__` method synthesised by pyright as this is inheriting from the stdlib's `enum.EnumType` which has a custom `__call__` which returns an instance of the current type (which isn't currently annotate-able)

erictraut:
I assume you're talking about `enum.EnumMeta`? There has never been a need to synthesize a `__call__` in the past for this stdlib class. Doing so would require a significant justification. Why do you think this is necessary?

Could you provide a minimal code sequence that demonstrates the behavior that you would like to see?

Pyright's current heuristics should restore the previous type evaluation behavior — and make it consistent with the behavior seen with other type checkers.

Gobot1234:
Currently if you normally inherit from Enum you get this behaviour which is correct
image
However this caused a regression for my Enum class which should look like a subclass of Enum and an instance of a subclass of EnumMeta/EnumType (it got renamed in 3.12) 
image

Gobot1234:
```py
import enum
from typing import Self


class EnumType(enum.EnumMeta):
    ...


class Enum(enum.Enum, metaclass=EnumType):
    def __new__(cls, value: object, extra: object) -> Self:
        ...


class Colours(enum.Enum):
    RED = 1
    BLUE = 2


Colours(1)  # fine


class MyColours(Enum):
    RED = 1
    BLUE = 2

MyColours(1)  # not fine
```
This is what the example looks like to the type checker, this doesn't work at runtime but I'd like to keep this as small as possible

erictraut:
Am I correct in assuming that `EnumType` implements a custom `__call__` method that injects the argument for the `extra` parameter in `Enum.__new__`?

The code above doesn't type check in mypy either (as I would expect). Have you come up with any solution that works with mypy?

Gobot1234:
The way that python handles `type.__call__` vs `object.__new__` meant that `Enum_.__new__` is never called as it is shadowed by `EnumType_.__new__` (which is `enum.EnumMeta` to the type checker *) at runtime which was meant to perform a dictionary lookup that returns the member with the same value as passed 
see: https://github.com/Gobot1234/steam.py/blob/a634e54f21607815bc869245ade72af348d95428/steam/enums.py#L144-L148 which, as a blackbox does the same thing as `enum.Enum`

\* If you want to know why this is useful, removing the if not TYPE_CHECKING guard
```py
class MyEnum(Enum):
    RED = 1

reveal_type(MyEnum(1))  # type is Enum, not MyEnum which it is/was with if not TYPE_CHECKING
```

erictraut:
You're doing something pretty unusual here. Pyright isn't going to "synthesize an `__init__` method" specifically for your library or your use case, so that's not an option. Clearly you're going to need to use some hackery if you want this to work with static type checkers.

You didn't answer my question about mypy. Have you found a way to make it work there?

Gobot1234:
I haven't managed to make this work with mypy, I don't really use mypy with my codebase as it doesn't support a fair few features that I use regularly. 

I agree what I'm doing does seem unusual but I would have expected that pyright, when resolving a call expression would check for the `type.__call__` method all the way through the MRO, and if it wasn't `type.__call__` to use that (`enum.EnumMeta`s `__call__`) and then if it is `type.__call__` to then go to my custom `Enum`'s `__new__` method

erictraut:
If this is for a public library, then I would think that it's important for it to work with other type checkers. If it's for a private code base, then you have much more freedom for workarounds — including, for example, not relying on a pseudo constructor call to access an enum element. Why not simply implement a class method that allows you to access an enum member by value?

There's no way for a type checker to know whether a custom metadata `__call__` will do the "normal thing" and call through to `__new__` or do something completely different. Pyright's current heuristics work with the vast majority of libraries. Yours is the only one that I've seen so far. Any defensible (principled) change to the heuristic that would make this work for your use case would break many other libraries, so that wouldn't make sense.

Gobot1234:
> If this is for a public library, then I would think that it's important for it to work with other type checkers

Running my code through mypy raises `AssertionError: Internal error: unresolved placeholder` currently so I don't want to go through the effort to make it work unfortunately.

> Why not simply implement a class method that allows you to access an enum member by value?

I was originally just following the enum spec so implemented the same interface.

> There's no way for a type checker to know whether a custom metadata `__call__` will do the "normal thing" and call through to `__new__` or do something completely different. Pyright's current heuristics work with the vast majority of libraries. Yours is the only one that I've seen so far. Any defensible (principled) change to the heuristic that would make this work for your use case would break many other libraries, so that wouldn't make sense.

Thanks for explaining and totally understood. I've figured out a way around this by just sticking the `__new__` in an `if not TYPE_CHECKING` block to further improve the hackiness of my code :wink:.

PR #5617, opened 2023-08-01
Fixed a bug that led to a false negative when determining whether two…

erictraut:
… base classes that are combined using multiple inheritance implement the same method in an incompatible way. The bug caused the check to be skipped if the child class also implemented the same method as the two base classes. This addresses https://github.com/python/mypy/issues/15790.

### DIFF
--- a/packages/pyright-internal/src/analyzer/checker.ts
+++ b/packages/pyright-internal/src/analyzer/checker.ts
@@ -5177,7 +5177,7 @@ export class Checker extends ParseTreeWalker {
             return getClassFieldsRecursive(specializedBaseClass);
         });
 
-        const childClassSymbolMap = getClassFieldsRecursive(classType);
+        const childClassSymbolMap = getClassFieldsRecursive(classType, /* skipInitialClass */ true);
 
         for (let symbolMapBaseIndex = 1; symbolMapBaseIndex < baseClassSymbolMaps.length; symbolMapBaseIndex++) {
             const baseSymbolMap = baseClassSymbolMaps[symbolMapBaseIndex];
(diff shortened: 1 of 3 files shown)


DiscordLiz:
I think it was premature to call this a bug. A subclass defined in this way is replaceably usable with each parent class.

DiscordLiz:
AB here is usable as a drop in replacement for both A and B

```py
class ParamA:
    ...

class ParamB:
    ...

class ParamAB(ParamA, ParamB):
    ...

class A:
    def foo(self, x: ParamA) -> ParamA:
        ...

class B:
    def foo(self, x: ParamB) -> ParamB:
        ...

class AB(A, B):
    @overload
    def foo(self, x: ParamA) -> ParamA:
        ...

    @overload
    def foo(self, x: ParamB) -> ParamB:
        ...

    def foo(self, x):
        if isinstance(x, ParamA):
            return ParamA()
        else:
            return ParamB()
```

DiscordLiz:
There is disucssion about the soundness of this and the need for `Not` as well to correctly express the actual type which has come up in the intersections pep discussions

DiscordLiz:
Copying the relevant comment portion: (note & for Intersection, ~ for Not)

a more expressive overload set for AB.foo would be

(overload) (ParamA & ParamB) -> Never
(overload) (ParamA & ~ParamB) -> ParamA
(overload) (ParamB & ~ParamA) -> ParamB

But if someone had an implementation that was

(overload) (ParamA & ParamB) -> ParamA & ParamB
(overload) (ParamA & ~ParamB) -> ParamA
(overload) (ParamB & ~ParamA) -> ParamB

it would also be valid.

As we don't currently have a way to express this, this should not be treated as incorrect, as it is possible to implement both of the behaviors the overload sets express.

PR #5627, opened 2023-08-01
Filter import suggestions when py.typed is available

rchiodo:
Addresses https://github.com/microsoft/pylance-release/issues/4528

Apply some extra filtering to import suggestions when py.typed is available.


erictraut (approved):
LGTM

PR #5644, opened 2023-08-03
Support zip imports from `.jar` files

debonte:
Addresses https://github.com/microsoft/pylance-release/issues/4648

The user is consuming Python libraries that are distributed in `.jar` files. Python's zip import feature supports any zip file regardless of file extension. We currently only support `.zip` and `.egg`.

This adds support for `.jar` files. It would be nice to support any extension, but `readdirEntriesSync` appears to be using the file extension check to avoid calling `yarnFS.isZip` on every file and I'm concerned about the perf implications of removing that check. We can reconsider in the future if users request support for a wider variety of extensions.


packages/pyright-internal/src/common/realFileSystem.ts
```diff
@@ -62,7 +66,7 @@ function getArchivePart(path: string) {
 }
 
 function hasZipOrEggExtension(p: string): boolean {
```
debonte, line 68: Thoughts on if/how to change the names of `hasZipOrEggExtension` and `isInZipOrEgg`?

`hasZipOrEggOrJarExtension` seems silly. I'm leaning towards removing the `OrEgg` part. Or just leaving the names as is.

rchiodo: My vote is remote the 'orEgg' part.

`hasZiplikeExtension`?

rchiodo (approved):

packages/pyright-internal/src/common/realFileSystem.ts
```diff
@@ -46,7 +47,10 @@ function getArchivePart(path: string) {
     if (idx <= 0) {
         idx = path.indexOf(DOT_EGG);
         if (idx <= 0) {
-            return null;
+            idx = path.indexOf(DOT_JAR);
```
erictraut, line 50: This technique for determining the file extension is not very robust. It doesn't properly handle file names with multiple dots. It also doesn't handle names like "x.zipx" or "a.jarxxxxx".

There is a function in pathUtils called `getFileExtension` that handles proper extraction of file extension. You can also use `stripFileExtension` if you want to strip the extension.

debonte: This function also supports paths within a zip file, treating the zip like a directory. For example: `c:\dir\file.zip\module.py`.

The second-to-last line verifies that the extension is either at the end of the path or is immediately followed by the path separator char.

So I believe it is properly handling the scenarios you mentioned and `getFileExtension` won't help here.

BertsLifee (approved):

packages/pyright-internal/src/common/realFileSystem.ts
debonte: @erictraut, based on my comment above are you ok with this change?

erictraut: Yep

PR #5656, opened 2023-08-06
Add support for finding definitions and hover text on TypedDict variable indexing

### LINKED ISSUE #5654: Show key type info when hovering over an index lookup on TypedDicts
## Current Behaviour
When hovering over an index operation into a `TypedDict` variable no type info for the key appears and definition is unavailable.

In contrast constructing a `TypedDict` object hovering over keys there does produce type info and definition.

```
from typing import TypedDict

class Test(TypedDict):
    field: str

test: Test = {
    'field': 'value',  # Hovering over 'field' gives type info and the definition is available
}
test['field']  # Hovering over 'field' in this case gives no info and no definition is found
```

## Expected Behaviour
Hovering over a `TypedDict` key within an indexing operation should show the key type info and the definition should be available.
Screenshot from 2023-08-06 14-12-01

## Motivating Use Case
Plain dictionaries often appear in legacy code bases and when migrating types into these code bases it is easier and less error prone to write `TypedDict`s to describe these dictionaries instead of migrating them into dataclasses or similar (IMO). Which do show type info when hovering over the property names.

Migrating this way works well to verify that all accesses and definitions are valid. However when writing code that indexes into one of these dictionaries, finding either the type, docstring or definition of a key is cumbersome. Since the developer must find the type of the indexed object the and manually scan through to the relevant key.

### Hover/Definition search for `TypedDict.get`
`.get` is often used when keys are marked `NotRequired` to skip verbose `'key' in obj` checks. It may improve usability if key definitions were easily available.

```python
from typing import TypedDict, NotRequired

class Test(TypedDict):
    field: NotRequired[str]

test: Test = {}
test.get('field')  # Hovering over 'field' could show type info for the 'field' key
```

### Hover/Definition search for `NamedTuple` indexing
The same functionality for `TypedDict` could be applied to `NamedTuple`. The use case is weaker however, since the field can be directly referenced by name instead of by index.

```python
from typing import NamedTuple

class Test(NamedTuple):
	first: int
	second: str

test = Test(1, 'second')
res = test[1]  # Hovering over 1 could show type info for the 'second' key
```

R-O-C-K-E-T:
Fixes #5654 

image

### DIFF
--- a/packages/pyright-internal/src/analyzer/typeEvaluator.ts
+++ b/packages/pyright-internal/src/analyzer/typeEvaluator.ts
@@ -19465,15 +19465,39 @@ export function createTypeEvaluator(importLookup: ImportLookup, evaluatorOptions
         return undefined;
     }
 
+    // Find the type of the dictionary that this string node is used as a key for.
+    // This can be in either a dictionary expression or an index operation.
+    function getDictionaryTypeForStringNode(node: StringNode): Type | undefined {
+        // If this a string key for a dictionary expression return the expected dictionary type
+        const expectedType = getExpectedType(node)?.type;
+        if (expectedType) {
+            return expectedType;
+        }
+
+        // If the string is indexing into an existing dictionary return the resolved dictionary type
+        if (
+            node.parent?.nodeType === ParseNodeType.StringList &&
+            node.parent.strings.length === 1 &&
+            node.parent.parent?.nodeType === ParseNodeType.Argument &&
+            node.parent.parent.parent?.nodeType === ParseNodeType.Index
+        ) {
+            const indexNode: IndexNode = node.parent.parent.parent;
+            const indexedType = getType(indexNode.baseExpression);
+            return indexedType;
+        }
+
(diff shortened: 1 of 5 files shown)


R-O-C-K-E-T:

erictraut:
@R-O-C-K-E-T, if you would like to add new features to pyright, we'd prefer that you propose the feature first in an enhancement request and/or a discussion topic. 

@rchiodo or @debonte, since this is a language server feature, could you please review this change and decide whether it's something you'd like to accept?

erictraut:
Actually, the change is in the core type evaluator code, and the implementation is not consistent with how I recommend implementing such a feature.

@R-O-C-K-E-T, I'm going to close the PR. If you'd like to see this feature, please propose it as an enhancement request in the pyright-release project. The pylance team will then decide if it's a feature they'd like to include in pylance. If it's something they'd like to include, I'll work with them to implement it in the core type checker in a way that's maintainable and complete.

PR #5786, opened 2023-08-22
Discard `TypeEvaluator` if cancellation leaves partially-constructed class type during `getTypeOfClass`

debonte:
Addresses https://github.com/microsoft/pylance-release/issues/4684

The bug repros when an `OperationCanceledException` is thrown within `getTypeOfClass`. There are two sets of `writeTypeCache` calls in `getTypeOfClass`. One set before `classType.details.fields` is assigned, and one after. If cancellation occurs between these two sets of `writeTypeCache` calls, we leave an incomplete `ClassType` in the type cache.

Fixed by catching `OperationCanceledException` that occur during that window and setting a new `isTypeCacheInvalid` boolean on the exception to `true` so when the exception is caught we know to discard the current `TypeEvaluator`.

Also replaced two calls to `TypeEvaluator.runWithCancellationToken` with calls to `_runEvaluatorWithCancellationToken` to ensure that we're calling `_createNewEvaluator` as needed for exceptions. Hover, for example, was calling `Program.run`.

### DIFF
--- a/packages/pyright-internal/src/analyzer/program.ts
+++ b/packages/pyright-internal/src/analyzer/program.ts
@@ -20,6 +20,7 @@ import { Diagnostic } from '../common/diagnostic';
 import { FileDiagnostics } from '../common/diagnosticSink';
 import { FileEditAction } from '../common/editAction';
 import { EditableProgram, Extensions, ProgramView } from '../common/extensibility';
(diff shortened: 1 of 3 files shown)


erictraut (approved):
LGTM

PR #5813, opened 2023-08-25
Make typeguard non experimental just to test mypy primer

rchiodo:
This is just an experiment for the https://github.com/python/peps/pull/3266. This is to verify backwards compatibility isn't an issue.


erictraut:
@rchiodo, I've fixed the issues you found with this run and then re-run the same experiment (enabling the new strict type guard experimental feature by default). Here are the results: https://github.com/microsoft/pyright/pull/5832#issuecomment-1695005098. I'm going to close this PR since it's now obsolete.

PR #5912, opened 2023-09-08
Skip real case path on pythonPath

rchiodo:
Alternative fix that maintains realCasePath for most cases, but skips it for pythonPath


erictraut (approved):

erictraut:
@AIGeneratedUsername, if you suspect a bug, could you please file a bug report with repro steps? A comment in a closed PR is likely to get lost.

PR #6001, opened 2023-09-21
Fix realCasePath changing empty paths to .

rchiodo:
Addresses https://github.com/microsoft/pylance-release/issues/4861

The root cause of this bug is that for a loose file (no workspace), the current execution root for command line options is just ''. Normalizing this changes it to '.' for some reason and that breaks all the other logic after that point.

The fix here is to not normalize when calling realCasePath. Nodefs.realCase can normalize paths on its own, so it's not necessary to normalize them ahead of time.

### DIFF
--- a/packages/pyright-internal/src/common/pathUtils.ts
+++ b/packages/pyright-internal/src/common/pathUtils.ts
@@ -536,7 +536,7 @@ export function stripFileExtension(fileName: string, multiDotExtension = false) 
 }
 
 export function realCasePath(pathString: string, fileSystem: ReadOnlyFileSystem): string {
-    return normalizePath(fileSystem.realCasePath(pathString));
+    return fileSystem.realCasePath(pathString);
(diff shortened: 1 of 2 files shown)


packages/pyright-internal/src/common/pathUtils.ts
```diff
@@ -536,7 +536,7 @@ export function stripFileExtension(fileName: string, multiDotExtension = false)
 }
 
 export function realCasePath(pathString: string, fileSystem: ReadOnlyFileSystem): string {
-    return normalizePath(fileSystem.realCasePath(pathString));
```
rchiodo, line 539: This normalizePath can change the empty string into a `.`. We shouldn't need to normalize since calling real case o n a path will have that same effect anyway.

debonte (approved):

erictraut (approved):

PR #6007, opened 2023-09-22
Create test for virtual environment

### LINKED ISSUE #5990: Add a test for virtual environments
All the current tests mimic the python environment at the moment. It would be nice to have tests that run inside of a true virtual environment so we can catch issues like:

https://github.com/microsoft/pyright/issues/5943

rchiodo:
This creates some new validation for virtual environments. It's only testing it using venv/venvPath and pythonPath at the moment.

Fixes #5990


rchiodo:
The validation make take a few iterations to get right. It's kinda hard to test locally.

erictraut (approved):
Nice! Thanks for doing this.

erictraut:
Looks like one of the tests is failing on Windows.

Issue #6008, opened 2023-09-23
Incorrect error on specific order of multi-pattern `case` clause


DanielRosenwasser:
I noticed a difference in the ordering between a match of `str(v) | bool(v) | float(v)` and `str(v) | float(v) | bool(v)`. The code is below.

Pyright incorrectly issues errors on code in the last match.

```python
SimpleValue = str | float | bool | None

def f(x: SimpleValue):
    print(f"Running with value {x}")

    match x:
        # No errors. ✅
        case str(v) | bool(v) | float(v):
            print("Match 1 happened!")
            print(v)
        case _:
            print("Match 1 didn't happen.")

    
    match x:
        # No errors. ✅
        case str() | float() | bool():
            print("Match 2 happened!")
        case _:
            print("Match 2 didn't happen.")

    match x:
        # Errors! ❌
        case str(v) | float(v) | bool(v):
            #                    ~~~~~~
            # Pattern will never be matched for subject type "None"
            # (reportUnnecessaryComparison)
            print("Match 3 happened!")

            print(v)
            #     ~
            # Argument type is partially unknown
            #     Argument corresponds to parameter "values" in function "print"
            #     Argument type is "str | bool | float | Unknown"
            # (reportUnknownArgumentType)
        case _:
            print("Match 3 didn't happen.")

f(True)
f(None)

# Output:
#   Running with value True
#   Match 1 happened!
#   True
#   Match 2 happened!
#   Match 3 happened!
#   True
#   Running with value None
#   Match 1 didn't happen.
#   Match 2 didn't happen.
#   Match 3 didn't happen.
```

Admittedly<sup>1</sup>, this only happens when you write useless code like `str(v) | float(v) | bool(v)`, or more-realistically, the still-useless `str(_) | float(_) | bool(_)`. If you remove the binding variable from these, the error goes away entirely.

**VS Code extension or command-line**

I'm running in VS Code via Pylance 2023.9.20 

______

<sup>1</sup>Be patient with me, this is my second time using pattern matching. 😅

DanielRosenwasser:
Actually, you can see this without placeholders if you have a clause like `case str() | bool() | float() | int()`.

erictraut:
Thanks for the bug report.

This is due to a small inconsistency between the Python type system and the Python runtime. The `bool` type is a subtype of `int` (both at runtime and in the type system).

```python
print(isinstance(True, int)) # True
```

[...]

PR #6013, opened 2023-09-24
Improved handling of types `float` and `complex`, which are special-c…

erictraut:
…ased in PEP 484 as "promotion types". The new logic now properly models the runtime behavior for `isinstance` and class pattern matching when used with these promotion types. This addresses #6008.

### DIFF
--- a/packages/pyright-internal/src/analyzer/checker.ts
+++ b/packages/pyright-internal/src/analyzer/checker.ts
@@ -3510,6 +3510,8 @@ export class Checker extends ParseTreeWalker {
             return transformPossibleRecursiveTypeAlias(subtype);
         });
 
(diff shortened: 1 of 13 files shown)


Avasam:
@erictraut Is that mypy_primer result expected? If so, why does a `isinstance` condition results in an unknown type in the else branch for floats unions?

erictraut:
Yes, this is expected. 

This code sample should make it clear why 

```python
def func(x: float | str):
    if isinstance(x, float):
        print('received a float!')
    else:
        reveal_type(x) # int | str

        # The following line is a type violation because `x` is
        # is possibly an `int`.
        return x.upper()

func(1)
```

Issue #5985, opened 2023-09-20
Pyright accepts an invalid ellipsis default value, outside of an interface or protocol


maxime-didier:
**Describe the bug**

The following unsound definition is accepted by pyright 1.1.337 in strict mode while mypy 1.4.1 in strict mode rejects it:

```
def the_bad(x: list[int] = ...) -> list[int]:
    return x
```

`mypy --strict` outputs the following for this definition:

```
error: Incompatible default for argument "x" (default has type "ellipsis", argument has type "list[int]")  [assignment]
```

Pyright should reject this program.

erictraut:
The current behavior is by design, so I don't consider this an error. I'll change this to an enhancement request and consider it in that light.

The ellipse is used in Python to represent a placeholder. In typing terms, it's used to represent `Any` or "unspecified".

Pyright allows `...` to be used for the [body of a function](https://microsoft.github.io/pyright/#/mypy-comparison?id=ellipsis-in-function-body) regardless of the return type annotation. That is, it doesn't enforce the return type in this case. 

I think that not enforcing types for `...` used as a default argument value is consistent with pyright's other handling of `...`.

An ellipsis is used frequently in code samples (e.g. on Stack Overflow, GitHub issues, and documentation sites) to demonstrate type-related concepts, and there's value in being able to simply paste these code samples into an editor or playground. 

On the other hand, if this is a common source of runtime bugs, I can see why it might be useful for a type checker to report an error here. The question is whether this is actually a common source of runtime bug. 

Out of curiosity, did you actually run into this case in real code and get bitten by it? Or did you simply notice the current behavior and think "I expected pyright to report this"?

Azureblade3808:
I'd vote for keeping Pyright's current behavior.

I sometimes use ellipsis as a convenient sentinel in experimental code. Ellipsis, as a built-in object, has several advantages over custom sentinels --
- It doesn't require definition and importing.
- I don't need to worry about whether it is still the same instance after an auto-reloading which happens a lot in experiments.

In other aspects, it works almost no differently from a custom sentinel, no matter good or bad.

maxime-didier:
@erictraut

I've been maintaining a python codebase and introducing typing annotations to it. I noticed this while discussing a coding pattern with a coworker.

[...]

PR #6015, opened 2023-09-24
Added check for the use of an ellipsis for a default argument value f…

erictraut:
…or a function that is not in a stub, not overloaded, and without a placeholder implementation. This addresses #5985.


erictraut:
@Gobot1234, this change will require a modification to your code base. You're using `...` in a manner that isn't legit.

BTW, while looking into the new errors, I ran across the comment "# I won't tell Eric if you won't. (This isn't part of PEP 696)". 😀

PR #6028, opened 2023-09-26
Fix real casing of UNC paths (including mounted paths)

debonte:
Addresses https://github.com/microsoft/pylance-release/issues/4862

In the user's repro steps, they have mounted a share as a drive letter. The recent real case path changes cause the drive letter-based path (ex. `z:\test`) to be changed to a UNC path (ex. `\\server\share\test`). And it seems that we had pre-existing issues with using UNC paths in file specs. 
- Fix `getRootLength` for `\\server\share\dir` to return `\\server\` rather than `\\server\share\`. This matches [TypeScript's behavior](https://github.com/microsoft/TypeScript/blob/55d8bed85c11672b2b6f9ee3e6e1d384bef6f4da/src/compiler/path.ts#L241). Also brought over the comments from the TS implementation to explain what `getRootLength` is doing.
- Fix `getWildcardRegexPattern` to double-escape both leading backslashes in `\\server\share`. The result should be `\\\\\\\\server\\\\share` but was `\\\\server\\\\share`.
- Fix the "try going up a level" logic in `realCasePath` to not go up another level if `getRootLength(path)` returns the length of `path`. In this case `path` is a root and we need to bail at this point to avoid stack overflow -- the "go up a level" logic would just call `realCasePath` with `path` again.

### DIFF
--- a/packages/pyright-internal/src/common/pathUtils.ts
+++ b/packages/pyright-internal/src/common/pathUtils.ts
@@ -110,6 +110,9 @@ export function getRootLength(pathString: string): number {
         if (pathString.charAt(2) === path.sep) {
             return 3; // DOS: "c:/" or "c:\"
         }
+        if (pathString.length === 2) {
+            return 2; // DOS: "c:" (but not "c:d")
+        }
     }
     return 0;
 }
--- a/packages/pyright-internal/src/common/pathUtils.ts
+++ b/packages/pyright-internal/src/common/pathUtils.ts
@@ -92,24 +92,23 @@ export function getDirectoryPath(pathString: string): string {
     return pathString.substr(0, Math.max(getRootLength(pathString), pathString.lastIndexOf(path.sep)));
 }
 
+/**
+ * Returns length of the root part of a path or URL (i.e. length of "/", "x:/", "//server/").
+ */
 export function getRootLength(pathString: string): number {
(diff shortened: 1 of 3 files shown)


rchiodo:
Should we add tests to validation.yml for this case like I did for pyright?
https://github.com/microsoft/pyright/blob/a3cdd524f9a5e04f76c1c79f2a724af1e5bf7b6d/.github/workflows/validation.yml#L147

I just realized those tests I added won't run in pyrx as their only in the pyright actions.

rchiodo (approved):

erictraut (approved):
These changes look reasonable to me. I presume that you looked at the corresponding typescript code to determine if the pyright code handles all of the relevant cases?

debonte:
> These changes look reasonable to me. I presume that you looked at the corresponding typescript code to determine if the pyright code handles all of the relevant cases?

I did look and believed this now matched [their `getEncodedRootLength` implementation](https://github.com/microsoft/TypeScript/blob/55d8bed85c11672b2b6f9ee3e6e1d384bef6f4da/src/compiler/path.ts#L164), minus the part specific to URLs.

However, I just added unit tests for `getRootLength` based on the examples in their comments and found a discrepancy where `getRootLength("x:")` returned 0 instead of 2. Fixed that.

Issue #6136, opened 2023-10-12
unittest.mock.Mock is not treated as Any in strict mode


MartinAltmayerTMH:
As far as I know, `Mock` objects should be treated as `Any` by type checkers (see e.g. #3611). However, the following code raises type warnings in strict mode.

```python
from unittest.mock import Mock

def test_something():
  item = Mock()
  item.start.assert_not_called()
```
The errors are
```
my_test.py
  my_test.py:6:3 - error: Type of "start" is unknown (reportUnknownMemberType)
  my_test.py:6:3 - error: Type of "assert_not_called" is unknown (reportUnknownMemberType)
```

And if this should be the expected behavior, what is the correct way to work with mock functions like `assert_not_called`?

Tested with pyright 1.1.331 on the command line.

erictraut:
This is included in pyright 1.1.332, which I just published. It will also be included in a future release of pylance.

MartinAltmayerTMH:
Thanks for the quick resolution!

PR #6137, opened 2023-10-12
Changed type evaluation behavior for accesses to attributes on a clas…

erictraut:
…s that derives from `Any`. Previously, these were evaluated as `Unknown`, but they are now evaluated as `Any`. This is related to #6136.


erictraut:

Issue #6010, opened 2023-09-23
Improper override not reported


erictraut:
This comes from a [mypy issue](https://github.com/python/mypy/issues/16165).

Here's a minimal repro:

```python
from __future__ import annotations
from typing import Self, TypeVar, Generic

T = TypeVar("T")

class Base:
    def __and__(self: Self, other: Self) -> Self:
        ...

class Derived(Generic[T], Base):
    def __and__(self: Derived[bool], other: Derived[bool]) -> Derived[bool]:
        return Derived[bool]()
```

erictraut:
This is included in pyright 1.1.332, which I just published. It will also be included in a future release of pylance.

PR #6138, opened 2023-10-12
Fixed a bug in the `reportIncompatibleMethodOverride` check that lead…

erictraut:
…s to a false negative if the base class uses a `Self` type and the override uses an incompatible specialized type. This addresses #6010.


erictraut:
@Gobot1234, this bug fix will affect the steam.py project. I confirmed that this is a legit typing issue that was previously going unreported in pyright.

Gobot1234:
> @Gobot1234, this bug fix will affect the steam.py project. I confirmed that this is a legit typing issue that was previously going unreported in pyright.

Thanks for the report!

PR #6155, opened 2023-10-13
Set `isThirdPartyImport` and `isThirdPartyPyTypedPresent` for interim files

debonte:
Intended to address https://github.com/microsoft/pylance-release/issues/4059

But I'm uncertain about:
1. Is it correct to set `isThirdPartyImport` to `true` on `SourceFile` and `SourceFileInfo` only when `moduleNameAndType.importType === ImportType.ThirdParty`? We could have done that before, so I'm not sure if I'm missing some complexity here.
2. When calling `_shouldWalkUp` if `current` is outside of the `root` directory, the return value of `_shouldWalkUp` is based on comparing the lengths of the two paths which seems wrong since they are unrelated. For example, if `current` is `%LOCALAPPDATA%\Programs\Python\Python311\Lib"` and `root` is the workspace root, their relative lengths are irrelevant. Should `pathUtils.containsPath` be getting called somewhere? Or am I using `_shouldWalkUp` incorrectly?


erictraut (approved):
This change looks reasonable to me. It's touching piece of code that deals with a lot of complexity (given all of the weird edge cases of the Python import system), so I can't say with full confidence that it won't introduce regressions in some edge case. Your changes don't appear to have any impact on the core import resolution logic (i.e. the logic used to model the behavior of `import` statements), so if there is a regression, it will presumably show up only in completion suggestions and the pytest fixture functionality.

packages/pyright-internal/src/analyzer/importResolver.ts
```diff
@@ -1197,16 +1204,45 @@ export class ImportResolver {
...
         }
 
+        if (importType === ImportType.ThirdParty) {
+            const root = this.getParentImportResolutionRoot(filePath, execEnv.root);
+
+            // Go up directories one by one looking for a py.typed file.
+            let current = ensureTrailingDirectorySeparator(getDirectoryPath(filePath));
+            while (this._shouldWalkUp(current, root, execEnv)) {
+                if (this.fileExistsCached(combinePaths(current, 'py.typed'))) {
+                    isThirdPartyPyTypedPresent = true;
```
heejaechang, line 1214: I think you need to check whether it is `py.typed` or `partial py.typed`? and only set it when it is `non partial py.typed`?

heejaechang: `getModuleNameInfoFromPath` already walk down from each `import root` to the `file path` given, so, rather than `move` up from `the file path`, we could return `pyTyped` data from `getmoduleNameInfoFromPath` as well along with `moduleName` ?

heejaechang:
> Is it correct to set isThirdPartyImport to true on SourceFile and SourceFileInfo only when moduleNameAndType.importType === ImportType.ThirdParty? We could have done that before, so I'm not sure if I'm missing some complexity here.

when I looked it last time, this one was when we set it as `third party` even if `importType` is not `ThirdParty`

https://github.com/microsoft/pyright/blob/main/packages/pyright-internal/src/analyzer/program.ts#L1320

so, `ThirdParty` is set `isThridPartyImport` is always true, but when `Local` is imported by `ThirdPartyImport`, then it is considered `ThirdPartyImport` even if it is `Local`.

packages/pyright-internal/src/analyzer/importResolver.ts
heejaechang: > I think you need to check whether it is py.typed or partial py.typed? and only set it when it is non partial py.typed?

actually, our vfs might already hide `partial py.typed` when it merge them, so, it might not visible to `import resolver`, we probably can test with `bundled stubs` we have in `pylance` since some of them are `partial py.typed`.

packages/pyright-internal/src/tests/importStatementUtils.test.ts
```diff
@@ -205,9 +225,24 @@ test('getTextEditsForAutoImportInsertions - multiple modules with existing impor
...
     `;
 
-    const module1 = { moduleName: 'sys', importType: ImportType.BuiltIn, isLocalTypingsFile: false };
-    const module2 = { moduleName: 'numpy', importType: ImportType.ThirdParty, isLocalTypingsFile: false };
-    const module3 = { moduleName: 'test', importType: ImportType.Local, isLocalTypingsFile: false };
+    const module1 = {
+        moduleName: 'sys',
+        importType: ImportType.BuiltIn,
+        isLocalTypingsFile: false,
+        isThirdPartyPyTypedPresent: false,
```
rchiodo, line 232: Are there any tests for when this should come out as true?

debonte:
> when I looked it last time, this one was when we set it as `third party` even if `importType` is not `ThirdParty`
> 
> https://github.com/microsoft/pyright/blob/main/packages/pyright-internal/src/analyzer/program.ts#L1320
> 
> so, when `ThirdParty` is set, `isThridPartyImport` is always true, but when `Local` is imported by `ThirdPartyImport`, then it is considered `ThirdPartyImport` even if it is `Local`.

Yeah, I saw that code, but since `_createInterimFile` is creating a file rather than analyzing an import, the "local import from 3rd party file" logic didn't seem relevant to me in this scenario.

packages/pyright-internal/src/tests/importStatementUtils.test.ts
debonte: Not yet. I'll try to add some. When I created this PR I was feeling unsure about the implementation and didn't want to invest time in tests before I got some feedback.

packages/pyright-internal/src/analyzer/importResolver.ts
debonte: > it might not visible to import resolver

The return type of `getPyTypedInfo` includes an `isPartiallyTyped` boolean. I can check that. I didn't realize that `isThirdPartyPyTypedPresent` was limited to non-partial `py.typed` files.

> we could return `pyTyped` data from `getmoduleNameInfoFromPath` as well along with `moduleName` ?

Ok, I'll give that a shot.

debonte: > `getModuleNameInfoFromPath` already walk down from each `import root` to the `file path` give

@heejaechang, where is this happening? I'm not seeing it.

heejaechang: ah. sorry, you are right. it just get module name from relative path and not actually walk down.

heejaechang:
> Yeah, I saw that code, but since _createInterimFile is creating a file rather than analyzing an import, the "local import from 3rd party file" logic didn't seem relevant to me in this scenario.

so, currently, if you have a `file/module` in `typings` or `extraPath` (ex, typings/pandas/core/accessor.py), and `pandas` package has absolute path `from pandas.core import accessor` rather than `relative path` somewhere, it could load one in `typings` rather than one in `pandas` package.

pyright supports this kind of `shadowing` so that user can provide per file `stub` in `typing` folder. 

if those files are marked as `user files` (match include/exclude in configOption)`, then we create `sourceFileInfo` when `program` is created. but if they are not and if they are brought into `program` by other `sourceFile` then they are marked as `third party`

now question is if that file is brought into the program by either `interimFile` or `setFileOpen`, how should we mark it. `isThridParty`? or not.

might be fine to just not mark it. I am just saying difference on if that file is loaded in program differently, it might have marked as `isThirdParty`

packages/pyright-internal/src/tests/importStatementUtils.test.ts
debonte: It wasn't clear to me that adding to `importStatementUtils.test.ts` was necessary/interesting. I added a new test that repros the user's scenario in `completions.importInterimFile.fourslash.ts`.

rchiodo: I was just thinking for completeness. You added this new logic, but there's no test for the 'true' case. If the 'true' case isn't important, do we even need the flag?

debonte: I only added `isThirdPartyPyTypedPresent` in this file because it's (now) a required field on `ModuleNameAndType`. I noticed that all of the tests in this file also have `isLocalTypingsFile` set to `false`, I'm assuming for the same reason.

I think the core problem here is that the auto-importer is using `ModuleNameAndType` but doesn't care about `isThirdPartyPyTypedPresent` (or `isLocalTypingsFile`?). I'll try introducing a new type instead.

packages/pyright-internal/src/languageService/autoImporter.ts
```diff
@@ -12,15 +12,15 @@ import { DeclarationType } from '../analyzer/declaration';
 import { ImportResolver, ModuleNameAndType } from '../analyzer/importResolver';
 import { ImportType } from '../analyzer/importResult';
 import {
+    ImportGroup,
```
heejaechang, line 15: sorting seems wrong?

packages/pyright-internal/src/tests/fourslash/completions.importInterimFile.fourslash.ts
```diff
@@ -0,0 +1,52 @@
...
+//// def selection_interval(): pass
+
+// @filename: altair/py.typed
+// @library: true
+//// # has to contain something for file to be written
+
+{
+    // Force interim file to be created
+    // @ts-ignore
+    await helper.verifyCompletion('included', 'markdown', {
```
heejaechang, line 29: why interim file is created here?

packages/pyright-internal/src/tests/fourslash/completions.importInterimFile.fourslash.ts
```diff
@@ -0,0 +1,52 @@
...
+        marker1: {
+            completions: [
+                { label: 'altair', kind: Consts.CompletionItemKind.Module, documentation: 'module docstring' },
+            ],
+        },
+    });
+
+    helper.replace(0, 9, 'import altair as alt\n\nalt.');
```
heejaechang, line 37: what these 2 number `0` and `9` mean? what does `replace` actually replace?

packages/pyright-internal/src/tests/fourslash/completions.importInterimFile.fourslash.ts
heejaechang: originally `interim` file meant to indicate a file that is not `rooted` by any other file so, it would go away when `removeUnncessaryFile` is called.

by `rooted`, it means a file is used (imported) by user files (isTracked) or open files transitively.

any file that are not imported by rooted files transitively such as `pytest fixtures in `conftest.py` that are not imported by any thing` or `module in string literal that are not imported by anything` and etc are created as `interim` file.

everything involved here seems explicitly imported? so not sure what would case interim file to be created?

it seems not using any shadow file as well which could also cause `interim` file to be created.

packages/pyright-internal/src/tests/importStatementUtils.test.ts
debonte: I switched to a new `ModuleImportInfo` type rather than adding `isThirdPartyPyTypedPresent` to `ModuleNameAndType`. I tried removing `isLocalTypingsFile` but it's actually used [here](https://github.com/microsoft/pyrx/blob/4d2c5e6f9da0c3634aff7e6a50bb66dff6706807/packages/pyright/packages/pyright-internal/src/analyzer/importStatementUtils.ts#L742).

packages/pyright-internal/src/tests/harness/fourslash/testState.ts
```diff
@@ -49,6 +50,9 @@ import {
     setTestingMode,
 } from '../../../common/pathUtils';
 import { convertOffsetToPosition, convertPositionToOffset } from '../../../common/positionUtils';
+import { ServiceProvider } from '../../../common/serviceProvider';
```
heejaechang, line 53: why ordering changed? is pyright and pylance uses different prettier setting?

rchiodo (approved):

packages/pyright-internal/src/tests/harness/fourslash/testState.ts
debonte: The `.prettierrc` in pyrx and pyright are identical. The new ordering looks correct to me. Not sure why the ordering was wrong before this.

packages/pyright-internal/src/tests/fourslash/completions.importInterimFile.fourslash.ts
debonte: Here's the callstack where it gets hit in this scenario. This occurs before the user has completely typed in the module name `altair` (test.py contains `import al`), so at that point it's not explicitly imported. When manually reproing the issue, I've noticed that it doesn't repro if you type slowly after the interim file is created, which I believe is because you're allowing `removeUnneccessaryFiles` to get called.

```text
Program._createInterimFileInfo (c:\src\pyright\packages\pyright-internal\src\analyzer\program.ts:1578)
Program.addInterimFile (c:\src\pyright\packages\pyright-internal\src\analyzer\program.ts:351)
<anonymous -- this is SourceMapper._boundSourceGetter> (c:\src\pyright\packages\pyright-internal\src\analyzer\program.ts:1212)
SourceMapper.findModules (c:\src\pyright\packages\pyright-internal\src\analyzer\sourceMapper.ts:65)
getModuleDocStringFromPaths (c:\src\pyright\packages\pyright-internal\src\analyzer\typeDocStringUtils.ts:203)
CompletionProvider.resolveCompletionItem (c:\src\pyright\packages\pyright-internal\src\languageService\completionProvider.ts:350)
resolveCompletionItem (c:\src\pyright\packages\pyright-internal\src\tests\harness\fourslash\testState.ts:1522)
```

packages/pyright-internal/src/tests/fourslash/completions.importInterimFile.fourslash.ts
debonte: Changed the args to me more meaningful.

packages/pyright-internal/src/tests/fourslash/completions.importInterimFile.fourslash.ts
```diff
@@ -34,7 +34,7 @@
         },
     });
 
-    helper.replace(0, 9, 'import altair as alt\n\nalt.');
+    helper.replace(helper.BOF, helper.getPosition('marker1'), 'import altair as alt\n\nalt.');
```
heejaechang, line 37: but how does it know which file? it always assumes active file? since there can be multiple open file? is the last file always active file?

heejaechang (approved):

packages/pyright-internal/src/tests/fourslash/completions.importInterimFile.fourslash.ts
debonte: Yes, currently it always targets the active file, which by default is the first file. That's how `replace` worked when I found it, and I didn't change that. Apparently this is the first test that has needed to edit files, so adding more flexibility seems unnecessary at the moment.

Issue #6160, opened 2023-10-14
1.1.329-1.1.331: More regressions in composite expressions involving overloads


rsokl:
Starting in 1.1.329, and persisting through 1.1.331, there are regressions in pyright's evaluation of composite expressions involving overloads. These turned out to be distinct from what I reported in #6071 and were not addressed by pyright 1.1.330.

I have tried to come up with a minimal example that doesn't involve hydra-zen but I have failed to do so despite lots of tinkering. 

Checking out [cfa9f37](https://github.com/mit-ll-responsible-ai/hydra-zen/commit/cfa9f37175eeab39f2a1f52754c58391947c0ccb) (currently `main` on hydra-zen) and using pyright 1.1.331 in basic mode, here is a minimal hydra-zen example. 

```python
from typing_extensions import assert_type
from hydra_zen import builds, instantiate

class A:
    ...
    
builds_A = builds(A)
assert_type(instantiate(builds_A), A)  # OK

assert_type(instantiate(builds(A)), A)  # "assert_type" mismatch: expected "A" but received "Partial[A]"
```

(apologies for not being able to deliver a simpler example)

erictraut:
Thanks for the bug report. 

This is caused by a heuristic that I added to overload matching with the intent of avoiding type errors when calling certain overloaded functions. There are cases where the normal overload matching behavior results in a call return type that is incompatible with the expected return type (when bidirectional type inference is being used) but at least one other overload produces a result that is compatible with the expected return type. 

Here is an example where `str.split` when used with `LiteralString` arguments.

```python
class str:
    ...
    @overload
    def split(self: LiteralString, sep: LiteralString | None = None, maxsplit: SupportsIndex = -1) -> list[LiteralString]: ...
    @overload
    def split(self, sep: str | None = None, maxsplit: SupportsIndex = -1) -> list[str]: ...


# Should this generate a type error?
x: list[str] = "a,b".split(",")
```

I've convinced myself that the heuristic isn't sound, so I've removed it entirely. This eliminates the false positives in your test cases, but it will unfortunately result in false positives in other (more common) use cases.

This will be fixed in the next release.

erictraut:
This is included in pyright 1.1.332, which I just published. It will also be included in a future release of pylance.

PR #6163, opened 2023-10-14
Fixed a regression in overload matching that resulted in false positi…

erictraut:
…ve errors in hydra-zen. This change involves removing a heuristic from overload matching that attempted to eliminate a false positive error for `x: list[str] = "a,b".split(",")`. This addresses #6160.

### DIFF
--- a/packages/pyright-internal/src/analyzer/typeEvaluator.ts
+++ b/packages/pyright-internal/src/analyzer/typeEvaluator.ts
@@ -246,7 +246,6 @@ import {
     preserveUnknown,
     removeParamSpecVariadicsFromFunction,
     removeParamSpecVariadicsFromSignature,
-    replaceTypeVarsWithAny,
     requiresSpecialization,
     requiresTypeArguments,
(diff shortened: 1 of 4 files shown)


erictraut:
@Dr-Irv, FYI, this bug fix may have an impact on `pandas-stubs`.

Dr-Irv:
> @Dr-Irv, FYI, this bug fix may have an impact on `pandas-stubs`. I'm not able to repro the above results from `mypy_primer` when I run pyright on `pandas-stubs` locally, so I'm guessing that the above issues are all caused by not having the full dependencies installed in the `mypy_primer` script. I'm going to therefore assume that the above is just noise and not indicative of a real issue. Just wanted to give you a heads up in case you want to independently confirm.

@erictraut is this released yet so we can test?

erictraut:
This change was included in pyright 1.1.332, which was released last week Tuesday. It was also in last week's prerelease version of Pylance. It will be in this week's production Pylance release (normally published every Wed).

Issue #6146, opened 2023-10-13
Renaming a symbol in a module in a symlink directory not taken into account


ilrico:
**Describe the bug**
When renaming a symbol (a method for instance Myclass.mymethod() to .my_method() ) in /mylib/mymodule.py, with mylib/ a symlink in my project, then VSCode will make the change in files this class is imported (main.py) but pyright will show an error "Member 'my_method' is unknown", and it will stay as that undefinitively.
BUT. If you rename symbol again, back to original name, the error will vanish instantantly.

More generally I has similar sync issues, also when refactoring a module in a symlink directory, whereas the same module refactoring if module directory is copied into the main project (not symlinked) does not trigger an error.

In all cases, <Developer: Reload window> solves the errors.

**Versions**
Pytight v1.1.331
VSCodium 1.81.1 on GNU/Linux Debian

**Out-of-scope**
As a sidenote (sorry to be out-of-scope, but I think it's important), since I suspect Eric will answer, I want to underline the amazing work support provided here. Such a large project with only 16 open issues, it's just incredible. And when you dig into issues you see an exceptionnal quality in answers/remarks. I cannot name another major project with this degree of excellence in its support.

ilrico:
I just had the same issue modifying an enum in the module located in the symlinked directory. <Reload window> solved it.

erictraut:
@ilrico, thanks for the kind words.

It's not clear whether this issue is related to the language server's "rename provider" (the functionality that implements semantic renaming), the virtual file system and cache used internally to pyright, or the internal dependency tracking that forces a file to be rechecked after a dependent file is modified. More investigation will be needed.

@heejaechang, would you like to investigate this? It's likely related to code that you've contributed to pyright. Feel free to transfer the issue to pylance-release if you'd prefer to track it there.

heejaechang:
let me take a look

PR #6180, opened 2023-10-16
Make sure we always normalize drive letter even for symbolic link.

heejaechang:
fixes https://github.com/microsoft/pyright/issues/6146

it is a regression from our change to use `realCase` in all places. added special casing to `symlink` case.


heejaechang:
@rchiodo can you take a look?

rchiodo (approved):

PR #6239, opened 2023-10-26
Support unpacked TypedDict in parameter hover

pakrym-stripe:
Fixes: https://github.com/microsoft/pylance-release/issues/5015

## Before:
Parameter has no docstring
image

## After
Doc string is now included when hovering over the parameter:
image

You can also navigate to key definition.

### DIFF
--- a/packages/pyright-internal/src/analyzer/typeEvaluator.ts
+++ b/packages/pyright-internal/src/analyzer/typeEvaluator.ts
@@ -19953,6 +19953,19 @@ export function createTypeEvaluator(importLookup: ImportLookup, evaluatorOptions
                                 .getDeclarations()
                                 .find((decl) => decl.type === DeclarationType.Parameter);
                         }
+
+                        const parameterDetails = getParameterListDetails(type);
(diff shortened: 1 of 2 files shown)


packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -19953,6 +19953,19 @@ export function createTypeEvaluator(importLookup: ImportLookup, evaluatorOptions
                                 .getDeclarations()
                                 .find((decl) => decl.type === DeclarationType.Parameter);
                         }
+
+                        const parameterDetails = getParameterListDetails(type);
```
pakrym-stripe, line 19957: It felt slightly heavy-handed to call the entire `getParameterListDetails` here but also the code to deduce is `unpackedKwargsTypedDictType` is not the most trivial.

erictraut: Looks reasonable to me.

erictraut:
Thanks for the contribution. The change looks good to me.

Issue #6250, opened 2023-10-28
python.analysis.exclude is not respected by Pyright


ppeetteerrs:
**Is your feature request related to a problem? Please describe.**
I think it makes sense for Pyright to append / override the default list of excluded filespecs with ones specified in VSCode's python.analysis.exclude. Currently VSCode users would need to duplicate the excluded list of file specs in pyrightconfig.json.

**Describe the solution you’d like**
Add the option to package.json and the corresponding append / override logic to configOptions.ts or service.ts. It should be a pretty trivial change.

erictraut:
`python.analysis.exclude` isn't currently a supported language server setting for either pyright or pylance. If you want to specify an `exclude` setting, you can do so in your `pyrightconfig.json` file or `pyproject.toml` file.

There has been some demand for adding this support. If you're interested, please upvote [this issue](https://github.com/microsoft/pylance-release/discussions/2581). I'll leave it up to the pylance team to decide whether to add such a feature.

If we were to implement this, it would not be an "append" operation. As with other config options that are exposed in both config files and as a language server setting, the config file would take precedence over the language server setting.

erictraut:
I just realized the other issue (the discussion topic that I linked to) is locked and marked as complete.

@debonte, it looks like `python.analysis.include`, `python.analysis.exclude`, and `python.analysis.ignore` settings were added to pylance. Why wasn't this implemented in the pyright code base? I'd think that we'd want this to be consistent across pyright and pylance. Was this just an oversight?

ppeetteerrs:
Thanks for the prompt response, really appreciate it!

debonte:
> Was this just an oversight?

Yes, an oversight on our part.

erictraut:
This is addressed in pyright 1.1.334, which I just published.

PR #6251, opened 2023-10-28
Added support for `python.analysis.include`, `python.analysis.exclude…

erictraut:
…` and `python.analysis.ignore` settings. These have been implemented in pylance for a while now, but they were never implemented previously in pyright. Also fixed a bug in the original implementation that caused the config file not to override the language server settings for `python.analysis.exclude` and `python.analysis.ignore` as it should. This addresses #6250.

### DIFF
--- a/packages/vscode-pyright/package.json
+++ b/packages/vscode-pyright/package.json
@@ -165,7 +165,7 @@
                     "items": {
                         "type": "string"
                     },
-                    "description": "Paths of directories or files that should be included. If no paths are specified, Pylance defaults to the workspace root directory. Paths may contain wildcard characters ** (a directory or multiple levels of directories), * (a sequence of zero or more characters), or ? (a single character).",
+                    "description": "Paths of directories or files that should be included. If no paths are specified, pyright defaults to the workspace root directory. Paths may contain wildcard characters ** (a directory or multiple levels of directories), * (a sequence of zero or more characters), or ? (a single character).",
                     "scope": "resource"
                 },
                 "python.analysis.exclude": {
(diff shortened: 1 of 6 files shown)


heejaechang (approved):

packages/vscode-pyright/package.json
```diff
@@ -159,6 +159,33 @@
...
+                    "scope": "resource"
+                },
+                "python.analysis.exclude": {
+                    "type": "array",
+                    "default": [],
+                    "items": {
+                        "type": "string"
+                    },
+                    "description": "Paths of directories or files that should not be included. These override the include directories, allowing specific subdirectories to be excluded. Note that files in the exclude paths may still be included in the analysis if they are referenced (imported) by source files that are not excluded. Paths may contain wildcard characters ** (a directory or multiple levels of directories), * (a sequence of zero or more characters), or ? (a single character). If no exclude paths are specified, Pylance automatically excludes the following: `**/node_modules`, `**/__pycache__`, `.git` and any virtual environment directories.",
```
debonte, line 177: Uses the term Pylance instead of Pyright
```suggestion
                    "description": "Paths of directories or files that should be included. If no paths are specified, Pyright defaults to the workspace root directory. Paths may contain wildcard characters ** (a directory or multiple levels of directories), * (a sequence of zero or more characters), or ? (a single character).",
                    "scope": "resource"
                },
                "python.analysis.exclude": {
                    "type": "array",
                    "default": [],
                    "items": {
                        "type": "string"
                    },
                    "description": "Paths of directories or files that should not be included. These override the include directories, allowing specific subdirectories to be excluded. Note that files in the exclude paths may still be included in the analysis if they are referenced (imported) by source files that are not excluded. Paths may contain wildcard characters ** (a directory or multiple levels of directories), * (a sequence of zero or more characters), or ? (a single character). If no exclude paths are specified, Pyright automatically excludes the following: `**/node_modules`, `**/__pycache__`, `.git` and any virtual environment directories.",
```

debonte (approved):

PR #6281, opened 2023-11-01
Supports Chinese Docstring

Aruelius:
`\w+` only matches one or more word characters (same as [a-zA-Z0-9_]+), when the docstring is chinese (or other) it will not be matched.

https://github.com/microsoft/pylance-release/issues/4840

### DIFF
--- a/packages/pyright-internal/src/analyzer/docStringConversion.ts
+++ b/packages/pyright-internal/src/analyzer/docStringConversion.ts
@@ -589,7 +589,7 @@ class DocStringConverter {
 
         // catch-all for styles except reST
         const hasArguments =
-            !line?.endsWith(':') && !line?.endsWith('::') && !!line.match(/^\s*.*?\w+(\s*\(.*?\))*\s*:\s*\w+/g);
+            !line?.endsWith(':') && !line?.endsWith('::') && !!line.match(/^\s*.*?\w+(\s*\(.*?\))*\s*:[\s\S]*/g);
 
         // reSt params. Attempt to put directives lines into their own paragraphs.
         const restDirective = DirectivesExtraNewlineRegExp.test(line); //line.match(/^\s*:param/);
--- a/packages/pyright-internal/src/analyzer/docStringConversion.ts
+++ b/packages/pyright-internal/src/analyzer/docStringConversion.ts
@@ -589,7 +589,7 @@ class DocStringConverter {
 
         // catch-all for styles except reST
         const hasArguments =
-            !line?.endsWith(':') && !line?.endsWith('::') && !!line.match(/^\s*.*?\w+(\s*\(.*?\))*\s*:\s*[\u4e00-\u9fa5\w+]*/g);
+            !line?.endsWith(':') && !line?.endsWith('::') && !!line.match(/^\s*.*?\w+(\s*\(.*?\))*\s*:\s*\p{L}+/gu);
 
         // reSt params. Attempt to put directives lines into their own paragraphs.
         const restDirective = DirectivesExtraNewlineRegExp.test(line); //line.match(/^\s*:param/);
--- a/packages/pyright-internal/src/analyzer/docStringConversion.ts
+++ b/packages/pyright-internal/src/analyzer/docStringConversion.ts
@@ -589,7 +589,7 @@ class DocStringConverter {
 
         // catch-all for styles except reST
         const hasArguments =
-            !line?.endsWith(':') && !line?.endsWith('::') && !!line.match(/^\s*.*?\w+(\s*\(.*?\))*\s*:[\s\S]*/g);
+            !line?.endsWith(':') && !line?.endsWith('::') && !!line.match(/^\s*.*?\w+(\s*\(.*?\))*\s*:\s*[\u4e00-\u9fa5\w+]*/g);
 
         // reSt params. Attempt to put directives lines into their own paragraphs.
         const restDirective = DirectivesExtraNewlineRegExp.test(line); //line.match(/^\s*:param/);


packages/pyright-internal/src/analyzer/docStringConversion.ts
```diff
@@ -589,7 +589,7 @@ class DocStringConverter {
 
         // catch-all for styles except reST
         const hasArguments =
-            !line?.endsWith(':') && !line?.endsWith('::') && !!line.match(/^\s*.*?\w+(\s*\(.*?\))*\s*:\s*\w+/g);
+            !line?.endsWith(':') && !line?.endsWith('::') && !!line.match(/^\s*.*?\w+(\s*\(.*?\))*\s*:\s*[\u4e00-\u9fa5\w+]*/g);
```
debonte, line 592: The `+` quantifier should still be applied to the entire group of characters.
```suggestion
            !line?.endsWith(':') && !line?.endsWith('::') && !!line.match(/^\s*.*?\w+(\s*\(.*?\))*\s*:\s*[\u4e00-\u9fa5\w]+/g);
```

debonte, line 592: Did some reading on this and I think adding the `u` flag combined with `\p{L}` is the proper way to handle Unicode letters. This accepts both of the following strings from the issue:
- `        b (str): a中文`
- `        b (str): 中文`

https://www.regular-expressions.info/unicode.html
https://stackoverflow.com/questions/280712/how-can-i-use-unicode-aware-regular-expressions-in-javascript

```suggestion
            !line?.endsWith(':') && !line?.endsWith('::') && !!line.match(/^\s*.*?\w+(\s*\(.*?\))*\s*:\s*\p{L}+/gu);
```

debonte, line 592: @bschnurr, do we have existing unit tests that we can expand to cover this?

debonte, line 592: Is this really the only regex that needs to be updated for us to properly support Chinese characters? I understand that it fixes this one scenario, but are there others we should update at the same time?

Aruelius: Hi, this one scenario is the biggest issue for me, for now, I have to add the `\n` newline character after docstring.
`    a: 中文\n`
And I think it need be fixed, I thank for you help, it's very helpful for me.

packages/pyright-internal/src/analyzer/docStringConversion.ts
```diff
@@ -589,7 +589,7 @@ class DocStringConverter {
 
         // catch-all for styles except reST
         const hasArguments =
-            !line?.endsWith(':') && !line?.endsWith('::') && !!line.match(/^\s*.*?\w+(\s*\(.*?\))*\s*:\s*\w+/g);
+            !line?.endsWith(':') && !line?.endsWith('::') && !!line.match(/^\s*.*?\w+(\s*\(.*?\))*\s*:\s*\p{L}+/gu);
```
heejaechang, line 592: probably we want to add some tests covering issue it is fixing to prevent regressions in future?

Aruelius: that will be good, thank you.

erictraut: @Aruelius, are you planning to add tests to this PR?

packages/pyright-internal/src/analyzer/docStringConversion.ts
```diff
@@ -589,7 +589,7 @@
 
         // catch-all for styles except reST
         const hasArguments =
-            !line?.endsWith(':') && !line?.endsWith('::') && !!line.match(/^\s*.*?\w+(\s*\(.*?\))*\s*:\s*\w+/g);
+            !line?.endsWith(':') && !line?.endsWith('::') && !!line.match(/^\s*.*?\w+(\s*\(.*?\))*\s*:[\s\S]*/g);
```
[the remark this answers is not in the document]
erictraut: @Aruelius, note the issue detected above. This will need to either be addressed or an explanation given for why it's OK.

packages/pyright-internal/src/analyzer/docStringConversion.ts
erictraut: @bschnurr, any thoughts here?

packages/pyright-internal/src/analyzer/docStringConversion.ts
bschnurr: I actually have a fix.. i used this line
`!line?.endsWith(':') && !line?.endsWith('::') && !!line.match(/^\s*.*?\S+(\s*\(.*?\))*\s*:\s*\S+/g);`

erictraut: @bschnurr, do you mean that you're going to update this PR? Or that you're going to fix the problem in a separate PR and we should close this one?

debonte: `\S` includes non-letter characters like `<`, `>`, `+`, `$`, `,`, `'`, etc, whereas `\p{L}` only includes characters that Unicode says are letters. Is `\S` what we want? I'm not familiar with the docstring format requirements, but going from `\w` to `\S` seems strange to me unless we should always have been including symbol characters.

Btw, my initial suggestion of `\p{L}` only includes letters, not numbers. So if we wanted to use that approach, I believe `[\p{L}\p{N}]` would be better.

bschnurr: I thought using `\S` would be the most flexible when it comes to user naming stuff

bschnurr:
Closing. new PR here with a test. https://github.com/microsoft/pyright/pull/6307
thank you

PR #6287, opened 2023-11-01
[PylanceBot] Pull Pylance with Pyright 1.1.334

PylanceBot:
    1. pylance loc update 20231031.1 https://github.com/microsoft/pyrx/pull/4266
    2. Fix URI mapping for wsl https://github.com/microsoft/pyrx/pull/4255
    3. Fix crash when calling `isSymbolicLink` on `stat` of zip/egg https://github.com/microsoft/pyrx/pull/4254
    4. Fix potential issue with query parameters in a URI https://github.com/microsoft/pyrx/pull/4245
    5. don't always look for pyTyped since it is expensive
    6. fixed build failures
  
    Co-authored-by: Bill Schnurr <bschnurr@microsoft.com>
    Co-authored-by: HeeJae Chang <hechang@microsoft.com>
    Co-authored-by: Erik De Bonte <erikd@microsoft.com>
    Co-authored-by: Rich Chiodo <rchiodo@microsoft.com>
    Co-authored-by: Stella Huang <stellahuang@microsoft.com>

### DIFF
--- a/packages/pyright-internal/src/common/pathUtils.ts
+++ b/packages/pyright-internal/src/common/pathUtils.ts
@@ -144,7 +144,7 @@ export function getPathSeparator(pathString: string) {
 
 export function getPathComponents(pathString: string) {
     const normalizedPath = normalizeSlashes(pathString);
-    const rootLength = getRootLength(normalizedPath, /*checkUri*/ isUri(normalizedPath));
+    const rootLength = getRootLength(normalizedPath, /* checkUri */ isUri(normalizedPath));
     const root = normalizedPath.substring(0, rootLength);
     const sep = getPathSeparator(pathString);
     const rest = normalizedPath.substring(rootLength).split(sep);
--- a/packages/pyright-internal/src/common/pathUtils.ts
+++ b/packages/pyright-internal/src/common/pathUtils.ts
@@ -92,19 +92,21 @@ export function forEachAncestorDirectory(
 
 export function getDirectoryPath(pathString: string): string {
     if (isUri(pathString)) {
-        return Utils.dirname(URI.parse(pathString).with({ fragment: '' })).toString();
+        return Utils.dirname(URI.parse(pathString).with({ fragment: '', query: '' })).toString();
     }
     return pathString.substr(0, Math.max(getRootLength(pathString), pathString.lastIndexOf(path.sep)));
 }
 
+const _schemePattern = /^\w[\w\d+.-]*$/;
+
 export function isUri(pathString: string) {
-    return pathString.indexOf(':') > 1;
+    return pathString.indexOf(':') > 1 && _schemePattern.test(pathString.split(':')[0]);
 }
 
 /**
  * Returns length of the root part of a path or URL (i.e. length of "/", "x:/", "//server/").
  */
-export function getRootLength(pathString: string): number {
+export function getRootLength(pathString: string, checkUri = true): number {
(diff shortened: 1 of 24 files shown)


packages/pyright-internal/src/common/pathUtils.ts
```diff
@@ -92,19 +92,21 @@ export function forEachAncestorDirectory(
 
 export function getDirectoryPath(pathString: string): string {
     if (isUri(pathString)) {
-        return Utils.dirname(URI.parse(pathString).with({ fragment: '' })).toString();
+        return Utils.dirname(URI.parse(pathString).with({ fragment: '', query: '' })).toString();
     }
     return pathString.substr(0, Math.max(getRootLength(pathString), pathString.lastIndexOf(path.sep)));
 }
 
+const _schemePattern = /^\w[\w\d+.-]*$/;
```
erictraut, line 100: If this is used only in one function, define it within that function. If it's used within multiple functions, move it to the top of the file and give it a better name so it's clear what it means within the global scope of this source file.

packages/pyright-internal/src/common/pathUtils.ts
```diff
@@ -142,7 +144,7 @@ export function getPathSeparator(pathString: string) {
 
 export function getPathComponents(pathString: string) {
     const normalizedPath = normalizeSlashes(pathString);
-    const rootLength = getRootLength(normalizedPath);
+    const rootLength = getRootLength(normalizedPath, /*checkUri*/ isUri(normalizedPath));
```
erictraut, line 147: Nit: "/* checkUri */" should have spaces for consistency.

packages/pyright-internal/src/languageServerBase.ts
```diff
@@ -1441,6 +1441,17 @@ export abstract class LanguageServerBase implements LanguageServerInterface, Dis
         };
     }
 
+    protected sendDiagnostics(params: PublishDiagnosticsParams[]) {
+        for (const param of params) {
+            if (param.diagnostics.length === 0) {
```
erictraut, line 1446: This looks wrong. It's important for zero-length diganostic lists to be reported to the client. This is how old diagnostics are cleared for that file.

erictraut (changes requested):

packages/pyright-internal/src/languageServerBase.ts
heejaechang: this still send diagnostics. this logic is to clean up `this.documentsWithDiagnsotics` map we maintain. we use this info so that when workspace is closed, we clean up diagnostics reported if there is any.

heejaechang: by the way, actual change is just renaming `_sendDiagnostics` to `sendDiagnostics`. actual logic didn't change.

packages/pyright-internal/src/common/pathUtils.ts
bschnurr, line 147:
```suggestion
    const rootLength = getRootLength(normalizedPath, /* checkUri */ isUri(normalizedPath));
```

packages/pyright-internal/src/common/pathUtils.ts
rchiodo: It's only used in this function but I didn't want it to recreate the regex every time the function was called. I'll move it to the top.

rchiodo (approved):

erictraut (approved):

PR #6323, opened 2023-11-04
Fix incorrect typehint for callbacks in multiprocessing.Pool

Serious-senpai:
Fix incorrect typehint according to https://github.com/python/typeshed/pull/10949


erictraut:
Typeshed changes should be made in the [typeshed project](https://github.com/python/typeshed), not in pyright. Pyright simply snapshots the official typeshed stubs on a roughly a biweekly basis.

PR #6369, opened 2023-11-07
Support boolean values in settings.json's diagnosticSeverityOverrides

StellaHuang95:
This PR adds support for accepting `true` and `false` in `settings.json`'s `diagnosticSeverityOverrides`.

It makes the `diagnosticSeverityOverrides` work the same as the corresponding config file options and the `# pyright` comment overrides, which already support `true` and `false` for all the `reportXXX` rules. The value `true`  will be treated as "error" and `false` will be treated as "none".


erictraut (approved):
Looks good to me. Thanks for doing this!

debonte (approved):

PR #6408, opened 2023-11-09
Potential fix for bind failure during source file check

rchiodo:
Addresses https://github.com/microsoft/pylance-release/issues/5097

Saw a potential spot where this problem might occur:

Binding is required but it doesn't happen because a file is skipped. Check is still called in that situation.
Also added more information when binding is required but the check happens anyway.

### DIFF
--- a/packages/pyright-internal/src/analyzer/program.ts
+++ b/packages/pyright-internal/src/analyzer/program.ts
@@ -1890,12 +1890,12 @@ export class Program {
                 return false;
             }
 
-            if (!this._disableChecker) {
+            if (!this._disableChecker && !fileToCheck.sourceFile.isFileDeleted()) {
                 // For ipython, make sure we check all its dependent files first since
                 // their results can affect this file's result.
                 const dependentFiles = this._checkDependentFiles(fileToCheck, chainedByList, token);
 
-                this._bindFile(fileToCheck);
+                this._bindFile(fileToCheck, undefined, /* skipFileNeededCheck */ fileToCheck.sourceFile.isBindingRequired());
                 if (this._preCheckCallback) {
                     const parseResults = fileToCheck.sourceFile.getParseResults();
                     if (parseResults) {
--- a/packages/pyright-internal/src/analyzer/program.ts
+++ b/packages/pyright-internal/src/analyzer/program.ts
@@ -1895,7 +1895,11 @@ export class Program {
                 // their results can affect this file's result.
                 const dependentFiles = this._checkDependentFiles(fileToCheck, chainedByList, token);
 
-                this._bindFile(fileToCheck, undefined, /* skipFileNeededCheck */ fileToCheck.sourceFile.isBindingRequired());
+                this._bindFile(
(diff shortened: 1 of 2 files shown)


packages/pyright-internal/src/analyzer/sourceFile.ts
```diff
@@ -135,6 +135,18 @@ class WriteableData {
     builtinsImport: ImportResult | undefined;
     // True if the file appears to have been deleted.
     isFileDeleted = false;
+
+    dump() {
+        return JSON.stringify(this, (key, value) => {
```
erictraut, line 140: I think this may be unsafe. There are data structures in the WritableData that have circularities, so JSON.stringify will fail. I recommend outputting specific fields that you're interested in. Also, maybe pick a better name than `dump`.

packages/pyright-internal/src/analyzer/program.ts
```diff
@@ -1890,12 +1890,16 @@ export class Program {
                 return false;
             }
 
-            if (!this._disableChecker) {
+            if (!this._disableChecker && !fileToCheck.sourceFile.isFileDeleted()) {
```
erictraut, line 1893: I wonder if it would be better to add the check for `isFileDeleted()` within `_bindFile`? I guess it depends on whether there are other places that call `_bindFile` that do not already check for this condition.

packages/pyright-internal/src/analyzer/program.ts
```diff
@@ -1890,12 +1890,16 @@ export class Program {
...
                 // For ipython, make sure we check all its dependent files first since
                 // their results can affect this file's result.
                 const dependentFiles = this._checkDependentFiles(fileToCheck, chainedByList, token);
 
-                this._bindFile(fileToCheck);
+                this._bindFile(
+                    fileToCheck,
+                    undefined,
+                    /* skipFileNeededCheck */ fileToCheck.sourceFile.isBindingRequired()
```
erictraut, line 1901: Why is this needed? I think I understand, but it's not obvious. Perhaps include a comment that explains why — to prevent it from being deleted in the future.

packages/pyright-internal/src/analyzer/program.ts
rchiodo: `_bindFile` actually already checks for this and skips binding if the file is deleted. I was thinking that maybe _bindFile should return a boolean indicating if it actually bound the file or not (instead of doing this check here), so callers could skip calling things that depend upon the bind happening.

packages/pyright-internal/src/analyzer/sourceFile.ts
rchiodo: Sure will do. The replacer should have prevented the circular printing but that would be safer.

packages/pyright-internal/src/analyzer/program.ts
rchiodo: Thanks. I'll add a comment.

erictraut (approved):

Issue #6470, opened 2023-11-16
Assigning param value via walrus operator `:=` in function with `@overload` signature results in `reportGeneralTypeIssues` error


alexander-beedie:
### Describe the bug

Taking advantage of the walrus operator when assigning a value to a parameter in a function that has an overloaded signature can result in a `reportGeneralTypeIssues` error.

### Code or Screenshots

Have condensed the original issue (raised in the polars repo ;) down to the following MWE:
```python
from typing import overload

@overload
def func(value: None = None) -> None: ...

@overload
def func(value: str) -> str: ...

def func(value: str | None = None) -> str | None:
    return value

# no issues
func( value='test' )

# error: 'No overloads for "func" match the provided arguments'
func( value=(param_value := 'test') )
```
Saving that code into a "test.py" file and executing `pyright test.py` shows the error:
```
pyright test.py
.../test.py:16:1 - error: No overloads for "func" match the provided arguments
1 error, 0 warnings, 0 informations 
```
Validating the same code with `mypy` seems fine:
```
mypy test.py  
Success: no issues found in 1 source file
```
**Also:** removing the overloads (but keeping the walrus assignment in the function call) works fine in _both_ linters.

erictraut:
Thanks for the bug report. This will be fixed in the next release.

erictraut:
This is addressed in pyright 1.1.337, which I just published. It will also be included in a future release of pylance.

PR #6471, opened 2023-11-16
Fixed a bug that resulted in a false positive error when using a walr…

erictraut:
…us operator in a call expression that targets an overloaded function. This addresses #6470.


alexander-beedie:
Now that's fast! Many thanks 🥳

Issue #6314, opened 2023-11-03
1.1.333, 1.1.334 regression - high CPU usage (~100% single core) in VS Code


tomasz-c:
**Describe the bug**

- update the pyrigtht extension to version 1.1.333 or 1.1.334,
- open the directory containing the git repository and python scripts in VS Code
- click on any python script to display it in VS Code,

This creates an electron process that consumes ~100% of the CPU (on one core). Then pyright doesn't work (it doesn't suggest the code). Closing VS Code does not end this process, it stays and consumes CPU.
If a directory without a git repository is opened, or the git extension in VS Code is disabled ("git.enabled": false), pyright works correctly. However, the cause seems to be pyright, as downgrading to 1.1.332 solves this problem.

**VS Code extension or command-line**
The issue occurs as an extension in VS Code
Tested in VS Code versions: 1.83.1 and 1.84.0 on Linux
I also have "python.languageServer" set to "None", but changing it to 'jedi' doesn't seem to change anything.

erictraut:
Thanks for reporting the issue, but you haven't provided any repro steps. Without a repro case, I won't be able to make any further progress on the issue. Is your code base public? If so, please provide the steps required to clone the repo and install the dependencies. If your code isn't public, you'll need to narrow down the problem to a particular piece of code and provide a minimal sample or give me temporary access to a repo that contains the relevant portions.

mephinet:
I can confirm the issue, but I've failed to single out a minimal setup that shows the issue in a reproducible way...

erictraut:
I've looked through all of the changes between 1.1.333 and 1.1.334. The one that stands out as a potential culprit is this the bug fix for #6234. Another possibility is the fix for #6224.

I still need a repro.

tomasz-c:
I thought this happens with any repository and python code, but it doesn't.
This can be reproduced with sample public code:

https://github.com/uwi-info3180/flask-sqlite

Open the flask-sqlite directory as 'Open folder' in VS Code and click on the 'run.py' file to load into the editor.

erictraut:
Hmm, I'm not able to repro the problem you're seeing either with or without the Git extension enabled. It's curious that you're able to repro it only with the Git extension disabled. That seems to indicate some interaction between Pyright and the Git extension, but I don't have any working theories.

@debonte, @rchiodo — let me know if you have any ideas.

mephinet:
[...]

PR #6477, opened 2023-11-16
Fix infinite loop in _tryWalkUp

### LINKED ISSUE #6465: Something in hover provider broke in a commit released in 1.1.333
Hover suggestion just shows "loading" when I hit Ctrl+Space in a custom Monaco setup.

Expected behavior: Should show something, whether a keyword, or a name, or something similar.

Bug Screenshot
image

The bug seems to occur on 1.1.333 and after.

So, I decided to set up a development instance of pyright, and integrate it into my custom Monaco setup. After some digging I've concluded that the offending commit is `* cfe56bfb3 pull-pylance-with-pyright-1.1.332 (#6195)`

I've been trying to track down the offending changes in the above commit, but I've been coming up short.
edit: I think the offending changes are in `packages/pyright-internal/src/common/pathUtils.ts`

Any additional information I can get for you, let me know. Thanks!

rchiodo:
Non file URIs don't ever indicate diskPathRoot, so double check we're not going to loop forever.
Fixes #6465 

Should also fix https://github.com/microsoft/pyright/issues/6314

### DIFF
--- a/packages/pyright-internal/src/analyzer/importResolver.ts
+++ b/packages/pyright-internal/src/analyzer/importResolver.ts
@@ -199,7 +199,7 @@ export class ImportResolver {
             );
 
             let success;
-            [success, current] = this._tryWalkUp(current);
+            [success, current] = this._(current);
             if (!success) {
                 break;
             }
@@ -576,7 +576,7 @@ export class ImportResolver {
             }
 
             let success;
-            [success, current] = this._tryWalkUp(current);
+            [success, current] = this._(current);
             if (!success) {
                 break;
             }
@@ -1247,7 +1247,7 @@ export class ImportResolver {
                 }
 
                 let success;
-                [success, current] = this._tryWalkUp(current);
+                [success, current] = this._(current);
                 if (!success) {
                     break;
                 }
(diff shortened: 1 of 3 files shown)


packages/pyright-internal/src/analyzer/importResolver.ts
```diff
@@ -199,7 +199,7 @@ export class ImportResolver {
             );
 
             let success;
-            [success, current] = this._tryWalkUp(current);
+            [success, current] = this._(current);
```
debonte, line 202: How does this work? What is `this._`?

rchiodo: What the hell? Where did that happen?

rchiodo: Uggh, that's what I get for using the github browser editor.

debonte (approved):

packages/pyright-internal/src/tests/samples/stubs/placeholder.txt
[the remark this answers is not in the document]
rchiodo: It's so the directory is created. You can't add an empty directory in git AFAICT. 

I'll add a bit of text explaining that.

erictraut (approved):

PR #6480, opened 2023-11-17
New workflow to let Pylance know when Pyright publishes

bschnurr:
Adding a new workflow help keep Pylance and Pyright in sync

### DIFF
--- a/.github/workflows/trigger_pylance.yaml
+++ b/.github/workflows/trigger_pylance.yaml
@@ -3,7 +3,6 @@ name: Trigger Pylance Dispatch on Release
 on:
   release:
     types: [published]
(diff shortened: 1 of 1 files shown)


.github/workflows/trigger_pylance.yaml
```diff
@@ -0,0 +1,16 @@
...
+on:
+  release:
+    types: [published]
+
+jobs:
+  triggerDispatch:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Repository Dispatch
+        uses: peter-evans/repository-dispatch@v1
```
bschnurr, line 12:
```suggestion
        uses: peter-evans/repository-dispatch@v2
```

PR #6519, opened 2023-11-21
Change all file paths to use a new `URI` class.

rchiodo:
This should allow Pylance (and Pyright) to support files that come from locations other than the local file system.

All locations that used to refer to a file path have now been changed to a URI with the following exceptions:

- command line arguments still use strings. They're converted to URIs when being consumed by the Pyright server.
- tests still use strings. This meant changing a lot less code in the tests. Conversion happens in the testState layer and in some of the other setup code
- realFileSystem converts URIs to their file path, so it's making an assumption that all URIs are actually files. This will be addressed at a future point (in Pylance actually).

Some other caveats:
- Empty string is not allowed as a filePath anymore (importResolver used this to distinguish unresolved paths). This has been replaced with an 'Empty' URI.
- Performance is slightly slower. For running against Pandas, Resolving imports now takes about 800ms vs 500ms. This could potentially be improved with more work on optimization.
- Optimization was done mostly with caching (or memoization). This _shouldnt_ have much memory overhead as the number of unique items is actually rather small compared to the number of calls, but it may need more testing on larger code bases to make sure.


packages/pyright-internal/src/common/uri/memoization.ts
```diff
@@ -0,0 +1,99 @@
...
+    return function (target: any, functionName: string, descriptor: PropertyDescriptor) {
+        const originalMethod = descriptor.value;
+        descriptor.value = function (this: any, ...args: any) {
+            // Call the function once to get the result.
+            const result = originalMethod.apply(this, args);
+
+            // Then we replace the original function with one that just returns the result.
+            this[functionName] = () => {
+                // Note that this poses a risk. The result is passed by reference, so if the caller
+                // modifies the result, it will modify the cached result.
```
rchiodo, line 52: I mitigated this risk in callers by changing return values to readOnly. One such location was getPathComponents. A lot of code modifies the returned array. That didn't work with this caching, so instead I changed getPathComponents to return a `readonly string[]` which forced all callers to make a copy if they wanted to modify it.

rchiodo: That copy of the array was still faster than having getPathComponents not be cached.

erictraut:
@rchiodo, I think it would be preferable not to output the `file://` part of the URI in errors output by the CLI. It's redundant since the CLI deals only with files, not other URI types.

rchiodo:
> @rchiodo, I think it would be preferable not to output the `file://` part of the URI in errors output by the CLI. It's redundant since the CLI deals only with files, not other URI types.

Will do.

rchiodo:
I'll also need to see why those imports aren't found. And potentially add tests or those cases.

erictraut:
Don't worry about the imports. These "mypy_primer" tests produce many errors. The import errors are expected because "mypy_primer" doesn't install dependencies for most of these code bases before it runs the type checker. The purpose of "mypy_primer" is to run before and after a change to see if there are any differences in the output.

rchiodo:
> Don't worry about the imports. These "mypy_primer" tests produce many errors. The import errors are expected because "mypy_primer" doesn't install dependencies for most of these code bases before it runs the type checker. The purpose of "mypy_primer" is to run before and after a change to see if there are any differences in the output.

Oh I initially thought these were bugs. I should have looked closer, it was just a difference in the output paths.

packages/pyright-internal/src/analyzer/importResolver.ts
```diff
@@ -1942,7 +1904,7 @@ export class ImportResolver {
                             }
                         } else if (innerEntry.isFile()) {
                             if (innerEntry.name.endsWith('.pyi')) {
-                                const strippedFileName = stripFileExtension(innerEntry.name);
+                                const strippedFileName = innerEntry.name.slice(0, -3);
```
debonte, line 1907: Shouldn't this be -4 instead of -3?

Also, shouldn't this be removing the extension in the same way that `_buildStdlibCache` does for `.py`. I'm not in love with either approach, but they should be the same?

packages/pyright-internal/src/analyzer/importResolver.ts
```diff
@@ -2429,44 +2397,52 @@ export class ImportResolver {
...
-            const initPyiPath = combinePaths(currentPath, dir, '__init__.pyi');
+            const initPyiPath = dir.initStubUri;
             if (this.fileExistsCached(initPyiPath)) {
-                suggestions.set(dir, initPyiPath);
+                suggestions.set(dirSuggestion, initPyiPath);
                 return;
             }
 
-            const initPyPath = combinePaths(currentPath, dir, '__init__.py');
+            const initPyPath = dir.initFileUri;
```
debonte, line 2438: Maybe `initPyUri` and `initPyiUri` would be more clear than `initFileUri` and `initStubUri`?

packages/pyright-internal/src/analyzer/importResolver.ts
```diff
@@ -2656,12 +2638,12 @@ export class ImportResolver {
...
     }
 
-    private _isNativeModuleFileName(moduleName: string, fileName: string): boolean {
+    private _isNativeModuleFileName(moduleName: string, fileUri: Uri): boolean {
         // Strip off the final file extension and the part of the file name
         // that excludes all (multi-part) file extensions. This allows us to
         // handle file names like "foo.cpython-32m.so".
-        const fileExtension = getFileExtension(fileName, /* multiDotExtension */ false).toLowerCase();
-        const withoutExtension = stripFileExtension(fileName, /* multiDotExtension */ true);
+        const fileExtension = fileUri.extname.toLowerCase();
```
debonte, line 2645: There were other places where `extname` was checked against literal lowercase values. Maybe `extname` should do the `toLowerCase` call internally?

packages/pyright-internal/src/analyzer/importResolver.ts
```diff
@@ -1891,17 +1853,17 @@ export class ImportResolver {
...
-            const readDir = (root: string, prefix: string | undefined) => {
+            const readDir = (root: Uri, prefix: string | undefined) => {
                 this.readdirEntriesCached(root).forEach((entry) => {
                     if (entry.isDirectory()) {
-                        const dirRoot = combinePaths(root, entry.name);
+                        const dirRoot = root.combinePaths(entry.name);
                         readDir(dirRoot, prefix ? `${prefix}.${entry.name}` : entry.name);
                     } else if (entry.name.includes('.py')) {
-                        const stripped = stripFileExtension(entry.name);
+                        const stripped = entry.name.split('.py')[0];
```
debonte, line 1866: Should we still be calling `stripFileExtension` here? `split` is going to do the wrong thing for weird filenames like `weird.python.py`.

packages/pyright-internal/src/analyzer/packageTypeVerifier.ts
```diff
@@ -99,7 +94,7 @@ export class PackageTypeVerifier {
             this._configOptions.evaluateUnknownImportsAsAny = true;
         }
 
-        this._execEnv = this._configOptions.findExecEnvironment('.');
+        this._execEnv = this._configOptions.findExecEnvironment(Uri.file('.'));
```
debonte, line 97: Should `Uri.file(".")` be a constant? Do we use it elsewhere?

packages/pyright-internal/src/common/uri/uri.ts
```diff
@@ -0,0 +1,206 @@
...
+    getPathComponents(): readonly string[];
+
+    getPath(): string;
+
+    getFilePath(): string;
+
+    getRelativePathComponents(to: Uri): readonly string[];
+    getShortenedFileName(maxDirLength?: number): string;
+
+    stripExtension(): Uri;
```
debonte, line 104: Can we organize these methods into related groups? For example, all of the `fooExtension` methods should be together.

packages/pyright-internal/src/common/uri/uri.ts
```diff
@@ -0,0 +1,206 @@
...
+    readonly key: string;
+
+    // Returns the scheme of the URI.
+    readonly scheme: string;
+
+    // Returns the last segment of the URI, similar to the UNIX basename command.
+    readonly basename: string;
+
+    // Returns the extension of the URI, similar to the UNIX extname command.
+    readonly extname: string;
```
debonte, line 26: I noticed that `extname` returns ".ext". Should it return "ext"? That might be a religious battle. I don't feel strongly either way, but I think it should be documented here.

packages/pyright-internal/src/common/uri/uri.ts
```diff
@@ -0,0 +1,206 @@
...
+
+export interface Uri {
+    // Unique key for storing in maps.
+    readonly key: string;
+
+    // Returns the scheme of the URI.
+    readonly scheme: string;
+
+    // Returns the last segment of the URI, similar to the UNIX basename command.
+    readonly basename: string;
```
debonte, line 23: The term `basename` means nothing to me. Maybe I just need to add this to my internal dictionary, but I wonder if there's a better name.

.NET chose to name their equivalent as [`GetFileName`](https://learn.microsoft.com/en-us/dotnet/api/system.io.path.getfilename?view=net-8.0#system-io-path-getfilename(system-string)). Maybe we should follow their lead?

GetFileName("c:\foo\") returns ""
GetFileName("c:\foo\bar") returns "bar"
GetFileName("c:\foo\bar.txt") returns "bar.txt"

debonte (commented):
I haven't even gotten through the `src\analyzer` files yet :), but here are my thoughts so far.

packages/pyright-internal/src/analyzer/binder.ts
```diff
@@ -2462,10 +2463,12 @@ export class Binder extends ParseTreeWalker {
...
         const resolvedPath =
-            aliasDecl?.path && aliasDecl.loadSymbolsFromPath
-                ? aliasDecl.path
-                : aliasDecl?.submoduleFallback?.path && aliasDecl.submoduleFallback.loadSymbolsFromPath
-                ? aliasDecl.submoduleFallback.path
+            aliasDecl?.uri && aliasDecl.loadSymbolsFromPath && !aliasDecl.uri.isEmpty()
+                ? aliasDecl.uri
+                : aliasDecl?.submoduleFallback?.uri &&
+                  aliasDecl.submoduleFallback.loadSymbolsFromPath &&
+                  !aliasDecl.submoduleFallback.uri.isEmpty()
```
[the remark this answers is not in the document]
rchiodo: There isn't. I wish. Would have made finding the bugs with truthiness a lot easier.

packages/pyright-internal/src/analyzer/declaration.ts
```diff
@@ -31,7 +32,7 @@ import {
     YieldNode,
 } from '../parser/parseNodes';
 
-export const UnresolvedModuleMarker = '*** unresolved ***';
+export const UnresolvedModuleMarker = Uri.parse('unresolved-module-marker://**/*');
```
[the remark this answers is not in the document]
rchiodo: No it doesn't change the URI that is generated. It just has to match on the equals.

packages/pyright-internal/src/analyzer/importResolver.ts
rchiodo: Yes this looks like a mistake. I'll put it back.

packages/pyright-internal/src/analyzer/importResolver.ts
```diff
@@ -370,7 +358,7 @@ export class ImportResolver {
         }
 
         // The "default" workspace has a root-less execution environment; ignore it.
-        if (execEnv.root) {
+        if (execEnv.root && Uri.isUri(execEnv.root)) {
```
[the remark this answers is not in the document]
rchiodo: No it can be a string too. At least the type is. It's likely a URI by this point though.

packages/pyright-internal/src/analyzer/importResolver.ts
```diff
@@ -163,36 +156,36 @@ export class ImportResolver {
...
 
-        let current = origin;
+        let current: Uri | undefined = origin;
         while (this._shouldWalkUp(current, root, execEnv)) {
             this._getCompletionSuggestionsAbsolute(
-                sourceFilePath,
+                sourceFileUri,
                 execEnv,
-                current,
+                current!,
```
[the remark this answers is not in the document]
rchiodo: Let me double check. it might still need the URI on the false case?

packages/pyright-internal/src/analyzer/declarationUtils.ts
```diff
@@ -78,7 +79,7 @@ export function areDeclarationsSame(
         return false;
     }
 
-    if (decl1.path !== decl2.path) {
+    if (!decl1.uri.equals(decl2.uri)) {
```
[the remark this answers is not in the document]
rchiodo: No it does not. There is no such thing in Typescript.

packages/pyright-internal/src/analyzer/importResolver.ts
rchiodo: Yeah this looks like a bug. I should put it back to the way it was before.

packages/pyright-internal/src/analyzer/importResolver.ts
rchiodo: Good idea. I'll change that.

packages/pyright-internal/src/analyzer/packageTypeVerifier.ts
rchiodo: This might be the only spot. But it is effectively a constant because of the caching that happens.

packages/pyright-internal/src/common/uri/uri.ts
rchiodo: I'll add a comment.

packages/pyright-internal/src/common/uri/uri.ts
rchiodo: filename makes sense to me. I'll change it to that.

packages/pyright-internal/src/analyzer/importResolver.ts
```diff
@@ -1351,10 +1321,11 @@ export class ImportResolver {
             for (let i = 0; i < moduleDescriptor.nameParts.length; i++) {
                 const isFirstPart = i === 0;
                 const isLastPart = i === moduleDescriptor.nameParts.length - 1;
-                dirPath = combinePaths(dirPath, moduleDescriptor.nameParts[i]);
+                const oldDirPath = dirPath;
+                dirPath = dirPath.combinePaths(moduleDescriptor.nameParts[i]);
 
                 if (useStubPackage && isFirstPart) {
-                    dirPath += stubsSuffix;
+                    dirPath = oldDirPath.combinePaths(`${moduleDescriptor.nameParts[i]}${stubsSuffix}`);
```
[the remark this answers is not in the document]
rchiodo: I need to double check, but I believe addPath would be appropriate here.

packages/pyright-internal/src/analyzer/declarationUtils.ts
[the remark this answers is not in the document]
rchiodo: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_equality

packages/pyright-internal/src/analyzer/importResolver.ts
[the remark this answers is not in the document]
debonte: If it's a string, it is guaranteed to be a stringified URI? If not, maybe I'm missing something, but this seems like a symptom of the problem you're trying to fix? That we don't know if a string is a URI or not?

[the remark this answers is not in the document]
rchiodo: No, it's not guaranteed to be a stringified URI. I believe this was a result of root sometimes being relative from the command line options. Perhaps I can fix that problem the same way I fixed the importResolver.tests.ts to check if the command line has a relative path or not.

packages/pyright-internal/src/analyzer/binder.ts
```diff
@@ -2515,11 +2518,11 @@ export class Binder extends ParseTreeWalker {
             .getDeclarations()
             .find((decl) => decl.type === DeclarationType.Alias && decl.firstNamePart === firstNamePartValue);
         let newDecl: AliasDeclaration;
-        let pathOfLastSubmodule: string;
+        let uriOfLastSubmodule: Uri;
         if (importInfo && importInfo.isImportFound && !importInfo.isNativeLib && importInfo.resolvedPaths.length > 0) {
-            pathOfLastSubmodule = importInfo.resolvedPaths[importInfo.resolvedPaths.length - 1];
+            uriOfLastSubmodule = importInfo.resolvedPaths[importInfo.resolvedPaths.length - 1];
```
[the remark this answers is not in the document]
rchiodo: Sure sounds reasonable.

packages/pyright-internal/src/analyzer/importResolver.ts
rchiodo: Will change.

packages/pyright-internal/src/analyzer/importResolver.ts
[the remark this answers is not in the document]
rchiodo: combinePaths wouldn't because it would add a directory separator before the suffix.

packages/pyright-internal/src/analyzer/importResolver.ts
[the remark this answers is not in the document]
rchiodo: No it doesn't. I'll change this to just return URI | undefined.

packages/pyright-internal/src/analyzer/importResolver.ts
rchiodo: Actually  no we depend upon casing (at least some o the tests fail if I change it) I believe a '.Py' is not the same as a '.py' file.

packages/pyright-internal/src/analyzer/importResolver.ts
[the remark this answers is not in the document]
debonte: We're still using `current!` at the top of the loop. Can't we avoid that by testing the return value of `_tryWalkUp`?

packages/pyright-internal/src/analyzer/program.ts
```diff
@@ -391,23 +382,22 @@ export class Program {
...
                 /* isThirdPartyImport */ false,
                 moduleImportInfo.isThirdPartyPyTypedPresent,
                 this._editModeTracker,
                 this._console,
                 this._logTracker,
-                options?.realFilePath,
                 options?.ipythonMode ?? IPythonMode.None
             );
-            const chainedFilePath = options?.chainedFilePath;
+            const chainedFilePath = options?.chainedUri;
```
debonte, line 400: Should `chainedUri` be `chainedFileUri`?

packages/pyright-internal/src/analyzer/service.ts
```diff
@@ -288,93 +284,81 @@ export class AnalyzerService {
 
         this._backgroundAnalysisProgram.setConfigOptions(configOptions);
 
-        this._executionRootPath = realCasePath(
-            combinePaths(commandLineOptions.executionRoot, configOptions.projectRoot),
-            this.fs
-        );
+        this._executionRootUri = configOptions.projectRoot;
```
debonte, line 287: What's the difference between `executionRoot` and `projectRoot`? Why aren't we combining them anymore?

packages/pyright-internal/src/analyzer/tracePrinter.ts
```diff
@@ -22,36 +22,33 @@ export type PrintableType = ParseNode | Declaration | Symbol | Type | undefined;
...
+    function printFileOrModuleName(fileUriOrModule: Uri | AbsoluteModuleDescriptor | undefined) {
+        if (fileUriOrModule) {
+            if (Uri.isUri(fileUriOrModule)) {
                 for (const root of roots) {
-                    if (filePathOrModule.startsWith(root)) {
-                        const subFile = filePathOrModule.substring(root.length);
-                        return stripFileExtension(subFile).replace(separatorRegExp, '.');
+                    if (fileUriOrModule.isChild(root)) {
+                        const subFile = root.getRelativePath(fileUriOrModule);
+                        return subFile!.split('.')[0].replace(separatorRegExp, '.');
```
debonte, line 45: Should we continue to use the `stripFileExtension` from `pathUtils` for this?

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -5359,7 +5360,7 @@ export function createTypeEvaluator(importLookup: ImportLookup, evaluatorOptions
                                 fileInfo.executionEnvironment.pythonVersion >= PythonVersion.V3_7 ||
                                 getAttrSymbol
                                     .getDeclarations()
-                                    .some((decl) => decl.path.toLowerCase().endsWith('.pyi'));
+                                    .some((decl) => decl.uri.extname.toLowerCase().endsWith('.pyi'));
```
debonte, line 5363: Is `extname...endsWith` correct? Is that necessary to handle foo.bar.pyi? Or is it superfluous?

Also makes me wonder if `Uri` should support a method like `hasFileExtension(".pyi")`

packages/pyright-internal/src/analyzer/typeStubWriter.ts
```diff
@@ -158,13 +159,13 @@ export class TypeStubWriter extends ParseTreeWalker {
...
 
-    constructor(private _stubPath: string, private _sourceFile: SourceFile, private _evaluator: TypeEvaluator) {
+    constructor(private _stubPath: Uri, private _sourceFile: SourceFile, private _evaluator: TypeEvaluator) {
         super();
 
         // As a heuristic, we'll include all of the import statements
         // in "__init__.pyi" files even if they're not locally referenced
         // because these are often used as ways to re-export symbols.
-        if (this._stubPath.endsWith('__init__.pyi')) {
+        if (this._stubPath.pathEndsWith('__init__.pyi')) {
```
debonte, line 168: Shouldn't this technically be checking `filename` rather than `pathEndsWith` so file names like `foo__init__.pyi` don't get picked up?

packages/pyright-internal/src/backgroundAnalysisBase.ts
```diff
@@ -339,20 +345,20 @@ export abstract class BackgroundAnalysisRunnerBase extends BackgroundThreadBase
...
             }
 
             case 'getDiagnosticsForRange': {
                 run(() => {
-                    const { filePath, range, cancellationId } = msg.data;
+                    const { fileUri, range, cancellationId } = msg.data;
                     const token = getCancellationTokenFromId(cancellationId);
 
-                    return this.handleGetDiagnosticsForRange(filePath, range, token);
+                    return this.handleGetDiagnosticsForRange(fileUri, range, token);
```
debonte, line 361: Where does `fileUri` get converted back from a string to a `Uri` object? Seems like it must be happening since `this.handleGetDiagnosticsForRange` takes a `Uri`, but isn't the `fileUri` that was sent in the `msg.data` a string?

packages/pyright-internal/src/analyzer/service.ts
rchiodo: URIs are never relative. So you can't combine them anymore. project root gets computed from the execution root now.

packages/pyright-internal/src/analyzer/program.ts
rchiodo: Sure, I don't think it matters much. I guess chainedFileUri indicates it's a file though. I'll change it.

packages/pyright-internal/src/analyzer/tracePrinter.ts
rchiodo: Yes another good catch.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
rchiodo: endswith is superfluous here. Will fix.

packages/pyright-internal/src/analyzer/typeStubWriter.ts
rchiodo: Good point.

packages/pyright-internal/src/backgroundAnalysisBase.ts
rchiodo: Oh this is a really good catch. This wouldn't have worked. It wasn't doing the conversion. I guess none of our tests use the background thread.

packages/pyright-internal/src/analyzer/importResolver.ts
debonte: VS Code recognizes `CAPITAL.PY` as a python file and Pylance analyzes it, for what that's worth.

.vscode/launch.json
```diff
@@ -8,9 +8,11 @@
             "program": "${workspaceRoot}/packages/pyright/index.js",
             "preLaunchTask": "npm: build:cli:dev",
             "args": [
-                "--verifytypes",
-                "pandas.core.reshape.concat",
+                "--stats",
```
heejaechang, line 11: is it supposed to be checked in?

rchiodo: I don't think it matters as that original one was specific to some other test case.

rchiodo: Although maybe that original one runs on everybody's machine. I can revert it.

packages/pyright-internal/src/common/uri/baseUri.ts
```diff
@@ -0,0 +1,307 @@
...
+        return Object.freeze(this.getPathComponentsImpl());
+    }
+
+    abstract getPath(): string;
+
+    abstract getFilePath(): string;
+
+    @cacheMethodWithArgs()
+    getRelativePathComponents(to: Uri): readonly string[] {
+        const fromComponents = this.getPathComponents();
```
rchiodo, line 203: This should probably return empty if the schemes don't match

packages/pyright-internal/src/analyzer/importResolver.ts
```diff
@@ -612,24 +596,30 @@ export class ImportResolver {
...
 
-    protected dirExistsCached(path: string): boolean {
-        const splitPath = this._splitPath(path);
-
-        if (!splitPath[0] || !splitPath[1]) {
-            if (!this.fileSystem.existsSync(path)) {
-                return false;
+    protected dirExistsCached(uri: Uri): boolean {
+        const parent = uri.getDirectory();
+        if (parent === uri) {
```
rchiodo, line 601: This should likely use equals

packages/pyright-internal/src/backgroundAnalysisBase.ts
```diff
@@ -136,7 +137,7 @@ export class BackgroundAnalysisBase {
         const cancellationId = getCancellationTokenId(token);
         this.enqueueRequest({
             requestType: 'getDiagnosticsForRange',
-            data: { filePath, range, cancellationId },
+            data: { fileUri: fileUri.toString(), range, cancellationId },
```
heejaechang, line 140: I see, you need this since uri is not pure `object` but has `class implementation`

rchiodo: If it was a pure object would this somehow just call 'toString' automatically?

packages/pyright-internal/src/analyzer/binder.ts
```diff
@@ -4016,14 +4019,14 @@ export class Binder extends ParseTreeWalker {
                 ? loaderActions.implicitImports.get(implicitImport.name)
                 : undefined;
             if (existingLoaderAction) {
-                existingLoaderAction.path = implicitImport.path;
+                existingLoaderAction.uri = implicitImport.path;
```
heejaechang, line 4022: implicitImport.path -> implicitImport.uri ?

packages/pyright-internal/src/analyzer/importResolver.ts
```diff
@@ -268,73 +257,72 @@ export class ImportResolver {
 
             const relativeStubPaths: string[] = [];
             for (const importRootPath of importRootPaths) {
```
heejaechang, line 259: importRootPaths -> importRootUris?

packages/pyright-internal/src/analyzer/importResolver.ts
```diff
@@ -431,7 +421,7 @@ export class ImportResolver {
         this._invalidateFileSystemCache();
         return true;
 
-        function addPaths(path?: string) {
+        function addPaths(path?: Uri) {
```
heejaechang, line 424: path -> uri?

packages/pyright-internal/src/analyzer/checker.ts
```diff
@@ -231,7 +231,7 @@ export class Checker extends ParseTreeWalker {
         const codeComplexity = AnalyzerNodeInfo.getCodeFlowComplexity(this._moduleNode);
 
         if (isPrintCodeComplexityEnabled) {
-            console.log(`Code complexity of module ${this._fileInfo.filePath} is ${codeComplexity.toString()}`);
+            console.log(`Code complexity of module ${this._fileInfo.fileUri} is ${codeComplexity.toString()}`);
```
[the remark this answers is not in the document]
rchiodo: There is a distinction I think that's subtle. Something called a 'fileUri' is expected to actually point to a file. Something called a 'uri' could be a directory or a file.

packages/pyright-internal/src/analyzer/importResolver.ts
```diff
@@ -474,47 +464,47 @@ export class ImportResolver {
             if (versionRange.max !== undefined && pythonVersion > versionRange.max) {
                 // Add excludes for both the ".pyi" file and the directory that contains it
                 // (in case it's using a "__init__.pyi" file).
-                const moduleDirPath = combinePaths(typeshedStdlibPath, ...moduleName.split('.'));
+                const moduleDirPath = typeshedStdlibPath.combinePaths(...moduleName.split('.'));
```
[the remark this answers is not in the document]
rchiodo: I think this one should actually stay 'path' as it's the path/directory to the typeshed stdlib. Stdlib path is the common way to talk about this?

rchiodo:
There seems to be no cache misses with the current code. I added a counter to the import resolver for when a cache miss happened in both this version and the current main. That counter came out the same for both.

I may have at one point had problems with the cache, but seems like those are fixed. Next step is to see if the memoization is worth the effort or not.

rchiodo:
Seems with the cache working, memoization is no longer necessary. It doesn't buy us any noticeable difference. I'm removing it.

erictraut:
> The memory usage is like 5-10% worse with the URIs than with strings

I'm really surprised by this. My intuition is that it should be in the noise — less than 0.1%. These URIs should not be created very often, especially with aggressive caching in the import resolution code paths. Any idea where these are getting created on a regular basis?

rchiodo:
> > The memory usage is like 5-10% worse with the URIs than with strings
> 
> I'm really surprised by this. My intuition is that it should be in the noise — less than 0.1%. These URIs should not be created very often, especially with aggressive caching in the import resolution code paths. Any idea where these are getting created on a regular basis?

Maybe I should have given more detail. Here's an example difference:

What's in main:

```
Heap stats: {
    "total_heap_size": 1512943616,
    "total_heap_size_executable": 12058624,
    "total_physical_size": 1512458160,
    "total_available_size": 1404477464,
    "used_heap_size": 767038960,
    "heap_size_limit": 2197815296,
    "malloced_memory": 532552,
    "peak_malloced_memory": 19791200,
    "does_zap_garbage": 0,
    "number_of_native_contexts": 1,
    "number_of_detached_contexts": 0,
    "total_global_handles_size": 8192,
    "used_global_handles_size": 4288,
    "external_memory": 795519
}
```

What's in this branch:

```
Heap stats: {
    "total_heap_size": 1539198976,
    "total_heap_size_executable": 12058624,
    "total_physical_size": 1538550136,
    "total_available_size": 1394446176,
    "used_heap_size": 776694152,
    "heap_size_limit": 2197815296,
    "malloced_memory": 270408,
    "peak_malloced_memory": 21914768,
    "does_zap_garbage": 0,
    "number_of_native_contexts": 1,
    "number_of_detached_contexts": 0,
    "total_global_handles_size": 8192,
    "used_global_handles_size": 4288,
    "external_memory": 795519
}
```

5-10% for that example is an exaggeration, it's more like 1%.

I added some instrumentation to see how many URIs get created for this example:

```
Analysis stats
Total files parsed and bound: 1867
Total files checked: 1547
Total URIs created: 201216
```

So 200K URIs.

rchiodo:
I'm working on adding more instrumentation to see what the most common call sites are. I'm guessing combinePath

packages/pyright-internal/src/common/uri/baseUri.ts
```diff
@@ -0,0 +1,288 @@
...
+        }
+    }
+
+    // Returns the extension of the URI, similar to the UNIX extname command.
+    get extname(): string {
+        return this.getExtnameImpl();
+    }
+
+    // Returns a URI where the path just contains the root folder.
+    get root(): Uri {
```
heejaechang, line 49: is there a reason why some uses `xxxImpl` and others just `abstract`? it looks like there is no common impl on top of `xxxImpl` so making them just `abstract` feels like get same effect?

packages/pyright-internal/src/common/uri/baseUri.ts
```diff
@@ -0,0 +1,288 @@
...
+    // Determines whether a Uri is a child of some parent Uri.
+    abstract isChild(parent: Uri, ignoreCase?: boolean): boolean;
+
+    abstract isLocal(): boolean;
+
+    isUntitled(): boolean {
+        return this.scheme === 'untitled';
+    }
+
+    equals(other: Uri | undefined, ignoreCase?: boolean): boolean {
```
heejaechang, line 127: do we still need `ignoreCase` ?

rchiodo: Oh I suppose not. Good point.

packages/pyright-internal/src/common/uri/baseUri.ts
```diff
@@ -0,0 +1,288 @@
...
+            if (relativeToComponents.length > 0) {
+                return ['.', ...relativeToComponents].join('/');
+            }
+        }
+        return undefined;
+    }
+
+    getPathComponents(): readonly string[] {
+        // Make sure to freeze the result so that it can't be modified.
+        return Object.freeze(this.getPathComponentsImpl());
```
heejaechang, line 180: so, this one having `xxxImpl` make sense since there is common code applied to `xxxImpl`

packages/pyright-internal/src/common/uri/baseUri.ts
rchiodo: This is a left over side effect of the memoization. The BaseUri would have the public function. It would be cached. The sub classes would implement the real implementation. I only did this for things that were called a whole lot.

I could do like you said, but didn't think it mattered much. Maybe that would be slightly faster.

rchiodo:
I was close. Most of it was getDirectory(), then combinePath():

```
Analysis stats
Total files parsed and bound: 1867
Total files checked: 1547
Total URIs created: 201216
URI call sites and count:
 FileUri.getDirectoryImpl  : 101100
 FileUri.combinePathsImpl  : 53829
 FileUri.addPath  : 33917
 Object.file  : 12362
 FileUri.getRootImpl  : 6
 __webpack_require__  : 1
 Object.parse  : 1
```

packages/pyright-internal/src/common/realFileSystem.ts
```diff
@@ -286,71 +301,81 @@ class RealFileSystem implements FileSystem {
...
     }
 
-    unlinkSync(path: string) {
+    unlinkSync(uri: Uri) {
+        const path = this._getNormalizedPath(uri);
         yarnFS.unlinkSync(path);
     }
 
-    realpathSync(path: string) {
+    realpathSync(uri: Uri) {
```
heejaechang, line 314: do we still need this?

packages/pyright-internal/src/common/uri/baseUri.ts
```diff
@@ -0,0 +1,288 @@
...
+            return this.key.toLowerCase() === other?.key.toLowerCase();
+        }
+        return this.key === other?.key;
+    }
+
+    abstract startsWith(other: Uri | undefined, ignoreCase?: boolean): boolean;
+
+    pathStartsWith(name: string): boolean {
+        // ignore path separators.
+        name = normalizeSlashes(name);
```
heejaechang, line 138: now, with everything changed to `Uri`, are there any case where we need to explicitly use `normalizeSlash` outside of `file system implementation`? 

in Uri, `/` is always used regardless of underlying file system?

rchiodo: Uri does not use `/` everywhere. Just the WebURI does. FileURI still uses whatever slash is appropriate for the file system.

packages/pyright-internal/src/common/realFileSystem.ts
rchiodo: The function? Not sure, but I think that would be a separate change. I'd rather not introduce more complexity if I can help it. I'd remove that separately.

packages/pyright-internal/src/common/uri/baseUri.ts
rchiodo: I'll simplify it to match what it would have been without memoization.

rchiodo:
I did some more instrumentation to see what the first caller of the functions were:

Here's like the top 15 spots:
```
Total URIs created: 201216
 FileUri.getDirectory  -> ImportResolver.fileExistsCached  : 37406
 FileUri.addPath  -> ImportResolver._resolveAbsoluteImport  : 33913
 FileUri.getDirectory  -> ImportResolver.dirExistsCached  : 21785
 FileUri.combinePaths  -> ImportResolver._resolveAbsoluteImport  : 15536
 FileUri.getDirectory  -> ImportResolver.getModuleNameInfoFromPath  : 15502
 FileUri.combinePaths  -> ImportResolver.getModuleNameInfoFromPath  : 15431
 FileUri.getDirectory  -> ImportResolver._resolveAbsoluteImport  : 14132
 Object.file  -> RealFileSystem.realCasePath  : 12348
 FileUri.combinePaths  -> ImportResolver._getPyTypedInfo  : 10427
 FileUri.getDirectory  -> ImportResolver.resolveImportInternal  : 10388
 FileUri.combinePaths  -> /home/rich/source/pyright/packages/pyright/dist/pyright-internal.js:92649:48 : 5496
 FileUri.combinePaths  -> getFileSystemEntriesFromDirEntries  : 4248
 FileUri.getDirectory  -> ImportResolver._getModuleNameForImport  : 1693
 FileUri.combinePaths  -> /home/rich/source/pyright/packages/pyright/dist/pyright-internal.js:29553:76 : 840
```

packages/pyright-internal/src/common/uri/fileUri.ts
```diff
@@ -0,0 +1,196 @@
...
+        if (!this._formattedString) {
+            this._formattedString =
+                this._originalString ||
+                URI.file(this._filePath).with({ query: this._query, fragment: this._fragment }).toString();
+        }
+        return this._formattedString;
+    }
+    override toUserVisibleString(): string {
+        return this._filePath;
+    }
```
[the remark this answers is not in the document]
rchiodo: toString() has the full uri, so like `file:///c:/user/source`, toUserVisibleString() just uses the file path `c:\user\source`. It was so the output of pyright CLI could remain the same between the two changes.

packages/pyright-internal/src/common/uri/fileUri.ts
```diff
@@ -0,0 +1,196 @@
...
+    @cacheMethodWithArgs()
+    override isChild(parent: Uri, ignoreCase?: boolean): boolean {
+        if (!FileUri.isFileUri(parent)) {
+            return false;
+        }
+
+        return this.startsWith(parent, ignoreCase) && parent._filePath.length < this._filePath.length;
+    }
+    override isLocal(): boolean {
+        return true;
```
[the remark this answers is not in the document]
rchiodo: isLocal means not on the web. It was originally used to detect if we were in a virtual workspace scenario or not.

packages/pyright-internal/src/common/uri/fileUri.ts
```diff
@@ -0,0 +1,196 @@
...
+    @cacheMethodWithArgs()
+    override startsWith(other: Uri | undefined, ignoreCase?: boolean): boolean {
+        if (!other || !FileUri.isFileUri(other)) {
+            return false;
+        }
+        if (other.isEmpty() !== this.isEmpty()) {
+            return false;
+        }
+        if (this.scheme !== other.scheme) {
+            return false;
```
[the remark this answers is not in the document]
rchiodo: Oh that's true. I'll get rid of this check.

packages/pyright-internal/src/common/uri/uriUtils.ts
```diff
@@ -0,0 +1,406 @@
...
+        return false;
+    }
+}
+
+export interface FileSystemEntries {
+    files: Uri[];
+    directories: Uri[];
+}
+
+export function forEachAncestorDirectory(
```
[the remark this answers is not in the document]
rchiodo: I didn't write this function, just copied it here to use URIs instead, so your analysis of how it works is more than I knew :)

packages/pyright-internal/src/common/uri/uriUtils.ts
```diff
@@ -0,0 +1,406 @@
...
+
+            wildcardRoot = wildcardRoot.combinePaths(component);
+        }
+    }
+
+    return wildcardRoot;
+}
+
+export function hasPythonExtension(uri: Uri) {
+    return uri.extname === '.py' || uri.extname === '.pyi';
```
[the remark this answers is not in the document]
rchiodo: They didn't before, so I didn't change them. A lot of the functions here were in pathUtils.ts before.

packages/pyright-internal/src/common/uri/webUri.ts
```diff
@@ -0,0 +1,199 @@
...
+export class WebUri extends BaseUri {
+    private static _cache = new Map<string, WebUri>();
+    private constructor(
+        private readonly _scheme: string,
+        private readonly _authority: string,
+        private readonly _path: string,
+        private readonly _query: string,
+        private readonly _fragment: string,
+        private readonly _originalString: string | undefined,
+        key: string
```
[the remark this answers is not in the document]
rchiodo: I suppose it could do that, but I was using the key in a cache at some point. I needed it before creating the object.

packages/pyright-internal/src/common/uri/webUri.ts
```diff
@@ -0,0 +1,199 @@
...
+    override getPath(): string {
+        return this._path;
+    }
+    override getFilePath(): string {
+        debug.fail(`${this} is not a file based URI.`);
+    }
+
+    protected override combinePathsImpl(...paths: string[]): Uri {
+        // Resolve and combine paths, never want URIs with '..' in the middle.
+        let combined = resolvePaths(this._path, ...paths).replace(/\\/g, '/');
```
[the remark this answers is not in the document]
rchiodo: No they can't, but a caller of this function can pass in `\` in the paths that are combined.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -5368,7 +5369,7 @@ export function createTypeEvaluator(importLookup: ImportLookup, evaluatorOptions
                                 fileInfo.executionEnvironment.pythonVersion >= PythonVersion.V3_7 ||
                                 getAttrSymbol
                                     .getDeclarations()
-                                    .some((decl) => decl.path.toLowerCase().endsWith('.pyi'));
+                                    .some((decl) => decl.uri.extname.toLowerCase() === '.pyi');
```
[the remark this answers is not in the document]
rchiodo: Sure, but it might be hard to find all of these locations.

packages/pyright-internal/src/common/uri/fileUri.ts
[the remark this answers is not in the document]
debonte: I see. I find this name pretty confusing. Please at least add a comment on `isLocal`. But maybe we can come up with a more self-explanatory name?

packages/pyright-internal/src/common/realFileSystem.ts
```diff
@@ -359,52 +384,37 @@ class RealFileSystem implements FileSystem {
...
-
-        if (driveLength === 0) {
-            return path;
-        }
-
-        // `vscode` sometimes uses different casing for drive letter.
-        // Make sure we normalize at least drive letter.
-        return combinePaths(fs.realpathSync.native(path.substring(0, driveLength)), path.substring(driveLength));
+    private _getNormalizedPath(uri: Uri) {
+        return uri.getFilePath();
```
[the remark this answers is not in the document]
rchiodo: Sure, will do

packages/pyright-internal/src/common/realFileSystem.ts
```diff
@@ -453,17 +463,13 @@ export class WorkspaceFileWatcherProvider implements FileWatcherProvider, FileWa
...
     private _tmpdir?: tmp.DirResult;
 
-    tmpdir() {
-        if (!this._tmpdir) {
-            this._tmpdir = tmp.dirSync({ prefix: 'pyright' });
-        }
-
-        return this._tmpdir.name;
+    tmpdir(): Uri {
+        return Uri.file(this._getTmpDir().name, /* isCaseSensitive */ true);
```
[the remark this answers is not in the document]
rchiodo: There's a chicken and egg issue. To compute whether or not a file system is case sensitive, you need a tmp object.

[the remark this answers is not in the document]
rchiodo: I suppose this could be used in a non case sensitive file system though. And then it might mess up appending things? But if you appended things, you'd probably be making import paths, so you'd want them to be case sensitive anyway.

packages/pyright-internal/src/common/uri/baseUri.ts
```diff
@@ -0,0 +1,310 @@
...
+        if (index > 0) {
+            return base.slice(0, index);
+        } else {
+            return base;
+        }
+    }
+
+    // Returns the extension of the URI, similar to the UNIX extname command.
+    @cacheProperty()
+    get extname(): string {
```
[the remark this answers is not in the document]
rchiodo: Will do.

packages/pyright-internal/src/common/uri/baseUri.ts
```diff
@@ -0,0 +1,310 @@
...
+    get packageUri(): Uri {
+        // This is assuming that the current path is a directory already.
+        return this.addPath('.py');
+    }
+
+    // Returns a URI where the path contains the path with .pyi appended.
+    @cacheProperty()
+    get packageStubUri(): Uri {
+        // This is assuming that the current path is a directory already.
+        return this.addPath('.pyi');
```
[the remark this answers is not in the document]
rchiodo: Will do

packages/pyright-internal/src/common/uri/baseUri.ts
```diff
@@ -0,0 +1,310 @@
...
+
+        const components = toComponents.slice(start);
+        const relative: string[] = [];
+        for (; start < fromComponents.length; start++) {
+            relative.push('..');
+        }
+        return [...relative, ...components];
+    }
+
+    getShortenedFileName(maxDirLength: number = 15): string {
```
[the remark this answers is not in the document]
rchiodo: I'll change the property to fileName

packages/pyright-internal/src/common/uri/emptyUri.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    static get instance() {
+        return EmptyUri._instance;
+    }
+
+    override get scheme(): string {
+        return '';
+    }
+    get isCaseSensitive(): boolean {
+        return true;
```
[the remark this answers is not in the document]
rchiodo: EmptyUri should never be combined with anything else. In fact it doesn't let itself be combined. So it doesn't really matter.

packages/pyright-internal/src/common/uri/emptyUri.ts
```diff
@@ -0,0 +1,100 @@
...
+        return this;
+    }
+    override addPath(extra: string): Uri {
+        return this;
+    }
+    override getDirectoryImpl(): Uri {
+        return this;
+    }
+    override isRoot(): boolean {
+        return false;
```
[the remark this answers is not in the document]
rchiodo: This doesn't really matter at the moment, but I agree it could be misused by mistake. I'll switch it to true.

packages/pyright-internal/src/common/uri/fileUri.ts
```diff
@@ -0,0 +1,196 @@
...
+
+export class FileUri extends BaseUri {
+    private _formattedString: string | undefined;
+    private constructor(
+        key: string,
+        private readonly _filePath: string,
+        private readonly _query: string,
+        private readonly _fragment: string,
+        private _originalString: string | undefined,
+        private _isCaseSensitive: boolean
```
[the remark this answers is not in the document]
rchiodo: Good point.

packages/pyright-internal/src/common/uri/fileUri.ts
```diff
@@ -0,0 +1,196 @@
...
+        return isDiskPathRoot(this._filePath);
+    }
+
+    @cacheMethodWithArgs()
+    override isChild(parent: Uri, ignoreCase?: boolean): boolean {
+        if (!FileUri.isFileUri(parent)) {
+            return false;
+        }
+
+        return this.startsWith(parent, ignoreCase) && parent._filePath.length < this._filePath.length;
```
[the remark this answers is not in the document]
rchiodo: StartsWith uses <=, not <, so it's required here.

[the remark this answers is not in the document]
rchiodo: But yes I'll switch the order.

erictraut:
> I did some more instrumentation to see what the first caller of the functions were

It looks like the vast majority of these are related to the `getDirectory` code path. Am I correct in assuming that a `Url` object is immutable once it's created? If so, then you should be able to cache the results of `getDirectory` and return it for all subsequent calls rather than constructing a new one. Looks like you're already doing this for `toString`, which caches `_formattedString`. My guess is that if you also cache the `_directory`, it will eliminate the vast majority of constructor calls.

packages/pyright-internal/src/common/uri/fileUri.ts
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    protected override getPathComponentsImpl(): string[] {
+        const components = getPathComponents(this._filePath);
+        // Remove the first one if it's empty. The new algorithm doesn't
+        // expect this to be there.
+        if (components.length > 0 && components[0] === '') {
+            components.shift();
+        }
+        return components.map((component) => component.replace(/\\/g, '/'));
```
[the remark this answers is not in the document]
rchiodo: I'll move it to a shared function.

rchiodo:
> > I did some more instrumentation to see what the first caller of the functions were
> 
> It looks like the vast majority of these are related to the `getDirectory` code path. Am I correct in assuming that a `Url` object is immutable once it's created? If so, then you should be able to cache the results of `getDirectory` and return it for all subsequent calls rather than constructing a new one. Looks like you're already doing this for `toString`, which caches `_formattedString`. My guess is that if you also cache the `_directory`, it will eliminate the vast majority of constructor calls.

Yeah the memoization was doing this for me before, but some of the values could be cached internally as well.

rchiodo:
Actually I think I'll add the memoization back. It's a much faster implementation of caching as it replaces the entire function with just the result. I can use this for all the properties.

packages/pyright-internal/src/common/uri/uriUtils.ts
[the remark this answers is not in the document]
rchiodo: I'll add another spot to use the new `hasExtension` function.

rchiodo:
After adding memoization back, the number of URIs came out to 16K instead of 200K. Memory usage so far is about the same as before. Although seems to be less than main in some aspects.

Here's an example run:

Main:

```
Heap stats: {
    "total_heap_size": 1087954944,
    "total_heap_size_executable": 21757952,
    "total_physical_size": 1086502560,
    "total_available_size": 1134131216,
    "used_heap_size": 1044040584,
    "heap_size_limit": 2197815296,
    "malloced_memory": 270408,
    "peak_malloced_memory": 22843392,
    "does_zap_garbage": 0,
    "number_of_native_contexts": 1,
    "number_of_detached_contexts": 0,
    "total_global_handles_size": 8192,
    "used_global_handles_size": 4288,
    "external_memory": 1278493
}
```

URI branch:

```
Heap stats: {
    "total_heap_size": 1126420480,
    "total_heap_size_executable": 21757952,
    "total_physical_size": 1124671088,
    "total_available_size": 1100920840,
    "used_heap_size": 1076603168,
    "heap_size_limit": 2197815296,
    "malloced_memory": 270408,
    "peak_malloced_memory": 24030776,
    "does_zap_garbage": 0,
    "number_of_native_contexts": 1,
    "number_of_detached_contexts": 0,
    "total_global_handles_size": 8192,
    "used_global_handles_size": 4288,
    "external_memory": 1268986
}
```

Within 5% of main. That may be the noise level though, so it's hard to tell. This isn't exactly the most precise method for tracking memory usage :)

packages/pyright-internal/src/common/uri/baseUri.ts
```diff
@@ -0,0 +1,310 @@
...
+        const reduced = [components[0]];
+        for (let i = 1; i < components.length; i++) {
+            const component = components[i];
+            if (!component || component === '.') {
+                continue;
+            }
+
+            if (component === '..') {
+                if (reduced.length > 1) {
+                    if (reduced[reduced.length - 1] !== '..') {
```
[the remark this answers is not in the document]
rchiodo: I copied this algorithm from path utils. It does seem like it doesn't eliminate '..' on the front, but I don't think that's possible to have because the front is always the root path. That's why it starts at 1.

I'll add a comment to that effect.

rchiodo:
I believe @debonte and @heejaechang are finished?

rchiodo:
I can't seem to reproduce the mypy primer difference locally. I get the same output from what's in main and from this branch. Might have to actually run this with mypy primer.

erictraut:
@rchiodo, yes, I'm able to repro the diff locally. It's a legit bug in the PR.

The problem shows up in the statement `import _md5`. The `_md5` refers to a native library (a `.so` file on my Mac). The actual file name is '_md5.cpython-311-darwin.so'. The bug was introduced in the method `_isNativeModuleFileName`. The old code was:

```ts
        // Strip off the final file extension and the part of the file name
        // that excludes all (multi-part) file extensions. This allows us to
        // handle file names like "foo.cpython-32m.so".
        const fileExtension = getFileExtension(fileName, /* multiDotExtension */ false).toLowerCase();
        const withoutExtension = stripFileExtension(fileName, /* multiDotExtension */ true);
```

With the old logic, `withoutExtension` produced the value "_md5". 

Here's the new logic in your PR:

```ts
        const fileExtension = fileUri.extension.toLowerCase();
        const withoutExtension = fileUri.fileNameWithoutExtension;
```

With the new logic, `withoutExtension` produces the value "_md5.cpython-311-darwin", which doesn't match the import name.

erictraut:
Looks like it's an easy fix. Just change the code in `_isNativeModuleFileName` to:

```ts
        const withoutExtension = fileUri.stripAllExtensions().fileName;
```

packages/pyright-internal/src/languageService/definitionProvider.ts
```diff
@@ -66,13 +67,13 @@ export function addDeclarationsToDefinitions(
             resolvedDecl.type === DeclarationType.Alias &&
             resolvedDecl.symbolName &&
             resolvedDecl.submoduleFallback &&
-            resolvedDecl.submoduleFallback.path
+            resolvedDecl.submoduleFallback.uri
```
erictraut, line 70: Missing ! .isEmpty()

packages/pyright-internal/src/readonlyAugmentedFileSystem.ts
```diff
@@ -76,129 +68,124 @@ export class ReadOnlyAugmentedFileSystem implements FileSystem {
...
-    isInZip(path: string): boolean {
-        return this.realFS.isInZip(path);
+    isInZip(uri: Uri): boolean {
+        return this.realFS.isInZip(uri);
     }
 
-    protected recordMovedEntry(mappedPath: string, originalPath: string, rootPath: string) {
-        this._entryMap.set(mappedPath, originalPath);
-        this._reverseEntryMap.set(originalPath, mappedPath);
+    protected recordMovedEntry(mappeduri: Uri, originaluri: Uri, rootPath: Uri) {
```
erictraut, line 158: Should be `mappedUri` and `originalUri` (capitalized `U`).

erictraut (changes requested):
Wow, that's an impressive PR!

I've found a few real bugs.

Some of my comments are questions. No need to write answers to me. These are questions for you, and if you have good answers, then you can resolve them.

Most of my comments are small nits.

I'm pretty anxious about this PR. I think this is going to result in a very long tail of regressions. We're going to be paying for this one for a long time. Yet, I think it's the right thing to do.

packages/pyright-internal/src/analyzer/service.ts
```diff
@@ -118,11 +112,12 @@ export class AnalyzerService {
 
     private _disposed = false;
     private _pendingLibraryChanges: RefreshOptions = { changesOnly: true };
+    private _isFileSystemCaseSensitve = true;
```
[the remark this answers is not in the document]
rchiodo: It's a cache of the main fs in the service, and whether or not it's case sensitive or not. Everywhere it's used, we're creating a URI that is assuming the URI comes from the fs in the service. I'll remove it and just call the check on the fileSystem where needed to avoid any confusion.

packages/pyright-internal/src/analyzer/service.ts
```diff
@@ -527,33 +526,43 @@ export class AnalyzerService {
...
                     projectRoot = configFilePath;
                     configFilePath = this._findConfigFile(configFilePath);
                     if (!configFilePath) {
-                        this._console.info(`Configuration file not found at ${projectRoot}.`);
+                        this._console.info(`Configuration file not found at ${projectRoot.toUserVisibleString()}.`);
                     }
                 }
             }
-        } else if (projectRoot) {
+        } else if (this.fs.existsSync(projectRoot)) {
```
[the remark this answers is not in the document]
rchiodo: Yeah this seems wrong. It should check if the executionRoot is empty, not if the project root exists.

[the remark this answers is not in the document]
rchiodo: The old code would create the project root from the realcasepath of the executionRoot. If this was the empty string it would just return the empty string and this if case would be false. To have the same logic I should check the executionRoot.

packages/pyright-internal/src/analyzer/sourceMapper.ts
```diff
@@ -645,14 +644,14 @@ export class SourceMapper {
...
-                    !decl.path ||
+                    decl.uri.isEmpty() ||
                     decl.node.nodeType !== ParseNodeType.ImportFrom ||
                     !decl.node.isWildcardImport
                 ) {
                     continue;
                 }
 
-                const uniqueId = `@${decl.path}/l/${symbolName}`;
+                const uniqueId = `@${decl.uri}/l/${symbolName}`;
```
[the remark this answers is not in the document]
rchiodo: Wow, good one. This was a potential bug.

packages/pyright-internal/src/backgroundAnalysis.ts
```diff
@@ -17,13 +17,14 @@ import { ConsoleInterface } from './common/console';
 import { FullAccessHost } from './common/fullAccessHost';
 import { Host } from './common/host';
 import { ServiceProvider } from './common/serviceProvider';
+import { getRootUri } from './common/uri/uriUtils';
 
 export class BackgroundAnalysis extends BackgroundAnalysisBase {
-    constructor(console: ConsoleInterface) {
+    constructor(console: ConsoleInterface, isCaseSensitive: boolean) {
```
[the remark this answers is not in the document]
rchiodo: It's not computable with just a filesystem though. It requires a `Tmp` implementation, which would have to be passed in here too. I guess I could do that instead?

packages/pyright-internal/src/backgroundAnalysisBase.ts
```diff
@@ -283,15 +289,16 @@ export abstract class BackgroundAnalysisRunnerBase extends BackgroundThreadBase
...
     protected constructor(serviceProvider: ServiceProvider) {
         super(workerData as InitializationData, serviceProvider);
 
         // Stash the base directory into a global variable.
         const data = workerData as InitializationData;
-        this.log(LogLevel.Info, `Background analysis(${threadId}) root directory: ${data.rootDirectory}`);
-
-        this._configOptions = new ConfigOptions(data.rootDirectory);
+        this.log(LogLevel.Info, `Background analysis(${threadId}) root directory: ${data.rootUri}`);
+        this.isCaseSensitive = serviceProvider.isFsCaseSensitive();
```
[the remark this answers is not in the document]
rchiodo: This is specific to the fs system in the service provider. Should I change the name to make that clearer, or do you think it better to not compute this in the service provider in the first place?

packages/pyright-internal/src/backgroundThreadBase.ts
```diff
@@ -62,6 +60,12 @@ export class BackgroundThreadBase {
         if (!this._serviceProvider.tryGet(ServiceKeys.tempFile)) {
             this._serviceProvider.add(ServiceKeys.tempFile, new RealTempFile());
         }
+
+        // Stash the base directory into a global variable.
```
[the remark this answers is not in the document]
rchiodo: It was previously, just this happened a couple of lines above. Line 51 in the old code. I can't do it now until the file system and tmp service are there, otherwise we can't figure out the case sensitivity.

packages/pyright-internal/src/languageServerBase.ts
```diff
@@ -379,16 +366,16 @@ export abstract class LanguageServerBase implements LanguageServerInterface, Dis
 
         this.fs = this.serverOptions.serviceProvider.fs();
 
-        this.uriParser = uriParserFactory(this.fs);
+        this._isCaseSensitive = this.serverOptions.serviceProvider.isFsCaseSensitive();
```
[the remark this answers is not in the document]
rchiodo: It's really just a cache of whether or not the file system is case sensitive. Maybe it would be better if this was a property on a file system?

packages/pyright-internal/src/languageService/analyzerServiceExecutor.ts
```diff
@@ -120,33 +118,25 @@ function getEffectiveCommandLineOptions(
...
         // The Python VS Code extension treats the value "python" specially. This means
         // the local python interpreter should be used rather than interpreting the
         // setting value as a path to the interpreter. We'll simply ignore it in this case.
-        if (!isPythonBinary(serverSettings.pythonPath)) {
-            commandLineOptions.pythonPath = combinePaths(
-                workspaceRootPath || languageServiceRootPath,
-                serverSettings.pythonPath
-            );
-        }
+        commandLineOptions.pythonPath = serverSettings.pythonPath.getFilePath();
```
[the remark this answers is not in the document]
rchiodo: It was a side effect of the pythonPath in the settings becoming a URI. It will never refer to just `python` or `python3` anymore. 

The code that sets the settings.pythonPath checks for this case already, so the old check of setting it when it wasn't `python` or `python3` isn't necessary (or really possible anymore). 

To be honest, I'm not 100% sure this is the correct thing to do though. Like if the pythonPath was set to a relative path, it would come in here as a rooted path, so it seems like it should work, but that means there's a dependency between that happening and this code. If it wasn't rooted, it seems like this should still use the file path though. I'll change just that.

packages/pyright-internal/src/languageServerBase.ts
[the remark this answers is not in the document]
rchiodo: I changed this to be stored in the file system itself.

packages/pyright-internal/src/analyzer/service.ts
```diff
@@ -175,6 +172,10 @@ export class AnalyzerService {
         return this._backgroundAnalysisProgram.importResolver.fileSystem;
     }
 
+    get tmp() {
```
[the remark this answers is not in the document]
rchiodo: Removed.

packages/pyright-internal/src/analyzer/program.ts
```diff
@@ -1417,17 +1410,17 @@ export class Program {
...
-            const oldFilePath = importInfo.sourceFile.getFilePath();
+            const oldFilePath = importInfo.sourceFile.getUri();
 
             // A previous import was removed.
-            if (!newImportPathMap.has(oldFilePath)) {
+            if (!newImportPathMap.has(oldFilePath.key)) {
                 importInfo.mutate((s) => {
                     s.importedBy = s.importedBy.filter(
-                        (fi) => fi.sourceFile.getFilePath() !== sourceFileInfo.sourceFile.getFilePath()
+                        (fi) => fi.sourceFile.getUri() !== sourceFileInfo.sourceFile.getUri()
```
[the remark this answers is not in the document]
rchiodo: Another bug found. Thanks.

packages/pyright-internal/src/analyzer/importResolver.ts
```diff
@@ -2656,12 +2620,12 @@ export class ImportResolver {
...
 
-    private _isNativeModuleFileName(moduleName: string, fileName: string): boolean {
+    private _isNativeModuleFileName(moduleName: string, fileUri: Uri): boolean {
         // Strip off the final file extension and the part of the file name
         // that excludes all (multi-part) file extensions. This allows us to
         // handle file names like "foo.cpython-32m.so".
-        const fileExtension = getFileExtension(fileName, /* multiDotExtension */ false).toLowerCase();
-        const withoutExtension = stripFileExtension(fileName, /* multiDotExtension */ true);
+        const fileExtension = fileUri.extension.toLowerCase();
+        const withoutExtension = fileUri.fileNameWithoutExtension;
```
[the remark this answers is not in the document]
rchiodo: Thanks a bunch for finding this. Was trying to figure out other ways on how to reproduce the problem and wasn't quite sure what I was going to do if I could never reproduce it.

packages/pyright-internal/src/analyzer/importResolver.ts
```diff
@@ -2628,13 +2585,20 @@ export class ImportResolver {
         return implicitImportMap;
     }
 
+    private _getPyTypedInfo(filePath: Uri): PyTypedInfo | undefined {
```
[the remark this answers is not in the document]
rchiodo: Added a comment.

erictraut (approved):

PR #6560, opened 2023-11-28
Include directories in the folder map for mapped partial stubs

rchiodo:
Addresses https://github.com/microsoft/pylance-release/issues/5139

Root cause was we weren't adding mapped entries for the directories for the stubs, just for the files. This means the directory had to exist in the real package in order for the mapped stubs to be found.

Specifically for the bug, the `keras` folder is not in the `site-packages\tensorflow` folder.

Thanks to @heejaechang for pointing out how this all works.

### DIFF
--- a/packages/pyright-internal/src/pyrightFileSystem.ts
+++ b/packages/pyright-internal/src/pyrightFileSystem.ts
@@ -15,13 +15,7 @@ import { getPyTypedInfo } from './analyzer/pyTypedUtils';
 import { ExecutionEnvironment } from './common/configOptions';
 import { FileSystem, MkDirOptions } from './common/fileSystem';
 import { stubsSuffix } from './common/pathConsts';
-import {
-    combinePaths,
-    ensureTrailingDirectorySeparator,
-    getPathComponents,
-    isDirectory,
-    tryStat,
-} from './common/pathUtils';
+import { combinePaths, ensureTrailingDirectorySeparator, isDirectory, tryStat } from './common/pathUtils';
 import { ReadOnlyAugmentedFileSystem } from './readonlyAugmentedFileSystem';
 
 export interface SupportPartialStubs {
@@ -157,23 +151,7 @@ export class PyrightFileSystem extends ReadOnlyAugmentedFileSystem implements IP
                         for (const partialStub of partialStubs) {
                             const originalPyiFile = combinePaths(partialStubPackagePath, partialStub);
                             const mappedPyiFile = combinePaths(packagePath, partialStub);
-
-                            this.recordMovedEntry(mappedPyiFile, originalPyiFile);
-
-                            // Record directory structure too.
-                            const stubParts = getPathComponents(partialStub);
-                            for (let i = 1; i < stubParts.length - 1; i++) {
-                                const originalDir = combinePaths(partialStubPackagePath, ...stubParts.slice(1, i + 1));
-                                const mappedDir = combinePaths(packagePath, ...stubParts.slice(1, i + 1));
-                                if (!this.isMovedEntry(originalDir)) {
-                                    this.recordMovedEntry(
-                                        mappedDir,
-                                        originalDir,
-                                        /* reversible */ true,
-                                        /* isFile */ false
-                                    );
-                                }
-                            }
+                            this.recordMovedEntry(mappedPyiFile, originalPyiFile, packagePath);
                         }
                     } catch {
                         // ignore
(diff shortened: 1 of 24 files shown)


packages/vscode-pyright/src/extension.ts
```diff
@@ -71,7 +71,9 @@ export async function activate(context: ExtensionContext) {
 
     const bundlePath = context.asAbsolutePath(path.join('dist', 'server.js'));
     const runOptions = { execArgv: [`--max-old-space-size=${defaultHeapSize}`] };
-    const debugOptions = { execArgv: ['--nolazy', '--inspect=6600', `--max-old-space-size=${defaultHeapSize}`] };
+    const debugOptions = {
+        execArgv: ['--nolazy', '--inspect=6600', `--inspect-brk`, `--max-old-space-size=${defaultHeapSize}`],
```
rchiodo, line 75: Whoops, I need to remove this. I added this for debugging.

erictraut (commented):
I don't have a deep enough understanding of the "mapped file" mechanism to effectively code review this change. If other members of the pylance team sign off on it, I'm fine with it.

bschnurr (approved):

packages/pyright-internal/src/pyrightFileSystem.ts
```diff
@@ -153,6 +159,21 @@ export class PyrightFileSystem extends ReadOnlyAugmentedFileSystem implements IP
                             const mappedPyiFile = combinePaths(packagePath, partialStub);
 
                             this.recordMovedEntry(mappedPyiFile, originalPyiFile);
+
+                            // Record directory structure too.
```
heejaechang, line 163: can this code move into the `recordMovedEntry`? I dont think we need entry in `_entryMap` but need to be in `folderInfo`, so you could change this part

```
const directory = ensureTrailingDirectorySeparator(getDirectoryPath(mappedPath));
        const folderInfo = getOrAdd(this._folderMap, directory, () => []);

        const name = getFileName(mappedPath);
        if (!folderInfo.some((entry) => entry.name === name)) {
            folderInfo.push({ name, isFile });
        }
```

to go through all `path components` of `directory` and make sure all those parts exist, otherwise create entry in `folderInfo` ?

and I think instead of using `isMovedEntry`, it could use `existsSync` since we don't care how it exists, but just care whether it exists or not.

rchiodo: Good idea. I'll try that.

packages/pyright-internal/src/readonlyAugmentedFileSystem.ts
```diff
@@ -161,19 +167,28 @@ export class ReadOnlyAugmentedFileSystem implements FileSystem {
...
         const folderInfo = getOrAdd(this._folderMap, directory, () => []);
 
         const name = getFileName(mappedPath);
         if (!folderInfo.some((entry) => entry.name === name)) {
-            folderInfo.push({ name, isFile });
+            folderInfo.push({ name, isFile: true });
+        }
+
+        // Add the directory entries for the sub paths as well.
+        const subPathEntries = getRelativePathComponentsFromDirectory(rootPath, mappedPath, false);
```
heejaechang, line 183: give directory instead of `mappedPath` ? since you already got one?

packages/pyright-internal/src/readonlyAugmentedFileSystem.ts
```diff
@@ -161,19 +167,28 @@ export class ReadOnlyAugmentedFileSystem implements FileSystem {
...
+        for (let i = 1; i < subPathEntries.length - 1; i++) {
```
heejaechang, line 184: then you won't need the `-1`

rchiodo: Okay, trying now.

packages/pyright-internal/src/readonlyAugmentedFileSystem.ts
heejaechang: `false` means `case insensitive`? shouldn't we give it based on file system's case sensitivity?

heejaechang (approved):
looks good to me except that `ignoreCase` flag.

packages/pyright-internal/src/readonlyAugmentedFileSystem.ts
heejaechang: so, talked with Rich. `false` is okay since `mappedPath` is always `rootPath` + `relative Path to the stub file`, so casing on `rootPath` part should always match.

PR #6625, opened 2023-12-02
Fix two typos in Spanish localization (package.nls.es.json).

JoseVteB:
There's a couple of **typos** in the **Spanish localization** that change the meaning of the addressed error.

In Spanish, when you say "_No se tiene acceso_ ..." or "_No se puede acceder_ ..." there's some kind permission implied. You are really saying "You're not allowed to ...".

However, the translated variables refer to "lack of access to ...". I suggest changing the translation in accordance with the rest of the variables of the "_unaccessed_" type.


JoseVteB:

erictraut:
Thanks for the contribution!

Issue #6607, opened 2023-11-30
Add new `standard` type checking mode


erictraut:
Currently pyright supports "basic" and "strict" type checking modes. Each of these indicates a default set of diagnostic rules and severity levels. I propose to introduce a new "standard" mode that sits between "basic" and "strict". It would include all diagnostic rules that are required for conformance with the new Python typing specification.

erictraut:
This is addressed in pyright 1.1.339, which I just published. It will also be included in a future release of pylance.

PR #6627, opened 2023-12-02
Added a new typeCheckingMode called "standard". It's a subset of "str…

erictraut:
…ict" but a superset of "basic". It is the new default mode, and it should cover all of the required checks for conformance with the Python typing standard. This addresses #66607.


wookayin:
#6607

Issue #6633, opened 2023-12-02
`reportUnknownMemberType` false negative in CLI when method has unknown return type


DetachHead:
**Describe the bug**
in the code below, the `reportUnknownMemberType` error is only reported in the language server in vscode, but not when running the pyright command

**Code or Screenshots**
```python
class Foo:
    def foo(self, asdf):
        return asdf


foo = Foo()

print(foo.foo(1)) # error: reportUnknownMemberType
```
```
> pyright .\asdf.py
0 errors, 0 warnings, 0 informations
```
![image](https://github.com/microsoft/pyright/assets/57028336/f89848f5-d2b0-4f89-b97d-0167f9ff0537)

**VS Code extension or command-line**
pylance: 2023.11.10
pyright pyi package: 1.1.338

as far as i can tell, this version of pylance is using the same version of pyright

erictraut:
The `reportUnknownMemberType` check isn't enabled by default. You need to enable it. If you want it to be enabled in the CLI, you need to do so in a configuration file. Refer to [this documentation](https://microsoft.github.io/pyright/#/configuration).

DetachHead:
@erictraut i forgot too include it but i just edited the issue too clarify that i do have that option enabled

erictraut:
Please post the contents of your `pyrightconfig.json` or `pyprotject.toml` file.

erictraut:
The code you've posted above does not generate any `reportUnknownMemberType` errors.

Code sample in [pyright playground](https://pyright-play.net/?code=MQAgDgngTglg5gCwC4C4RQKZgPZSQVQDsBrQ7Ad0IFkMBbAIwygBUIwMBeJKAVwwCh%2BAYwA2AQwDOEkADFs2FPxDKQAEwwAzEBvkAKCRhEaANCEmqNASkUrb6DEh5RCZiRcE7sIDrL2XBYLCESLqeAHSeugCMlv5AA)

erictraut:
This behavior changed in pyright 1.1.338. The version of pylance that you're using is based on an older version.

erictraut:
Refer to [this documentation](https://github.com/microsoft/pylance-release/blob/main/USING_WITH_PYRIGHT.md).

DetachHead:
[...]

PR #6644, opened 2023-12-04
add `importStrategy` option to vscode extension to allow using the version of pyright installed in the project

DetachHead:
after the discussion in https://github.com/microsoft/pyright/issues/6633#issuecomment-1837113310 about adding this option to pylance, i decided to try adding it to the pyright extension instead.

specifically, it looks for the [pyright pipi package](https://pypi.org/project/pyright/) in the project and falls back to the bundled one if it can't be found.

currently there are a couple issues however:

- [ ] the setting currently requires the vscode window to be reloaded to take effect (`Developer: Reload Window`)
- [ ] when i try to use my new `fromEnvironment` option, errors are not displayed, even though the language server from the pypi package seems to be running, and other language server commands such as organize imports seem to work:
  image
  image
  
  i can't see any errors in the log either:
  ```
  [Info  - 6:09:37 PM] Pyright language server 1.1.337 starting
  [Info  - 6:09:37 PM] Server root directory: C:\Users\user\.cache\pyright-python\1.1.337\node_modules\pyright\dist
  [Info  - 6:09:37 PM] Starting service instance "project"
  Received pythonPath from Python extension: c:\Users\user\Documents\project\.venv\Scripts\python.exe
  [Info  - 6:09:42 PM] Setting pythonPath for service "project": "c:\Users\user\Documents\project\.venv\Scripts\python.exe"
  [Info  - 6:09:42 PM] Loading pyproject.toml file at C:\Users\user\Documents\project\pyproject.toml
  [Info  - 6:09:42 PM] Assuming Python version 3.12
  [Info  - 6:09:42 PM] Found 671 source files
  ```

  any idea what could be causing this?

### DIFF
--- a/packages/vscode-pyright/package-lock.json
+++ b/packages/vscode-pyright/package-lock.json
@@ -9,6 +9,7 @@
             "version": "1.1.338",
             "license": "MIT",
             "dependencies": {
+                "@vscode/python-extension": "^1.0.5",
                 "vscode-jsonrpc": "8.1.0",
                 "vscode-languageclient": "8.1.0",
                 "vscode-languageserver": "8.1.0",
@@ -192,6 +193,16 @@
             "resolved": "https://registry.npmjs.org/@types/vscode/-/vscode-1.78.1.tgz",
             "integrity": "sha512-wEA+54axejHu7DhcUfnFBan1IqFD1gBDxAFz8LoX06NbNDMRJv/T6OGthOs52yZccasKfN588EyffHWABkR0fg==",
             "dev": true
+        },
+        "node_modules/@vscode/python-extension": {
+            "version": "1.0.5",
+            "resolved": "https://registry.npmjs.org/@vscode/python-extension/-/python-extension-1.0.5.tgz",
+            "integrity": "sha512-uYhXUrL/gn92mfqhjAwH2+yGOpjloBxj9ekoL4BhUsKcyJMpEg6WlNf3S3si+5x9zlbHHe7FYQNjZEbz1ymI9Q==",
+            "license": "MIT",
+            "engines": {
+                "node": ">=16.17.1",
+                "vscode": "^1.78.0"
+            }
         },
         "node_modules/@webassemblyjs/ast": {
             "version": "1.11.6",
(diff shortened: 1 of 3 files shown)


erictraut:
I'm not interested in adding feature this to pyright, so I'm closing this PR.

The approach you're taking here is a big hack — not something we would support. Also, the pypi wrapper is a community-maintained piece of code, and we're not going to take a dependency on it.

erictraut:
Am I correct in assuming that you run your CI script on a server (e.g. as part of a GitHub action)? If so, then the only way to get 100% consistent results between your local dev environment and your CI is to pin the version of the type checker you're using in both places.

Pyright includes basic language server features, and we have historically published a VS Code extension for pyright, but we recommend against using this if you're a VS Code user. Instead, we recommend you use Pylance. We will not add any functionality in the Pyright VS Code extension that is not also available in Pylance, and in general, most new language server features are developed only in Pylance. Any solution of the sort you're proposing should be discussed and resolved with the Pylance team.

DetachHead:
> If so, then the only way to get 100% consistent results between your local dev environment and your CI is to pin the version of the type checker you're using in both places.

that's the problem. as far as i'm aware there's no actual way to pin the version used locally by vscode. the closest thing you can do to that is disable extension auto updates and have users manually install specific extension versions (https://github.com/microsoft/pylance-release/issues/5201#issuecomment-1839083165) which isn't ideal

> Pyright includes basic language server features, and we have historically published a VS Code extension for pyright, but we recommend against using this if you're a VS Code user. Instead, we recommend you use Pylance. We will not add any functionality in the Pyright VS Code extension that is not also available in Pylance, and in general, most new language server features are developed only in Pylance. Any solution of the sort you're proposing should be discussed and resolved with the Pylance team.

currently there's a message in [the pyright extension description that says "For most VS Code users, we recommend using the Pylance extension rather than Pyright"](https://marketplace.visualstudio.com/items?itemName=ms-pyright.pyright), but that seems to imply that there are reasons not to use pylance, and i figured this could potentially be added to that list of reasons.

but if that's not the case, is it fair to say that the pyright vscode extension is deprecated? in that case, i think the extension description should be updated to clearly say so

PR #6683, opened 2023-12-08
Fix additional URI conversion problems for background thread

insilications:
Fixes additional `Uri` conversion problems for the background thread, specifically `DiagnosticRelatedInfo`, which was missed in #6678 after the implementation of the `Uri` class (#6519).


insilications:
It seems that this also needs a `fileUri: fileUri.toString()` conversion as well. I will test it later:

https://github.com/microsoft/pyright/blob/a2e462e6b07d4299410350e288201d0034f81035/packages/pyright-internal/src/backgroundAnalysisBase.ts#L88

PR #6709, opened 2023-12-11
Include import aliases in document/workspace symbols

debonte:
Address https://github.com/microsoft/pylance-release/issues/4977

It's unclear to me why aliases are currently excluded from document/workspace symbols requests. Simplicity? But this ask to include them seems reasonable to me.

### DIFF
--- a/packages/pyright-internal/src/languageService/symbolIndexer.ts
+++ b/packages/pyright-internal/src/languageService/symbolIndexer.ts
@@ -10,13 +10,14 @@ import { CancellationToken, CompletionItemKind, SymbolKind } from 'vscode-langua
 
 import { AnalyzerFileInfo } from '../analyzer/analyzerFileInfo';
 import * as AnalyzerNodeInfo from '../analyzer/analyzerNodeInfo';
-import { Declaration, DeclarationType } from '../analyzer/declaration';
+import { AliasDeclaration, Declaration, DeclarationType } from '../analyzer/declaration';
 import { getLastTypedDeclaredForSymbol, isVisibleExternally } from '../analyzer/symbolUtils';
 import { throwIfCancellationRequested } from '../common/cancellationUtils';
 import { getSymbolKind } from '../common/lspUtils';
 import { convertOffsetsToRange } from '../common/positionUtils';
 import { Range } from '../common/textRange';
 import { Uri } from '../common/uri/uri';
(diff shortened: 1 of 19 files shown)


packages/pyright-internal/src/languageService/symbolIndexer.ts
```diff
@@ -153,3 +154,14 @@ function collectSymbolIndexDataForName(
 
     indexSymbolData.push(data);
 }
+
+function shouldAliasBeIndexed(declaration: AliasDeclaration) {
+    if (
```
rchiodo, line 159: The logic here is a little hard to follow for me. 

So an alias should be indexed if:
- ParseNodeType.ImportAs or
- ParseNodeType.ImportFromAs and
- It is defined

Would the true version of this condition make more sense when reading it?

erictraut: Yeah, I had a hard time following the logic here as well. Perhaps a comment would help clarify? It looks like you're proposing to include only those import statements that include an `as` clause (and therefore define an "alias" for the imported symbol)?

erictraut:
Yeah, I don't know why it was written this way. I can tell I didn't write this code because it's using "yoda" conventions, which is not something I do. If I had written this code, it would have been `declaration.type === DeclarationType.Alias`. I'm guessing that Jake Bailey wrote this initially.

erictraut (approved):
The logic LGTM, but could benefit from a comment.

packages/pyright-internal/src/languageService/symbolIndexer.ts
debonte: Inverted the node type logic and added a comment.

rchiodo (approved):

heejaechang:
2 things. for `document symbol` it might be fine to include `alias` but for `workspace symbol`, it would cause hundreds of same symbols to show up? for example, `pd` in every files that uses `pandas` will now show up in `go to symbol`.

..

second, change here won't affect `pylance`, `pylance` has its own implementation since it uses `indexing`. this change will only affect `pyright`

heejaechang:
if we are going to make the same change in `pylance`, there are other considerations to make before including `alias` in user files in indices since we do `alias deduplication` in `pylance` to make sure we don't show multiple same import aliases from multiple files or between different packages.

Issue #6722, opened 2023-12-13
Pyright v1.1.340 crashes with an internal error


agronholm:
I'm running `pyright --verifytypes` against https://github.com/agronholm/anyio and it started failing with an internal error as soon as v1.1.340 was published. After downgrading to v1.1.339 it works fine.
<details>
<summary>Run pyright against AnyIO</summary>

Run pyright --verifytypes anyio

```
added 1 package, and audited 2 packages in 2s

found 0 vulnerabilities
Module name: "anyio"
Package directory: "file:///home/runner/work/anyio/anyio/src/anyio"
Module directory: "file:///home/runner/work/anyio/anyio/src/anyio"
Path of py.typed file: "file:///home/runner/work/anyio/anyio/src/anyio/py.typed"

Public modules: 1
   anyio
   error: An internal error occurred while verifying types: "TypeError: Cannot read properties of undefined (reading 'startsWith')
      at Object.file (/home/runner/.cache/pyright-python/1.1.340/node_modules/pyright/dist/pyright-internal/src/common/uri/uri.ts:167:33)
      at forEach (/home/runner/.cache/pyright-python/1.1.340/node_modules/pyright/dist/pyright-internal/src/analyzer/packageTypeVerifier.ts:592:38)
      at Map.forEach (<anonymous>)
      at b._getTypeKnownStatusForSymbolTable (/home/runner/.cache/pyright-python/1.1.340/node_modules/pyright/dist/pyright-internal/src/analyzer/packageTypeVerifier.ts:515:21)
      at b._verifyTypesOfModule (/home/runner/.cache/pyright-python/1.1.340/node_modules/pyright/dist/pyright-internal/src/analyzer/packageTypeVerifier.ts:398:22)
      at forEach (/home/runner/.cache/pyright-python/1.1.340/node_modules/pyright/dist/pyright-internal/src/analyzer/packageTypeVerifier.ts:185:30)
      at Array.forEach (<anonymous>)
      at b.verify (/home/runner/.cache/pyright-python/1.1.340/node_modules/pyright/dist/pyright-internal/src/analyzer/packageTypeVerifier.ts:184:35)
      at verifyPackageTypes (/home/runner/.cache/pyright-python/1.1.340/node_modules/pyright/dist/pyright-internal/src/pyright.ts:493:33)
      at processArgs (/home/runner/.cache/pyright-python/1.1.340/node_modules/pyright/dist/pyright-internal/src/pyright.ts:370:16)
[...]

PR #6726, opened 2023-12-13
Fix usage of invalid path in packageTypeVerifier

rchiodo:
`module` just happens to be an object in Javascript so that this used to work, but `module` is not defined at the location this is being used. I believe the intent was to use the declPath.

Addresses https://github.com/microsoft/pyright/issues/6722


debonte (approved):
Yikes.

erictraut:
Yikes indeed!

erictraut (approved):

Issue #6725, opened 2023-12-13
Pyright 1.1.340 doesn't include the scheme in the URI for the rename LSP action


tmke8:
It works fine with 1.1.339.

When using the rename functionality in neovim ([vim.lsp.buf.rename](https://neovim.io/doc/user/lsp.html#vim.lsp.buf.rename()), so, `textDocument/rename`) with the pyright LS, I get this error:

```
URI must contain a scheme: /home/tmke8/to/my/project/main.py
```

I'm guessing it's missing `file://`? If it helps I can try to investigate the problem further.

**VS Code extension or command-line**
As language server with neovim.

rchiodo:
Thanks for the issue. Can you outline the repro steps more? I'm not familiar with neovim. I'm really not sure how to get pyright into neovim as it doesn't seem to be listed in the neovim plugins:
https://dotfyle.com/neovim/plugins/trending?page=1&q=pyright

erictraut:
@tmke8, are you using the [coc-pyright](https://github.com/fannheyward/coc-pyright) extension for vim? Could you provide the full JSON for the LSP command that is being sent to the pyright language server?

My hypothesis is that the command includes a malformed URI, and older versions of pyright were permissively handling it. If my hypothesis is correct, then you'll need to work with the maintainers of coc-pyright to get this fixed.

tmke8:
I'm not using coc-pyright, I'm using the standard configuration from https://github.com/neovim/nvim-lspconfig :

```lua
require('lspconfig').pyright.setup {}
```
and then
```lua
vim.keymap.set('n', '<space>rn', vim.lsp.buf.rename, opts)
```
somewhere else in the neovim config file.

I'll see whether I can log the JSON somehow.

tmke8:
<details><summary>These should be the logs:</summary>

```
[DEBUG][2023-12-13 18:53:54] .../vim/lsp/rpc.lua:284	"rpc.send"	{
  id = 8,
  jsonrpc = "2.0",
  method = "textDocument/rename",
  params = {
    newName = "ACSSurveyYear",
    position = {
      character = 6,
      line = 22
    },
    textDocument = {
      uri = "file:///home/tmk/dev/python/conduit/conduit/fair/data/datasets/acs.py"
    }
  }
}
[DEBUG][2023-12-13 18:53:54] .../vim/lsp/rpc.lua:387	"rpc.receive"	{
  id = 8,
  jsonrpc = "2.0",
  result = {
[...]

PR #6728, opened 2023-12-13
Fix URI return value on renames

rchiodo:
Addresses https://github.com/microsoft/pyright/issues/6725

Root cause was using the fileuri.key value as the return value for the uri of a document. However the tests were running the same logic on the expected changes so it wasn't verifying the URIs were actually URIs, just that the changes were the same.

### DIFF
--- a/packages/pyright-internal/src/tests/harness/fourslash/workspaceEditTestUtils.ts
+++ b/packages/pyright-internal/src/tests/harness/fourslash/workspaceEditTestUtils.ts
@@ -91,6 +91,11 @@ export function verifyDocumentEdits(
                             return false;
                         }
 
+                        if (!actualEdit.textDocument.uri.includes(':')) {
+                            // Not returning a URI, so fail.
+                            return false;
+                        }
+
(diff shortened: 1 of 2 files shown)


debonte (approved):

packages/pyright-internal/src/tests/harness/fourslash/workspaceEditTestUtils.ts
```diff
@@ -91,6 +91,11 @@ export function verifyDocumentEdits(
                             return false;
                         }
 
+                        if (!actualEdit.textDocument.uri.includes(':')) {
```
erictraut, line 94: Under what circumstance would this ever occur? Or are you just being super paranoid here?

erictraut (approved):

packages/pyright-internal/src/tests/harness/fourslash/workspaceEditTestUtils.ts
rchiodo: Just being super paranoid. It's why the test didn't fail before. This ensures we don't make this same mistake again.

Issue #6737, opened 2023-12-14
Error: Debug Failure. EmptyUri.getFilePath() should not be called


westernwontons:
**Describe the bug**
Opening any Python file after updating to version 1.1.340 results in `Error: Debug Failure. EmptyUri.getFilePath() should not be called` and causes `pyright` to crash immediately.

**Code or Screenshots**
Opening an empty Python file is enough to cause the crash.

## LSP Logs

```
[ERROR][2023-12-14 18:28:08] ...lsp/handlers.lua:556	"Notification workspace/didChangeConfiguration defines parameters by name but received parameters by position"
[ERROR][2023-12-14 18:28:08] ...lsp/handlers.lua:556	"Notification workspace/didChangeConfiguration defines 1 params but received 0 arguments"
[ERROR][2023-12-14 18:28:08] .../vim/lsp/rpc.lua:675	"rpc"	"/Users/notimportant/.local/share/nvim/mason/bin/pyright-langserver"	"stderr"	"Error: Debug Failure. EmptyUri.getFilePath() should not be called.\n    at l.getFilePath (/Users/notimportant/.local/share/nvim/mason/packages/pyright/node_modules/pyright/dist/pyright-internal/src/common/uri/emptyUri.ts:58:15)\n    at Function.runWithOptions (/Users/notimportant/.local/share/nvim/mason/packages/pyright/node_modules/pyright/dist/pyright-internal/src/languageService/analyzerServiceExecutor.ts:35:36)\n    at S.updateOptionsAndRestartService (/Users/notimportant/.local/share/nvim/mason/packages/pyright/node_modules/pyright/dist/pyright-internal/src/languageServerBase.ts:525:33)\n    at S.updateSettingsForWorkspace (/Users/notimportant/.local/share/nvim/mason/packages/pyright/node_modules/pyright/dist/pyright-internal/src/languageServerBase.ts:509:14)\n    at async Promise.all (index 0)\n"
```

## Node version
v20.10.0

## Update

It seems that Pyright crashes unconditionally if `pyproject.toml` is not present in the working directory.

rchiodo:
@westernwontons thanks for the issue. Can you reproduce with just the command line? It doesn't seem to repro for me.

rchiodo:
Never mind, I think the callstack is enough information to figure out what the problem is. Thanks.

erictraut:
This is addressed in pyright 1.1.341, which I just published.

PR #6738, opened 2023-12-14
Fix emptyUri being used as a file path

rchiodo:
Addresses https://github.com/microsoft/pyright/issues/6737.

The old logic would have just passed an empty path here, so make `EmptyUri.getFilePath` return an empty path.


debonte (approved):

rchiodo:
Pylance LSP tests are catching this as I'm working on porting the URI changes there. Another case where an LSP based test would have likely caught this issue.

Issue #6742, opened 2023-12-14
CLI with -w raises TypeError on file change


scanny:
**Describe the bug**
After upgrading from `v1.1.325` to `v1.1.340`, running `pyright -w {target-path}` raises the following on the first file write (so second "watched" run, initial run works fine):
```
Watching for file changes...
TypeError: n.getFilePath is not a function
    at oA.<anonymous> (/Users/scanny/.cache/pyright-python/1.1.340/node_modules/pyright/dist/pyright-internal/src/analyzer/service.ts:1255:40)
    at oA.emit (node:events:514:28)
    at oA.emitWithAll (/Users/scanny/.cache/pyright-python/1.1.340/node_modules/pyright/dist/pyright-internal/node_modules/chokidar/index.js:541:32)
    at awfEmit (/Users/scanny/.cache/pyright-python/1.1.340/node_modules/pyright/dist/pyright-internal/node_modules/chokidar/index.js:607:14)
    at /Users/scanny/.cache/pyright-python/1.1.340/node_modules/pyright/dist/pyright-internal/node_modules/chokidar/index.js:728:9
    at FSReqCallback.oncomplete (node:fs:201:5)
```

Downgrading to 1.1.325 eliminates this behavior.

**VS Code extension or command-line**
This is using the CLI from `zsh 5.8` on a Mac.

Let me know if there's anything else you need :)

rchiodo:
Thanks for the issue. This was caused by the change to add URI support. Fix should be in the next release.

erictraut:
This is addressed in pyright 1.1.341, which I just published.

scanny:
Confirmed fixed on my Intel Mac running Big Sur and pyright v1.1.341.

Thanks @erictraut! :)

Issue #6740, opened 2023-12-14
1.1.340: schema for --outputjson is not according to spec


Josverl:
**Finding**
When running **pyright 1.1.340** the schema for json outputs appears to have changed 
Specifically the `generalDiagnostics`  `file: string` attribute appears to be missing and replaced by a `uri` attribute with a different schema. 

the problem does not appear for 1.1.339
`npx pyright@1.1.339 --project /workspaces/micropython-stubs/tests/quality_tests/feat_stdlib --outputjson` 
Returns the expected schema.

* Expected Schema : https://microsoft.github.io/pyright/#/command-line?id=json-output

**Code or Screenshots**

> `npx pyright --project /workspaces/micropython-stubs/tests/quality_tests/feat_stdlib --outputjson`

``` json
{
    "version": "1.1.340",
    "time": "1702587966554",
    "generalDiagnostics": [
        {
            "uri": {
                "_key": "/workspaces/micropython-stubs/tests/quality_tests/feat_stdlib/check_os/check_uname.py",
                "_filePath": "/workspaces/micropython-stubs/tests/quality_tests/feat_stdlib/check_os/check_uname.py",
                "_query": "",
                "_fragment": "",
                "_isCaseSensitive": true,
                "_formattedString": "file:///workspaces/micropython-stubs/tests/quality_tests/feat_stdlib/check_os/check_uname.py"
            },
            "severity": "information",
            "message": "Type of \"os_uname\" is \"uname_result\"",
            "range": {
                "start": {
                    "line": 23,
                    "character": 12
                },
                "end": {
                    "line": 23,
                    "character": 20
                }
            }
        }
    ],
    "summary": {
        "filesAnalyzed": 17,
        "errorCount": 0,
        "warningCount": 0,
        "informationCount": 1,
        "timeInSec": 1.347
    }
}
```

**VS Code extension or command-line**
- Linux codespaces-0599fe 6.2.0-1018-azure # 18~22.04.1-Ubuntu SMP Tue Nov 21 19:25:02 UTC 2023 x86_64 GNU/Linux
- or Windows 11 
- pyright 1.1.340 via npx ( same issue via PyPI) 
- node: v20.10.0

rchiodo:
Thanks for the issue. This was caused by the change to add URI support. Fix should be in the next release.

erictraut:
This is addressed in pyright 1.1.341, which I just published.

rnovacek:
[...]

PR #6743, opened 2023-12-14
Fix pyright output schema and file watching

rchiodo:
Addresses https://github.com/microsoft/pyright/issues/6742
Addresses https://github.com/microsoft/pyright/issues/6740


packages/pyright-internal/src/common/realFileSystem.ts
```diff
@@ -453,7 +453,7 @@ export class WorkspaceFileWatcherProvider implements FileWatcherProvider, FileWa
         // for a file change. It is event handler's job to filter those out.
         this._fileWatchers.forEach((watcher) => {
             if (watcher.workspacePaths.some((dirPath) => fileUri.pathStartsWith(dirPath))) {
-                watcher.eventHandler(eventType, fileUri);
+                watcher.eventHandler(eventType, fileUri.getFilePath());
```
debonte, line 456: Why do we want to convert uris to strings here?

rchiodo: Cause the eventHandler type takes a string now (I was wrong in thinking I could change it). It's external but this code is mimicking it.

debonte (approved):

Issue #6845, opened 2023-12-29
Overlapping overload not reported when parameter type includes tuples with overlapping unions


insilications:
This code should produce an "overlapping override" error, but it doesn't. Mypy currently flags this condition.

```python
from typing import overload

class A:
    @overload
    def select(self, p1: tuple[str | tuple[int | float, str], str], /) -> str: ...
    @overload
    # def select(self, p1: tuple[str | tuple[int | float, str] | int | float, str | int | float], /) -> int: ...
    def select(self, p1: tuple[tuple[int | float, str] | int | float, str | int | float], /) -> int: ...
```

erictraut:
This is a duplicate of #6825, which I opened recently.

Issue #6825, opened 2023-12-28
Overlapping overload not reported when parameter type includes overlapping unions


erictraut:
This code should produce an "overlapping override" error, but it doesn't. Mypy currently flags this condition.

```python
class A:
    @overload
    def select(self, p1: int | set[int], /) -> str:
        ...

    @overload
    def select(self, p1: int | list[int], /) -> int:
        ...
```

erictraut:
This is addressed in pyright 1.1.344, which I just published. It will also be included in a future release of pylance.

PR #6846, opened 2023-12-29
Fixed a bug in assignTupleTypeArguments that prevented detection of o…

insilications:
…verlapping unions

This is a tentative fix of the specific case of #6845 regarding detection of overlapping unions. Further improvements on detection of overlapping unions are tracked by #6825


erictraut:
I appreciate your attempt to help with PRs, but it's much faster for me to fix issues myself rather than try to understand and validate fixes submitted by other people.

This particular change isn't the correct fix for the bug you reported, so I'm going to close it.

PR #6848, opened 2023-12-29
Fixed overlapping overload not reported when parameter type includes …

insilications:
…overlapping unions. This addresses #6825


insilications:
This is currently breaking overlapping overload detection of overloads that return the same type. Will look into it later.

erictraut:
I appreciate your attempt to help with PRs, but it's much faster for me to fix issues myself rather than try to understand and validate fixes submitted by other people.

This particular change isn't the correct fix for the bug you reported, so I'm going to close it.

PR #6855, opened 2023-12-29
Improved overload overlap logic to detect partial overlaps when param…

erictraut:
…eter types include unions that intersect. This addresses #6825.


insilications:
Thanks for the fix, Eric! At first, I hesitated to create a new AssignTypeFlags, but during my last attempt to fix my PR, I ended up propagating a new AssignTypeFlags to also enable the partial overload overlap check.

PR #6884, opened 2024-01-02
Make sure web uris aren't read from disk

rchiodo:
Follow up to https://github.com/microsoft/pyright/pull/6790

Pyright cannot handle non file URIs, hence the original reason for the assert. However the old pre URI change Pyright would just silently fail on these files. This makes it more explicit.

### DIFF
--- a/packages/pyright-internal/src/common/realFileSystem.ts
+++ b/packages/pyright-internal/src/common/realFileSystem.ts
@@ -21,6 +21,7 @@ import {
     nullFileWatcherProvider,
 } from './fileWatcher';
 import { getRootLength } from './pathUtils';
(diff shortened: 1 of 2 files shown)


erictraut (approved):
LGTM

debonte:
Will this prevent us from calling `WebUri.getFilePath`? Just wondering if the assert should be added back.

rchiodo:
> Will this prevent us from calling `WebUri.getFilePath`? Just wondering if the assert should be added back.

I'm not sure. I could also change isInZip to not call getFilePath for web uris, but I feel like if we want to maintain the old behavior, we should not assert. The old code would just silently fail for web uri stuff.

debonte (approved):

Issue #6887, opened 2024-01-03
PyRight language server crashes?


apollo13:
**Describe the bug**

When changing the "Python > Analysis: Log Level" setting from "Trace" to "Error" I see:
```
[Info  - 11:20:52 AM] Setting pythonPath for service "cloud-init": "/bin/python"
[Info  - 11:20:52 AM] Loading pyproject.toml file at /home/florian/sources/cloud-init/pyproject.toml
[Info  - 11:20:52 AM] Pyproject file "/home/florian/sources/cloud-init/pyproject.toml" has no "[tool.pyright]" section.
[Info  - 11:20:52 AM] Assuming Python version 3.10
Assuming Python platform Linux
DataCloneError: ()=>n could not be cloned.
    at new DOMException (node:internal/per_context/domexception:53:5)
    at Worker.postMessage (node:internal/worker:343:5)
    at c.enqueueRequest (/home/florian/.var/app/com.visualstudio.code/data/vscode/extensions/ms-pyright.pyright-1.1.344/dist/pyright-internal/src/backgroundAnalysisBase.ts:228:26)
    at c.setConfigOptions (/home/florian/.var/app/com.visualstudio.code/data/vscode/extensions/ms-pyright.pyright-1.1.344/dist/pyright-internal/src/backgroundAnalysisBase.ts:58:14)
    at t.BackgroundAnalysisProgram.setConfigOptions (/home/florian/.var/app/com.visualstudio.code/data/vscode/extensions/ms-pyright.pyright-1.1.344/dist/pyright-internal/src/analyzer/backgroundAnalysisProgram.ts:82:35)
    at w.setOptions (/home/florian/.var/app/com.visualstudio.code/data/vscode/extensions/ms-pyright.pyright-1.1.344/dist/pyright-internal/src/analyzer/service.ts:282:41)
    at Function.runWithOptions (/home/florian/.var/app/com.visualstudio.code/data/vscode/extensions/ms-pyright.pyright-1.1.344/dist/pyright-internal/src/languageService/analyzerServiceExecutor.ts:44:27)
    at b.updateOptionsAndRestartService (/home/florian/.var/app/com.visualstudio.code/data/vscode/extensions/ms-pyright.pyright-1.1.344/dist/pyright-internal/src/languageServerBase.ts:525:33)
[...]

PR #6889, opened 2024-01-03
Potential fix for config options crash

rchiodo:
Addresses https://github.com/microsoft/pyright/issues/6887.

### DIFF
--- a/packages/pyright-internal/src/backgroundThreadBase.ts
+++ b/packages/pyright-internal/src/backgroundThreadBase.ts
@@ -93,6 +93,13 @@ export class BackgroundThreadBase {
     }
 }
 
+export function createJsonObjectFrom(configOptions: ConfigOptions): any {
+    return {
+        ...JSON.parse(JSON.stringify(configOptions)),
+        defineConstant: configOptions.defineConstant, // This is the only field that can't be stringified.
+    };
+}
+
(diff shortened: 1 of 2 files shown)


debonte (approved):

erictraut (approved):

packages/pyright-internal/src/backgroundThreadBase.ts
```diff
@@ -93,6 +93,13 @@ export class BackgroundThreadBase {
     }
 }
 
+export function createJsonObjectFrom(configOptions: ConfigOptions): any {
+    return {
+        ...JSON.parse(JSON.stringify(configOptions)),
+        defineConstant: configOptions.defineConstant, // This is the only field that can't be stringified.
```
heejaechang, line 99: not sure why Map is not automatically serialized.

rchiodo: Map is serialized when sending the data, but not when calling JSON.stringify. See this here:
https://stackoverflow.com/questions/29085197/how-do-you-json-stringify-an-es6-map

heejaechang: so, issue is our `Uri` type is not serializable?

Issue #6869, opened 2024-01-01
Protocol matching doesn't reject explicit ClassVar in class when protocol is not marked as ClassVar


erictraut:
```python
from typing import ClassVar, Protocol, Sequence


class Proto(Protocol):
    val1: Sequence[int]

class Concrete:
    val1: ClassVar[Sequence[int]] = [0]

x: Proto = Concrete()  # Should generate an error
```

erictraut:
This is addressed in pyright 1.1.345, which I just published. It will also be included in a future release of pylance.

PR #6920, opened 2024-01-05
Fixed bug in protocol matching logic to reject a protocol match if a …

erictraut:
…variable is a `ClassVar` in the protocol but not in the concrete class (or vice versa). This addresses #6869.


erictraut:
@rsokl, this bug fix in pyright (which is required to pass the newly-created type checker conformance tests) triggers some new type violations in the hydra-zen source base. See the changes above. Looks like an easy fix. Just wanted to give you a heads up.

Issue #6917, opened 2024-01-05
`TypeVar` constrained to `Literal[True]` and `Literal[False]` is resolved to `bool`


Viicos:
Following https://github.com/microsoft/pyright/discussions/6914:

Code sample in [pyright playground](https://pyright-play.net/?pythonVersion=3.13&code=GYJw9gtgBALgngBwJYDsDmUkQWEMoAqiApgGoCGIANFADJIzEjkA2NA4sSk0gMYBQ-AlAC8hEhRAAKAEQEZNeo2YsA2gRABXYgF1FDJq1UAxVgGddASkH9eLcmbNQAglM7cQfdTssAuflCBUAAmxMBQAPoRqAxRUhYswDQomiwsvoSiUKYsFpZQALQAfFAAcmDc-kHVUAgOZjbkWa4paSI5eYIgxABuxKwR8AjEUuT5UADEUADKABZgqaFQAEbELqpKhmodujr8QA)

```python
from typing import TypeVar, Literal, Generic

T = TypeVar("T", Literal[True], Literal[False])


class A(Generic[T]):
    def __init__(self, null: T = False) -> None:
        pass


a = A(null=False)

reveal_type(a)  # Should be A[Literal[False]]

```

erictraut:
Thanks for the bug report. This will be addressed in the next release.

Viicos:
Thanks for the fix! I see in the PR that you are special casing literals. Should this be added to the spec in some way? (for the record mypy is not inferring the case without an explicit `null` value provided correctly: [playgroud](https://mypy-play.net/?mypy=latest&python=3.12&gist=947b9311a5826e9a6f557673bf86adfc))

erictraut:
> Should this be added to the spec in some way?

Maybe, although in this case I don't think there's any ambiguity because the constraints imposed on the TypeVar (in this case, value constraints) require that the solved type is a literal. The issue with mypy in this case is not because it doesn't solve `T` as a literal but because it doesn't support specifying a default argument for a generic parameter. I'm somewhat surprised that the maintainers of mypy haven't addressed this already since it's one of [the most upvoted mypy feature requests](https://github.com/python/mypy/issues/3737).

erictraut:
This is addressed in pyright 1.1.345, which I just published. It will also be included in a future release of pylance.

PR #6929, opened 2024-01-06
Fixed a bug that led to incorrect type evaluation when a constrained …

erictraut:
…TypeVar includes literal types as constraints. This addresses #6917.

### DIFF
--- a/packages/pyright-internal/src/tests/samples/constrainedTypeVar19.py
+++ b/packages/pyright-internal/src/tests/samples/constrainedTypeVar19.py
@@ -1,0 +1,18 @@
+# This tests the handling of a constrained TypeVar with literal types
+# in the constraints.
+
+from typing import TypeVar, Literal, Generic
+
+T = TypeVar("T", Literal[True], Literal[False])
+
+
+class A(Generic[T]):
+    def __init__(self, null: T = False) -> None:
+        pass
+
+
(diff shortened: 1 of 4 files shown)


packages/pyright-internal/src/tests/samples/constrainedTypeVar19.py
```diff
@@ -0,0 +1,18 @@
...
+
+class A(Generic[T]):
+    def __init__(self, null: T = False) -> None:
+        pass
+
+
+A(null=bool())  # Type error
+
+reveal_type(A(null=False), expected_text="A[Literal[False]]")
+reveal_type(A(), expected_text="A[Literal[False]]")
```
Viicos, line 17: May I ask: is `pyright` making use of the default value of the `__init__` signature in this case? If so, does it means this removes the need for a `TypeVar` default?

erictraut: Defaults for `TypeVar` are not yet officially supported. They're proposed in PEP 696, and pyright has provisional support for them.

I don't see a need for a default `TypeVar` value in this case. Pyright supports the use of a default argument value associated with a parameter whose type is `T`. Mypy doesn't support this, however, so if you want your code to work across type checkers, you might want to consider other options like overloads.

PR #6954, opened 2024-01-10
Port some of the URI fixes made in pylance back to pyright

rchiodo:
Instead of waiting for me to finish all the URI changes in pylance, I wanted to push this part of the changes to Pyright. 

I was debugging the pyright extension without this change and was noticing there were still issues serializing data to the background thread. This change fixes those problems.

### DIFF
--- a/packages/pyright-internal/package.json
+++ b/packages/pyright-internal/package.json
@@ -36,6 +36,7 @@
         "@types/command-line-args": "^5.2.0",
         "@types/fs-extra": "^11.0.1",
         "@types/jest": "^29.5.3",
+        "@types/lodash": "^4.14.202",
         "@types/node": "^17.0.45",
         "@types/tmp": "^0.2.3",
         "jest": "^29.6.1",
--- a/packages/pyright-internal/src/backgroundThreadBase.ts
+++ b/packages/pyright-internal/src/backgroundThreadBase.ts
@@ -96,20 +96,20 @@ export class BackgroundThreadBase {
 // Exposed here so it can be reused by a caller that wants to add more cases.
 export function serializeReplacer(key: string, value: any) {
     if (Uri.isUri(value)) {
-        return { uri_str: value.toString(), case_sensitive: value.isCaseSensitive };
+        return { __serialized_uri_str: value.toString(), __serialized_case_sensitive: value.isCaseSensitive };
     }
     if (value instanceof Map) {
-        return { map_val: [...value] };
+        return { __serialized_map_val: [...value] };
     }
     if (value instanceof Set) {
-        return { set_val: [...value] };
+        return { __serialized_set_val: [...value] };
     }
     if (value instanceof RegExp) {
-        return { regexp_val: { source: value.source, flags: value.flags } };
+        return { __serialized_regexp_val: { source: value.source, flags: value.flags } };
     }
     if (value instanceof ConfigOptions) {
         const entries = Object.entries(value);
-        return { config_options: entries.reduce((obj, e, i) => ({ ...obj, [e[0]]: e[1] }), {}) };
+        return { __serialized_config_options: entries.reduce((obj, e, i) => ({ ...obj, [e[0]]: e[1] }), {}) };
     }
 
     return value;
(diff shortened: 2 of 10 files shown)


packages/pyright-internal/src/common/uri/uriUtils.ts
```diff
@@ -404,3 +404,17 @@ export function getRootUri(isCaseSensitive: boolean): Uri | undefined {
     }
     return undefined;
 }
+
+export function encodeUri(fs: ReadOnlyFileSystem, uri: Uri): string {
+    // Convert to a URI string that the LSP client understands (mapped files are only local to the server).
+    return fs.getOriginalUri(uri).toString();
+}
+
+export function isUri(path: string): boolean {
```
debonte, line 413: Where does this get used? It reminds me of our old problematic logic to try to differentiate between URIs and paths.

rchiodo: Nowhere yet (it's in the pylance changes). It will be used in the config options when trying to determine if the execution root on the command line is a URI or not.

rchiodo: Specifically, it was necessary when opening a virtual workspace.

packages/pyright-internal/src/backgroundThreadBase.ts
```diff
@@ -93,69 +92,73 @@ export class BackgroundThreadBase {
...
 }
 
-export function createJsonObjectFrom(configOptions: ConfigOptions): any {
-    return {
-        ...JSON.parse(JSON.stringify(configOptions)),
-        defineConstant: configOptions.defineConstant, // This is the only field that can't be stringified.
-    };
+// Function used to serialize specific types that can't automatically be serialized.
+// Exposed here so it can be reused by a caller that wants to add more cases.
+export function serializeReplacer(key: string, value: any) {
```
erictraut, line 97: It looks like you're relying on these special names being unique and never appearing elsewhere in our code. That's probably an OK assumption because you're using a non-standard (Python-style) naming convention here, but it might be safer to use an even more obscure naming convention like `__serialization_mapValue` or something like that.

packages/pyright-internal/package.json
```diff
@@ -36,6 +36,7 @@
         "@types/command-line-args": "^5.2.0",
         "@types/fs-extra": "^11.0.1",
         "@types/jest": "^29.5.3",
+        "@types/lodash": "^4.14.202",
```
erictraut, line 39: Hmm, I generally try to avoid lodash. It has grown very large over the years, and because it's so intertangled, tree shaking isn't able to remove the unused parts. Please confirm that you're not bloating the bundle size by a large amount by including this. If you are, it would be better to just write your own version of the function(s) you need.

packages/pyright-internal/src/common/uri/uriUtils.ts
debonte: Maybe it shouldn't be added in this PR then? It would be easier to review if the usage scenarios were more clear.

packages/pyright-internal/package.json
rchiodo: This is just the types and it's not included in the actual bundle. It was used in the config.test.ts

packages/pyright-internal/src/backgroundThreadBase.ts
rchiodo: Good idea. I'll change it to that.

packages/pyright-internal/src/common/uri/uriUtils.ts
rchiodo: I'll remove it.

packages/pyright-internal/package.json
erictraut: But I noticed that you're calling `deepCopy` from `lodash` in another part of this PR.

rchiodo: In config.test.ts.

erictraut: Ah, got it. So this isn't in the production bundle.

debonte (approved):

PR #6967, opened 2024-01-12
[PylanceBot] Pull Pylance with Pyright

PylanceBot:
This is pulling back the rest of the changes I made for URI support in Pylance.

There's also some refactoring changes that Heejae made.


packages/pyright-internal/src/common/realFileSystem.ts
```diff
@@ -405,19 +405,19 @@ class RealFileSystem implements FileSystem {
...
-            const path = uri.getFilePath();
-            const stat = fs.lstatSync(path);
-            if (stat.isSymbolicLink()) {
-                return uri;
-            }
-
             // realpathSync.native will return casing as in OS rather than
             // trying to preserve casing given.
-            const realCase = fs.realpathSync.native(path);
+            const realCase = fs.realpathSync.native(uri.getFilePath());
```
erictraut, line 410: Just want to confirm that it was intentional that you removed the symlink logic here.

debonte (approved):

bschnurr (approved):

packages/pyright-internal/src/common/uri/memoization.ts
```diff
@@ -55,7 +55,7 @@ export function cacheStaticFunc() {
     return function cacheStaticFunc_Fast(target: any, functionName: string, descriptor: PropertyDescriptor) {
         const originalMethod = descriptor.value;
         descriptor.value = function (...args: any) {
-            const key = `${functionName}+${args.map((a: any) => a?.toString()).join(',')}`;
+            const key = `${functionName}+${args?.map((a: any) => a?.toString()).join(',')}`;
```
erictraut, line 58: If there are no args, this will presumably result in a key that has `undefined` in it. Won't that same key result from a function that is passed a single argument of type `undefined`? Just want to make sure this isn't a cache namespace collision.

erictraut (approved):

packages/pyright-internal/src/common/realFileSystem.ts
rchiodo: I believe this was on purpose. We don't want the original  path, we want the symlinked path, otherwise VS code will open another copy of the file. 

@debonte can correct me if I'm wrong. I believe this was a bug fix that he made.

packages/pyright-internal/src/common/uri/memoization.ts
rchiodo: I don't believe this is possible if the function was defined with arguments, args will always be defined in that case. So for a function `foo`, the argument part of the key is undefined when:
- There are no arguments ever
- There's one argument and it's undefined

But you can't have the same situation for the same function because it either has an argument or it doesn't (in the definition of the function). I believe that means there should be no cache collision.

packages/pyright-internal/src/common/realFileSystem.ts
debonte: Yes, this is intentional. Here's the explanation from the original pyrx PR:

> RealFileSystem.realCasePath currently only skips real-casing if the file/directory that the input path points to is a symbolic link. But it needs to also skip it if the parent, grandparent, etc directories are symbolic links.
>
> Rather than stat-ing all ancestor directories, I'm checking to see if the real-cased path differs from the original by more than just case. If it does, I'm returning the original path instead.

packages/pyright-internal/src/common/uri/memoization.ts
rchiodo: I tried this with some test code just to make sure:

```typescript
function cacheStaticFunc() {
    return function cacheStaticFunc_Fast(target: any, functionName: string, descriptor: PropertyDescriptor) {
        const originalMethod = descriptor.value;
        descriptor.value = function (...args: any) {
            const key = `${functionName}+${args?.map((a: any) => a?.toString()).join(',')}`;
            console.log(`Calling ${functionName} with args: ${key}`);
            return originalMethod.apply(this, args);
        };
        return descriptor;
    };
}

class Test {

  @cacheStaticFunc()
  static foo() {
   console.log('foo');
}
  @cacheStaticFunc()
  static bar(arg1?: string, arg2?: string) {
   console.log('bar');
}
}

Test.foo();
Test.bar();
Test.bar(undefined, undefined);
Test.bar('foo');
```

Produces the following output:
```
Calling foo with args: foo+
foo
Calling bar with args: bar+
bar
Calling bar with args: bar+,
bar
Calling bar with args: bar+foo
bar
```

rchiodo: Which I believe is following what I thought. If you pass undefined to a function with a single argument, it's the same as not passing an argument. Which internally (in the function) makes sense. That argument would be undefined if you passed nothing.

PR #7017, opened 2024-01-17
Fix basedpyright config

DetachHead:


DetachHead:
oops, github keeps defaulting to the upstream repo when these PRs are only meant for my fork

Issue #7016, opened 2024-01-17
extraPaths in language server no longer work on pyright >= 1.1.340


FindDefinition:
**Describe the bug**

I'm using pyright as language server via a websocket server and pyright server run in `--stdio` mode. recently I found that my dev project (installed via `pip install -e .`) can't be resolved by pyright langserver. previously I use `extraPaths` to tell pyright check my dev project and it works well in pyright <= 1.1.339. 

packages in standard python directory (miniconda3) have no problem.

I test some pyright versions by binary search and get following result:

version <= 1.1.339: works
version == 1.1.340: hang
version >= 1.1.341: not working

**Code or Screenshots**
I can provide language server input and output jsonrpc messages:
1.1.347:
```
[...]

PR #7037, opened 2024-01-18
Fix combining paths with empty uri

rchiodo:
Addresses #7016

### DIFF
--- a/packages/pyright-internal/src/common/uri/emptyUri.ts
+++ b/packages/pyright-internal/src/common/uri/emptyUri.ts
@@ -14,7 +14,7 @@ const EmptyKey = '<empty>';
 export class EmptyUri extends FileUri {
     private static _instance = new EmptyUri();
     private constructor() {
-        super(EmptyKey, '', '', '', undefined, false);
+        super(EmptyKey, '', '', '', undefined, /* isCaseSensitive */ true);
     }
 
     static get instance() {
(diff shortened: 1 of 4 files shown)


packages/pyright-internal/src/common/uri/uriUtils.ts
```diff
@@ -428,3 +429,7 @@ export function encodeUri(fs: ReadOnlyFileSystem, uri: Uri): string {
     // Convert to a URI string that the LSP client understands (mapped files are only local to the server).
     return fs.getOriginalUri(uri).toString();
 }
+
+export function resolveFilePath(root: Uri, filePath: string) {
```
rchiodo, line 433: Whoops don't need this.

erictraut (approved):

packages/pyright-internal/src/common/uri/emptyUri.ts
```diff
@@ -104,10 +100,6 @@ export class EmptyUri extends BaseUri {
         return 0;
     }
 
-    override combinePaths(...paths: string[]): Uri {
```
rchiodo, line 107: Essentially combinePaths with an empty string has to create a file URI

rchiodo: If you look at the logic in combinePaths for fileUris, it's the same as what we used to do in 'resolvePaths'

PR #7047, opened 2024-01-19
Update schema

InSyncWithFoo:
This is basically a copy from [downstream](https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/partial-pyright.json), most content of which was originally copied from this file and [`configuration.md`](https://github.com/microsoft/pyright/blob/main/docs/configuration.md). I'm not sure about `overrides`; I could have sworn I saw it mapping to an empty object at the very end of the schema, but after checking recent commits of this repo, now I think I just hallucinated.

I changed the formatting for `title` and `description` a bit; inline code are now delimited by backticks (similar to Markdown, partly because I wanted to make the transpilation to Markdown easier). [`markdownDescription`](https://code.visualstudio.com/docs/languages/json#_json-schemas-and-settings) is for VSCode and [`x-intellij-html-description`](https://www.jetbrains.com/help/pycharm/json.html#ws_json_show_doc_in_html) is for PyCharm; the latter is automatically generated from the former using [`marked`](https://www.npmjs.com/package/marked)`.parseInline()`. Regardless, it still doubles the maintenance efforts required, so this might be [considered](https://xkcd.com/1172/) a breaking change.

### DIFF
--- a/packages/vscode-pyright/schemas/pyrightconfig.schema.json
+++ b/packages/vscode-pyright/schemas/pyrightconfig.schema.json
@@ -27,10 +27,12 @@
       "$id": "#/properties/include",
       "type": "array",
       "title": "Files and directories included in type analysis",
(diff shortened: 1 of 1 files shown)


erictraut:
Yeah, sorry, but I don't think I want to accept this change.

Issue #7137, opened 2024-01-27
Add multi-root workspaceFolder support to path variable expansion


Seairth:
Currently, pyright does not support path variable expansion for tokens in the form of `${workspaceFolder:name}`, which is documented by VSCode for multi-root workspaces.  Following from the invitation at https://github.com/microsoft/pylance-release/issues/4405#issuecomment-1912811225, I added the feature. The current commit can be found at https://github.com/Seairth/pyright/commit/37fff578e2cb4d558def86b4edebab883c848051.

Overall changes:
* Re-ordered the `expandPathVariables` arguments and added an additional argument for the list of workspaces
* Enhanced `expandPathVariables` to iterate over the workspaces and perform variable expansion for tokens in the form of `${workspaceFolder:name}`
* Updated existing tests that used `expandPathVariables`
* Added new tests for `expandPathVariables` (including tests for existing functionality)

All tests pass.  I have also built the VSCode extension and verified the new functionality.

Before I submit a PR, I want to make sure the commit meets "Contributing" requirements.

erictraut:
Thanks for the contribution. It looks good to me, but I'd like the pylance team members to review it as well. Please go ahead and submit the PR.

erictraut:
This is addressed in pyright 1.1.350, which I just published.

PR #7138, opened 2024-01-27
Added multi-root workspaceFolder support in path variable expansion

Seairth:
See #7137


Seairth:

Seairth:
Hmm.  I see that amending the commit reset the PR.  For future reference, what is the preferred approach for addressing issues?

Seairth:
I ran prettier on the files, and it resolved one of the warnings.  I do not know why `npm run check` is reporting a warning about `server.ts`, since running prettier on the file yields no changes.

rchiodo (approved):
Thanks for adding the workspace:name support. This is great :)

Seairth:
> Thanks for adding the workspace:name support. This is great :)

I'm glad I could contribute back, even if a small way.

PR #7172, opened 2024-01-31
Add test(s) that validate Pyright can talk over LSP

### LINKED ISSUE #6677: Provide tests at the LSP level in Pyright
Pylance has a series of tests we run on submission that run the language server in a worker thread and then exercise the language server by passing messages to the worker thread. 

This request is to add these sort of tests to Pyright as well. Theoretically this would have caught this issue:
#6676

rchiodo:
Fixes #6677 

This should make sure we don't break the background thread in any submissions.

### DIFF
--- a/packages/pyright-internal/src/tests/languageServer.test.ts
+++ b/packages/pyright-internal/src/tests/languageServer.test.ts
@@ -1,3 +1,11 @@
+/*
+ * languageServer.test.ts
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ *
+ * Tests to verify Pyright works as the backend for a language server.
+ */
+
 import assert from 'assert';
 import {
     CancellationToken,
--- a/packages/pyright-internal/src/tests/languageServer.test.ts
+++ b/packages/pyright-internal/src/tests/languageServer.test.ts
@@ -3,8 +3,6 @@ import {
     CancellationToken,
     CompletionRequest,
     ConfigurationItem,
-    DidOpenTextDocumentNotification,
-    HoverRequest,
     InitializedNotification,
     InitializeRequest,
     MarkupContent,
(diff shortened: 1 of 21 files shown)


packages/pyright-internal/src/tests/lsp/main.ts
```diff
@@ -0,0 +1,11 @@
+/*
+ * node.ts
```
erictraut, line 2: File name is wrong in comment.

packages/pyright-internal/tsconfig.json
```diff
@@ -1,7 +1,9 @@
 {
     "extends": "../../tsconfig.json",
     "compilerOptions": {
-        "outDir": "./out"
+        "outDir": "./out",
+        "paths": {
```
erictraut, line 5: Is this change needed?

erictraut (approved):
Thanks for writing these tests. This is a good addition to the test suite.

Looks good overall. Just a few nits for code style conformance.

packages/pyright-internal/tsconfig.json
rchiodo: Yeah otherwise tsconfigResolveAliases throws an exception.

packages/pyright-internal/src/tests/languageServer.test.ts
```diff
@@ -0,0 +1,185 @@
+import assert from 'assert';
```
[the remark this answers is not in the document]
rchiodo: Will do

packages/pyright-internal/src/tests/lsp/main.ts
rchiodo: Will do

Loki419420 (commented):
#7179

PR #7216, opened 2024-02-06
Clarify that "exclude" overrides include directories *and* files

ljw1004:
The exclude directive overrides include directories *and* include files.

I verified by experiment, and also by reading the source code https://github.com/microsoft/pyright/blob/main/packages/pyright-internal/src/common/uri/uriUtils.ts#L54


erictraut:
Thanks for the contribution.

PR #7223, opened 2024-02-07
Docs: give concrete example of venv

ljw1004:
I found the venv documentation hard to follow, and had to debug-step through pyright code to discover what it was looking for. Here's a simple example to hopefully get the reader started.


ljw1004:
Drat I'm not very good at git yet. I'll close this and try again.

PR #7227, opened 2024-02-08
[doc] add example of venv

ljw1004:
I couldn't figure out how to use venv / venvPath from the docs, and had to step through with a debugger to figure out what it was looking for. I'm submitted this doc update in case other people might likewise benefit.

(On the other hand, the "venv" technique is recommended against, so maybe this documentation shouldn't be added!)


erictraut:
Thanks for taking the time to propose this change, but I'd prefer to leave it out of the documentation. As you pointed out, the venv mechanism is left in place for historical and backward compatibility reasons, but it's not the configuration technique that I recommend. It has many shortcomings, and there are much better (more robust) solutions.

Issue #7215, opened 2024-02-06
Exclude Jupyter notebooks completely from pyright


debonte:
### Discussed in https://github.com/microsoft/pyright/discussions/7195

<div type='discussions-op-text'>

<sup>Originally posted by **medihack** February  3, 2024</sup>
In my Django project I use some Jupyter notebooks in a separate `notebooks` folder. I want to completely exclude those notebooks from being scanned by pyright. That is why I tried the following setting in my `pyproject.toml`:

```toml
[tool.pyright]
exclude = ["**/notebooks", "**/.ipynb*"]
ignore = ["**/migrations", "**/notebooks", "**/.ipynb*"]
typeCheckingMode = "basic"
reportUnnecessaryTypeIgnoreComment = true
```

Unfortunately, the VS Code extension (v1.1.349) does not respect this setting and still reports some `reportUndefinedVariable` errors (that are defined in some previous cells). How do I completely turn off pyright (VS Code extension) for specific folders or files?</div>

debonte:
@medihack, this works for me on Linux, but is broken on Windows. I assume you are using Windows?

This is a regression caused by our recent URI refactoring. `WebUri.matchesRegex` is failing to match notebook cell URIs against the ignore specs because the `WebUri._path` has a leading slash. For example, `/c:/test/PyrightExperiments/notebooks/test.ipynb` which doesn't match `/^(c\:\/test\/PyrightExperiments(\/[^/][^/]*)*?\/[^/]*\.ipynb[^/]*)($|\/)/i`.

@rchiodo, I'm not sure what the right fix is here. The only things that occurred to me would be special-casing either Windows or notebook cells, which I'm assuming we'd like to avoid.

Btw, @medihack, once this is fixed, if you want to ignore all `.ipynb` files, you're missing a `*` -- rather than `"**/.ipynb*"` you want `"**/*.ipynb*"`. Also, the trailing `*` is unnecessary in Pyright. I think it may still be necessary in Pylance, because internally we tack on a `.py` extension after `.ipynb`, but users shouldn't need to know that, so I'd say that's a bug.

rchiodo:
> WebUri.matchesRegex is failing to match notebook cell URIs against the ignore specs because the WebUri._path has a leading slash.

I have two questions

1) why is the notebook a weburi if it's at `c:`
2) why would it have a leading slash?

debonte:
> 1. why is the notebook a weburi if it's at `c:`

[...]

PR #7231, opened 2024-02-08
Restrict vscode-pyright document selector to `file` and `untitled` schemes

debonte:
Address https://github.com/microsoft/pyright/issues/7215

This will prevent VS Code from communicating with Pyright about Python documents with URI schemes other than `file` and `untitled`. For example: `vscode-notebook-cell`.

I considered filtering out notebook cells in the textDocument didOpen, didChange, and didClose handlers, but then VS Code will still send requests (ex. hover) for the filtered out documents. So it seems best to tell VS Code more clearly what Pyright can handle.

This matches Pylance's document selector minus notebooks and virtual file systems. Pyright doesn't support virtual file systems, right?


rchiodo (approved):

erictraut:
I have some concerns about this approach. I think I recall that some editors use a URI other than "untitled" for unsaved documents. This would break those clients.

Perhaps we could blacklist notebooks rather than whitelisting just a few specific url types?

rchiodo:
> Pyright doesn't support virtual file systems, right?

No, it does not. It would need a way to send a request to read to a client. (The sync server support)

debonte:
> I think I recall that some editors use a URI other than "untitled" for unsaved documents. This would break those clients.

This change is in `vscode-pyright`. I was assuming that other editors would not use that package at all. Is that not true?

erictraut (approved):

erictraut:
> This change is in vscode-pyright. I was assuming that other editors would not use that package at all. Is that not true?

That should be fine then. Thanks.

PR #7332, opened 2024-02-25
Debugging

ljw1004:
I got a bit irritated in my edit-build-debug inner loop with the time to run webpack.

This commit adds a new launch target called "Pyright CLI (pyright-internal)" which bypasses webpack and just runs the package straight from the packages/pyright-internal/out directory where tsc has built it.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -17,6 +17,21 @@
             ]
         },
         {
+            "name": "Pyright CLI (pyright-internal)",
+            "type": "node",
+            "request": "launch",
+            "program": "${workspaceRoot}/packages/pyright-internal/debug.js",
+            "args": [
+                "--typeshedpath",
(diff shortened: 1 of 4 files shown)


debonte:
I wonder if you could make this work like "Pyright extension (watch)" instead so it would do an incremental build whenever you saved a file, but otherwise no build would be required on relaunch.

ljw1004:
@microsoft-github-policy-service agree company="Meta"

ljw1004:
> I wonder if you could make this work like "Pyright extension (watch)" instead so it would do an incremental build whenever you saved a file, but otherwise no build would be required on relaunch.

Oh that's beautiful. I hadn't understood this "preLaunchTask / isBackground:true / beginsPattern+endsPattern" until now. Thanks for the pointer. I'll create another PR for it.

PR #7375, opened 2024-02-29
Drop support for node 12

rchiodo:
This commit https://github.com/microsoft/pyright/commit/e955fab18e5a2994c83d5909bbb644705dd4eb75 brought in a dependency change that requires node 14 or higher.

This change ups the minimum engine required to 14 so that we can use this dependency.

### DIFF
--- a/docs/ci-integration.md
+++ b/docs/ci-integration.md
@@ -90,7 +90,7 @@ vercomp () {
 # Node version check
 echo "Checking node version..."
 NODE_VERSION=`node -v | cut -d'v' -f2`
-MIN_NODE_VERSION="10.15.2"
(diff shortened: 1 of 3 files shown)


rchiodo:
Oh and I did verify that node 14 works. At least I can run it like so:

```
(main) C:\Users\rchiodo\source\repos\pyright\packages\pyright>nvm use 14
Now using node v14.21.3 (64-bit)

(main) C:\Users\rchiodo\source\repos\pyright\packages\pyright>node --version
v14.21.3

(main) C:\Users\rchiodo\source\repos\pyright\packages\pyright>node index.js --version
pyright 1.1.352
```

erictraut (approved):
LGTM

PR #7380, opened 2024-02-29
Send unnecssary/deprecated diagnostics when clients don't support tags

the-mikedavis:
Previously if an LSP client didn't declare tagSupport in the client capabilities for publishDiagnostic, pyright didn't send diagnostics for unnecessary or deprecated diagnostics. I think it's still valuable to send these regardless of support for tags. When a client doesn't support tags, we can just skip attaching them.

### DIFF
--- a/packages/pyright-internal/src/languageServerBase.ts
+++ b/packages/pyright-internal/src/languageServerBase.ts
@@ -1516,21 +1516,19 @@ export abstract class LanguageServerBase implements LanguageServerInterface, Dis
                 diag.category === DiagnosticCategory.UnusedCode ||
                 diag.category === DiagnosticCategory.UnreachableCode
             ) {
-                vsDiag.tags = [DiagnosticTag.Unnecessary];
-                vsDiag.severity = DiagnosticSeverity.Hint;
-
-                // If the client doesn't support "unnecessary" tags, don't report unused code.
-                if (!this.client.supportsUnnecessaryDiagnosticTag) {
-                    return;
+                // If the client doesn't support "unnecessary" tags, don't attach them.
+                if (this.client.supportsUnnecessaryDiagnosticTag) {
+                    vsDiag.tags = [DiagnosticTag.Unnecessary];
                 }
+
+                vsDiag.severity = DiagnosticSeverity.Hint;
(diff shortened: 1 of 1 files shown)


the-mikedavis:

erictraut:
I don't agree with this change. These tagged hints are not meant to be presented as regular diagnostics, and they shouldn't be treated as such. There is no way to suppress them (e.g. through `# type: ignore`). They are meant to be subtle visual hints to the user. If the client doesn't support these tags, then they should not be generated.

the-mikedavis:
Can I suggest using an LSP extension for sending these hints, or the semantic tokens part of the spec instead? I don't use Python or Pyright with any regularity so I'm a little out of my depth here, but if these are not regular diagnostics then I believe it's improper to send them through textDocument/publishDiagnostic. Using tags this way is against the spirit of the spec in my interpretation:

```typescript
export interface Diagnostic {
  // ..

	/**
	 * Additional metadata about the diagnostic.
	 *
	 * @since 3.15.0
	 */
	tags?: DiagnosticTag[];

  // ..
}
```

I believe tags are meant to be additional context about a diagnostic - the diagnostic is still a diagnostic.

Other clients like Neovim already present them as regular diagnostics with extra styling, and I think there is no way to supress the diagnostics there (though, again, I'm out of my water here - I'm not familiar with `# type: ignore`ing). Neovim's presentation, from a downstream bug report (https://github.com/helix-editor/helix/discussions/9765):

neovim-hint-deprecated

It's not a high priority suggestion at all as supporting tags in clients is very straightward. But I believe semantic tokens (specifically the [`SemanticTokenModifiers.deprecated`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_semanticTokens) member) is a better fit for these hints.

kas2020-commits:
> [tagged hints]... are meant to be subtle visual hints to the user.

@erictraut if that's the case then what is a user supposed to do if they're trying to enforce the `deprecateTypingAliases` code standard on a code base? The only recourse available is to set `typeCheckingMode` to strict which seems unnecessarily excessive for a simple deprecation warning.

PR #7388, opened 2024-03-02
Show `str` method docstrings when called on `LiteralString`

debonte:
Address https://github.com/microsoft/pylance-release/issues/5566

Pylance is failing to get the docstring for `str` methods when they are called on `LiteralString` objects. See https://github.com/microsoft/pylance-release/issues/5566#issuecomment-1974835169. It's [this `isInheritedFromBuiltin` check](https://github.com/microsoft/pyright/blob/c64f9b3f1f2dbe89af8346dfab2e69a5af1460bb/packages/pyright-internal/src/analyzer/typeDocStringUtils.ts#L82) that prevents the docstring from being used.

I changed`isInheritedFromBuiltin` to return `false` if both the base class and derived class are builtins. From the comments where this function is called, I believe the intent is to avoid showing docstrings from builtin types on library/user types that inherit from them. But when the derived type is also a builtin, I'm thinking we should still show the docstring.

I was also thinking about changing `isSameGenericClass` to think that `str` and `LiteralString` were the "same", but that caused a lot of test failures. Mainly false negatives I think, which in hindsight makes sense.

### DIFF
--- a/packages/pyright-internal/src/analyzer/typeDocStringUtils.ts
+++ b/packages/pyright-internal/src/analyzer/typeDocStringUtils.ts
@@ -64,6 +64,7 @@ function isInheritedFromBuiltin(type: FunctionType | OverloadedFunctionType, cla
         type.details.moduleName === 'builtins' &&
         !!classType &&
         !!type.boundToType &&
+        !ClassType.isBuiltIn(type.boundToType) &&
         !ClassType.isSameGenericClass(classType, type.boundToType)
     );
 }
--- a/packages/pyright-internal/src/analyzer/typeDocStringUtils.ts
+++ b/packages/pyright-internal/src/analyzer/typeDocStringUtils.ts
@@ -61,11 +61,10 @@ function isInheritedFromBuiltin(type: FunctionType | OverloadedFunctionType, cla
     // Functions that are bound to a different type than where they
     // were declared are inherited.
     return (
-        type.details.moduleName === 'builtins' &&
-        !!classType &&
+        !!type.details.methodClass &&
+        ClassType.isBuiltIn(type.details.methodClass) &&
         !!type.boundToType &&
-        !ClassType.isBuiltIn(type.boundToType) &&
-        !ClassType.isSameGenericClass(classType, type.boundToType)
+        !ClassType.isBuiltIn(type.boundToType)
     );
 }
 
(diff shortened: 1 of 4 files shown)


packages/pyright-internal/src/analyzer/typeDocStringUtils.ts
```diff
@@ -64,6 +64,7 @@ function isInheritedFromBuiltin(type: FunctionType | OverloadedFunctionType, cla
         type.details.moduleName === 'builtins' &&
         !!classType &&
         !!type.boundToType &&
+        !ClassType.isBuiltIn(type.boundToType) &&
         !ClassType.isSameGenericClass(classType, type.boundToType)
```
debonte, line 68: If we take this change, I think the `isSameGenericClass` call is no longer needed. Presumably if `type` is a builtin, but `type.boundToType` is not, they are not the same class.

packages/pyright-internal/src/analyzer/typeDocStringUtils.ts
```diff
@@ -64,6 +64,7 @@ function isInheritedFromBuiltin(type: FunctionType | OverloadedFunctionType, cla
         type.details.moduleName === 'builtins' &&
```
debonte, line 64: Should this be calling `isBuiltIn` rather than checking the `moduleName`? Not sure if this is intentional or just old.

packages/pyright-internal/src/analyzer/types.ts
```diff
@@ -1189,6 +1189,14 @@ export namespace ClassType {
             return true;
         }
 
+        // Handle LiteralString as a special case.
+        if (
+            (ClassType.isBuiltIn(classType, 'str') && ClassType.isBuiltIn(type2, 'LiteralString')) ||
+            (ClassType.isBuiltIn(classType, 'LiteralString') && ClassType.isBuiltIn(type2, 'str'))
+        ) {
+            return true;
```
debonte, line 1197: Based on the test errors, this seems to be wrong.

packages/pyright-internal/src/analyzer/typeDocStringUtils.ts
erictraut: I'm not familiar with this code. I think Jake wrote it. The change looks reasonable to me.

heejaechang (approved):

heejaechang:
tests?

debonte:
> tests?

Added tests and also simplified `isInheritedFromBuiltin` to something that I believe is still correct. @heejaechang , can you take another look?

heejaechang (approved):

PR #7434, opened 2024-03-09
Save some memory in token creation

rchiodo:
While investigating this issue:
https://github.com/microsoft/pylance-release/issues/5440

Erik and I noticed that the `escapedValue` for a bunch of different tokens was not actually allocated from the original string. It was an entirely new string, meaning a new allocation.

It seems that it doesn't need to be that way. It can just point to the original string.

I verified this uses less memory by snapshotting a run of pylance with a file with a lot of string literals in it.

In the original version, we end up with a parse tree like so:

image

You can see that it's allocated 18.9 MB for the tokens themselves.

If we expand into these tokens, you can see one of the largest ones has its own copy of the giant literal string at the top of the file:

image

That token has 32K allocated to hold that large literal string.

After this change, the amount of memory for the same run goes down by 800K:

image

And if we find that same token with the escaped value, it's now a `sliced` value, which is really just a reference to the original string:

image

### DIFF
--- a/packages/pyright-internal/src/analyzer/parseTreeUtils.ts
+++ b/packages/pyright-internal/src/analyzer/parseTreeUtils.ts
@@ -277,7 +277,7 @@ export function printExpression(node: ExpressionNode, flags = PrintExpressionFla
             let escapedString = node.token.escapedValue;
             if ((flags & PrintExpressionFlags.DoNotLimitStringLength) === 0) {
                 const maxStringLength = 32;
-                escapedString = escapedString.substring(0, maxStringLength);
+                escapedString = escapedString.slice(0, maxStringLength);
             }
 
             if (node.token.flags & StringTokenFlags.Triplicate) {
--- a/packages/pyright-internal/src/parser/stringTokenUtils.ts
+++ b/packages/pyright-internal/src/parser/stringTokenUtils.ts
@@ -9,7 +9,6 @@
  */
 
 import { Char } from '../common/charCodes';
-import { maxStringTokenLength } from './tokenizer';
 import { FStringMiddleToken, StringToken, StringTokenFlags } from './tokenizerTypes';
 
 export const enum UnescapeErrorType {
@@ -93,14 +92,11 @@ export function getUnescapedString(stringToken: StringToken | FStringMiddleToken
     const addInvalidEscapeOffset = () => {
         // Invalid escapes are not reported for raw strings.
         if (!isRaw) {
-            // If this is the last character of a truncated string, don't report.
-            if ((stringToken.flags & StringTokenFlags.ExceedsMaxSize) === 0 || strOffset < maxStringTokenLength) {
-                output.unescapeErrors.push({
-                    offset: strOffset - 1,
-                    length: 2,
-                    errorType: UnescapeErrorType.InvalidEscapeSequence,
-                });
-            }
+            output.unescapeErrors.push({
+                offset: strOffset - 1,
+                length: 2,
+                errorType: UnescapeErrorType.InvalidEscapeSequence,
+            });
         }
     };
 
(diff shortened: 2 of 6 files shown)


packages/pyright-internal/src/analyzer/parseTreeUtils.ts
```diff
@@ -277,7 +277,7 @@ export function printExpression(node: ExpressionNode, flags = PrintExpressionFla
             let escapedString = node.token.escapedValue;
             if ((flags & PrintExpressionFlags.DoNotLimitStringLength) === 0) {
                 const maxStringLength = 32;
-                escapedString = escapedString.substring(0, maxStringLength);
```
rchiodo, line 280: These `slices` aren't really necessary, it's just that `substring` has been deprecated. I figured I'd switch on the files I was looking at. I can put these back if we want to not bother with these changes.

packages/pyright-internal/src/parser/tokenizer.ts
```diff
@@ -1572,17 +1569,22 @@ export class Tokenizer {
         const isTriplicate = (flags & StringTokenFlags.Triplicate) !== 0;
         const isFString = (flags & StringTokenFlags.Format) !== 0;
         let isInNamedUnicodeEscape = false;
-        let escapedValueParts: number[] = [];
+        const start = this._cs.position;
+        let escapedValueLength = 0;
+        const getEscapedValue = () => this._cs.getText().slice(start, start + escapedValueLength);
```
rchiodo, line 1574: This is where the real change happens. Instead of building the escaped value from the the char codes, it can just be built from the literal string itself.

packages/pyright-internal/src/parser/tokenizer.ts
```diff
@@ -1655,41 +1660,35 @@ export class Tokenizer {
...
                     this._cs.moveNext();
                 }
             } else {
-                escapedValueParts.push(this._cs.currentChar);
+                escapedValueLength++;
                 this._cs.moveNext();
             }
         }
 
-        // String.fromCharCode.apply crashes (stack overflow) if passed an array
```
rchiodo, line 1683: This is also unnecessary now since we're just slicing into the original string.

packages/pyright-internal/src/parser/stringTokenUtils.ts
```diff
@@ -142,7 +146,7 @@ export function getUnescapedString(stringToken: StringToken | FStringMiddleToken
     while (true) {
         let curChar = getEscapedCharacter();
         if (curChar === Char.EndOfText) {
-            return completeUnescapedString(output);
+            return completeUnescapedString(output, escapedString);
```
rchiodo, line 149: Found this one too. Netted another 200K in memory savings with the test app I'm using

packages/pyright-internal/src/parser/tokenizer.ts
```diff
@@ -1655,41 +1660,35 @@ export class Tokenizer {
...
-        // that is too long. Cut off the extra characters in this case to avoid
-        // the crash. It's unlikely that the full string value will be used as
-        // a string literal or a docstring, so this should be fine.
-        if (escapedValueParts.length > maxStringTokenLength) {
-            escapedValueParts = escapedValueParts.slice(0, maxStringTokenLength);
-            flags |= StringTokenFlags.ExceedsMaxSize;
```
erictraut, line 1689: Are you removing the generation of ExceedsMaxSize? If this is no longer a limitation, let's completely remove ExceedsMaxSize from the StringTokenFlags enumeration along with any references to it.

erictraut (approved):
A relatively simple change that results in pretty significant memory savings. Love it! 

I left one small comment. Looks like we can remove some additional code.

debonte (approved):

packages/pyright-internal/src/parser/tokenizer.ts
rchiodo: Thanks, will do.

Issue #7159, opened 2024-01-29
IntelliSense is not reactive with sympy


denismazzucato:
## Environment data

-   Language Server version: 2023.12.1
-   OS and version: macOS Sonoma 14.2
-   Python version: 3.12.0
-   Conda version 23.10.0
-   Sympy version: 1.12

## Code Snippet

```python
from sympy import Symbol

x = Symbol("x")
```

## Expected behavior

Hovering over the the variable x should show that x is a Symbol in no-time, or auto-completion should appear instantly when typing anywhere in the code above.

## Actual behavior

It takes about 5 seconds to show the type information when hover with the cursor on x, and about 12 seconds to show any auto-completion.

The problem is using Symbol from sympy as all the other features seems to not run any delay. For example, the code below does not halt IntelliSense:
```python
import sympy as sym

x = sym.var("x")
```

## Logs

[pylance-sympy-trace.log](https://github.com/microsoft/pylance-release/files/14086270/pylance-sympy-trace.log)

erictraut:
The `sympy` library claims to be a "typed library" by virtue of the fact that it includes a "py.typed" marker file (as outlined in [PEP 561](https://peps.python.org/pep-0561/#packaging-type-information)). However, sympy unfortunately contains almost no type information. That means pyright will attempt to infer type information (function return types, etc.) from the code. 

You would normally be able to disable this behavior in pyright by setting the `useLibraryCodeForTypes` option to false, but this setting applies only to libraries that are not marked as "py.typed". When a library claims to be typed, pyright honors that designation and acts accordingly. 

I ran `pyright --verifytypes sympy` to determine the "type completeness" of the library. Most "py.typed" libraries are 50% or more "type complete". Sympy is currently just 6%. It really has no right to claim that it's a typed library.

```
Symbols exported by "sympy": 35039
  With known type: 2107
  With ambiguous type: 1808
  With unknown type: 31124

Type completeness score: 6%
```

I'm not sure why sympy maintainers decided to add a "py.typed" marker file when the library is not typed. Perhaps they didn't understand the implications of doing so.

For most libraries, pyright can infer return types relatively quickly. Sympy, however, contains many functions and methods that have complex code flow graphs and large numbers of variables, which makes type inference very costly.

If this library is important to you, I recommend reaching out to the maintainers to ask them to 1) remove the "py.typed" marker, since the library is clearly not typed, or 2) add more type annotations to the library's public interface.

[...]

PR #7477, opened 2024-03-13
Potential method to shorten evaluation time for certain packages

rchiodo:
See this comment here:
https://github.com/microsoft/pyright/issues/7159#issuecomment-1992769774

After some investigation I found that sympy is just evaluating a lot more types. This change allows a limit be set for expanding types. Internally in Pylance I was thinking we'd have a setting like so:

```json
"python.analysis.minimizedPackages": ["sympy"]
```

This would be applied to the `type` argument on the filter to see what module the type comes from. If a package is in this list, the user will still get some information, but the package will analyze much faster.

The alternative would be to provide type stubs for these packages, but that can take a long time. This is more like a shortcut for when we know certain packages are slow to analyze.

### DIFF
--- a/packages/pyright-internal/src/analyzer/backgroundAnalysisProgram.ts
+++ b/packages/pyright-internal/src/analyzer/backgroundAnalysisProgram.ts
@@ -20,6 +20,7 @@ import { Uri } from '../common/uri/uri';
 import { AnalysisCompleteCallback, analyzeProgram } from './analysis';
 import { ImportResolver } from './importResolver';
 import { MaxAnalysisTime, OpenFileOptions, Program } from './program';
+import { Type } from './types';
 
 export enum InvalidatedReason {
     Reanalyzed,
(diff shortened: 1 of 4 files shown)


rchiodo:
After using this setting for sympy, sympy analysis times are back to what they were before https://github.com/microsoft/pyright/commit/6753a5d30535ba9a5708ff62494f841519fc54ce

erictraut (commented):
Please DO NOT commit this change. This will produce inconsistent and non-deterministic type evaluation results. I understand the desire to speed up sympy analysis, but this is NOT the right approach.

rchiodo:
Okay I'll look into generating stubs instead.

erictraut:
Stubs are one answer, but there might be other solutions. It sounds like you have done some analysis on the problem. If you're able to write up the notes about what you've found and include them in [this issue](https://github.com/microsoft/pyright/issues/7159), I might be able to suggest other potential solutions. I haven't had time to do an in-depth analysis of the problem yet (and probably won't have time in the near future), but if you've already done the analysis, we might be able to tag-team.

rchiodo:
Thanks. I did add some comments but I'll try to add some more detail.

PR #7526, opened 2024-03-20
push pylance to pyright

bschnurr:
- Upgrade to latest prerelease LSP 3.18
- Remove LS root uri


bschnurr:
holding off. more uri changes coming.

PR #7539, opened 2024-03-21
Update Unicode range tables to match Unicode 15.1

debonte:
Addresses https://github.com/microsoft/pylance-release/issues/5656

The user's specific complaint was that CJK Ideograph characters were not recognized by the tokenizer as Python identifier chars despite being part of Unicode's `Lo` (other letters) category. The reason was that the `unicodeLoSurrogate` table only contained the endpoints of the CJK ideograph ranges. See the exerpt below from our current `unicodeLoSurrogate` table which, for example, indicates that codes `55360 56320` (U+20000) and `55401 57053` (U+2A6DD) are in the `Lo` category, but in fact all of the characters between those two codes are also `Lo` and are therefore also valid identifier characters.

```typescript
export const unicodeLoSurrogate: UnicodeSurrogateRangeTable = {
    ...
    55360: [56320],
    55401: [57053, 57088],
    55405: [57140, 57152],
    55406: [56349, 56352],
    55411: [56993, 57008],
    55418: [57312],
```

It's likely that this mistake happened because the CJK Ideographs are listed in the unicode metadata files using [an old range representation](https://www.unicode.org/reports/tr44/#Code_Point_Ranges) that is still supported for backwards compatibility. In this legacy approach, rather than declaring a range as `AAAA..BBBB`, the first and last characters of a range are listed on separate lines which makes them look like individual characters, but the presence of the "First" and "Last" keywords within the angle brackets indicates that they are actually the endpoints of a character range:

```text
20000;<CJK Ideograph Extension B, First>;Lo;0;L;;;;;N;;;;;
2A6DF;<CJK Ideograph Extension B, Last>;Lo;0;L;;;;;N;;;;;
2A700;<CJK Ideograph Extension C, First>;Lo;0;L;;;;;N;;;;;
2B739;<CJK Ideograph Extension C, Last>;Lo;0;L;;;;;N;;;;;
2B740;<CJK Ideograph Extension D, First>;Lo;0;L;;;;;N;;;;;
2B81D;<CJK Ideograph Extension D, Last>;Lo;0;L;;;;;N;;;;;
2B820;<CJK Ideograph Extension E, First>;Lo;0;L;;;;;N;;;;;
2CEA1;<CJK Ideograph Extension E, Last>;Lo;0;L;;;;;N;;;;;
2CEB0;<CJK Ideograph Extension F, First>;Lo;0;L;;;;;N;;;;;
2EBE0;<CJK Ideograph Extension F, Last>;Lo;0;L;;;;;N;;;;;
2EBF0;<CJK Ideograph Extension I, First>;Lo;0;L;;;;;N;;;;;
2EE5D;<CJK Ideograph Extension I, Last>;Lo;0;L;;;;;N;;;;;
```

I wrote a Python script that transforms the official Unicode character database into the `unicode.ts` format. By default it uses Unicode 15.1 which is the version that CPython is currently using. This fixed the CJK Ideograph issue and also a number of other inconsistencies between our tables and Unicode 15.1.

FYI, I found https://ucsx.org/convert-form very useful for translating between UCS character codes and their UTF-16 equivalents.


packages/pyright-internal/src/parser/unicode.ts
```diff
@@ -291,6 +291,8 @@ export const unicodeLu: UnicodeRangeTable = [
     4295,
     4301,
     [5024, 5109],
+    [7312, 7354],
```
debonte, line 294: @erictraut, I'd like to make two changes to the file format if you're OK with it:
1. Change character codes from decimal to hex.
2. Add comments on surrogate table lines with the UCS character codes in hex. Or even the actual characters?

This would save a lot of numeric translation hassle when a human tries to understand what characters are in here vs. not. And should be an easy change to the generator.

Thinking I should submit this as a separate change though.

build/generateUnicodeTables.py
```diff
@@ -0,0 +1,198 @@
+import sys
```
erictraut, line 1: Please include a header comment. Otherwise, looks good to me.

PR #7565, opened 2024-03-26
Use a list for the configuration options

CaselIT:
I think that using a list for the various configuration options makes the file significantly easier to read


CaselIT:

erictraut:
I'm willing to give it a try.

Thanks for the contribution.

CaselIT:
thanks!

PR #7601, opened 2024-04-01
Create memory compressed method for storing tokens

rchiodo:
This bug here led us to do a bunch of analysis of the memory in use by Pylance/Pyright:

https://github.com/microsoft/pylance-release/issues/5440

While analyzing the memory in use, it seemed like we were taking up a lot of memory with each token. Here's an example from opening the aws_cdk/__init__.py file

image

That's one of the larger tokens but most of them are around that size. 40 bytes for the class of the token itself, and 116 bytes if you include all the data it was storing. 

That seems like an awful lot of memory to hold onto a token.

## The Change 

This change implements a compressed token storage that uses a single Int32Array to store all of the token data and one more to store the indices into the data array for each token.

That makes the data for a single object take only the storage needed for its data. So for a default token this would be:

- 4 Bytes for type and comments length (comments length is limited to 2 Bytes or 32k possible comments hanging onto a token)
- 4 Bytes for start position (4GB is the limit now for the size of files we can open, we'd crash long before this anyway)
- 4 Bytes for length
- 4 Bytes for the index

Given that a simple token before was at a minimum of 28 bytes, we save at least 12 bytes per token. 

Here's a grab of having both methods in place at the same time.

image

The red arrow is pointing to the total size using the old class based method.

The green arrow is pointing to the total size using the new Int32Array method

That was a degenerate case where we'd get a savings of 60%, but other files in the same workspace have at least a 30% improvement:

(Improvement in Builtins)
image

## Perf differences

Switching things to this method isn't free though. Creating tokens from the array is slow. At least slower than just having the token objects already in memory. 

That's why we implemented a two stage approach. Using the original Token object approach while parsing and binding, and then switch to the Int32Array method for storage and later lookup.

Creation of the array does add some overhead too. In a bunch of tests, it added about 10-15% overhead time to parsing.

Here's the output from a number of runs:

[normal.txt](https://github.com/microsoft/pyright/files/14829209/normal.txt)

[compressed.txt](https://github.com/microsoft/pyright/files/14829210/compressed.txt)

I also did a pprof comparison and the offending difference was all in the compress method.


rchiodo:
I also ported this change to Pylance and tried running the example in the bug without using a custom node executable. We no longer run out of memory. It maxes out at about 2.5GB instead of going to 4 using our current implementation.

packages/pyright-internal/src/parser/tokenCollection.ts
```diff
@@ -0,0 +1,192 @@
...
+    constructor(private readonly _tokens: Token[]) {
+        super(_tokens);
+    }
+
+    toArray(): Token[] {
+        return this._tokens;
+    }
+}
+
+function* generateCompressedData(
```
rchiodo, line 34: I used a generator here because in order to generate an Int32Array, you need to either give it the initial size (which would require two passes through the token list) or give it an iterator.

packages/pyright-internal/src/parser/tokenCollection.ts
```diff
@@ -0,0 +1,192 @@
...
+        currentPosition += tokenArray.length;
+        yield* tokenArray;
+    }
+}
+
+// Special TokenCollection that is optimized for memory usage but is a lot
+// slower to fetch actual tokens.
+export class TokenCollectionCompressed implements ITokenCollection {
+    private _tokenData: Int32Array;
+    private _numberData: (number | bigint)[] = [];
```
rchiodo, line 53: NumberTokens are special cased in that their data can be larger than 4GB. So I store the number value in a separate array.

erictraut:
Thanks for the analysis and the proposed solution. There are tradeoffs (complexity + speed vs. storage), so I'd like to talk through the motivation and pros/cons of this change before merging it. Let's discuss in our sync meeting tomorrow.

I'm surprised that token storage is taking up as much memory as it apparently does. Some random thoughts... Could we go further than this and completely deallocate the token collection after parsing is complete? Critical tokens could be referenced in the parse tree, and the rest could just be discarded. If necessary, we could reconstruct the token list on demand, e.g. if they're needed for semantic token highlighting and the user happens to open that file.

rchiodo:
Closing as we switched to the not storing the tokens anymore.

PR #7605, opened 2024-04-02
Share heap usage numbers to support cache cleanup in VS code

rchiodo:
VS code uses pointer compression which seems to share the heap with all the worker threads. This means the `getHeapStatistics` function is not returning the correct amount of memory in use for a thread.

This change shares this number with the other threads so they can take total usage into account when deciding whether or not to clean up the heap.

### DIFF
--- a/packages/pyright-internal/src/analyzer/cacheManager.ts
+++ b/packages/pyright-internal/src/analyzer/cacheManager.ts
@@ -8,7 +8,7 @@
  * if memory usage approaches the max heap space.
  */
 
-import { HeapInfo } from 'v8';
+import type { HeapInfo } from 'v8';
 import { Worker } from 'worker_threads';
 import { AnalysisRequest } from '../backgroundAnalysisBase';
 import { ConsoleInterface } from '../common/console';
@@ -55,6 +55,8 @@ export class CacheManager {
         if (msg.requestType === 'cacheUsageBuffer') {
             const index = parseInt(msg.data || '0');
             const buffer = msg.sharedUsageBuffer;
+            // Index of zero is reserved for the main thread so if
+            // the index isn't passed, don't save the shared buffer.
             if (buffer && index) {
                 this._sharedUsageBuffer = buffer;
                 this._sharedUsagePosition = index;
--- a/packages/pyright-internal/src/analyzer/cacheManager.ts
+++ b/packages/pyright-internal/src/analyzer/cacheManager.ts
@@ -8,6 +8,9 @@
  * if memory usage approaches the max heap space.
  */
 
+import { HeapInfo } from 'v8';
+import { Worker } from 'worker_threads';
+import { AnalysisRequest } from '../backgroundAnalysisBase';
 import { ConsoleInterface } from '../common/console';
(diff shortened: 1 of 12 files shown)


packages/pyright-internal/src/analyzer/cacheManager.ts
```diff
@@ -8,6 +8,9 @@
  * if memory usage approaches the max heap space.
  */
 
+import { HeapInfo } from 'v8';
```
erictraut, line 11: Will this break browser support for vscode.dev?

packages/pyright-internal/src/analyzer/cacheManager.ts
```diff
@@ -24,12 +27,41 @@ export interface CacheOwner {
...
+                view[index] = 0;
+            });
+        }
+    }
+
+    handleCachedUsageBufferMessage(msg: AnalysisRequest) {
+        if (msg.requestType === 'cacheUsageBuffer') {
+            const index = parseInt(msg.data || '0');
+            const buffer = msg.sharedUsageBuffer;
+            if (buffer && index) {
```
erictraut, line 58: Isn't 0 a legal index? Or is an index of 0 reserved for the main thread?

This `if` statement will evaluate to falsy if index is 0.

packages/pyright-internal/src/backgroundAnalysis.ts
```diff
@@ -19,18 +19,22 @@ import { ServiceProvider } from './common/serviceProvider';
 import { getRootUri } from './common/uri/uriUtils';
 
 export class BackgroundAnalysis extends BackgroundAnalysisBase {
-    constructor(serviceProvider: ServiceProvider) {
+    constructor(serviceProvider: ServiceProvider, workerIndex = BackgroundThreadIndex) {
```
erictraut, line 22: This seems fragile. If more than one BackgroundAnalysis object is instantiated and the caller forgets to pass in a unique index, the two will now step on each other's data in the shared buffer. Is there a way to make this more robust, like dynamically generating a unique ID or index during construction of the BackgroundAnalysis?

erictraut (approved):
Looks good to me. Just a couple of questions to consider. No need to respond or make changes if you have good answers for them.

packages/pyright-internal/src/analyzer/cacheManager.ts
rchiodo: I don't know yet.  I was going to handle this when we port this to Pylance. If that's the case, I can just import the type. Actually I'll just do that now.

packages/pyright-internal/src/analyzer/cacheManager.ts
rchiodo: 0 is the index reserved for the main thread. I'll add a comment.

packages/pyright-internal/src/backgroundAnalysis.ts
rchiodo: Oh good point. I'll generate it from a static. This is always called on the main thread.

PR #7623, opened 2024-04-05
Changed file content caching behavior. After parsing a file, we no lo…

erictraut:
…nger cache the file (text) contents unless the file is open in the editor. This should reduce memory usage significantly for large projects.

### DIFF
--- a/packages/pyright-internal/src/analyzer/sourceFile.ts
+++ b/packages/pyright-internal/src/analyzer/sourceFile.ts
@@ -96,7 +96,6 @@ class WriteableData {
     parseTreeNeedsCleaning = false;
 
     parserOutput: ParserOutput | undefined;
-    parsedFileContents: string | undefined;
     tokenizerLines: TextRangeCollection<TextRange> | undefined;
     tokenizerOutput: TokenizerOutput | undefined;
 
@@ -408,7 +407,6 @@ export class SourceFile {
         this._writableData.parserOutput = undefined;
         this._writableData.tokenizerLines = undefined;
         this._writableData.tokenizerOutput = undefined;
-        this._writableData.parsedFileContents = undefined;
         this._writableData.moduleSymbolTable = undefined;
         this._writableData.isBindingNeeded = true;
     }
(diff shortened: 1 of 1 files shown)


rchiodo:
Yeah I don't think it actually makes any difference. 

Here's a snapshot with a really large file open.

image

8MB from tokens.

I close the file (which cleans up the tokens and the client document) and I get this:

image

It's still 3MB for parse nodes.

That seems to make this part make sense. The amount of space taken up by strings:

Here's with the large file open:

image


Here's (same folder) with the file closed:

image

Sliced strings seem to make up the difference.

rchiodo:
Let me try the same thing without your change. If my hypothesis is correct, it's total memory should be really close.

rchiodo:
Hmm, I'm not sure. The string data in the old version is charged against the SourceFile object:

image

But if that was all that was holding onto it, then the string size between the two different runs should diff by 3MB, but they don't:

Here's the size of all strings in the old version after closing the large file:

image

It's only 1MB bigger than the one that throws away the `parsedFileContents`.

rchiodo:
Do you think it would be worthwhile to try having parse nodes never store a string (just start/length)?

erictraut:
I wasn't aware of the "(sliced string)" mechanism in V8. It looks like the runtime aggressively creates slices that reference the original string. This is done for many methods on String including `substr`, `substring`, `slice`, and `match`. I guess this makes sense in most cases, but it means that any substring created using these means will hold an implicit reference to the original string. That means our parse tree will inevitably hold a reference to the original text unless we find every place that we call one of these String methods and replace them with some other call that defeats this runtime "optimization".

https://bugs.chromium.org/p/v8/issues/detail?id=2869
https://www.reddit.com/r/javascript/comments/17v3efm/exploring_v8s_strings_implementation_and/

erictraut:
> Do you think it would be worthwhile to try having parse nodes never store a string (just start/length)?

This would be a huge change. It would affect both the parser and the tokenizer. Both of them currently store references to substrings.

rchiodo:
I tried out something that's a really large project. This isn't quite as precise as with closing and opening files, but the differences between the old

image

and this change:
image

show no changes.

rchiodo:
Tokenizer gets cleaned up though (so it shouldn't matter if it references strings)? Parse nodes all have a start/length, could that be used as is? Maybe this isn't the same as the string content though when the node has a string.

erictraut:
> Tokenizer gets cleaned up though (so it shouldn't matter if it references strings)?

Some parse nodes reference tokens, so that's still a problem.

Looking at the token structures and parse node structures, there are relatively few that retain references to strings: string literal tokens, string literal nodes, name nodes, and comments (associated with any node). We could try to defeat V8's slicing behavior in these cases.

erictraut:
@rchiodo, can you tell me what tool you're using to get those snapshots? Is this the chrome:inspect tool in the Chrome or Edge debuggers? Or some other tool?

rchiodo:
> @rchiodo, can you tell me what tool you're using to get those snapshots? Is this the chrome:inspect tool in the Chrome or Edge debuggers? Or some other tool?

Yep, edge://inspect.

I build the vscode-pyright extension, launch VS code with the test code open and then wait on the edge://inspect page for the pyright server.js to show up:

image

You have to add the port we use for debugging to the `Discover network targets` though:

image

packages/pyright-internal/src/analyzer/sourceFile.ts
```diff
@@ -550,17 +548,36 @@ export class SourceFile {
             return undefined;
         }
 
-        assert(this._writableData.parserOutput !== undefined && this._writableData.parsedFileContents !== undefined);
+        assert(this._writableData.parserOutput !== undefined);
```
heejaechang, line 551: we probably need to audit every places `getParseResults` is used. I believe we have many places where if `getParseResults` returned  `a value` once, we assume `getparseResults` will always return a value. now that assumption is not valid since it could return `undefined` next time `getParseResults` is called.

ex)
const parseResults = xxxx.getParseResults();
if (!parseResults) {
return ....
}

after this point,  I believe we have some code that will do `xxx.getParseresults()!` instead of passaround `parseResults` on every calls.

heejaechang: we might add `getParserOutput` for consumers that only care about parse tree but not tokens. 

that said, for it to be useful, we probably need some data from `TokenizerOutput`? I think the memory heavy parts is `// List of all tokens. tokens: TextRangeCollection<Token>;` but other data are cheap to hold onto memory?

I wonder when `Diagnostic mode` is set to `workspace`, how UX will be like? it probably need to re-tokenize file for every diagnostics exist in closed files? not to get `token` but for `lines` and `pyright/type ignores` ?

it might be better to drop only `tokens` but keep `TokenizerOutput` ?

PR #7652, opened 2024-04-09
Skip unnecessary py.typed file exist checks.

heejaechang:
In `importResolver`, it already checks if a `py.typed` file exists before calling `getPyTypedInfo`, using the file system cache, so there's no need to recheck it inside `getPyTypedInfo`. This minor improvement results in a 500~1000ms speed boost for certain cases where the `py.typed` file is checked across many folders, such as in `indexer`. 

It would be beneficial to extend the cache to the file system itself so all code could benefit from it, but for now, the improvement is specific to the `importResolver`, where the cache exists.

image

vs

image

### DIFF
--- a/packages/pyright-internal/src/analyzer/pyTypedUtils.ts
+++ b/packages/pyright-internal/src/analyzer/pyTypedUtils.ts
@@ -36,6 +36,10 @@ export function getPyTypedInfo(fileSystem: FileSystem, dirPath: Uri): PyTypedInf
  * Retrieves information about a py.typed file. The pyTypedPath provided must be a valid path.
  */
 export function getPyTypedInfoForPyTypedFile(fileSystem: FileSystem, pyTypedPath: Uri) {
+    // This function intentionally doesn't check whether the given py.typed path exists or not,
+    // as filesystem access is expensive if done repeatedly.
+    // The caller should verify the file's validity before calling this method and use a cache if possible
+    // to avoid high filesystem access costs.
     let isPartiallyTyped = false;
 
     // Read the contents of the file as text.
--- a/packages/pyright-internal/src/analyzer/pyTypedUtils.ts
+++ b/packages/pyright-internal/src/analyzer/pyTypedUtils.ts
@@ -16,6 +16,9 @@ export interface PyTypedInfo {
     isPartiallyTyped: boolean;
 }
 
+/**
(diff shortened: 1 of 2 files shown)


packages/pyright-internal/src/analyzer/pyTypedUtils.ts
```diff
@@ -21,13 +21,17 @@ export function getPyTypedInfo(fileSystem: FileSystem, dirPath: Uri): PyTypedInf
...
     const pyTypedPath = dirPath.pytypedUri;
-
     if (!fileSystem.existsSync(pyTypedPath) || !isFile(fileSystem, pyTypedPath)) {
         return undefined;
     }
 
+    return getPyTypedInfoUnsafe(fileSystem, pyTypedPath);
+}
+
+export function getPyTypedInfoUnsafe(fileSystem: FileSystem, pyTypedPath: Uri) {
```
erictraut, line 32: Why is this function named "unsafe"? That's normally a big red flag for callers that the function should be used only under certain circumstances. There should at least be a comment here explaining the name and the circumstances under which this function is safe to call.

heejaechang: @erictraut if you prefer other name I can use that other name. any suggestion? I can make it like `getPyTypedInfoWithoutCheckingPyTypePath` or something like that.

erictraut: Choosing a name like this is OK if there's a comment that explains the thinking behind the name. Of course, it's even better if the function's name is sufficiently clear that a comment is no longer needed.

heejaechang: @erictraut can you check again? I added comments and changed name to `getPyTypedInfoForPyTypedFile`

packages/pyright-internal/src/analyzer/pyTypedUtils.ts
```diff
@@ -16,18 +16,32 @@ export interface PyTypedInfo {
     isPartiallyTyped: boolean;
 }
 
+/**
```
erictraut, line 19: Nit: Please be consistent with other function comments used throughout the pyright code base and use `//` comments rather than block comments. Same below.

PR #7654, opened 2024-04-10
Don't use Uri.Empty for default workspace.

heejaechang:
fixes https://github.com/microsoft/pylance-release/issues/5740

It broke due to this change https://github.com/microsoft/pyrx/pull/4878

During this update, we modified how Uri handles case sensitivity, identifying a specific problem where an `EmptyUri` was defaulting to case sensitivity: false. 

However, `EmptyUri` implemented functions like resolvePaths/combinePaths that ended up using incorrect case sensitivity.  To address this, we changed how `EmptyUri` behaves and changed the default workspace not to use `EmptyUri`. However, it appears that we missed its usage in this spot.

### DIFF
--- a/packages/pyright-internal/src/analyzer/service.ts
+++ b/packages/pyright-internal/src/analyzer/service.ts
@@ -518,7 +518,7 @@ export class AnalyzerService {
             ? optionRoot
             : isString(optionRoot)
             ? Uri.file(optionRoot, this.serviceProvider, /* checkRelative */ true)
-            : Uri.empty();
+            : Uri.file('<default workspace root>', this.serviceProvider);
 
         const executionRoot = this.fs.realCasePath(executionRootUri);
         let projectRoot = executionRoot;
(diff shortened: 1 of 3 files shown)


packages/pyright-internal/src/analyzer/service.ts
```diff
@@ -523,7 +523,7 @@ export class AnalyzerService {
             ? optionRoot
             : isString(optionRoot)
             ? Uri.file(optionRoot, this.serviceProvider, /* checkRelative */ true)
-            : Uri.empty();
+            : Uri.file('<default workspace root>', this.serviceProvider);
```
debonte, line 526: Any concerns about `<` and `>` not being legal characters in file paths? At least in Windows file paths.

They are also stream redirection operators for stdin/stdout in some contexts. I can't think of a situation where it would be problematic, but it makes me a little nervous.

How did this work before the URI refactoring?

heejaechang: it used to just `''` string. `'<default workspace root>'` is intentionally random text that won't exist but some operation path related operation with `absolute path` will work with correct case sensitivity.

rchiodo: Won't this root get combined with other file paths? Whereas the empty would never combine?

heejaechang: I updated PR with feedback you gave offline.

rchiodo (approved):

PR #7663, opened 2024-04-10
Pull Pylance with Pyright 1.1.358

debonte:
- Localization update
- Pyright changes in support of "implement abstract class" code action -- https://github.com/microsoft/pylance-release/issues/5362


packages/pyright-internal/src/languageService/completionProvider.ts
```diff
@@ -405,30 +405,6 @@ export class CompletionProvider {
         return this.program.configOptions;
     }
 
-    protected isSimpleDefault(node: ExpressionNode): boolean {
```
bschnurr, line 408: This was moved to `parseTreeUtils.ts`

PR #7744, opened 2024-04-22
Add Installation Instructions for PyCharm

max-muoto:
There exist two Pyright extensions for PyCharm. One simply runs the CLI, and routes error messages into PyCharm, while the later actually uses the LSP API. The former only exists because usage of LSP API is limited to professional edition users. Both come from the same creator.

This adds notes in the documentation to these two plugins, since the lack of a PyCharm plugin initially made it quite difficult for engineers on my team to get effective live type-checking.


erictraut:
Thanks for the contribution.

Issue #7820, opened 2024-05-01
Pyright fails to bind function arguments to type var tuple


LeeeeT:
You've successfully defeated the [bug with the `curry` function](https://github.com/microsoft/pyright/issues/7588) and passed to the next level. On Level 2, your enemy is a sibling of `curry`, the `uncurry` function, which again causes a false positive.

```python
from collections.abc import Callable

def curry[First, *Rest, Result](function: Callable[[First, *Rest], Result]) -> Callable[[*Rest], Callable[[First], Result]]:
    return lambda *rest: lambda first: function(first, *rest)

def uncurry[First, *Rest, Result](function: Callable[[*Rest], Callable[[First], Result]]) -> Callable[[First, *Rest], Result]:
    return lambda first, *rest: function(*rest)(first)

@curry
@curry
def f(a: str, b: int, c: bool) -> None:
    pass

reveal_type(f)  # (bool) -> ((int) -> ((str) -> None))

uncurried = uncurry(uncurry(f))
```

```
file.py:16:21 - error: Argument of type "(int, *int) -> ((str) -> None)" cannot be assigned to parameter "function" of type "(*Rest@uncurry) -> ((First@uncurry) -> Result@uncurry)" in function "uncurry"
  Type "(int, *int) -> ((str) -> None)" is incompatible with type "(*Union[*Rest@uncurry], *Union[*Rest@uncurry]) -> ((str) -> None)"
file.py:16:29 - error: Argument of type "(bool) -> ((int) -> ((str) -> None))" cannot be assigned to parameter "function" of type "(*Rest(1)@uncurry) -> ((First(1)@uncurry) -> Result(1)@uncurry)" in function "uncurry"
  Type "(bool) -> ((int) -> ((str) -> None))" is incompatible with type "(*Union[*Rest@uncurry]) -> ((Union[*Rest@uncurry]) -> ((str) -> None))"
    Function return type "(int) -> ((str) -> None)" is incompatible with type "(Union[*Rest@uncurry]) -> ((str) -> None)"
      Type "(int) -> ((str) -> None)" is incompatible with type "(Union[*Rest@uncurry]) -> ((str) -> None)"
```

It, again, works without error if you apply `uncurry` once at a time like this:

```python
applied_once = uncurry(f)
applied_twice = uncurry(applied_once)

reveal_type(applied_once)  # (int, bool) -> ((str) -> None)
reveal_type(applied_twice)  # (str, int, bool) -> None
```

[...]

PR #7917, opened 2024-05-14
Fixed bug that resulted in false positive when evaluating a function …

erictraut:
…call involving a TypeVarTuple under certain circumstances. This addresses #7820.


erictraut:
@dbratti, just a heads up that this bug fix in pyright will require you to make a small change to your code. In particular, the implementation of function `starcompose` in the file `core/compose.py` will need to be changed from `Callable[[Any], Any]` to `Callable[..., Any]` (both for the `fns` parameter and the return type). The type `Any` is not compatible with an unpacked TypeVarTuple, but pyright previously didn't flag that as an error.

PR #7956, opened 2024-05-20
Update GitHub Actions from "actions" org to resolve Node.js deprecations

boidolr:
`github/codeql-action` needs to be updated as well, but might need some configuration changes.

PR #7978, opened 2024-05-22
Update GitHub Actions from "actions" org & update mypy_primer for changed upload-artifacts behaviour

boidolr:
Follow up to https://github.com/microsoft/pyright/pull/7956, this time handling upload-artifacts separately.
With updated version of actions/upload-artifacts uploads to the same name are no longer possible. Therefore the changes to the mypy_primer workflows are necessary. Handling of the migration is described in the document belonging to upload-artifacts [MIGRATION.md](https://github.com/actions/upload-artifact/blob/main/docs%2FMIGRATION.md).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -46,7 +46,7 @@ jobs:
           npm run package
           mv pyright-*.vsix ${{ env.VSIX_NAME }}
 
-      - uses: actions/upload-artifact@v3
+      - uses: actions/upload-artifact@v4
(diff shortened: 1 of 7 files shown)


erictraut:
Thanks for the update.

Out of curiosity, why do you care about pyright's CI scripts?

boidolr:
I care about stuff working, and things like this are an easy thing to contribute. So not particularly bound to the ci scripts.

PR #7982, opened 2024-05-23
Use typeshed version info for stdlib module list

### LINKED ISSUE #7981: Version-aware `reportShadowedImports`
**Is your feature request related to a problem? Please describe.**

With `reportShadowedImports` enabled, Pyright gives me the following diagnostic when checking a file called `parser.py`:
```
"/PATH/TO/parser.py" is overriding the stdlib module "parser" (reportShadowedImports)
```

However, `parser` was removed in Python 3.10, and I'm using 3.12:
```js
// console.log(execEnv.pythonVersion);
PythonVersion {
  _major: 3,
  _minor: 12,
  _micro: 0,
  _releaseLevel: 'final',
  _serial: 0
}
```

**Describe the solution you’d like**

Only emit a diagnostic if the "shadowed" module actually exists in the current Python version. This would make `reportShadowedImports` much more useful. And since we already store version info for stdlib modules in typeshed, it should be an easy fix!

ujan-r:
Fixes #7981

### DIFF
--- a/packages/pyright-internal/src/analyzer/importResolver.ts
+++ b/packages/pyright-internal/src/analyzer/importResolver.ts
@@ -338,7 +338,7 @@ export class ImportResolver {
 
     isStdlibModule(module: ImportedModuleDescriptor, execEnv: ExecutionEnvironment): boolean {
         if (!this._stdlibModules) {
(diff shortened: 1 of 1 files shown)


ujan-r:

rchiodo (approved):

PR #8032, opened 2024-05-31
Add the remaining builtin comparison methods

YoungMind1:


erictraut:
Please do not modify the typeshed stubs in the pyright repo. These stubs are copied from the typeshed project periodically, and the source of truth lives here: https://github.com/python/typeshed.

YoungMind1:
hah! interesting. Thanks for the explanation

PR #8037, opened 2024-06-01
Fix typo in type-stubs.md

jrheard:
Another edited version that could work would be '[...] file, all of the symbols imported from these modules will be of type “Unknown”', but I figured you folks would probably prefer the "Pyright will treat" version. Doesn't matter much to me either way, obviously feel free to go with whatever wording you like - I just noticed that the current wording seemed like a mistake and wanted to suggest a fix!


erictraut:
Thanks for the contribution.

rpreslar4765 (approved):

PR #8103, opened 2024-06-09
Unify behavior between `pyrightconfig.json` and `pyproject.toml`

### LINKED ISSUE #8102: Different behavior between `pyrightconfig.json` and `pyproject.toml`
**Describe the bug**
Basically, everything works fine if I use `pyrightconfig.json`, and run `pyright` with no arguments
```json
{
    "ignore": [
        "test/*"
    ],
}
```
But If I do the same with I do the same with `pyproject.toml` the analysis takes much longer.
```toml
[tool.pyright]
ignore = ["test/*"]

[tool.pytest.ini_options]
log_cli = true
log_cli_level = "INFO"
addopts = "--capture=no --cov-report html --cov-report json --cov=."

[tool.coverage.run]
branch = true

[tool.coverage.report]
exclude_also = [
    "if __name__ == .__main__.:",
    "if TYPE_CHECKING:",
    "raise NotImplementedError",
    "@(abc\\.)?abstractmethod",
]
```

**Code or Screenshots**
Running the command with verbose: `pyright --verbose` reveals something interesting
First with `pyrightconfig.json`, pyright auto-excludes `<project>\venv` and finds 68 source files.
```
Loading configuration file at <project>\pyrightconfig.json
Assuming Python version 3.11.9.final.0
No include entries specified; assuming <project>
Auto-excluding **/node_modules
Auto-excluding **/__pycache__
Auto-excluding **/.*
Search paths for file:///<project>
  c:\Users\Auguste\.cache\pyright-python\1.1.366\node_modules\pyright\dist\typeshed-fallback\stdlib
  <project>
  <project>\typings
  c:\Users\Auguste\.cache\pyright-python\1.1.366\node_modules\pyright\dist\typeshed-fallback\stubs\...
  c:\Users\Auguste\AppData\Local\Programs\Python\Python311\DLLs
  c:\Users\Auguste\AppData\Local\Programs\Python\Python311\Lib
  c:\Users\Auguste\AppData\Local\Programs\Python\Python311
  <project>\venv
  <project>\venv\Lib\site-packages
  <project>\venv\Lib\site-packages\win32
  <project>\venv\Lib\site-packages\win32\lib
  <project>\venv\Lib\site-packages\pythonwin
Auto-excluding <project>\venv
Found 68 source files
...
...
...
```
But with `pyproject.toml`, pyright does not exclude venv and finds 4741.
```
Loading pyproject.toml file at <project>\pyproject.toml
Assuming Python version 3.11.9.final.0
Search paths for file:///<project>
  c:\Users\Auguste\.cache\pyright-python\1.1.366\node_modules\pyright\dist\typeshed-fallback\stdlib
  <project>
  <project>\typings
  c:\Users\Auguste\.cache\pyright-python\1.1.366\node_modules\pyright\dist\typeshed-fallback\stubs\...
  c:\Users\Auguste\AppData\Local\Programs\Python\Python311\DLLs
  c:\Users\Auguste\AppData\Local\Programs\Python\Python311\Lib
  c:\Users\Auguste\AppData\Local\Programs\Python\Python311
  <project>\venv
  <project>\venv\Lib\site-packages
  <project>\venv\Lib\site-packages\win32
  <project>\venv\Lib\site-packages\win32\lib
  <project>\venv\Lib\site-packages\pythonwin
Found 4741 source files
...
...
...
```

So the question is, why the different behavior?

**command-line**
`pyright`

augustelalande:
Resolved #8102

Basically, there is a check here
https://github.com/microsoft/pyright/blob/238f70049f47dccd16d06bc0fd795a189cf0ae8d/packages/pyright-internal/src/analyzer/service.ts#L646-L660
which initializes include and exclude if no `pyrightconfig.json` is found. Note it does not check for `pyproject.toml`.

Later if neither config defined an include or exclude they also get initialized, and `autoExcludeVenv` gets set.
https://github.com/microsoft/pyright/blob/238f70049f47dccd16d06bc0fd795a189cf0ae8d/packages/pyright-internal/src/analyzer/service.ts#L689-L698

However, in the case of `pyproject.toml`, because exclude has already been set, `autoExcludeVenv` does not get set, leading to a difference in behavior between `pyrightconfig.json` and `pyproject.toml`.

This adds an extra check for `pyproject.toml` to avoid this.

### DIFF
--- a/packages/pyright-internal/src/analyzer/service.ts
+++ b/packages/pyright-internal/src/analyzer/service.ts
@@ -643,7 +643,7 @@ export class AnalyzerService {
             });
         }
 
-        if (!configFilePath && commandLineOptions.executionRoot) {
+        if (!(configFilePath || pyprojectFilePath) && commandLineOptions.executionRoot) {
             if (commandLineOptions.includeFileSpecs.length === 0) {
                 // If no config file was found and there are no explicit include
                 // paths specified, assume the caller wants to include all source
--- a/packages/pyright-internal/src/analyzer/service.ts
+++ b/packages/pyright-internal/src/analyzer/service.ts
@@ -643,7 +643,9 @@ export class AnalyzerService {
             });
         }
 
-        if (!(configFilePath || pyprojectFilePath) && commandLineOptions.executionRoot) {
+        const configs = this._getExtendedConfigurations(configFilePath ?? pyprojectFilePath);
+
+        if (!configs && commandLineOptions.executionRoot) {
             if (commandLineOptions.includeFileSpecs.length === 0) {
                 // If no config file was found and there are no explicit include
                 // paths specified, assume the caller wants to include all source
@@ -662,8 +664,6 @@ export class AnalyzerService {
         configOptions.disableTaggedHints = !!commandLineOptions.disableTaggedHints;
 
         configOptions.initializeTypeCheckingMode(commandLineOptions.typeCheckingMode ?? 'standard');
-
-        const configs = this._getExtendedConfigurations(configFilePath ?? pyprojectFilePath);
 
         if (configs) {
             for (const config of configs) {
--- a/packages/pyright-internal/src/common/configOptions.ts
+++ b/packages/pyright-internal/src/common/configOptions.ts
@@ -1128,7 +1128,7 @@ export class ConfigOptions {
         // Read the "include" entry.
         if (configObj.include !== undefined) {
             if (!Array.isArray(configObj.include)) {
-                console.error(`Config "include" entry must must contain an array.`);
+                console.error(`Config "include" entry must contain an array.`);
             } else {
                 this.include = [];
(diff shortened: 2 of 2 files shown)


augustelalande:

erictraut:
Thanks for the bug report and the PR. I'm going to close the PR because I'd like to fix the problem in a more surgical manner than you did in this PR.

augustelalande:
@erictraut FYI I initially implement the fix the same way as you did here https://github.com/microsoft/pyright/commit/4319e03b7114d9aa7f6dedd80eca16ee263cdf18 but I think this may cause issues if a `pyproject.toml` is present but does not contain a pyright section.

erictraut:
Ah, good catch. Then I think a better solution is to restructure the code so we fill in default include and exclude values only after loading any config files — and only after we confirm that the config files didn't provide any explicit include or exclude values.

PR #8118, opened 2024-06-11
Discard `TypeEvaluator` if cancellation leaves partially-constructed function type in cache

debonte:
This is the function equivalent of the class cancellation fix in https://github.com/microsoft/pyright/pull/5786.

If `getTypeOfFunctionDecorated` or `getTypeOfLambdaWithExpectedType` are cancelled before they complete, which would leave a partially evaluated function type in the type cache, we now set `isTypeCacheInvalid` to `true` on the `OperationCanceledException`. This will cause `Program._runEvaluatorWithCancellationToken` to discard the `TypeEvaluator` and create a new instance.

I believe this will address https://github.com/microsoft/pylance-release/issues/4295.

### DIFF
--- a/packages/pyright-internal/src/analyzer/typeEvaluator.ts
+++ b/packages/pyright-internal/src/analyzer/typeEvaluator.ts
@@ -14510,8 +14510,8 @@ export function createTypeEvaluator(
             return { type: functionType, isIncomplete, typeErrors };
         } catch (e) {
             if (OperationCanceledException.is(e)) {
-                // If the work was canceled before the class types were updated, the
-                // class type in the type cache is in an invalid, partially-constructed state.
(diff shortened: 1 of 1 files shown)


erictraut (approved):

erictraut:
You hit the same MacOS test failure that I've been seeing sporadically. It tends to go away if you run the test again, but that's unnerving. I've never been able to repro it locally on my Mac. I've reviewed the test code to try to determine potential causes, and I'm not able to find the problem. It could be related to a timeout in the test framework. I somewhat suspect that it's related to my other recent change that I asked you to code review. This error started at about the same time.

erictraut:
@debonte, is the PR ready from your standpoint? If so, let's merge it so we can get it into today's pyright release.

PR #8200, opened 2024-06-21
add an azure devops build

bschnurr:
new build pipeline for azure devops


AdamYoblick (approved):

build/azure-pipelines.yml
debonte, line 1: Maybe we should put this in an `azuredevops` subdirectory within `build` to match the pyrx location?

bschnurr: done

PR #8186, opened 2024-06-20
Docs: fix broken markdown link in config.md

jvacek:

PR #8201, opened 2024-06-21
fix: usage of `deprecated` version of `Node.js`

### LINKED ISSUE #8199: `validation.yml` uses `actions/cache@v3`, which uses a `deprecated` version of `Node.js`

hamirmahal:
fixes #8199.

### DIFF
--- a/.github/workflows/validation.yml
+++ b/.github/workflows/validation.yml
@@ -30,7 +30,7 @@ jobs:
         shell: bash
         run: |
           echo "dir=$(npm config get cache)" >> $GITHUB_OUTPUT
-      - uses: actions/cache@v3
+      - uses: actions/cache@v4
         with:
           path: ${{ steps.npm-cache.outputs.dir }}
           key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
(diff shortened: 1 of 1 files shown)


hamirmahal (commented):
`actions/cache@v4` is the latest major version at the time of this writing.

debonte (approved):
Thanks @hamirmahal!

hamirmahal:
You're welcome @debonte!

debonte:
@erictraut, is there something wrong with the CLA thing? I see that https://github.com/microsoft/pyright/pull/8186 didn't run the CLA check either.

erictraut:
@debonte, I don't know about the license/cla action. It was added automatically by the team at Microsoft that oversees OSS projects. They may have some insights.

erictraut:
This PR looks OK to me, but I'm not sure what problem this is fixing. I've never seen an issue with our CI scripts relating to the v3 version of this action. The phrase "if it's ain't broken, don't fix it" comes to mind. Is the concern that it will stop working at some point in the future? @hamirmahal, can you explain why you are proposing that we make this change?

debonte:
We update these regularly in pyrx for what it's worth. I'd be inclined to take it, if only to reduce the warning noise on the action summary pages so we can see real issues more easily.

hamirmahal:
@erictraut Like Erik mentioned, updating to `Node 20` will reduce noise and make actual issues more noticeable.

You can sometimes get a speed improvement for free by updating `Node`, and even if GitHub Actions are free for open source repositories, there's still the environmental impact of runners unnecessarily taking a few seconds longer to run by being on an older, slower version of `Node`. If there is a speed improvement, developers also benefit by getting feedback on their changes more quickly.

[GitHub recommends Actions users use the latest versions](https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/). [Node 16 has reached its end of life](https://github.com/nodejs/Release/#end-of-life-releases).

PR #8259, opened 2024-06-28
Explore the removal of `Sequence[str]` as a base class to `str`.

erictraut:


erictraut:
This relates to:
https://github.com/python/typing/issues/256
https://github.com/python/mypy/issues/11001
https://github.com/microsoft/pyright/discussions/8222

PR #8293, opened 2024-07-02
inital pyright azure devops release pipeline

bschnurr:
New Publishing workflow 

- Build get tagged triggers this pipeline
- the pipeline creates a signed Vsix and publish as draft to github releases page with initial rough version of release notes
- Eric converts github draft release to published
- someone on the pylance team approves the pipeline run and it continues to publish to marketplace

Alternative:
I can make it auto publish to marketplace after 2hrs or a set amount of time. If Eric doesn't convert the release draft to published, in github, then the marketplace step should fail.  The benefit being the process is not blocked by Pylance, for instance if it was a weekend or holiday


build/azuredevops/azure-pipelines-release.yml
```diff
@@ -0,0 +1,144 @@
...
+                  gitHubConnection: 'Github-Pylance' # The name of the GitHub service connection
+                  repositoryName: 'microsoft/pyright' # The name of your GitHub repository
+                  action: 'create'
+                  isDraft: true
+                  isPreRelease: false
+                  addChangeLog: true
+                  title: 'Published $(Build.SourceBranchName)'
+                  assets: |
+                    $(Pipeline.Workspace)/$(ARTIFACT_NAME_VSIX)/$(VSIX_NAME)
+      - stage: WaitForValidation
```
debonte, line 92: Will Eric have access to the AzDO page to tell the release to proceed? I'm not sure if he currently does any manual testing of the VSIX before publishing to Marketplace. He may prefer to skip the manual validation step and go straight to publishing.

bschnurr: No he wont have access.. it will be up to us for final button click

debonte: Ok, is that an intentional process change for security reasons?

bschnurr: not intentional... i just dont think he has azure devops access. Maybe there is a way to grant it

debonte: Ok. So I'm wondering if removing the manual validation step would be simpler and closer to what he's used to.

bschnurr: i think i need to leave it though.  it can be changed to 'pass' when the timeout expires... Eric needs time to change the draft release to publish and edit the notes.  In azure devops there is no build trigger or github published.

debonte: I see. Could the GitHub release be created as a non-draft? For what it's worth, the GitHub release text can be changed after the release is published.

On the other hand, maybe publishing to marketplace just because a tag was pushed is too risky and a manual step somewhere along the line would be best.

build/azuredevops/azure-pipelines-release.yml
```diff
@@ -0,0 +1,143 @@
...
+                  notifyUsers: 'plseng@microsoft.com'
+                  instructions: 'Please test the latest draft release of Pyright.'
+                  onTimeout: 'reject'
+
+      - stage: PublishExtension
+        dependsOn:
+          - WaitForValidation
+        jobs:
+          - job: publish_extension
+            displayName: Publish extension to marketplace
```
debonte, line 112: Are you planning to eventually automate publishing to npm also?

bschnurr: when Eric mentioned that today. i was going to take a look.

Neevan960:
hi

PR #8295, opened 2024-07-02
Fix severity overrides being skipped when there's a pyrightconfig.json

rchiodo:
In these two issues, the severity override being set in the user's settings.json is not being applied:

https://github.com/microsoft/pylance-release/issues/6087
https://github.com/microsoft/pylance-release/issues/6085

The reason for this was the commandline settings were never being applied when there's a pyrightconfig.json. I modified the logic to apply the command line options first and then let the pyrightconfig.json files to override those values.

I also found an issue in the test code that prevented the LSP tests from cleaning up server instances.

### DIFF
--- a/packages/pyright-internal/src/analyzer/service.ts
+++ b/packages/pyright-internal/src/analyzer/service.ts
@@ -640,6 +640,7 @@ export class AnalyzerService {
         configOptions.disableTaggedHints = !!commandLineOptions.disableTaggedHints;
 
         configOptions.initializeTypeCheckingMode(commandLineOptions.typeCheckingMode ?? 'standard');
+        configOptions.applyDiagnosticOverrides(commandLineOptions.diagnosticSeverityOverrides);
 
         const configs = this._getExtendedConfigurations(configFilePath ?? pyprojectFilePath);
 
@@ -653,8 +654,6 @@ export class AnalyzerService {
                     commandLineOptions
                 );
             }
-        } else {
-            configOptions.applyDiagnosticOverrides(commandLineOptions.diagnosticSeverityOverrides);
         }
 
         // If no include paths were provided, assume that all files within
--- a/packages/pyright-internal/src/tests/lsp/languageServer.ts
+++ b/packages/pyright-internal/src/tests/lsp/languageServer.ts
@@ -135,7 +135,8 @@ class TestServer extends PyrightServer {
     constructor(
         connection: Connection,
         fs: FileSystem,
-        private readonly _supportsBackgroundAnalysis: boolean | undefined
+        private readonly _supportsBackgroundAnalysis: boolean | undefined,
+        private readonly _name: string
     ) {
         super(connection, _supportsBackgroundAnalysis ? 1 : 0, fs);
     }
@@ -190,7 +191,7 @@ async function runServer(
         // Create a host so we can control the file system for the PyrightServer.
         const disposables: Disposable[] = [];
         const host = createTestHost(testServerData);
-        const server = new TestServer(connection, host.fs, testServerData.backgroundAnalysis);
+        const server = new TestServer(connection, host.fs, testServerData.backgroundAnalysis, testServerData.testName);
 
         // Listen for the test messages from the client. These messages
         // are how the test code queries the state of the server.
@@ -325,7 +326,7 @@ class ServerStateManager {
             if (serverIndex >= 0) {
                 try {
                     instance.disposables[serverIndex].dispose();
-                    instance.disposables = instance.disposables.splice(serverIndex, 1);
+                    instance.disposables.splice(serverIndex, 1);
                 } catch (e) {
                     // Dispose failures don't matter.
                 }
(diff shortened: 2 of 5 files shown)


packages/pyright-internal/src/analyzer/service.ts
```diff
@@ -640,6 +640,7 @@ export class AnalyzerService {
         configOptions.disableTaggedHints = !!commandLineOptions.disableTaggedHints;
 
         configOptions.initializeTypeCheckingMode(commandLineOptions.typeCheckingMode ?? 'standard');
+        configOptions.applyDiagnosticOverrides(commandLineOptions.diagnosticSeverityOverrides);
```
rchiodo, line 643: This is the fix here. Apply diagnostic overrides from the command line options first, then apply the json from the pyrightconfig.json files.

packages/pyright-internal/src/tests/lsp/languageServer.ts
```diff
@@ -325,7 +326,7 @@ class ServerStateManager {
             if (serverIndex >= 0) {
                 try {
                     instance.disposables[serverIndex].dispose();
-                    instance.disposables = instance.disposables.splice(serverIndex, 1);
+                    instance.disposables.splice(serverIndex, 1);
```
rchiodo, line 329: This was a bug in the cleanup. `splice` doesn't return the modified array, it modifies it inline and returns the deleted items. This bug was preventing the connection from being disposed, meaning all test servers were still listening to messages.

packages/pyright-internal/src/tests/lsp/languageServerTestUtils.ts
```diff
@@ -211,7 +211,7 @@ export function createFileSystem(projectRoot: string, testData: FourSlashData, o
         mountedPaths.set(bundledStubsFolder, bundledStubsFolderPathTestServer);
     }
 
-    const vfsInfo = createVfsInfoFromFourSlashData(projectRoot, testData);
+    const vfsInfo = createVfsInfoFromFourSlashData(projectRoot, testData, /* ignoreRawConfigJson */ true);
```
rchiodo, line 214: This change makes it possible to read the pyrightconfig.json as a file instead of having it be part of the vfsInfo.

packages/pyright-internal/src/analyzer/service.ts
erictraut: Hmm, is this the behavior we want? Shouldn't a config file always override all of the diagnostic severity settings (both defaults and specified)? In other words, if there is a config file, the diagnostic severities provided by the language server settings should be ignored, right? Otherwise, developers working on a project will see different behaviors based on their own local settings.

Let's discuss when we meet later today.

rchiodo: >  In other words, if there is a config file, the diagnostic severities provided by the language server settings should be ignored, right?

This wasn't the case before. They were combined before AFAIK

rchiodo: Yeah in the old code:

- pyrightconfig has an explicit value = this value is used regardless of settings.json
- pyrightconfig doesn't have an explicit value = value in settings.json is used
- settings.json doesn't have an explicit value = default value for typechecking mode is used

erictraut: It's even more complicated than what's listed above because typeCheckingMode can be specified in both places.

Let's discuss what behavior we want. I'm not convinced the old behavior is desirable. The current behavior (what's checked in) is more IMO defensible and consistent with what devs will typically desire.

Is there a precedent set by TypeScript here?

rchiodo: I can't find similar settings in typescript. Tsconfig.json doesn't have settings that can be overridden in the VS code settings.json file. 

Maybe command line arguments? Is that the equivalent of settings.json?

erictraut: I don't think command-line arguments are equivalent to settings.json. Command-line arguments should always take precedence whereas settings.json (especially when it's the user settings.json), should not override project-specific settings.

rchiodo: Hmm, command line doesn't seem the equivalent. For tsc it trumps everything.

I have a tsconfig.json like so:

```json5
{
    "compilerOptions": {
        "allowUnreachableCode": false,
    }
}
```
And then some code like so:

```typescript
function fn(n: number) {
  if (n > 5) {
    return true;
  } else {
    return false;
  }
  return true;
}
```

Running `tsc`, it outputs an error on the last line. 

If I run `tsc --allowUnreachableCode`, it does not output an error.

rchiodo:
After some discussion we decided to:

- typeCheckingMode and diagnosticSeverityOverrides should not be settable in settings.json if a pyrightconfig.json or a pyproject.toml with a [tool.pyright] section was present. They're like compiler flags and would be weird to override them in a user file.
- However if pyrightconfig.json or pyproject.toml are not present, the current VS code behavior should continue to work. The user can set these values in their settings.json

We're going to try this out and see how many people actually have this situation - where they're setting these values and they also have a pyrightconfig.json

PR #8371, opened 2024-07-11
Modified parse node definitions so they are all "monomorphic" (the sa…

erictraut:
…me fields with the same layout). This helps the V8 engine optimize performance.

### DIFF
--- a/packages/pyright-internal/src/analyzer/analyzerNodeInfo.ts
+++ b/packages/pyright-internal/src/analyzer/analyzerNodeInfo.ts
@@ -81,65 +81,67 @@ export type ScopedNode = ModuleNode | ClassNode | FunctionNode | LambdaNode | Co
 // Cleans out all fields that are added by the analyzer phases
 // (after the post-parse walker).
 export function cleanNodeAnalysisInfo(node: ParseNode) {
-    const analyzerNode = node as AnalyzerNodeInfo;
-    delete analyzerNode.scope;
-    delete analyzerNode.declaration;
-    delete analyzerNode.flowNode;
-    delete analyzerNode.afterFlowNode;
-    delete analyzerNode.fileInfo;
-    delete analyzerNode.codeFlowExpressions;
-    delete analyzerNode.codeFlowComplexity;
-    delete analyzerNode.dunderAllInfo;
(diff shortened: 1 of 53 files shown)


DanielRosenwasser:
You mentioned elsewhere you weren't seeing quite the same speed improvement results as in https://github.com/microsoft/TypeScript/pull/58928. One thing to note is that @dragomirtitian actually made three broader changes in that PR, and managed to untangle them into 3 separate PRs:

- Monomorphic signature objects: https://github.com/microsoft/TypeScript/pull/59192
- Monomorphic type objects: https://github.com/microsoft/TypeScript/pull/59191
- Monomorphic syntax nodes: https://github.com/microsoft/TypeScript/pull/59190

According to https://github.com/microsoft/TypeScript/pull/58928#issuecomment-2216956874, the win in monomorphizing types was the best one.

From this PR, it looks like you mostly changed syntax nodes. If you're seeing a \~5% improvement on average that probably checks out, though there might be other factors at play here too.

erictraut:
@DanielRosenwasser, thanks for the additional info. I didn't realize that the changes in typescript went beyond syntax nodes. Yes, I suspect that making the type objects monomorphic will have a bigger impact than the parse nodes. It's a heavier lift to make that change, but it could be worth it.

packages/pyright-internal/src/analyzer/analyzerNodeInfo.ts
```diff
@@ -81,108 +81,110 @@ export type ScopedNode = ModuleNode | ClassNode | FunctionNode | LambdaNode | Co
...
+    if (analyzerNode) {
+        delete analyzerNode.scope;
+        delete analyzerNode.declaration;
+        delete analyzerNode.flowNode;
+        delete analyzerNode.afterFlowNode;
+        delete analyzerNode.fileInfo;
+        delete analyzerNode.codeFlowExpressions;
+        delete analyzerNode.codeFlowComplexity;
+        delete analyzerNode.dunderAllInfo;
+        delete analyzerNode.typeParameterSymbol;
```
dmichon-msft, line 94: This categorically deoptimizes the internal maps. If possible set them to `undefined` instead.

PR #8539, opened 2024-07-25
Update unreachableCode message to include "for type checking"

luabud:
We talked about https://github.com/microsoft/pylance-release/issues/5647 in the Pylance team and we thought adjusting the "unreachable code" message would help avoid confusion, since it makes it more clear it's about type checking and not runtime.


erictraut:
I've been thinking about the user feedback on this issue. I think that rewording the message is an OK option, but it doesn't get to the heart of the feedback. I have another idea I'd like to explore. Please hold off on merging this change until I have a chance to prototype the alternative idea I have in mind.

PR #8540, opened 2024-07-25
push pylance to pyright 2024.8

heejaechang:
Here's the list of changes to include:

- add `system memory info` in `mem stats`
- Fix early break in loop in `ImportResolver._getModuleNameForImport`
- invalidate imports when they are actually changed.
- rename workspace name if it is changed after workspace is created
- some refactoring on `find configuration file`
- some other misc changes.


packages/pyright-internal/src/analyzer/typeUtils.ts
```diff
@@ -448,6 +448,15 @@ export interface MapSubtypesOptions {
     retainTypeAlias?: boolean;
 }
 
+export interface MapSubtypesOptions {
```
debonte, line 451: This type already exists on line 436 just above here.

heejaechang: oh. I thought I fixed those.

packages/pyright-internal/src/analyzer/serviceUtils.ts
debonte, line 1: Looks like this is part of https://github.com/microsoft/pyrx/pull/5431/files which shouldn't be merged into Pyright?

packages/pyright-internal/src/analyzer/typeUtils.ts
heejaechang: ah, this is another merge conflict issues.

rchiodo (approved):

packages/pyright-internal/src/analyzer/serviceUtils.ts
heejaechang: @rchiodo ?

rchiodo: It looks correct to me. We need it actually or pyright won't build. It was a refactor of moving those functions out of service.ts

debonte (approved):

Issue #2672, opened 2021-12-07
Add reportXXX rule for unreachable code


smackesey:
Pyright flags unreachable code with a `hint` diagnostic. I would like to override this severity using `python.analysis.diagnosticSeverityOverrides`. However, I don't see a corresponding `reportXXX` rule for unreachable code that I see for other diagnostic categories. Can this be added?

erictraut:
This has been discussed before. The tagged hint isn't meant to be displayed like a diagnostic. A few clients incorrectly display it as such. A client can specify whether it supports this particular tagged hint. If it does, it should display them appropriately (as a subtle "dimmed" text). Some clients request this tagged hint and then display it in manner that makes it appear as a normal diagnostic. You should talk to the maintainer of your client and ask them to add a configuration option to disable this tagged hint.

smackesey:
Respectfully, the reason this has been discussed many times before (I have seen it in other issues) is because your stance here is confusing for me and other users who think of PyRight as a language server supporting Language Server Protocol. You seem to be insisting on a definition of "diagnostic" that is different than LSP's definition, despite the fact that PyRight uses LSP. I am sure that issues around hint diagnostics will continue to crop up in other contexts so long as you maintain this stance.

[Diagnostics](https://microsoft.github.io/language-server-protocol/specification#diagnostic) are a specific concept in Language Server Protocol. They can have 1 of 4 severities: `hint`, `info`, `warning`, `error`. The job of the language server is to simply send diagnostics, which Pyright does and does well. Among the many kinds of diagnostics PyRight sends is an `"Code is unreachable"` diagnostic with severity of `hint`.

It is strange to say "the tagged hint isn't meant to be displayed as a diagnostic", because it is literally communicated to the editor a diagnostic. A generic editor-embedded LSP client will simply display it as such, using whatever behavior has been configured to display hint diagnostics. Further, from the perspective of LSP, there is no "correct" behavior for showing hint or any other kind of diagnostic-- that's up to the client. Just because some editors (VSCode) display "hint" diagnostics categorically differently than error/warning diagnostics, doesn't mean every editor needs to do so. Further, there is nowhere in LSP where the server specifies how particular diagnostics are to be displayed. Am I wrong here? Or does PyRight communicate this styling info through some nonstandard structure? If PyRight doesn't communicate this styling info, then on what grounds can you say editors are behaving "incorrectly" when they display the communicated `hint` diagnostic like any other `hint` diagnostic from any other server?

[...]

PR #8542, opened 2024-07-25
Added a new configuration option `identifyUnreachableCode`

erictraut:
Added a new configuration option `identifyUnreachableCode`. It is off by default when `typeCheckingMode` is "off" but otherwise on by default. When disabled, it causes pyright not to identify code blocks that are determined to be unreachable via type analysis. Code blocks that are determined to be unreachable via non-type information are still displayed as such.

This builds on an idea from Heejae, who suggested that we distinguish between code that is unreachable based on "syntax" versus "type analysis". Those aren't quite the right designations, but I think the idea is on the right track. I've added some [new documentation](https://microsoft.github.io/pyright/#/type-concepts-advanced?id=reachability) that explains the difference between unreachable code based on "non-type" vs "type" information. Examples are provided.

I've updated pyright's code flow engine to distinguish between these two types of unreachability. With that in place, we can now provide additional user control over unreachable code based on type analysis.

Here's what it looks like with "typeCheckingMode" set to "off". Note that some unreachable code blocks (those that are based on non-type analysis) are still displayed as "grayed out".

image

And what it looks like with "typeCheckingMode" set to "basic" or above. Note that additional code blocks are now displayed as grayed out.

image

@luabud, @heejaechang, @debonte, @rchiodo, let me know what you think of this approach


docs/configuration.md
```diff
@@ -59,6 +59,8 @@ Relative paths specified within the config file are relative to the config file
 
 - <a name="deprecateTypingAliases"></a> **deprecateTypingAliases** [boolean]: PEP 585 indicates that aliases to types in standard collections that were introduced solely to support generics are deprecated as of Python 3.9. This switch controls whether these are treated as deprecated. This applies only when pythonVersion is 3.9 or newer. The default value for this setting is `false` but may be switched to `true` in the future.
 
+- <a name="identifyUnreachableCode"></a> **identifyUnreachableCode** [boolean]: If enabled, code that is determined to be unreachable by type analysis is reported using a tagged hint. This setting does not affect code that is determined to be unreachable regardless of type analysis. such code is always reported as such. This setting also has no effect when when using the command-line version of pyright because it never emits tagged hints for unreachable code.
```
rchiodo, line 62: I find the name confusing. This really means unreachable by static type analysis, right? I think it needs some way to distinguish this from unreachable without type analysis.

`identifyTypeAnalysisUnreachableCode`? Maybe that's too long.

rchiodo: I was also thinking there should be a different error message for this class of errors.

erictraut: > Maybe that's too long.

Yeah, I think "identifyTypeAnalysisUnreachableCode" is too long and cumbersome. IMO, the name "identifyUnreachableCode " plus the documentation make it clear enough.

> I was also thinking there should be a different error message for this class of errors.

These are never reported as errors, and they never show up in the "problems" section. Or are you talking about the "hint" message that appears when you hover over the gray text? Yeah, I think it makes sense to use different wording for the two cases. We could use a variant of the wording from @luabud's PR.

rchiodo: Yeah the message that shows up in VS code. The warning text. Hopefully that being different will alleviate some of the concerns we've seen lately.

debonte, line 62: > regardless of type analysis. such code is always reported as such.

There's either a punctution or capitalization issue here. Also overuse of the word "such" IMO. :)
```suggestion
- <a name="identifyUnreachableCode"></a> **identifyUnreachableCode** [boolean]: If enabled, code that is determined to be unreachable by type analysis is reported using a tagged hint. This setting does not affect code that is determined to be unreachable regardless of type analysis; such code is always reported as unreachable. This setting also has no effect when when using the command-line version of pyright because it never emits tagged hints for unreachable code.
```

rchiodo (approved):

docs/configuration.md
debonte: I agree that this name is confusing. The main way that users complain about the current behavior is by saying, "You are reporting X as unreachable, but it isn't". So enabling this feature via "identifyUnreachableCode=true` seems weird from that perspective.

Also, even when this feature is disabled we'd be identifying some code as unreachable.

How about changing this from a bool setting to an enum?

erictraut: I don't think it's a good idea to introduce an enumerated setting. This would be inconsistent with the framework we've established for config settings. All current config settings are either diagnostic severities or booleans, and they have well-defined override semantics in the config file and in `# pyright` comments.

Also, "syntactic" isn't an accurate description because unreachability based on non-type information includes non-syntactic cues. Refer to the documentation for more details.

If users want to disable all tagged hints, they have the option of doing so.

Users who are complaining about this are not interested in type analysis. They have type checking disabled, so they're (somewhat) understandably confused when pylance is reporting something that is based on type analysis. From a type analysis perspective, this code is unreachable. To users of mypy, this would not be confusing at all.

erictraut: The name `enableReachabilityHints` is worth considering as a boolean. The verb "enable" is more consistent with existing settings than "identify". I'm somewhat reluctant to use the word "hints" here, but maybe that's OK.

Here are some other options to consider: "enableReachableCode", "enableCodeReachability", "enableUnreachableAnalysis", "enableUnreachableCodeAnalysis"? These aren't quite accurate because unreachable code analysis is always enabled; this switch simply determines whether we report the results of the analysis.

Here's another consideration. There has been [some demand](https://github.com/microsoft/pyright/issues/2672) for a way to have pyright emit a real diagnostic for unreachable code detected through type analysis. Mypy has an option like this. I've been reluctant to add this because it's an option that we couldn't enable by default even in strict mode, but I could see us adding it in the future if there's sufficient demand. If we do, it should have a name like `reportUnreachable` or `reportUnreachableCode`. It would be nice if the name we choose for this feature had a form that starts with "report".

erictraut: @luabud, any thoughts on this overall approach or on the name of the config setting?

luabud: I actually agree `identifyUnreachableCode` should be fine given the documentation is clear, but IMO the least confusing ones are `enableReachabilityAnalysis` or even  `analyzeReachability`  for consistency with ` analyzeUnannotatedFunctions` (although to your point, it may be better to not be consistent in this case since the unreachable code analysis is actually happening)

luabud (approved):

docs/configuration.md
erictraut: OK, thanks for all the feedback. I'm going to switch the name to "enableReachabilityAnalysis".

I'll close this PR and create a new one with all of the changes squashed into one.

PR #8659, opened 2024-08-05
update ja translation for 'unaccessed'

pinterior:
Currently, 'not accessed' is translated to 'アクセスできません'.
but 'アクセスできません' means 'cannot  access'.
- en: https://github.com/microsoft/pyright/blob/9a5643fc5f6e8f089f031e278014d9dedafb217a/packages/pyright-internal/src/localization/package.nls.en-us.json#L556-L560
- ja: https://github.com/microsoft/pyright/blob/9a5643fc5f6e8f089f031e278014d9dedafb217a/packages/pyright-internal/src/localization/package.nls.ja.json#L567-L571

the intended meaning is the class (or other identifier) is not used from anywhere.
a more accurate translation would be '参照されていません'. ('not referenced' in English)

direct translation or 'not accessed' will be 'アクセスされていません' but this is a bit unfriendly.


debonte:
Thanks for your contribution! Unfortunatley we are unable to accept external localization string contributions. I'll submit your suggestion as a bug to our localization team and get it fixed that way.

debonte:
Issue has been filed. For internal folks, it's at https://dev.azure.com/ceapex/CEINTL/_workitems/edit/989130

debonte:
@pinterior, the updated strings have been merged into Pylance and will show up in Pyright in the near future. This is what the loc team chose, which seems to be slightly different than what you proposed but still translates to "not referenced":

        "unaccessedClass": "クラス \"{name}\" は参照されていません",
        "unaccessedFunction": "関数 \"{name}\" は参照されていません",
        "unaccessedImport": "インポート \"{name}\" は参照されていません",
        "unaccessedSymbol": "\"{name}\" は参照されていません",
        "unaccessedVariable": "変数 \"{name}\" は参照されていません",

PR #8702, opened 2024-08-08
Fix fileInfo null error

KacieKK:
Address https://github.com/microsoft/pylance-release/issues/6234

check `hasAnalyzerInfo` before moving forward

### DIFF
--- a/packages/pyright-internal/src/analyzer/analyzerNodeInfo.ts
+++ b/packages/pyright-internal/src/analyzer/analyzerNodeInfo.ts
@@ -228,11 +228,6 @@ export function isCodeUnreachable(node: ParseNode): boolean {
     return false;
 }
 
-export function hasAnalyzerInfo(node: ParseNode): boolean {
-    const info = getAnalyzerInfo(node);
-    return info?.fileInfo? true : false;
-}
-
 function getAnalyzerInfo(node: ParseNode): AnalyzerNodeInfo | undefined {
     return node.a as AnalyzerNodeInfo | undefined;
 }
--- a/packages/pyright-internal/src/analyzer/analyzerNodeInfo.ts
+++ b/packages/pyright-internal/src/analyzer/analyzerNodeInfo.ts
@@ -228,6 +228,11 @@ export function isCodeUnreachable(node: ParseNode): boolean {
     return false;
 }
 
+export function hasAnalyzerInfo(node: ParseNode): boolean {
+    const info = getAnalyzerInfo(node);
+    return info?.fileInfo? true : false;
+}
+
 function getAnalyzerInfo(node: ParseNode): AnalyzerNodeInfo | undefined {
     return node.a as AnalyzerNodeInfo | undefined;
 }
--- a/packages/pyright-internal/src/analyzer/program.ts
+++ b/packages/pyright-internal/src/analyzer/program.ts
@@ -2032,6 +2032,10 @@ export class Program {
             }
 
             // We might not have the file info if binding failed for whatever reasons.
+            if (!AnalyzerNodeInfo.hasAnalyzerInfo(parseResults.parseTree)) {
+                continue;
+            }
+            
             const fileInfo = AnalyzerNodeInfo.getFileInfo(parseResults.parseTree);
             if (fileInfo && fileInfo.accessedSymbolSet) {
                 dependentFiles.push(parseResults);
--- a/packages/pyright-internal/src/analyzer/program.ts
+++ b/packages/pyright-internal/src/analyzer/program.ts
@@ -2032,10 +2032,11 @@ export class Program {
             }
 
             // We might not have the file info if binding failed for whatever reasons.
-            if (!AnalyzerNodeInfo.hasAnalyzerInfo(parseResults.parseTree)) {
+            // Check whether the file has been bound
+            if (file.sourceFile.isBindingRequired()) {
                 continue;
             }
-            
+
             const fileInfo = AnalyzerNodeInfo.getFileInfo(parseResults.parseTree);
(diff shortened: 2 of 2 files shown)


packages/pyright-internal/src/analyzer/program.ts
```diff
@@ -2032,6 +2032,10 @@ export class Program {
             }
 
             // We might not have the file info if binding failed for whatever reasons.
+            if (!AnalyzerNodeInfo.hasAnalyzerInfo(parseResults.parseTree)) {
```
erictraut, line 2035: Do we understand the root cause of this problem? I don't understand why a parse tree would ever _not_ have a fileInfo in its root node. This looks like it's just masking a deeper problem.

heejaechang: this is only place we ever set `fileInfo` - https://github.com/microsoft/pyright/blob/main/packages/pyright-internal/src/analyzer/sourceFile.ts#L816

so, one case would be if `binding` failed. 

or this call (https://github.com/microsoft/pyright/blob/main/packages/pyright-internal/src/analyzer/program.ts#L2019) return without actually calling `binding`

heejaechang: this crash is from `telemetry` so unfortunately we don't have repro steps.

packages/pyright-internal/src/analyzer/analyzerNodeInfo.ts
```diff
@@ -228,6 +228,11 @@ export function isCodeUnreachable(node: ParseNode): boolean {
     return false;
 }
 
+export function hasAnalyzerInfo(node: ParseNode): boolean {
+    const info = getAnalyzerInfo(node);
+    return info?.fileInfo? true : false;
```
heejaechang, line 233: I think we should jsut check whether `info` is there, not `fileInfo` or change the function name to `hasFileInfo`?

packages/pyright-internal/src/analyzer/program.ts
```diff
@@ -2032,6 +2032,10 @@ export class Program {
...
+                continue;
+            }
+            
             const fileInfo = AnalyzerNodeInfo.getFileInfo(parseResults.parseTree);
             if (fileInfo && fileInfo.accessedSymbolSet) {
```
heejaechang, line 2040: and `getFileInfo` can't return `undefined`, so you shouldn't need to check `fileInfo` in `if statement`?

packages/pyright-internal/src/analyzer/analyzerNodeInfo.ts
erictraut: The `fileInfo` subfield will appear only in the top-most parse node in a tree, so this isn't the correct check.

In any case, I don't think this is the right way to address this issue. If we're trying to protect against situations where binding failed, we need to do so in a more sound manner rather than just adding a spot check in this one place for some symptom of the underlying condition.

There are many places in the code where we assume that a bound file has a valid fileInfo at the topmost parse node in its tree. If this assumption is violated, it's potentially going to cause other crashes everywhere else it is accessed.

Please do not check in this change.

heejaechang: we are open to any other suggestion. another way is making sure that `fileinfo` is always set even when exception case.

packages/pyright-internal/src/analyzer/program.ts
erictraut: > so, one case would be if binding failed.

This would be a problem only if we catch an exception very early in the `bind` method, before we even construct and run the `Binder`. I think that's really unlikely unless there's some bad data structure corruption issue, in which case we're in a pretty bad situation and will likely crash anyway.

packages/pyright-internal/src/analyzer/analyzerNodeInfo.ts
erictraut: I think we need to understand the root cause better. Think about ways we can gather more information through telemetry or in the crash data. 

In the exception case, we could verify that `fileInfo` is set and assert (and therefore force a crash) if it's not. That would allow us to validate your theory and potentially get us one step closer to root causing the problem. I don't think we should just mask the problem and move on.

packages/pyright-internal/src/analyzer/program.ts
heejaechang: the other case is this

> this call (https://github.com/microsoft/pyright/blob/main/packages/pyright-internal/src/analyzer/program.ts#L2019) return without actually calling binding

basically, we call `_checkTypes` and assume the file would have bound. but in certain case, due to this, it won't.

https://github.com/microsoft/pyright/blob/e8f0965e5b0429232af0c8881781f0b5045b404e/packages/pyright-internal/src/analyzer/program.ts#L1899-L1912

For example, `isCheckingRequired` would return `false` even if `dump cache` dumped binding info (basically parse tree). and `binder` has not been called yet but parser did.

another approach to fix it would be not attaching data to parse node so they doesn't go away when parse tree go away like how data produced by checker do.

or force `_checkTypes` to run if `binding` information is not there.

erictraut: In that case, the file will not be marked as bound (i.e. `isBindingNeeded` will still be true). It sounds like we should check whether the file has been bound. This is a much cleaner and less fragile approach than checking for certain side effects of binding.

```python
            if (file.sourceFile.isBindingRequired()) {
                continue;
            }
```

heejaechang: ya, that will work as well.

erictraut (approved):
LGTM

PR #8704, opened 2024-08-08
Document different types of 'compiler' flags and ensure merging of flags is consistent

rchiodo:
This is the first stage in enforcing that a pyrightconfig.json is the one source of truth.

I updated the configuration docs a bit (although I don't think my update is entirely necessary) to mention the fact that a pyrightconfig.json trumps settings.

I then reworked the logic in the config setup to make sure that:
- pyrightconfig.json settings apply first. 
- command line settings override pyrightconfig.json, but only if not coming from VS code (from settings.json)
- command line settings are applied if pyrightconfig.json is not present
- typeCheckingMode defaults to 'standard' when a pyrightconfig.json is present. typeCheckingMode defaults to 'off' if coming from VS code, and defaults to 'standard' for the command line if there's no pyrightconfig.json

There was some circular problems where the initialize from json was always assuming the command line had already set things but then went and applied new settings from the command line. Hopefully this is more linear. Config always first. Command line always after.


rchiodo:
Oh I also tested this with Pylance. Having a pyrightconfig.json behaves as you'd expect. It trumps everything in the settings.json.

Next steps after this:
- Inject this into pylance
- Implement a warning/error provider for Pylance in the settings.json
- Blog/Wiki page about how settings are merged that the warning/error can point to.
- Fix the 
image
 menu in Pylance to show the correct state.

packages/pyright-internal/src/analyzer/service.ts
```diff
@@ -600,71 +600,54 @@ export class AnalyzerService {
...
                     config.configFileDirUri,
                     this.serviceProvider,
-                    host,
-                    commandLineOptions
+                    host
                 );
             }
+
+            // When not in VS code mode, command line options override config file options.
+            if (!commandLineOptions.fromVsCodeExtension) {
```
erictraut, line 624: It occurs to me that "fromVsCodeExtension" would be more accurately called "fromLanguageServer" since it also applies when using the pyright LSP. We can change that in a subsequent PR if you don't want to do it here.

The comments below also refer to "VS code extension", and they should more accurately refer to "language server".

docs/configuration.md
```diff
@@ -6,14 +6,14 @@ Pyright settings can also be specified in a `[tool.pyright]` section of a “pyp
 
 Relative paths specified within the config file are relative to the config file’s location. Paths with shell variables (including `~`) are not supported. Paths within a config file should generally be relative paths so the config file can be shared by other developers who contribute to the project.
 
-## Main Configuration Options
+## Environment Options
```
erictraut, line 9: I really like the way you've (re)organized these options. These headings provide a lot of clarity.

erictraut (approved):
I really like how this turned out. It not only makes things more consistent but it also cleans up the code pretty significantly. Thanks for doing this!

I presume that you've done some basic manual tests with the CLI? We don't have much test coverage for CLI options, so please do a few basic smoke tests before committing.

packages/pyright-internal/src/analyzer/service.ts
rchiodo: Good idea. Thanks. I'll change it.

rchiodo:
> I presume that you've done some basic manual tests with the CLI? We don't have much test coverage for CLI options, so please do a few basic smoke tests before committing.

I only did a couple, but I'll make sure to do some more. I was assuming the existing config tests were testing it a bunch, well because they didn't work on my first implementation :).

Issue #8622, opened 2024-07-31
A regression allows to violate the Liskov substitution principle for arguments when overriding methods


LeeeeT:
For the following code, pyright 1.1.373 correctly reports an incompatibility with the base class `A` when overriding the method `f` in the child class `B`:
```python
class A:
    def f(self, v: int): ...


class B[T](A):
    def f(self, v: T): ...  # 1.1.373: Type "int" is incompatible with type "T@B"
```

On version 1.1.374, the above code type checks without any errors. Here's a clear example on how this false negative can hide a serious bug:
```python
class A:
    def f(self, v: int): ...


class B[T](A):
    def f(self, v: T): ...  # 1.1.374: no errors


class C(B[str]):
    def f(self, v: str):
        v.lower()


def test(a: A):
    a.f(1)


test(C())
```

The type of the base class `A` is completely neglected in the child class `C` resulting in a runtime error.

erictraut:
This is addressed in pyright 1.1.376.

PR #8711, opened 2024-08-09
Fixed bug that results in a false negative `reportIncompatibleMethodO…

erictraut:
…verride` when the child class method uses a type parameter that is scoped to the child class. This addresses #8622.

### DIFF
--- a/packages/pyright-internal/src/analyzer/checker.ts
+++ b/packages/pyright-internal/src/analyzer/checker.ts
@@ -6770,7 +6770,7 @@ export class Checker extends ParseTreeWalker {
         const baseClass = baseClassAndSymbol.classType;
         const childClassSelf = ClassType.cloneAsInstance(selfSpecializeClass(childClassType));
 
-        const baseType = partiallySpecializeType(
+        let baseType = partiallySpecializeType(
             this._evaluator.getEffectiveTypeOfSymbol(baseClassAndSymbol.symbol),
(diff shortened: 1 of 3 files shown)


erictraut:
@Dr-Irv, looks like you removed some `# pyright: ignore` comments from pandas-stubs [recently](https://github.com/pandas-dev/pandas-stubs/commit/458ecb4bdfaa1b0bcc114f93f0c9d6ef1fa2b5f3). These will unfortunately need to be added back once pyright 1.1.376 is released. A recent regression in pyright resulted in a false negative for these cases, which explains why these errors disappeared in 1.1.374. They are legit type errors, and this fix will bring them back. Just wanted to give you a heads up. Apologies for the churn.

Dr-Irv:
> @Dr-Irv, looks like you removed some `# pyright: ignore` comments from pandas-stubs [recently](https://github.com/pandas-dev/pandas-stubs/commit/458ecb4bdfaa1b0bcc114f93f0c9d6ef1fa2b5f3). These will unfortunately need to be added back once pyright 1.1.376 is released. A recent regression in pyright resulted in a false negative for these cases, which explains why these errors disappeared in 1.1.374. They are legit type errors, and this fix will bring them back. Just wanted to give you a heads up. Apologies for the churn.

Thanks for the heads up.  Can you ping here or somewhere else once 1.1.376 is released?

Issue #8719, opened 2024-08-09
`diagnosticMode` is no longer defaulting to "openFilesOnly"


erictraut:
With the latest changes in "main", I noticed that the `diagnosticMode` setting is no longer defaulting to "openFilesOnly". The Pyright VS Code extension is now analyzing all of the files in my projects.

erictraut:
@rchiodo, looks like this is fallout from your settings change. Do you have time to look into this today, or should I?

erictraut:
@rchiodo, it looks like this affects more than just `diagnosticMode`. A bunch of functionality is broken in the VS Code extension for pyright. It's pretty much unusable currently. I'm going to revert your change for now.

rchiodo:
I can look now. Was OOF for a bit.

rchiodo:
I'm not reproducing. Do you have an example layout? I only see errors for the opened files.

erictraut:
I'm running debug also. Did you un-revert (i.e. reapply) your commit?

rchiodo:
Reproing now. Not sure what the problem was. 

It seems there are some settings that should always be applied. 'checkOpenFilesOnly' is one of them. I'll need to double check all of the options in 'CommandLineOptions' to make sure they're not one of the json allowed settings.

rchiodo:
Maybe it would be good to separate the parts of the CommandLineOptions into two sections. 'Json/pyproject.toml' configurable parts and those that are command line only.

erictraut:
Yeah, sounds reasonable. Maybe move all of the Json/pyproject.toml configurable fields into a sub-structure that hangs off CommandLineOptions?

rchiodo:
Yeah my thoughts exactly.

erictraut:
This is addressed in pyright 1.1.376.

PR #8726, opened 2024-08-09
Reapply the compiler flags change but separate out the 'language server' specific settings

rchiodo:
This is a reapplication of https://github.com/microsoft/pyright/pull/8704 but with also taking into account flags that have to always be set even when a pyrightconfig.json is present because these flags are not settable by a pyrightconfig.json.

I tested vscode-pyright with this change and it takes the diagnostic mode into account correctly now and watches for file changes. This wasn't working with the original submission.

Addresses https://github.com/microsoft/pyright/issues/8719

### DIFF
--- a/packages/pyright-internal/src/common/commandLineOptions.ts
+++ b/packages/pyright-internal/src/common/commandLineOptions.ts
@@ -31,11 +31,9 @@ export function getDiagnosticSeverityOverrides() {
 
 export type DiagnosticSeverityOverridesMap = { [ruleName: string]: DiagnosticSeverityOverrides };
 
-// Some options can be specified from a source other than the pyright config file.
-// This can be from command-line parameters or some other settings mechanism, like
-// that provided through a language client like the VS Code editor. These options
-// are later combined with those from the config file to produce the final configuration.
-export class CommandLineOptions {
+// Options that can be specified in a JSON config file. This list should match what is
+// defined in the pyrightconfig.schema.json file.
+export class CommandLineConfigOptions {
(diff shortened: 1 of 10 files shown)


erictraut (approved):
Looks good

packages/pyright-internal/src/common/commandLineOptions.ts
```diff
@@ -145,14 +136,35 @@ export class CommandLineOptions {
...
     disableTaggedHints?: boolean;
+}
+
+// Some options can be specified from a source other than the pyright config file.
+// This can be from command-line parameters or some other settings mechanism, like
+// that provided through a language client like the VS Code editor. These options
+// are later combined with those from the config file to produce the final configuration.
+export class CommandLineOptions {
+    // Settings that are possible to set in a config.json file.
+    configSettings: CommandLineConfigOptions = new CommandLineConfigOptions();
```
rchiodo, line 149: I'm not super happy with this name. config is used so much here. 'environmentSettings' maybe? As that's what the configuration.md calls them mostly now. Although it also includes type checking settings :(

erictraut: Hmm, "configSettings" seems good to me.

rchiodo: typeCheckingSettings?

rchiodo: Okay I'll just leave it.

Issue #8727, opened 2024-08-09
PythonPath from vscode-pyright is not settable if there is a pyrightconfig.json


rchiodo:
Found another issue with the recent config changes. Python path is not settable from the settings.json. Seem obvious in hindsight, but I think this should be allowed.

rchiodo:
I should have a fix shortly. I think this one setting should special cased? Or maybe in both places.

erictraut:
This is addressed in pyright 1.1.376.

PR #8728, opened 2024-08-10
Special case pythonPath to also come from the languageServer

rchiodo:
In https://github.com/microsoft/pyright/pull/8726 a pyrightconfig.json being present means all settings settable from a pyrightConfig.json would not be applied. There is a special case however where they need to be - the `pythonPath`. This is how a language server specifies which python it's using.

Addresses https://github.com/microsoft/pyright/issues/8727


rchiodo:
@erictraut, if you're okay with this, feel free to submit. I'll be OOF until Tuesday.

PR #8786, opened 2024-08-20
[Dont check in] test for get type of overload function

bschnurr:
image

behaviour change due to refactoring of overloads. https://github.com/microsoft/pyright/commit/2b603b9144698964f09f308bb70e14b2a60f11b1

previously "priv" contained overloads from "functionResult?.functionType.priv"

expected data to be assigned to "priv" from the commit before the overload change 2 weeks ago.
image


bschnurr:
@erictraut fyi. not sure this is a bug, but its affecting Pylance. Rich is trying to work around it. i'll update later tonight with more info.

PR #8799, opened 2024-08-21
Keep GitHub Actions up to date with GitHub's Dependabot

cclauss:
Automatest the creation of pull requests like:
* #4400
* #8201

Fixes software supply chain safety warnings like at the bottom right of
https://github.com/microsoft/pyright/actions/runs/10493143806

* [Keeping your actions up to date with Dependabot](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot)
* [Configuration options for the dependabot.yml file - package-ecosystem](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,13 @@
+# Keep GitHub Actions up to date with GitHub's Dependabot...
+# https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot
+# https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    groups:
+      github-actions:
+        patterns:
+          - "*"  # Group all Actions updates into a single larger pull request
+    schedule:
+      interval: weekly


.github/dependabot.yml
```diff
@@ -0,0 +1,13 @@
+# Keep GitHub Actions up to date with GitHub's Dependabot...
+# https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot
+# https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem
+version: 2
+updates:
+  - package-ecosystem: github-actions
```
debonte, line 6: I'm concerned that adding `dependabot.yml` with only the `github-actions` ecosystem is going to disable the existing Dependabot behavior that we've enabled in our repo settings (ex. for npm packages).

cclauss: Where are the `npm` settings?

debonte: As far as I know, there is no config file checked in at the moment. The Dependabot settings are configured via the repo settings. You won't (shouldn't) have access to this page, but I'm including the URL so you can maybe understand what I'm talking about by looking at a repo that you own -- https://github.com/microsoft/pyright/settings/security_analysis

debonte: To be clear, I'm not against the change you're proposing, but I think we need to somehow generate a `dependabot.yml` file that represents our current settings plus your additional settings.

cclauss: Understood.  I will first try to figure out how to set up Dependabot via `settings/security_analysis`.

Issue #8797, opened 2024-08-21
False positive `reportIndexIssue` for typings in `v1.1.376` when `pyproject.toml` contains `[tool.pyright]` section (even if this section is empty)


chegoryu:
**Describe the bug**

I gets false positive `Subscript for class "list" will generate runtime exception; enclose type annotation in quotes (reportIndexIssue)` error with pyright `v1.1.376`+ when  `[tool.pyright]` is present in `pyproject.toml` on this code (even if this section is empty):

```python
def f(_: int) -> list[int]:
    return []

```

Works fine with `v.1.1.375`:
```shell
$ cat pyproject.toml 
[tool.pyright]
$ PYRIGHT_PYTHON_FORCE_VERSION=v1.1.375 python3.11 -m pyright --pythonpath /usr/bin/python3.11 test.py --project pyproject.toml 
WARNING: there is a new pyright version available (vv1.1.375 -> v1.1.377).
Please install the new version or set PYRIGHT_PYTHON_FORCE_VERSION to `latest`


up to date, audited 2 packages in 599ms

found 0 vulnerabilities
0 errors, 0 warnings, 0 informations
```

False positive error with `v1.1.376`:

```shell
$ cat pyproject.toml 
[tool.pyright]
$ PYRIGHT_PYTHON_FORCE_VERSION=v1.1.376 python3.11 -m pyright --pythonpath /usr/bin/python3.11 test.py --project pyproject.toml 
WARNING: there is a new pyright version available (vv1.1.376 -> v1.1.377).
Please install the new version or set PYRIGHT_PYTHON_FORCE_VERSION to `latest`


up to date, audited 2 packages in 778ms

found 0 vulnerabilities
/home/echunaev/junk/test/test.py
  /home/echunaev/junk/test/test.py:1:18 - error: Subscript for class "list" will generate runtime exception; enclose type annotation in quotes (reportIndexIssue)
1 error, 0 warnings, 0 informations
```

Removing `[tool.pyright]` section solves the issue:

```shell
$ rm pyproject.toml && touch pyproject.toml
$ PYRIGHT_PYTHON_FORCE_VERSION=v1.1.376 python3.11 -m pyright --pythonpath /usr/bin/python3.11 test.py --project pyproject.toml 
WARNING: there is a new pyright version available (vv1.1.376 -> v1.1.377).
Please install the new version or set PYRIGHT_PYTHON_FORCE_VERSION to `latest`


up to date, audited 2 packages in 604ms

[...]

PR #8802, opened 2024-08-21
Ensure default version and platform only once

rchiodo:
Addresses https://github.com/microsoft/pyright/issues/8797


erictraut (approved):

rchiodo:
Okay that had a lot more impact than I thought. Must have missed something.

rchiodo:
Defaults were implicitly being used in the execution environments. Need to update to take that into account.

PR #8818, opened 2024-08-23
Handle undefined python versions

rchiodo:
While testing our latest release, we had an undefined PythonVersion happen at this point here:
https://github.com/microsoft/pyrx/blob/c1a1f8f062d3391f32b2c1ecc0e25082cf6a794c/packages/pyright/packages/pyright-internal/src/analyzer/importResolver.ts#L2002

Which caused a crash.

This change makes PythonVersion handle undefined.

### DIFF
--- a/packages/pyright-internal/src/common/pythonVersion.ts
+++ b/packages/pyright-internal/src/common/pythonVersion.ts
@@ -50,7 +50,12 @@ export namespace PythonVersion {
         };
     }
 
-    export function isEqualTo(version: PythonVersion, other: PythonVersion) {
+    export function isEqualTo(version: PythonVersion | undefined, other: PythonVersion | undefined) {
+        if (version === undefined && other === undefined) {
+            return true;
+        } else if (version === undefined || other === undefined) {
+            return false;
+        }
         if (version.major !== other.major || version.minor !== other.minor) {
             return false;
         }
@@ -76,7 +81,14 @@ export namespace PythonVersion {
         return true;
     }
 
-    export function isGreaterThan(version: PythonVersion, other: PythonVersion) {
+    export function isGreaterThan(version: PythonVersion | undefined, other: PythonVersion | undefined) {
+        if (version && !other) {
+            return true;
+        } else if (!version && other) {
+            return false;
+        } else if (!version || !other) {
+            return false;
+        }
         if (version.major > other.major) {
             return true;
         } else if (version.major < other.major) {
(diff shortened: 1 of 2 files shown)


packages/pyright-internal/src/common/pythonVersion.ts
```diff
@@ -76,7 +81,14 @@ export namespace PythonVersion {
         return true;
     }
 
-    export function isGreaterThan(version: PythonVersion, other: PythonVersion) {
+    export function isGreaterThan(version: PythonVersion | undefined, other: PythonVersion | undefined) {
+        if (version && !other) {
+            return true;
+        } else if (!version && other) {
```
erictraut, line 87: Nit: this "if" statement isn't needed. It's already covered by the "else if" statement below and returns the same result.

packages/pyright-internal/src/common/pythonVersion.ts
```diff
@@ -118,38 +130,38 @@ export namespace PythonVersion {
...
 
-    export function isLessOrEqualTo(version: PythonVersion, other: PythonVersion) {
+    export function isLessOrEqualTo(version: PythonVersion | undefined, other: PythonVersion | undefined) {
         return !isGreaterThan(version, other);
     }
 
-    export function toMajorMinorString(version: PythonVersion): string {
-        return `${version.major}.${version.minor}`;
+    export function toMajorMinorString(version: PythonVersion | undefined): string {
+        return `${version?.major}.${version?.minor}`;
```
debonte, line 146: Looks like `toString(undefined)` and `toMajorMinorString(undefined)` will return `"."`. That seems weird to me. I'd expect `undefined` or empty string.

packages/pyright-internal/src/common/pythonVersion.ts
```diff
@@ -76,7 +81,14 @@ export namespace PythonVersion {
         return true;
     }
 
-    export function isGreaterThan(version: PythonVersion, other: PythonVersion) {
+    export function isGreaterThan(version: PythonVersion | undefined, other: PythonVersion | undefined) {
```
erictraut, line 84: I'm a little uncomfortable defining boolean comparison operators for non-existent Python versions. Isn't the correct answer "undefined" in this case? Or are you thinking that a non-existent version is equivalent to a "negative" version and is therefore always considered less than all existent versions?

packages/pyright-internal/src/common/pythonVersion.ts
erictraut, line 146: I don't think we should be allowed to print a non-existent version. Or if we do, it shouldn't appear as `undefined.undefined`.

rchiodo: Good point. I'll make it print an empty string.

packages/pyright-internal/src/common/pythonVersion.ts
rchiodo: That was my thought. Undefined version is always less than a defined version.

packages/pyright-internal/src/common/pythonVersion.ts
rchiodo: Good point, will remove.

rchiodo:
It might be that the PythonVersion being undefined is the bug in the first place. We're doing more investigating.

rchiodo:
There was a bug on our side with serializing PythonVersions, and where the bug comes from. Closing for now.

PR #8837, opened 2024-08-26
Fixed a bug that resulted in incorrect evaluation when converting a c…

erictraut:
…onstructor to a callable and the constructed class has a metaclass with a `__call__` method whose `cls` parameter is annotated with `type[T]`. This addresses https://github.com/microsoft/pylance-release/issues/6297.


rchiodo:
Thanks :)

Issue #8881, opened 2024-09-03
Dataclass field with `converter` reported as "(function)" rather than "(variable)"


debonte:
Reported as a comment on https://github.com/microsoft/pylance-release/issues/5391 by @my1e5. Appears to be a regression in 1.1.373.

**Describe the bug**
Hovering over a reference to dataclass attribute that was declared using the `converter` field specifier claims that it is a function rather than a variable.

**Code or Screenshots**
```python
import attrs

@attrs.define
class A:
    x: tuple[int, ...] = attrs.field(converter=tuple)

A([1,2,3]).x
```

## Working correctly in `2024.5.102`
![no_bug](https://github.com/user-attachments/assets/c4284985-878d-4bc5-8939-09b218dc687f)

## Not working in `2024.7.103`
![bug](https://github.com/user-attachments/assets/ff8f2ad7-f236-4262-8e31-7ba9123c5072)

erictraut:
@debonte, you mentioned in another thread that you thought this was a regression between 1.1.372 and 1.1.373. I'm seeing the same behavior for both if converters are enabled. I think what happened is that prior to 1.1.373, converters didn't work without `enableExperimentalFeatures`. The OP probably didn't have this enabled. Without this, the type evaluator assumed that `x` was a regular variable declaration and produced type errors for use of the converter. The OP probably had type checking disabled and didn't see these errors.

If you recall, the support for converters (which you implemented) relies on a synthesized symbol type. That means there's no declaration for these symbols. When the hover provider receives no declarations from `getDeclarationsForNameNode`, it falls back on some (tenuous) default assumptions and displays "function" as the declaration type. So, this is arguably a bug in the hover provider. 

We'll need to rethink the interface to `getDeclarationsForNameNode`. It should probably return additional information when there is a synthesized type. I'll take a crack at that.

erictraut:
The highlight color is provided by pylance's "semantic token provider". This functionality isn't part of pyright, so the pylance team will need to address that aspect separately if it's still an issue.

erictraut:
This is addressed in pyright 1.1.379.

PR #8883, opened 2024-09-03
Fixed a bug that results in confusing hover information for an attrs …

erictraut:
…field that uses a converter. This addresses #8881.


erictraut:
@debonte, with this change, the hover text now displays the following. It's technically correct but may be confusing for some users. I'll let you and the pylance team decide whether further changes are needed. I think the hover provider now has the information it needs.

image

debonte (approved):

PR #8969, opened 2024-09-11
Pull Pylance with Pyright 1.1.380

StellaHuang95:
- Loc update
- [Handle extra path between multi root workspace roots.](https://github.com/microsoft/pylance-release/issues/6318)
- [Fix problem with settings.json watcher when multiple workspaces](https://github.com/microsoft/pylance-release/issues/6338)


rchiodo (approved):

rchiodo:
In case it's not obvious (took me a while to find it), the difference in the mypy primer is the casing of the word 'Protocol'

packages/pyright-internal/src/tests/samples/typeForm4.py
```diff
@@ -2,29 +2,12 @@
...
-    Required,
-    Type,
-    TypeAlias,
-    TypeGuard,
-    TypeVar,
-    Union,
-    Unpack,
-)
 import typing
+from typing import (Annotated, Any, Callable, Concatenate, Final, Generic,
```
debonte, line 6: This is strange. I'm not sure why this happened. Maybe this is an older version of typeForm4.py that we're pushing back? Maybe this is related to the `parent` commit in our `.gitrepo` being out of date?

I think we should undo this.

debonte:
> In case it's not obvious (took me a while to find it), the difference in the mypy primer is the casing of the word 'Protocol'

Yeah, I made those changes because I felt that it was clear we were referring to a `Protocol` object not the concept and it allowed me to lock that substring, requiring the localized versions to also include `Protocol`.

Happy to back this out if there are concerns.

rchiodo:
No concerns from me. Sounds like a good change.

packages/pyright-internal/src/tests/samples/typeForm4.py
StellaHuang95: sure

debonte (approved):

PR #9020, opened 2024-09-17
Update Unicode range tables to match Unicode 16.0

debonte:
[Python 3.14](https://github.com/python/cpython/blob/main/Doc/whatsnew/3.14.rst#unicodedata) [upgraded](https://github.com/python/cpython/pull/124017) from Unicode 15.1 to 16.0. If you feel it's too early to take this, feel free to close it and suggest better timing.

I'm not sure why the generator script was outputting all capital hexadecimal chars, whereas the generated `unicode.ts` had lowercase in the literals and uppercase in the comments. I changed the generator to match to minimize the diffs.


erictraut:
Thanks! I think it's fine to incorporate this now.

PR #9077, opened 2024-09-25
Pull Pylance with Pyright 1.1.382

StellaHuang95:
- Loc update
- Dependencies update
- Add alias symbol in user files to indices so it shows up in add/auto imports. Part of https://github.com/microsoft/pylance-release/issues/5434


packages/pyright-internal/jest.config.js
```diff
@@ -21,6 +21,7 @@ module.exports = {
             },
         ],
     },
+    testTimeout: 6000,
```
debonte, line 24: Was this an intentional change?

StellaHuang95: This is from the package update change, @bschnurr might know?

debonte (approved):

PR #9295, opened 2024-10-22
Specify return type for createHost

ljw1004:
Currently the return type of PyrightServer.createHost is left blank, inferred by typescript to be FullAccessHost. Therefore anyone who inherits from PyrightServer must also return FullAccessHost.

We should specify the return type explicitly. That way someone can inherit from it and return NoAccessHost, say.


erictraut:
@heejaechang, this is a hook you added a while ago. The proposed change looks innocuous to me, but I just want to make sure it doesn't break pylance. Could you verify?

heejaechang (approved):

Issue #9386, opened 2024-11-05
Pydantic Field alias - wrong invalid string reporting


yann-combarnous:
**Describe the bug**
For Pydantic field alias, valid strings are reported as being invalid.

**Code or Screenshots**
```
class Resource(BaseModel):
    odataType: str = Field(alias="@odata.type")  # type: ignore[reportGeneralTypeIssues]
```

alias is reported as `str | None`, but pyright raises `reportGeneralTypeIssues`

pydantic = "2.9.2"

**VS Code extension or command-line**
Using VSCode Pylance, also reported using jakebailey/pyright-action, using pylance-version: latest.

This is a new issue, was not reported prior to last week.

Viicos:
In Pydantic, you can also construct a model using the `model_validate_*` classmethods:

```python
class Resource(BaseModel):
    odataType: str = Field(alias="@odata.type")

resource = Resource.model_validate({'@odata.type': '...'})
```

Where it makes sense to use invalid Python identifiers as keys.

Here are a couple options that could be considered:
- Do not error when the alias is not a valid identifier. After all, users can still do `Resource(**{'@odata.type': '...'})`. This is similar to unpacked typed dictionaries, where the following currently works with no pyright error ([playground](https://pyright-play.net/?pythonVersion=3.12&strict=true&locale=en-us&enableExperimentalFeatures=true&code=GYJw9gtgBALgngBwJYDsDmUkQWEMoAqiApgCYAiSAxjADRQCqKCAhlQNYBQ3B5UAvIRIVqMABQByXhPoBvCQAEwpFjBYA6eAmISAXFADOMEAF8AlN06liwKMACuKKmIBUL9gHcWINAf1NWDgBtXgBdM311KMsHJ1cXeSUVNU0SPSgJKPUJc04gA)):

```python
from typing import TypedDict, Unpack


TD = TypedDict('TD', {'@odata.type': str})


def func(**kwargs: Unpack[TD]): ...


func(**{'@odata.type': '...'})
```
- Add a new kind of error so that these can easily be ignored at the project level. However, I don't know what's the policy when it comes to adding new/changin error codes in pyright.

@yann-combarnous: alternatively, you can set the `Field` as `Annotated` metadata. Pyright won't error, but won't be able to pick up other `Field` arguments, such as default values, etc:

```python
from typing import Annotated

class Resource(BaseModel):
    odataType: Annotated[str, Field(alias="@odata.type")] = "a"
[...]

PR #9394, opened 2024-11-05
Don't honor dataclass field aliases if they aren't valid identifiers

debonte:
What do you think of this approach? In https://github.com/microsoft/pyright/issues/9386#issuecomment-2457620152 you seemed to be suggesting fixing this in the completion provider, but it's also an issue for signature help and hover. So it seemed cleaner to handle this in the type synthesis code for dataclasses and TypedDicts. Unless this is going to cause problems elsewhere?

### DIFF
--- a/packages/pyright-internal/src/analyzer/dataClasses.ts
+++ b/packages/pyright-internal/src/analyzer/dataClasses.ts
@@ -25,6 +25,7 @@ import {
     ParseNodeType,
     TypeAnnotationNode,
 } from '../parser/parseNodes';
+import { Tokenizer } from '../parser/tokenizer';
 import * as AnalyzerNodeInfo from './analyzerNodeInfo';
 import { getFileInfo } from './analyzerNodeInfo';
 import { ConstraintSolution } from './constraintSolution';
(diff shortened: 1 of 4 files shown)


erictraut:
I don't think this is the correct approach. As it was pointed out in the other thread, it's still fine to call the constructor using an unpacked dictionary, and the keys of this dictionary can be arbitrary strings. They are not limited to valid identifiers.

```python
from pydantic import BaseModel, Field

class Resource(BaseModel):
    odataType: str = Field(alias="@odata.type")

Resource(**{"@odata.type": "int"})
```

Your proposed change would introduce a false positive in the above code.

debonte:
> Your proposed change would introduce a false positive in the above code.

I'm not seeing a diagnostic with that code, even in strict.

erictraut:
Oops, the example I provided doesn't generate an error because the literal types are not retained when evaluating the type of the dict expression. Here's a different example:

```python
TD1 = TypedDict("TD1", {"@odata.type": str})

def func(td1: TD1):
    Resource(**td1)
```

My point is that your change does not correctly model the type and its runtime behaviors. The current code does, and I don't think we should change it to fix shortcomings in the LS provider code. The right place to fix this is in the completion provider and the signature help provider.

I don't think there's a need to change this in the hover provider. It's arguably showing the correct type for the `__init__` method even though the signature looks a bit odd and cannot be created through a `def` statement. There are other cases like this, such as when a ParamSpec specialization creates a signature that has multiple parameters with the same name.

PR #9398, opened 2024-11-06
Skip keyword-only params with invalid names in signature help

debonte:
Updates the signature help provider to handle the scenarios mentioned in https://github.com/microsoft/pyright/issues/9386. Signature help should not suggest to the user that they can use a kwarg to initialize the value of a dataclass field with an alias that is not a valid identifier.

I think showing position-only parameters with invalid names is fine. And position-or-keyword parameters are also ok to show given that the completion provider was already updated to not suggest these param names.


packages/pyright-internal/src/languageService/signatureHelpProvider.ts
```diff
@@ -245,21 +249,29 @@ export class SignatureHelpProvider {
                 paramName = params[params.length - 1].name || '';
             }
 
-            parameters.push({
-                startOffset: label.length,
-                endOffset: label.length + paramString.length,
-                text: paramString,
-            });
+            const isKeywordOnly =
```
erictraut, line 252: I don't think this code works for a number of edge cases. It's mixing actual parameters (those that are in the `FunctionType.shared.parameters` array) with virtual parameters (those that are returned by `getParamListDetails`). I can say from personal experience that these should never be intermixed. The indexes of the actual parameters are not the same as the indexes of the virtual parameters. The `getParamListDetails` function does things like remove `/` and `*` separators, expand unpacked tuples for `*args`, and expand unpacked TypedDicts for `**kwargs`. So the list of virtual parameters can be quite different from the list of actual parameters. Prior to this PR, this function focused completely on actual parameters, so it was internally consistent. With this proposed change, there are cases that won't be handled correctly.

As a potential fix, you could call `getParamListDetails` and then search the list of virtual parameters for one whose "kind" is keyword-only and whose name matches that of the actual parameter. That way, you won't be depending on the indexes of the two lists being the same.

debonte: Thanks for catching that. Does this look right now?

packages/pyright-internal/src/languageService/signatureHelpProvider.ts
```diff
@@ -245,21 +249,28 @@ export class SignatureHelpProvider {
...
+                paramListDetails.params.find((param) => param.param.name === paramName)?.kind === ParamKind.Keyword;
```
erictraut, line 253: This check isn't quite correct. The problem is that this code finds only the first parameter with a matching name. Usually parameter names must be unique, but there are cases where you can have multiple parameters with the same name, and they can have different "kinds". I think this check would therefore be more robust: `paramListDetails.params.some(param => param.param.name === paramName && param.kind === ParamKind.Keyword).

erictraut (approved):

PR #9463, opened 2024-11-14
Dynamically import smol-toml package

heejaechang:
`module` change in `tsconfig.json` in https://github.com/microsoft/pyright/commit/6be5f8cfb989d0c77c313a1fd1a0286cc1b26c1e#diff-ac3a7b49798141e77c54e96bca892a082469cc83bb335322b35c01e020b060c8R1142 is causing issues in pylance due to it supporting multiple environments such as webpack, ts-node, browser and etc.

this is another approach to use `smol-toml`. confirmed it works as expected with webpack, ts-node and etc.


erictraut (changes requested):
This change appears to break pyright. Refer to the mypy_primer output. All of the projects that use pyproject.toml are reporting errors now.

heejaechang:
@erictraut can you take a look again?

packages/pyright-internal/src/common/shared.ts
```diff
@@ -0,0 +1,19 @@
+/*
+ * shared.ts
```
erictraut, line 2: Hmm, this isn't a very good name for a module. It's already located in a "common" directory, which implies that it's shared. The file name should describe what it does. Consider choosing a better name.

erictraut (approved):
Looks OK, but it would be better if the "shared.ts" file had a more appropriate and descriptive name.

packages/pyright-internal/src/common/shared.ts
heejaechang: any suggestion? how about `sharedSetup.ts` ?

erictraut: How about `asyncInitialization.ts`?

PR #9502, opened 2024-11-24
Update "mypy comparison" documentation based on recent improvements in mypy

kveretennicov:
It now uses union over join in more situations. See https://mypy-play.net/?mypy=latest&python=3.13&gist=c8e017a31980f3c4fc6fc66ed2331da9


erictraut:
Thanks for the contribution.

PR #9543, opened 2024-12-04
Pull Pylance with Pyright 1.1.390

bschnurr:


rohith-coder-45:
the modules from pylance is not working even though the required module is downloaded

packages/pyright-internal/src/common/tomlUtils.ts
```diff
@@ -19,7 +19,9 @@ type TomlPrimitive =
 // maintain existing module resolution to support multiple environments.
 let TOML: any;
 const loadTomlModule = (async () => {
-    TOML = await import('smol-toml');
+    // Use a magic comment to prevent webpack from creating an extra chunk for the dynamic import by default.
+    // An extra chunk will still be created if explicitly configured in the webpack config.
+    TOML = await import(/* webpackMode: "eager" */ 'smol-toml');
```
jamierpond, line 24: This phrasing is slightly confusing, it could almost be read like it's an instruction and not the addition that you've made when viewed outside the scope of this PR?

Maybe explain more about why the comment works, since that does seem magic or like it shouldn't work lol, maybe a link to the associated documentation?

prasanna143304:
every thing was fine

Issue #4706, opened 2023-02-28
Regression with default match pattern for Never type on 1.1.196


bellini666:
Take this simple example with `reportUnnecessaryComparison` set to `warning`

```python
from typing import assert_never

def get_foo() -> int | str:
    ...

foo = get_foo()

match foo:
    case int():
        ...
    case str():
        ...
    case _:  # Pattern will never be matched for subject type "Never" (reportUnnecessaryComparison)
        assert_never(foo)
```

This was working fine on previous versions.

erictraut:
This is working as designed. It's a new feature introduced in 1.1.296. The `reportUnnecessaryComparison` diagnostic check now applies to unnecessary `case` statements that will never be matched. In your code sample, the `case _:` is completely unnecessary because it will never be hit. There's also no need to use `assert_never` with match statements if you use `reportMatchNotExhaustive`. 

If you don't like this behavior, you can disable `reportUnnecessaryComparison`.

bellini666:
I understand and like the `reportUnnecessaryComparison` diagnostic for `case` statements that will never be matched (e.g. in the example code, if I had a `case float()` or `case anything` it would be useful).

But I don't think it should apply to `case _`, as it prevents that kind of defensive code. In that specific example, if `get_foo` happens to return something that is not an `int` or `str`, it would be nice to get a runtime error (before it existed I used to `raise AssertionError` at that line)

For `1.1.296` my only options are:

1) To disable `reportUnnecessaryComparison`, which I don't want to do because it provides me a lot of benefits that I don't want to lose
2) Add `type: ignore` on all those `case _` lines, which is acceptable but really annoying
3) Rely on `reportMatchNotExhaustive`, but I would also have to add a `if not isinstance(foo, (int, str))` before or after the `match` if I want to ensure the type at runtime, which would also raise another issue for `reportUnnecessaryIsInstance`

erictraut:
The `reportUnnecessaryComparison` assumes that static types are correct and accurately reflect the runtime types. If you don't want to make that assumption and feel the need to add additional defensive runtime checks, then it probably makes sense to disable `reportUnnecessaryComparison` globally in your code.

bellini666:
> If you don't want to make that assumption and feel the need to add additional defensive runtime checks

For my own code I have almost all reports enabled and do my best to avoid having to rely on `type: ignore` specifically to be able to have that assumption.

[...]

PR #9582, opened 2024-12-13
Added exemption to the `reportUnnecessaryComparison` rule for simple …

erictraut:
…wildcard patterns in a `case` clause. This addresses #4706.


alfapppro (commented):
revisar cambios

PR #9666, opened 2025-01-07
pylance-jan7

bschnurr:
- updated test harness to check open files are parsed and bound
- updated packages


.gitrepo
```diff
@@ -0,0 +1,12 @@
+; DO NOT EDIT (unless you know what you are doing)
```
rchiodo, line 1: This seems weird? Why did this end up in the root folder?

bschnurr: hmm. i'll revert

Issue #9405, opened 2024-11-07
False Liskov negative in generic subclass with different type variable


LeeeeT:
```python
class Map[T]:
    def map(self, x: T) -> T:
        return x

class Sub[T](Map[T]):
    def map[U: str](self, x: U) -> U:
        x.lower()
        return x

def f(v: Map[int]):
    v.map(1)

f(Sub[int]())
```

This code type checks but crashes at run time because pyright doesn't perform bound check for `U` in this case.

Related to #9347.

erictraut:
This is addressed in pyright 1.1.392.

Issue #9417, opened 2024-11-07
Pyright doesn't allow overriding methods with polymorphic functions


LeeeeT:
```python
class A:
    def f(self, v: str) -> str: ...

class B(A):
    def f[T](self, v: T) -> T: ...  # "object*" is not assignable to "str"
```

The above is expected to pass type checking. Pyright allows assigning `(T@f) -> T@f` to `(str) -> str`, so there should be no problems with these functions being methods.

Related to https://github.com/microsoft/pyright/issues/9347.

erictraut:
This is addressed in pyright 1.1.389

PR #9707, opened 2025-01-14
Fixed bug that results in an "incompatible method override" false neg…

erictraut:
…ative when overriding a method that uses class-scoped type parameters with a method that uses method-scoped type parameters. This addresses #9405.

### DIFF
--- a/packages/pyright-internal/src/tests/samples/methodOverride1.py
+++ b/packages/pyright-internal/src/tests/samples/methodOverride1.py
@@ -521,6 +521,7 @@ class Base7(Generic[T]):
 
 
 class Derived7_1(Base7[T]):
+    # This should generate an error.
     def method1(self, x: S) -> S:
         return x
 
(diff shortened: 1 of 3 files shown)


packages/pyright-internal/src/tests/samples/methodOverride1.py
```diff
@@ -521,10 +521,20 @@ def method1(self, x: T) -> T:
 
 
 class Derived7_1(Base7[T]):
+    # This should generate an error.
```
LeeeeT, line 524: Could you please explain why this should generate an error?

erictraut: Consider the following:

```python
d = Derived7_1[int]()
d.method1("")

def func(b: Base7[int]):
    b.method1(1)

func(d)
```

For an instance of `Derived7_1[int]`, the specialized type of the base class method is `(int) -> int`, and the specialized type of the derived class method is `(str) -> str`. It's a Liskov violation.

LeeeeT: `Derived7_1.method1` extends `Base7.method1` by making it polymorphic. `(S@method1) -> S@method1` is a super type of `(T) -> T` for any concrete `T` that comes from the generic specialization, isn't it? I don't know what the intended behavior is, but it feels inconsistent with what was implemented in https://github.com/microsoft/pyright/issues/9417. That's why I decided to point that out.

PR #9715, opened 2025-01-15
Pull Pylance with Pyright 1.1.392

bschnurr:


heejaechang:
diff from mypy primer doesn't look like due to this change.

rchiodo (approved):

PR #9734, opened 2025-01-21
Pull Pylance with Pyright 1.1.392

bschnurr:
pull pylance to pyright.

change includes

1. fix for temp folder. now, only main worker thread creates temp folder, and any other worker thread or child process will reuse that folder so it can be auto deleted when process exits.
2. some refactoring on import resolver so it can be reused
3. some tweaks to plumbing data around.


heejaechang:
I don't believe any of mypy primer change is due to this change.

rchiodo (approved):

Issue #9702, opened 2025-01-14
Inconsistent behavior between pylance and pyright


xixixao:
**Describe the bug**
Pyright CLI is reporting a type error, Pylance does not (even reporting "unused type: ignore" if that rule is enabled).

**Code**

```python
from upath import UPath

UPath("az://power-europe/processed", account_name="sfcommodgwcdataftr", anon=False)
```

This is from
`universal-pathlib = "0.1.4"`

Pyright reports
```
./test.py
  ./test.py:3:38 - error: No parameter named "account_name" (reportCallIssue)
  ./test.py:3:73 - error: No parameter named "anon" (reportCallIssue)
2 errors, 0 warnings, 0 informations
```

**VS Code extension or command-line**
Pylance: 2024.12.1
Pyright: 1.1.389

**Repro**

Assuming you have `poetry` and Python 3.11 installed:

```
git clone https://github.com/xixixao/example-pyright-issue-upath
cd example-pyright-issue-upath
poetry env use python3.11
poetry install
poetry shell
PYRIGHT_PYTHON_PYLANCE_VERSION="latest-release" pyright test.py
```

erictraut:
Pylance is built on top of pyright, so if you use the same version and configure them the same, they will output the same results. For a list of possible reasons why you're seeing a difference, refer to [this documentation](https://github.com/microsoft/pylance-release/blob/main/USING_WITH_PYRIGHT.md).

If you can't figure it out from the documentation, feel free to respond back here, and I'll try to help you out further.

I'm going to close in the meantime because this isn't an actionable pyright bug.

xixixao:
@erictraut 

1. I have confirmed the pyright version is the same
2. I have checked the bundled stubs, and they do not seem to include the library in question. I have also followed the instructions and copied the stubs over and added them to config.
3. I have added

```
[tool.pyright]
autoSearchPaths=false
extraPaths=[] # Include paths from PYTHONPATH env var and .env definition
typeCheckingMode="standard"
reportShadowedImports="warning"
```

to `pyproject.toml`.

I am still observing the divergent behavior.

I have updated the example repo with the stubs and the config, could you try it out?

erictraut:
Both pyright and pylance report the same error for me with the code snippet above. I'm using the latest version of pylance (2024.12.1) and the latest version of pyright (1.1.392). I'm using the same virtual environment in both cases.

Are you seeing any typing-related errors from pylance? For example, does the following code produce an error?

```python
[...]

PR #9735, opened 2025-01-21
Default python version not being picked up from the environment

rchiodo:
Addresses https://github.com/microsoft/pyright/issues/9702

Root cause was always checking for a pythonPath entry before setting the default version. This is incorrect in situations where a pythonPath is not specified, it should default to the current environment.

The original change that caused this problem was supposed to fix an issue where the default version was being set before the python path, but given that the default version is the last thing to be set, this isn't necessary.

Added a test to make sure this doesn't regress in the future.


erictraut (approved):
LGTM

PR #9850, opened 2025-02-07
Changed the type of an `async def` function from `typing.Coroutine` t…

erictraut:
…o `types.CoroutineType` for improved type accuracy.

### DIFF
--- a/packages/pyright-internal/src/analyzer/checker.ts
+++ b/packages/pyright-internal/src/analyzer/checker.ts
@@ -815,7 +815,10 @@ export class Checker extends ParseTreeWalker {
                         node
                     );
 
-                    if (isClassInstance(returnType) && ClassType.isBuiltIn(returnType, 'Coroutine')) {
+                    if (
+                        isClassInstance(returnType) &&
+                        ClassType.isBuiltIn(returnType, ['Coroutine', 'CoroutineType'])
+                    ) {
                         this._evaluator.addDiagnostic(
                             DiagnosticRule.reportUnusedCoroutine,
                             LocMessage.unusedCoroutine(),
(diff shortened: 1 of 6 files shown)


s3rius:
Hello! I have a question regarding to this PR. I have an issue in my project because of this change. Here's an example code that has an issue: 

```python
from typing import (
    Any,
    Callable,
    Coroutine,
    Generic,
    ParamSpec,
    TypeVar,
    overload,
    reveal_type,
)


async def meme() -> str: ...  # type: ignore[empty-body]


_FuncParams = ParamSpec("_FuncParams")
_RetType = TypeVar("_RetType")
_T1 = TypeVar("_T1")


class Meme(Generic[_FuncParams, _RetType]):
    def __init__(self, func: Callable[_FuncParams, _RetType]) -> None:
        self.func = func

    @overload
    async def call(
        self: "Meme[_FuncParams, Coroutine[Any, Any, _T1]]",
        *args: _FuncParams.args,
        **kwargs: _FuncParams.kwargs,
    ) -> _T1: ...
    @overload
    async def call(
        self: "Meme[_FuncParams, _RetType]",
        *args: _FuncParams.args,
        **kwargs: _FuncParams.kwargs,
    ) -> _RetType: ...
    async def call(self, *args: Any, **kwargs: Any) -> Any: ...


reveal_type(Meme(meme).call())
```

This revealed type on previous versions was evaluated as  `CoroutineType[Any, Any, str]`, but after this PR it is being evaluated as `CoroutineType[Any, Any, CoroutineType[Any, Any, str]]`. 

Is this behavior intended? Because this value is an actual coroutine and it seems like `CoroutineType` should be used interchangeably with `Coroutine`.

erictraut:
Yes, this new behavior is intended.

`CoroutineType` is a subclass of `Coroutine`. The runtime constructs a `CoroutineType` instance for an `async def` statement. The `CoroutineType` has a few attributes that are not available on the parent class `Coroutine`, so it makes sense for a type checker to use the narrower (more accurate) type in this case.

You can eliminate the error in your code by changing the `self` annotation on the first overload to `Meme[_FuncParams, CoroutineType[Any, Any, _T1]]`.

s3rius:
Thanks for your quick and well explanatory response. In that case I have another question. Should I use `CorutineType` in all places? And in which places should I use `Coroutine`?

s3rius:
Okay, thanks.

PR #9978, opened 2025-02-25
Changed the interpretation of an async function with an empty body (n…

erictraut:
…o yield statement) in a stub file or protocol definition that has a return type annotation of `AsyncGenerator`. Previously, this was assumed to be a generator even though a `yield` statement wasn't present. It is now assumed to be a `Coroutine`. This behavior is not spelled out in the typing spec, but pyright's new behavior is now consistent with mypy's.

### DIFF
--- a/packages/pyright-internal/src/analyzer/typeEvaluator.ts
+++ b/packages/pyright-internal/src/analyzer/typeEvaluator.ts
@@ -19030,24 +19030,6 @@ export function createTypeEvaluator(
                 );
             }
 
-            // If the return type is explicitly annotated as a generator, mark the
-            // function as a generator even though it may not contain a "yield" statement.
-            // This is important for generator functions declared in stub files, abstract
-            // methods or protocol definitions.
-            if (fileInfo.isStubFile || ParseTreeUtils.isSuiteEmpty(node.d.suite)) {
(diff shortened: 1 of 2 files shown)


mikenerone:
So what's the intended way to annotate this? How do I fix this case?
```python
class MyProtocol(Protocol):

    @asynccontextmanager
    async def yield_values(self) -> AsyncGenerator[SomeComplexClass, None]:
        ...
```
The best I came up with was to make the body this:
```python
        if TYPE_CHECKING:
            yield cast(SomeComplexClass, None)
```
This is acceptable, but it requires configs to assuage the wrath of both `coverage` (uncovered lines) and `flake8/ruff` (unreferenced arguments in a function that has a body - [ARG002](https://docs.astral.sh/ruff/rules/unused-method-argument/)), so I'm hoping there's a cleaner solution that I'm missing.

erictraut:
@mikenerone, this approach might address your needs:

```python
from contextlib import AbstractAsyncContextManager
from typing import Protocol

class SomeComplexClass: ...

class MyProtocol(Protocol):
    def yield_values(self) -> AbstractAsyncContextManager[SomeComplexClass]: ...
```

Issue #10018, opened 2025-03-03
Toplevel `__init__.py` imports considered invalid if partial stubs are available and do not include it


zmievsa:
**Describe the bug**
Previously, we used https://github.com/zmievsa/tortoise-orm-stubs as type stubs for https://github.com/tortoise/tortoise-orm and everything worked well. However, in versions 1.1.395 and 1.1.396 we started seeing a new bug: toplevel library `__init__.py` is not defined in stubs so its definitions were always taken from tortoise-orm directly but now direct imports from that file show `unknown import symbol` errors.

**Code or Screenshots**
First, install tortoise-orm==0.21.7 and tortoise-orm-stubs==1.0.2

```python
from tortoise import Tortoise
```

You will get the following error only on latest versions:
```
error: "Tortoise" is unknown import symbol (reportAttributeAccessIssue)
```

**VS Code extension or command-line**
Command-line: 1.1.396. Vscode doesn't have this problem.
python 3.12, macos and ubuntu

erictraut:
I'm not able to repro the problem with the steps provided. I've installed the two specified versions of these libraries (both the core library and the stub library) and pasted the code into a new source file, but I don't see any type checking error. 

Can you confirm that you're able to repro with the above steps with a clean project and venv? Is it possible that there are additional packages installed in your environment that could affect import resolution? Or custom pyright settings?

What happens if you right-click on `Tortoise` in the above code sample and choose `Go To Declaration`? For me, it takes me to the `site-packages/tortoise/__init__.py` file, which shows that import resolution is working correctly.

If I then type `from tortoise.fields import BigIntField` and right-click on `BigIntField` and choose `Go To Declaration`, it takes me to the `site-packages/tortoise-stubs/fields/__init__.pyi` file, which is what I'd expect.

zmievsa:
```bash
❯ uv --version
uv 0.5.26 (Homebrew 2025-01-30)
❯ uv init --python=3.12
Initialized project `testiks`
❯ uv add pyright tortoise-orm tortoise-orm-stubs
Using CPython 3.12.9 interpreter at: /opt/homebrew/opt/python@3.12/bin/python3.12
Creating virtual environment at: .venv
Resolved 10 packages in 141ms
Prepared 2 packages in 63ms
Installed 9 packages in 170ms
 + aiosqlite==0.21.0
 + iso8601==2.1.0
 + nodeenv==1.9.1
 + pypika-tortoise==0.5.0
 + pyright==1.1.396
[...]

PR #10020, opened 2025-03-03
Fix problem with partial stub service not being part of Pyright CLI

rchiodo:
A recent push from Pylance broke partial stub support in the pyright CLI.

The root cause we separated out the partial stub support into a separate class (we wanted to use the linking of directories in a file system for other things too), but we didn't ensure that the partial stub support was always available. 

Addresses: https://github.com/microsoft/pyright/issues/10018

### DIFF
--- a/packages/pyright-internal/src/common/serviceProviderExtensions.ts
+++ b/packages/pyright-internal/src/common/serviceProviderExtensions.ts
@@ -8,16 +8,16 @@
 import { CacheManager } from '../analyzer/cacheManager';
 import { ISourceFileFactory } from '../analyzer/programTypes';
 import { IPythonMode, SourceFile, SourceFileEditMode } from '../analyzer/sourceFile';
-import { ServiceKeys } from './serviceKeys';
+import { PartialStubService, SupportPartialStubs } from '../partialStubService';
 import { CaseSensitivityDetector } from './caseSensitivityDetector';
 import { ConsoleInterface } from './console';
+import { DocStringService, PyrightDocStringService } from './docStringService';
 import { FileSystem, TempFile } from './fileSystem';
+import { CommandService, WindowService } from './languageServerInterface';
 import { LogTracker } from './logTracker';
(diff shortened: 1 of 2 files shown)


packages/pyright-internal/src/common/serviceProviderExtensions.ts
```diff
@@ -77,6 +77,10 @@ ServiceProvider.prototype.console = function () {
     return this.get(ServiceKeys.console);
 };
 ServiceProvider.prototype.partialStubs = function () {
+    const result = this.tryGet(ServiceKeys.partialStubs);
+    if (!result) {
+        this.add(ServiceKeys.partialStubs, new PartialStubService(this.fs()));
```
rchiodo, line 82: Partial stub support used to be part of the file system, so it shouldn't be optional.

debonte, line 82: This looks good. I'm curious though why in other functions (ex. `docStringService` below we create a new instance if needed, but don't save it in the service provider.

debonte (approved):

packages/pyright-internal/src/common/serviceProviderExtensions.ts
```diff
@@ -77,6 +77,10 @@ ServiceProvider.prototype.console = function () {
     return this.get(ServiceKeys.console);
 };
 ServiceProvider.prototype.partialStubs = function () {
+    const result = this.tryGet(ServiceKeys.partialStubs);
```
debonte, line 80: Should we add a test?

rchiodo: That would be a lot of work? There's no test for running pyright from the command line. There are tests for the language server (which didn't have this bug) and there's tests to make sure partial stub service works, but there's no easy way to have a test where we validate the CLI is behaving as expected. Other than the mypy_primer. Which looks like it's showing a difference.

erictraut (commented):
LGTM

PR #10032, opened 2025-03-05
Mark `Pyright-Release` pipeline as a "release job"

debonte:
Recent Pyright-Release pipeline runs show the following error (ex. [1.1.396](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=11104436&view=results)) which needed to be fixed by the end of March:

```##[error]1ES PT Non-Blocking Error: Job 'create_release' uses the release tasks ["GitHubRelease@1"] but is not marked as a release job. Release tasks will soon be disallowed outside of release jobs. This is currently a non-blocking error (does not break pipeline runs), and will be turned as blocking error by the end of Mar 2025. More details: https://aka.ms/1espt/release-tasks-enforcement```

After adding the required `releaseJob` metadata I was getting the following error:

```/v1/Steps/UserProvidedReleaseSteps.yml@1ESPipelineTemplates (Line: 265, Col: 7): Unexpected value 'step `-download: artifact` is not allowed. Please use `pipelineArtifact` input or `1ES.DownloadPipelineArtifact@1` task instead. See https://aka.ms/1espt/inputs for more details.'```

...which I fixed by switching from the AzDO [built-in download step](https://learn.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/steps-download?view=azure-pipelines) to the 1ES template's [pipeline artifacts](https://aka.ms/1espt/inputs).

This closely matches Pylance's release pipeline.

### DIFF
--- a/build/azuredevops/azure-pipelines-release.yml
+++ b/build/azuredevops/azure-pipelines-release.yml
@@ -128,7 +128,6 @@ extends:
                   targetPath: $(Pipeline.Workspace)/$(ARTIFACT_NAME_VSIX)
             steps:
               - checkout: none
-              - download: current
               - task: GitHubRelease@1 #https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/github-release-v1?view=azure-pipelines
                 displayName: 'Create GitHub Release'
                 inputs:
--- a/build/azuredevops/azure-pipelines-release.yml
+++ b/build/azuredevops/azure-pipelines-release.yml
@@ -23,7 +23,7 @@ extends:
   parameters:
     sdl:
       codeSignValidation:
-        enabled: true
+        enabled: false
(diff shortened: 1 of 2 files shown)


bschnurr (approved):

bschnurr:
also thanks

debonte:
@erictraut, I can't see any way that this change would cause a diff for `mypy_primer`. Not sure what's going on there. Maybe the saved output from `mypy_primer` is somehow out of date?

erictraut:
@debonte, sympy has a circular dependency somewhere, and mypy_primer is using the "--threads" feature of pyright to run in parallel. The "--threads" feature is deterministic as long as there are no cycles. I'll eventually need to investigate and see if I can either suggest a fix in sympy or come up with a way to make pyright more deterministic in the presence of cycles. Either will probably take a lot of time, and it's not a high priority at the moment. So you can just ignore these for now.

PR #10055, opened 2025-03-11
Implement support for pull diagnostics in Pyright

rchiodo:
## Description

Currently whenever a file is opened or changed, Pyright analyzes the changed file (and its dependencies) on the BG thread and then sends a notification to the client with the list of diagnostics found. This is called 'push diagnostics'. https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_publishDiagnostics

This works fine as long as the associated file is actually open in the client. In some cases, the file may be open, but the user hasn't clicked on a tab or even the file may have been closed but we weren't told about it. 

See this issue for reference (which we already fixed in Pylance but not Pyright)
See https://github.com/microsoft/pylance-release/issues/5402

This causes issues with the problems list being out of sync with the actual state of the client. 

Additionally, Pyright has no notion of which files are more important than others to analyze (like maybe the one with the cursor in it).

This can be solved by using pull diagnostics instead:
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_pullDiagnostics

Pull diagnostics lets the client decide what files it wants the list of diagnostics for. This allows it to prioritize say the actively editing file over other open tabs. It lets it remove diagnostics for files that are no longer open but maybe without having the LS server close the file (so that the next open will get diagnostics faster).

## Implementation

We already have an 'analyzeFile' and 'analyze' message for a per file diagnostics request, so implementation was just calling those methods. 

I also added two ways that pull will not be used. 

- diagnosticMode = 'workspace'. In this mode I left Pyright to analyze all of the files in the workspace itself and push notifications. 
- disablePullDiagnostics - this setting is a safety valve in case this breaks something. Users can enable this setting and we'll go back to the old method for diagnostics.

## Testing

I changed the languageServer.test.ts to have more tests that used pull diagnostics and tested myself with diagnosticMode on and off.

### DIFF
--- a/packages/pyright-internal/src/analyzer/service.ts
+++ b/packages/pyright-internal/src/analyzer/service.ts
@@ -9,12 +9,7 @@
  */
 
 import * as JSONC from 'jsonc-parser';
-import {
-    AbstractCancellationTokenSource,
-    CancellationToken,
-    Connection,
-    DiagnosticRefreshRequest,
-} from 'vscode-languageserver';
+import { AbstractCancellationTokenSource, CancellationToken } from 'vscode-languageserver';
 import { parse } from '../common/tomlUtils';
 
 import { IBackgroundAnalysis, RefreshOptions } from '../backgroundAnalysisBase';
@@ -97,6 +92,7 @@ export interface AnalyzerServiceOptions {
     skipScanningUserFiles?: boolean;
     fileSystem?: FileSystem;
     usingPullDiagnostics?: boolean;
+    onInvalidated?: (reason: InvalidatedReason) => void;
 }
 
 interface ConfigFileContents {
@@ -138,12 +134,7 @@ export class AnalyzerService {
     private _disposed = false;
     private _pendingLibraryChanges: RefreshOptions = { changesOnly: true };
 
-    constructor(
-        instanceName: string,
-        serviceProvider: ServiceProvider,
-        options: AnalyzerServiceOptions,
-        private readonly _connection?: Connection
-    ) {
+    constructor(instanceName: string, serviceProvider: ServiceProvider, options: AnalyzerServiceOptions) {
         this._instanceName = instanceName;
 
         this._executionRootUri = Uri.empty();
@@ -234,19 +225,14 @@ export class AnalyzerService {
         backgroundAnalysis?: IBackgroundAnalysis,
         fileSystem?: FileSystem
     ): AnalyzerService {
-        const service = new AnalyzerService(
-            instanceName,
-            this._serviceProvider,
-            {
-                ...this.options,
-                serviceId,
-                backgroundAnalysis,
-                skipScanningUserFiles: true,
-                fileSystem,
-                usingPullDiagnostics: this.options.usingPullDiagnostics,
-            },
-            this._connection
-        );
+        const service = new AnalyzerService(instanceName, this._serviceProvider, {
+            ...this.options,
+            serviceId,
+            backgroundAnalysis,
+            skipScanningUserFiles: true,
+            fileSystem,
+            usingPullDiagnostics: this.options.usingPullDiagnostics,
+        });
 
         // Cloned service will use whatever user files the service currently has.
         const userFiles = this.getUserFiles();
@@ -488,6 +474,10 @@ export class AnalyzerService {
     }
 
     invalidateAndForceReanalysis(reason: InvalidatedReason) {
+        if (this.options.onInvalidated) {
+            this.options.onInvalidated(reason);
+        }
+
         this._backgroundAnalysisProgram.invalidateAndForceReanalysis(reason);
     }
 
@@ -500,12 +490,9 @@ export class AnalyzerService {
     }
 
     protected runAnalysis(token: CancellationToken) {
-        // In pull diagnostics mode, ask VS code to refresh the diagnostics. This is how
-        // we tell VS code that something happened that might affect diagnostics.
-        if (this.options.usingPullDiagnostics) {
-            this._connection?.sendRequest(DiagnosticRefreshRequest.type);
-        } else {
-            // Otherwise perform the analysis ourselves.
+        // In pull diagnostics mode, the service doesn't perform analysis on its own.
+        // Instead the client deliberately asks for diagnostics on a file-by-file basis.
+        if (!this.options.usingPullDiagnostics) {
             const moreToAnalyze = this._backgroundAnalysisProgram.startAnalysis(token);
             if (moreToAnalyze) {
                 this._scheduleReanalysis(/* requireTrackedFileUpdate */ false);
--- a/packages/pyright-internal/src/analyzer/service.ts
+++ b/packages/pyright-internal/src/analyzer/service.ts
@@ -9,7 +9,12 @@
  */
 
 import * as JSONC from 'jsonc-parser';
-import { AbstractCancellationTokenSource, CancellationToken } from 'vscode-languageserver';
+import {
+    AbstractCancellationTokenSource,
+    CancellationToken,
+    Connection,
+    DiagnosticRefreshRequest,
+} from 'vscode-languageserver';
 import { parse } from '../common/tomlUtils';
 
 import { IBackgroundAnalysis, RefreshOptions } from '../backgroundAnalysisBase';
@@ -91,6 +96,7 @@ export interface AnalyzerServiceOptions {
     serviceId?: string;
     skipScanningUserFiles?: boolean;
     fileSystem?: FileSystem;
+    usingPullDiagnostics?: boolean;
 }
 
 interface ConfigFileContents {
@@ -132,7 +138,12 @@ export class AnalyzerService {
     private _disposed = false;
     private _pendingLibraryChanges: RefreshOptions = { changesOnly: true };
 
-    constructor(instanceName: string, serviceProvider: ServiceProvider, options: AnalyzerServiceOptions) {
+    constructor(
+        instanceName: string,
+        serviceProvider: ServiceProvider,
+        options: AnalyzerServiceOptions,
+        private readonly _connection?: Connection
+    ) {
         this._instanceName = instanceName;
 
         this._executionRootUri = Uri.empty();
@@ -223,13 +234,19 @@ export class AnalyzerService {
         backgroundAnalysis?: IBackgroundAnalysis,
         fileSystem?: FileSystem
     ): AnalyzerService {
-        const service = new AnalyzerService(instanceName, this._serviceProvider, {
-            ...this.options,
-            serviceId,
-            backgroundAnalysis,
-            skipScanningUserFiles: true,
-            fileSystem,
-        });
+        const service = new AnalyzerService(
+            instanceName,
+            this._serviceProvider,
+            {
+                ...this.options,
+                serviceId,
+                backgroundAnalysis,
+                skipScanningUserFiles: true,
+                fileSystem,
+                usingPullDiagnostics: this.options.usingPullDiagnostics,
+            },
+            this._connection
+        );
 
         // Cloned service will use whatever user files the service currently has.
         const userFiles = this.getUserFiles();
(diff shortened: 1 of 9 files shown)


erictraut:
Is there any chance this will break the language server when it's running with clients other than VS Code? Does it fall back on push diagnostics if the client doesn't support the pull mechanism?

rchiodo:
>Is there any chance this will break the language server when it's running with clients other than VS Code? 
Well there's always that chance but it's implementing the LSP spec for pull diagnostics. As long as those other editors make the same pull calls, it should work. There's nothing VS code specific about the handling of the pull requests.

> Does it fall back on push diagnostics if the client doesn't support the pull mechanism?

Yes, the client has to have these settings in the initialization params:

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnosticClientCapabilities

If those aren't set, push diagnostics will still be used. We have to return this setting back in the result to indicate, yes we support pull diagnostics:

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnosticOptions

Do you have any ideas on which IDEs would be a good test bed? I should try that to make sure.

erictraut:
> Do you have any ideas on which IDEs would be a good test bed?

I'm not sure. I think the [vim LSP](https://microsoft.github.io/pyright/#/installation?id=vim) is the most popular (based purely on the quantity of issues I've received), so that might be a good candidate.

erictraut (approved):
LGTM. It might be worth doing a quick manual test of the CLI with watch mode (--watch). We don't have any tests for this, and it could potentially be affected by some of the code paths you touched in this PR

rchiodo:
> LGTM. It might be worth doing a quick manual test of the CLI with watch mode (--watch). We don't have any tests for this, and it could potentially be affected by some of the code paths you touched in this PR

Okay will do. Thanks for the feedback.

rchiodo:
neovim works with the new changes.

rchiodo:
watch mode works too (still produces errors on file change)

packages/pyright-internal/src/analyzer/service.ts
```diff
@@ -483,9 +500,16 @@ export class AnalyzerService {
     }
 
     protected runAnalysis(token: CancellationToken) {
-        const moreToAnalyze = this._backgroundAnalysisProgram.startAnalysis(token);
-        if (moreToAnalyze) {
-            this._scheduleReanalysis(/* requireTrackedFileUpdate */ false);
+        // In pull diagnostics mode, ask VS code to refresh the diagnostics. This is how
+        // we tell VS code that something happened that might affect diagnostics.
+        if (this.options.usingPullDiagnostics) {
+            this._connection?.sendRequest(DiagnosticRefreshRequest.type);
```
heejaechang, line 506: won't this cause all open files to get pull diagnostics request for every char changes?

rchiodo: I believe VS code is smart enough to only queue up 1 request after a change happens, but let me double check

rchiodo: No, you're right. It causes double requests. We don't actually compute anything on the second request (as the previous result id matches) but it's still sending them.

rchiodo: Oh actually this is already called during the timer

packages/pyright-internal/src/languageServerBase.ts
```diff
@@ -1061,6 +1096,85 @@ export abstract class LanguageServerBase implements LanguageServerInterface, Dis
...
+                // Then reanalyze the file (this should go to the background thread so this thread can handle other requests).
+                await workspace.service.analyzeFile(uri, token);
+
+                // Then pick up the diagnostics created.
+                serverDiagnostics = sourceFile
+                    ? await workspace.service.getDiagnosticsForRange(uri, sourceFile.getRange(), token)
+                    : [];
+
+                // If any text edits came in, make sure we reanalyze the file.
+                const sourceFileAfter = workspace.service.getSourceFile(uri);
```
debonte, line 1140: I'm confused about the `SourceFile` objects in play within this region of the code. We check if `sourceFile` is defined before the loop, but then within the loop we re-get it. And then re-get it at the bottom of the loop -- why would the `SourceFile` be a different object after analysis?

packages/pyright-internal/src/languageServerBase.ts
```diff
@@ -1061,6 +1096,85 @@ export abstract class LanguageServerBase implements LanguageServerInterface, Dis
...
+
+                // Then reanalyze the file (this should go to the background thread so this thread can handle other requests).
+                await workspace.service.analyzeFile(uri, token);
+
+                // Then pick up the diagnostics created.
+                serverDiagnostics = sourceFile
+                    ? await workspace.service.getDiagnosticsForRange(uri, sourceFile.getRange(), token)
+                    : [];
+
+                // If any text edits came in, make sure we reanalyze the file.
```
debonte, line 1139: It's not clear to me how we're doing what this comment says we're doing. Wouldn't we need to check the file content version (or whatever it's called) to make this happen?

packages/pyright-internal/src/languageServerBase.ts
```diff
@@ -1061,6 +1096,85 @@ export abstract class LanguageServerBase implements LanguageServerInterface, Dis
...
+                const sourceFileAfter = workspace.service.getSourceFile(uri);
+                diagnosticsVersionAfter = sourceFileAfter?.getDiagnosticVersion() ?? UncomputedDiagnosticsVersion;
+            }
+
+            // Then convert the diagnostics to the LSP format.
+            const lspDiagnostics = this._convertDiagnostics(workspace.service.fs, serverDiagnostics).filter(
+                (d) => d !== undefined
+            ) as Diagnostic[];
+
+            result.resultId = diagnosticsVersion.toString();
```
debonte, line 1149: Should this be `diagnosticsVersionAfter`?

packages/pyright-internal/src/languageServerBase.ts
```diff
@@ -1061,6 +1096,85 @@ export abstract class LanguageServerBase implements LanguageServerInterface, Dis
...
+            result.items = lspDiagnostics;
+        } else {
+            (result as any).kind = 'unchanged';
+            result.resultId = diagnosticsVersion.toString();
```
debonte, line 1153: If `sourceFile` is `undefined` I believe will set `resultId` to `"-1"` here. Is that ok? It seemed to be that the logic above assumes that `-1` is a secret value that we only use internally and will never be sent by LSP as the `previousResultId`.

packages/pyright-internal/src/analyzer/service.ts
rchiodo: I changed it to be a callback and only happen during 'invalidateAndForceReanalysis'. This should handle the cases that VS code doesn't already cause an update for

packages/pyright-internal/src/languageServerBase.ts
```diff
@@ -602,6 +628,15 @@ export abstract class LanguageServerBase implements LanguageServerInterface, Dis
             },
         };
 
+        if (this.client.hasPullDiagnosticsCapability) {
+            result.capabilities.diagnosticProvider = {
+                identifier: 'pyright',
+                documentSelector: null,
+                interFileDependencies: true,
+                workspaceDiagnostics: initializationOptions.diagnosticMode === 'workspace',
```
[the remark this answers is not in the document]
rchiodo: Yeah that's true. This will always be false.

packages/pyright-internal/src/languageServerBase.ts
rchiodo: Yes, it's checking the diagnostics version. That's how it checks. Diagnostics version would be zero if an update occurred so it should go back around.

This code here causes that (inside the program.setFileOpened call)

```
            // Reset the diagnostic version so we force an update to the
            // diagnostics, which can change based on whether the file is open.
            // We do not set the version to undefined here because that implies
            // there are no diagnostics currently reported for this file.
            sourceFileInfo.diagnosticsVersion = 0;
```

packages/pyright-internal/src/languageServerBase.ts
rchiodo: I don't think so. They're the same as long as the loop exits.

packages/pyright-internal/src/languageServerBase.ts
rchiodo: LSP says it's a string, so it can be anything. Next time through it just passes it to us, so I think -1 is fine here.

packages/pyright-internal/src/languageServerBase.ts
```diff
@@ -543,6 +561,14 @@ export abstract class LanguageServerBase implements LanguageServerInterface, Dis
             !!capabilities.textDocument?.completion?.completionItem?.resolveSupport?.properties.some(
                 (p) => p === 'additionalTextEdits'
             );
+        this.client.hasPullDiagnosticsCapability =
```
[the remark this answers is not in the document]
rchiodo: I'll rename them.

packages/pyright-internal/src/languageServerBase.ts
debonte: Ah, right, of course. For what it's worth, I found this loop extremely confusing. I stared at it for a long time trying to figure out what was going on here. I don't have specific recommendations on how to improve it unfortunately. There was some similar code that I reviewed in pyrx recently that I also found very difficult to wrap my head around. Might just be me.

packages/pyright-internal/src/languageServerBase.ts
rchiodo: Diagnostics version also updates when things are invalidated because of say an imported file changing

packages/pyright-internal/src/languageServerBase.ts
rchiodo: I forgot about the isCheckingRequired test. I think you're right. It should be -2.

erictraut:
@rchiodo, I just merged this PR by mistake. I thought I was pushing a different PR that I was working on. If this PR wasn't ready to merge, I can back it out. Sorry for any inconvenience.

rchiodo:
> @rchiodo, I just merged this PR by mistake. I thought I was merging a different PR that I was working on. If this PR wasn't ready to merge, I can back it out. Sorry for any inconvenience.

No worries. I can just submit follow up changes if @debonte has anymore feedback.

erictraut:
@rchiodo, with this change merged, I'm now seeing failures in local tests. Are you seeing these failures? Any suggestions for how to fix them?

 FAIL  src/tests/languageServer.test.ts (23.186 s)
  ● Basic language server tests › Diagnostics pull › background thread diagnostics open mode

    Unhandled method test/getOpenFiles

      at handleResponse (node_modules/vscode-jsonrpc/lib/common/connection.js:595:48)
      at handleMessage (node_modules/vscode-jsonrpc/lib/common/connection.js:432:20)
      at Immediate.<anonymous> (node_modules/vscode-jsonrpc/lib/common/connection.js:402:30)

  ● Basic language server tests › Diagnostics pull › Diagnostic severity overrides test

    thrown: "Exceeded timeout of 15000 ms for a test.
    Add a timeout value to this test to increase the timeout, if this is a long-running test. See https://jestjs.io/docs/api#testname-fn-timeout."

      246 |             });
      247 |
    > 248 |             test('Diagnostic severity overrides test', async () => {
          |             ^
      249 |                 const code = `
      250 | // @filename: test.py
      251 | //// def test([|/*marker*/x|]): ...

      at src/tests/languageServer.test.ts:248:13
      at src/tests/languageServer.test.ts:155:9
          at Array.forEach (<anonymous>)
      at src/tests/languageServer.test.ts:154:19
      at Object.<anonymous> (src/tests/languageServer.test.ts:36:1)

image

rchiodo:
Did you build the test server? `npm run test` is supposed to rebuild it. That first error would be exactly that. The test server didn't update

erictraut:
Ah, that explains it. I'm in the habit of running `npx jest` to run the tests. Yes, `npm run test` fixed the problem. Thanks!

PR #10080, opened 2025-03-13
Fix german translation error

flokoenig:
The original translation has a different meaning and hints at a completely different error.
I changed the meaning for unaccessed* from "can not be accessed"/"kann nicht zugegriffen werden" to "is not accessed"/"wird nicht zugegriffen".


debonte:
@flokoenig, thanks for the contribution! Our localized strings are maintained by an internal team at Microsoft. If we merge this PR, your changes will get overwritten by the existing strings the next time we sync with the loc team. I'll take your changes and submit them as a bug against the loc team. Expect to see the changes show up in Pylance soon (1-2 weeks) and then Pyright about a week later.

debonte:
Loc bug (only accessible by Microsoft employees) is here: https://dev.azure.com/ceapex/CEINTL/_workitems/edit/1038945

debonte:
FYI, the loc team approved the change and we've received the updated strings in Pylance's repo. They'll ship in Pylance's next release.

Issue #10198, opened 2025-03-29
Regression in tests


erictraut:
A recent merge from pylance has introduced a regression when running local tests.

To repro, run `npm run test` from the command line.

Here's the error:

```python
 ● Console

    console.error
      An internal error occurred while parsing file "/Users/eric/source/pyright/packages/pyright-internal/tests/samples/test_file1.py": Error: Global service provider not initialized for [object Object]
          at ServiceProvider.get (/Users/eric/source/pyright/packages/pyright-internal/src/common/serviceProvider.ts:85:19)
          at ServiceProvider.Object.<anonymous>.serviceProvider_1.ServiceProvider.console (/Users/eric/source/pyright/packages/pyright-internal/src/common/serviceProviderExtensions.ts:82:17)
          at ServiceProvider.Object.<anonymous>.serviceProvider_1.ServiceProvider.partialStubs (/Users/eric/source/pyright/packages/pyright-internal/src/common/serviceProviderExtensions.ts:87:83)
          at ImportResolver.get partialStubs [as partialStubs] (/Users/eric/source/pyright/packages/pyright-internal/src/analyzer/importResolver.ts:128:37)
          at ImportResolver.ensurePartialStubPackages (/Users/eric/source/pyright/packages/pyright-internal/src/analyzer/importResolver.ts:400:19)
          at ImportResolver._resolveImportStrict (/Users/eric/source/pyright/packages/pyright-internal/src/analyzer/importResolver.ts:945:14)
          at ImportResolver.resolveImportInternal (/Users/eric/source/pyright/packages/pyright-internal/src/analyzer/importResolver.ts:546:35)
          at ImportResolver.resolveImport (/Users/eric/source/pyright/packages/pyright-internal/src/analyzer/importResolver.ts:161:21)
          at resolveAndAddIfNotSelf (/Users/eric/source/pyright/packages/pyright-internal/src/analyzer/sourceFile.ts:1380:49)
          at SourceFile._resolveImports (/Users/eric/source/pyright/packages/pyright-internal/src/analyzer/sourceFile.ts:1405:36)
          at /Users/eric/source/pyright/packages/pyright-internal/src/analyzer/sourceFile.ts:704:47
          at TimingStat.timeOperation (/Users/eric/source/pyright/packages/pyright-internal/src/common/timing.ts:44:28)
          at /Users/eric/source/pyright/packages/pyright-internal/src/analyzer/sourceFile.ts:703:48
[...]

PR #10200, opened 2025-03-29
Fix test failure from Pylance merge

rchiodo:
Fixes https://github.com/microsoft/pyright/issues/10198

Console interface didn't need to be passed to the PartialStubsService. It's never used.


erictraut (approved):
LGTM

PR #10259, opened 2025-04-07
Fix logic typo in mypy-comparison.md

Alkarex:
If I am not mistaken, the documentation says the opposite of what it should be.


erictraut:
The current wording is correct. The proposed change would make it incorrect, so I'm going to reject this PR.

PR #10263, opened 2025-04-07
Fix "Bind called unnecessarily" assert when `__builtins__.pyi` imports the file being bound

debonte:
Addresses https://github.com/microsoft/pylance-release/issues/7095

`_bindFile` is called for `xxx/yyy.py`. It calls `_bindImplicitImports` which has the side-effect of binding `xxx/yyy.py`. Then we call `sourceFile.bind()` on `xxx/yyy.py` itself, which asserts.


rchiodo (approved):

packages/pyright-internal/src/analyzer/program.ts
```diff
@@ -1791,6 +1791,10 @@ export class Program {
             // Bind all of the implicit imports first. So we don't recurse into them.
             if (!isImplicitImport) {
                 this._bindImplicitImports(fileToBind);
+
+                if (!fileToBind.sourceFile.isBindingRequired()) {
```
erictraut, line 1795: This solution looks good to me.

The need for this check isn't immediately obvious, however. Since there's no regression test for this, a comment explaining why it's needed is probably a good idea to prevent it from being accidentally modified or deleted in the future.

erictraut (approved):

Issue #10406, opened 2025-05-02
Server doesn't advertise supported commands.


xebcam:
**Describe the bug**
Neovim checks the capability executeCommandProvider.commands list prior to
execution of any command, which in this case given that pyright doesn't
advertise any commands neovim can't run commands like
'pyright.organizeimports'

**VS Code extension or command-line**
pyright = v1.1.400
neovim = v0.11.1

erictraut:
These commands were implemented for the pyright VS Code extension, and some of them rely on logic found in both the server and client portions of the extension. They won't work if you simply register the command handlers on the server side. The "order imports" and "create stub" commands might work, but I haven't tried them.

The "order imports" command was added to pyright prior to isort, ruff, and other formatters that do automatic import ordering. I don't recommend using the one in pyright because it's not very configurable and doesn't match the sorting rules in these more popular tools. I should probably remove it, but I've left in in just in case some VS Code users still rely on it.

In any case, these commands were not designed to work with the general pyright LSP, so I don't think it makes sense to register them in the language server.

PR #10407, opened 2025-05-02
Initializing PyrightServer with supportedCommands (fix for #10406)

xebcam:
This fixes issue #10406

### DIFF
--- a/packages/pyright-internal/src/server.ts
+++ b/packages/pyright-internal/src/server.ts
@@ -15,6 +15,7 @@ import {
     WorkDoneProgressServerReporter,
 } from 'vscode-languageserver';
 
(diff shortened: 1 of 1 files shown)


erictraut:
This change breaks the pyright VS Code extension, which registers commands in the language server client [here](https://github.com/microsoft/pyright/blob/0ed54e79436eed15ca112e48f3ad0cd042bc528a/packages/vscode-pyright/src/extension.ts#L189). With this change, the commands get registered twice, which results in an LSP error.

Also, I don't think these commands will execute correctly if you simply register the handlers from the server. There is logic in the client that adds arguments that are needed by the server's handlers. The logic would need to be modified to support this from a server-side registration.

PR #10411, opened 2025-05-03
typed-libraries.md - split paragraphs into sublists for readibility

Andrej730:
No other changes besides changing the paragraphs structure.
Sublists are helping breaking down dense paragraphs to make them more readable.

Before:
image

After:
image


erictraut:
I'd prefer to keep this documentation as is for a couple of reasons. First, it's more grammatically and syntactically correct. Second, this section has been largely subsumed by (and for the most part copied verbatim in) [this section](https://typing.python.org/en/latest/spec/distributing.html#library-interface-public-and-private-symbols) of the Python typing spec, and I don't want the two to diverge any more than necessary.

Andrej730:
But it's just not very readable for anyone coming across this.
It is copied to typing spec but in a more readable way, copied version is breaking the paragraph to the same list I've suggested:

image

Issue #10473, opened 2025-05-19
false positive reportMatchNotExhaustive when all Literals are covered by types


syncosmic:
I hit this with a realistic use case in VS Code involving enums in a PySide application, but [here's an MRE](https://pyright-play.net/?strict=true&code=MTAEBcE8AcFNQJYDsBmsBO7YBNQCNJRpJ0EBzAC3FAEMBnUAWgD5QAKZagH1DvHQCUAKGywUoAMYB7VAFc6OAPqdYZDBwBciJOAEahoQ6CzhZ6JIkTiEoADygArKFgAbBb34dhQgLY1wEhSSMijySipq6GwADHoGRiCgskii6HxSUrj0wT7QLrDgsPGGEvTwfFF0ekSkOmwARADq8GRS1DQetWT1wkaSZdrgXlrQtUNNLW20FhEYPUJCiVBw2miYOPiExKSU7QwAMgiF6DQuANoA5NEXADSgAIwAuiJiwXIK2Ip4Gfk0SJqDOJ9ExmCz1aL1KyWAC80NA0Wcbng9wWfgCQWk7yU3ykv3%2BsX0fTAyVS6UytAY0ly%2BUKxX67nB9RGYwazVArWoAC8MFIKbROshur0jKV3PdmZxWZNqDJ4Nk-oNVHNvGjAm9Qh8vj9YH8YkCEkQSOQqKBsFJYHQkBdqFhTghuRAKP5HfBlvBArAJABrSlSakFVyEaQANwwoEOx1Ol2udyedLAAGEygwAO5HCjIUCqoJ8fywHywHSm3lIKawAAeTvk4AQoZchCdKXytBcLlAwdOsgt8aMAFUkI3sPlcG6tPUIxgo1dbg9HvUe4YAJLiCtVvi1%2BCDlyCxAMUvUFQpHB3GjYXD1UXwRQjeh0ed9S%2BdNhVCV1CbsqYdCo7tijC2yPw8HrUBuXQKQBHmB8BklBBqlGSV3w5aZFUidg-zoACaCAwhZQggQgA) that works in pyright-play (strict mode, pyright 1.1.4000, python 3.13). The error takes a moment to pop up.

Expected behavior as I interpret it would be for the final `match` statement not to throw `reportMatchNotExhaustive`, since the other two correctly don't.

[...]

PR #10479, opened 2025-05-20
Fixed bug in type narrowing logic for class patterns in a `match` sta…

erictraut:
…tement. Negative type narrowing was incorrect when the subject is a literal type and the class is the corresponding non-literal class. This addresses #10473.


syncosmic:
Amazing! Thanks for the elegant and lightning-fast fix!

Issue #10487, opened 2025-05-21
regression: "unecessary #type ignore" reported in stdlib


bukzor:
I've pinpointed this regression to 1.401.

```
$ pyright --verbose lib/sudo-gcp/sudo_gcp/__main__.py
Loading pyproject.toml file at ./pyproject.toml
Execution environment: python
  Extra paths:
    ./venv/lib/python3.13/site-packages
  Python version: 3.13
  Python platform: Linux
Found 1 source file
pyright 1.1.401
  ./node_modules/.pnpm/pyright@1.1.401/node_modules/pyright/dist/typeshed-fallback/stdlib/_typeshed/_type_checker_internals.pyi
  ./node_modules/.pnpm/pyright@1.1.401/node_modules/pyright/dist/typeshed-fallback/stdlib/_typeshed/_type_checker_internals.pyi:52:91 - error: Unnecessary "# type: ignore" comment (reportUnnecessaryTypeIgnoreComment)
1 error, 0 warnings, 0 informations
```

By contrast, in 1.1.400

```
[2025-05-21 Wed PM 02:09-0500] buck@macbookpro.lan $ pyright --verbose lib/sudo-gcp/sudo_gcp/__main__.py
Loading pyproject.toml file at ./pyproject.toml
Execution environment: python
  Extra paths:
    ./venv/lib/python3.13/site-packages
  Python version: 3.13
  Python platform: Linux
Found 1 source file
pyright 1.1.400
0 errors, 0 warnings, 0 informations
```

bukzor:
Minimal reproduction:

```
cat > main.py <<EOF
def main():
    pass
EOF

cat >pyproject.toml <<EOF
[tool.pyright]
reportUnnecessaryTypeIgnoreComment = "error"
EOF


virtualenv .venv
./.venv/bin/pip install pyright
./.venv/bin/pyright .
```

srittau:
In fact, we can't update pyright in typeshed due to this bug. The `# type: ignore[misc]` statement in this location is necessary for mypy, and pyright should not warn about it when running our test cases. See python/typeshed#14121. 

Also related #10484

pjonsson:
This issue got 20 thumbs up so I'm guessing a reasonable amount of people are affected by it. I had a memory that Pyright releases were automated, but I seem to have misremembered, what is the planned date for the next release of Pyright?

erictraut:
I plan to release a new version tomorrow (Tuesday), and it will include this fix.

[...]

Issue #10484, opened 2025-05-21
Implicit override warnings reported for _type_checker_internals.pyi


correctmost:
**Describe the bug**
`reportImplicitOverride` warnings are reported for an internal `_type_checker_internals.pyi` file when type checking.

**Code or Screenshots**

test.py:
```python
x=1
```

pyrightconfig.json:
```json
{
  "reportImplicitOverride": true
}
```

Running `npx pyright test.py` produces these warnings:

```
~/.npm/_npx/110e52990071af13/node_modules/pyright/dist/typeshed-fallback/stdlib/_typeshed/_type_checker_internals.pyi
  ~/.npm/_npx/110e52990071af13/node_modules/pyright/dist/typeshed-fallback/stdlib/_typeshed/_type_checker_internals.pyi:40:9 - error: Method "items" is not marked as override but is overriding a method in class "Mapping[str, object]" (reportImplicitOverride)
  ~/.npm/_npx/110e52990071af13/node_modules/pyright/dist/typeshed-fallback/stdlib/_typeshed/_type_checker_internals.pyi:41:9 - error: Method "keys" is not marked as override but is overriding a method in class "Mapping[str, object]" (reportImplicitOverride)
  ~/.npm/_npx/110e52990071af13/node_modules/pyright/dist/typeshed-fallback/stdlib/_typeshed/_type_checker_internals.pyi:42:9 - error: Method "values" is not marked as override but is overriding a method in class "Mapping[str, object]" (reportImplicitOverride)
```

**VS Code extension or command-line**
Command-line version 1.1.401

jorenham:
For what it's worth; this does not occur when running pyright using https://github.com/RobertCraigie/pyright-python. But I did encounter these same errors with [`pyright-action`](https://github.com/RobertCraigie/pyright-python), for example in https://github.com/scipy/scipy-stubs/actions/runs/15169287965/job/42655058438?pr=521.

erictraut:
This is addressed in pyright 1.1.402.

PR #10494, opened 2025-05-23
Fixed recent regression that results in spurious diagnostics being re…

erictraut:
…ported for the `_type_checker_internals.pyi` stub file. This addresses #10484 and #10487.


debonte:
FWIW, this is not sufficient to address the analysis-related telemetry issue that I mentioned in email.

PR #10538, opened 2025-06-04
Fix `reportImplicitOverride` error from `_type_checker_internals.pyi`

Geson-anko:
## Bug Fix

I've fixed unintended error `reportImplicitOverride` from `_type_checker_internals.pyi`.

#10484


erictraut:
Thanks, but this isn't the right place to make such a change. This file is vendored from the [typeshed project](https://github.com/python/typeshed). If you think you've found a bug, please submit a bug report or PR there.

PR #10595, opened 2025-06-11
Fixed bug where `isUserCode` returned true for `stdlib/string/templatelib.pyi`

debonte:
After pulling the changes in 1.1.402 into Pylance, I was seeing a test failure where we enumerated the user code subtypes of `object` and some of the results were in `stdlib/string/templatelib.pyi` (not user code). This was because `addTrackedFile` was always passing false for `isTypeshedFile`.

Fix expands upon https://github.com/microsoft/pyright/pull/10494/files. I also added a unit test that should detect this sort of issue in the future.


erictraut (approved):
LGTM. Thanks for fixing this.

PR #10697, opened 2025-07-11
Bump typeshed stubs to `d5af6bee35dff9a73e7e42c5645e6de38088574d`

engnatha:
Some type stubs for networkx were incorrect and fixed upstream through https://github.com/python/typeshed/issues/14391. This pull request brings in those changes to pyright. This will resolve errors with pyright/pylance that need to be ignored with a `# pyright: ignore` directive.


engnatha:
I wasn't sure the proper way to pull in stubs. It looks like they are copies of the upstream library, so I've done that here. I couldn't find any documentation for why `tensorflow` was excluded, but I've included it here in case that is desirable.

engnatha:
@microsoft-github-policy-service agree company="Astranis Space Technologies"

erictraut:
Thanks for the contribution, but there's no need to manually update the typeshed stubs. I do this regularly (typically every two to four weeks) during the normal course of publishing pyright. I have an established process for updating the typeshed stubs. It's a little involved because changes in some stubs (like `typing.pyi`) can affect core type checking functionality, so it's important for me to be aware of any such changes.

PR #10729, opened 2025-07-21
Fix inconsistent union constraint inference for Awaitable[T] | T

### LINKED ISSUE #10719: Inconsistent choice of `Union` variant when multiple choices are valid
In the following code, `extract` takes a function that returns `Awaitable[T] | T`. When I give it a function that returns `Awaitable[int]` pyright in theory can **choose** `T` to be either `int` or `Awaitable[int]`. Pyright seems to always prefer `int` except for the `f3` case where `int` is wrapped in two generic classes (`A` and `B`) and `async` syntax is used. This discrepancy is unintuitive, I couldn't find any information on how pyright makes its choice in such ambiguous cases.

```python
from collections.abc import Awaitable, Callable

class A[T]: pass
class B[T]: pass

async def f1() -> int: ...
async def f2() -> A[int]: ...
async def f3() -> A[B[int]]: ...

def g1() -> Awaitable[int]: ...
def g2() -> Awaitable[A[int]]: ...
def g3() -> Awaitable[A[B[int]]]: ...

def extract[T](x: Callable[[], Awaitable[T] | T]) -> T: ...

reveal_type(extract(f1))  # okay: int
reveal_type(extract(f2))  # okay: A[int]
reveal_type(extract(f3))  # weird: Awaitable[A[B[int]]]

reveal_type(extract(g1))  # okay: int
reveal_type(extract(g2))  # okay: A[int]
reveal_type(extract(g3))  # okay: A[B[int]]
```

We can explicitly tell pyright the desired order (first try `Awaitable[T]` and then `T`) by using overloads:

```python
from collections.abc import Awaitable, Callable
from typing import overload, reveal_type

class A[T]: pass
class B[T]: pass

async def f1() -> int: ...
async def f2() -> A[int]: ...
async def f3() -> A[B[int]]: ...

def g1() -> Awaitable[int]: ...
def g2() -> Awaitable[A[int]]: ...
def g3() -> Awaitable[A[B[int]]]: ...

@overload
def extract[T](x: Callable[[], Awaitable[T]]) -> T: ...

@overload
def extract[T](x: Callable[[], T]) -> T: ...

def extract[T](x: Callable[[], Awaitable[T] | T]) -> T: ...

reveal_type(extract(f1))  # int
reveal_type(extract(f2))  # A[int]
reveal_type(extract(f3))  # A[B[int]]

reveal_type(extract(g1))  # int
reveal_type(extract(g2))  # A[int]
reveal_type(extract(g3))  # A[B[int]]
```

But this approach can lead to a lot of overloads everywhere a `Union` with overlapping variants is used. I would prefer pyright to resolve constraints in the same order the `Union` variants are written in the code which would effectively give the same result as overloads.

For reference, mypy doesn't make a choice in ambiguous cases at all. It simply sets `T` to `Never` and gives a type error.

nbom-25i:
Fixes #10719 
 
Resolves issue where Pyright inconsistently inferred types for union
constraints like `Awaitable[T] | T`, particularly affecting complex
generic types like `A[B[int]]`.

Changes:
- Preserve union member declaration order in validateCallArgs by using sortSubtypes: false
- Fix CoroutineType handling in getTypeOfAwaitable to extract return type from index 2 instead of 0 for `CoroutineType[YieldT, SendT, ReturnT]`
- Enhance union assignment scoring to prefer earlier union members when constraint scores are equal

Before: `extract(f3) where f3: async () -> A[B[int]]` returned
`CoroutineType[Any, Any, A[B[int]]]`
After:  `extract(f3) correctly returns A[B[int]]`

This ensures consistent behavior where union members are tried in
their declared order (Awaitable[T] first, then T), matching the
expected overload resolution behavior.

### DIFF
--- a/packages/pyright-internal/src/analyzer/typeEvaluator.ts
+++ b/packages/pyright-internal/src/analyzer/typeEvaluator.ts
@@ -3021,6 +3021,12 @@ export function createTypeEvaluator(
                         specializedType.priv.typeArgs &&
                         specializedType.priv.typeArgs.length > 0
                     ) {
+                        // For CoroutineType, the return type is the third type argument (index 2)
+                        // CoroutineType[YieldT, SendT, ReturnT] where ReturnT is what we want
+                        if (ClassType.isBuiltIn(specializedType, ['Coroutine', 'CoroutineType']) && 
+                            specializedType.priv.typeArgs.length >= 3) {
+                            return specializedType.priv.typeArgs[2];
+                        }
                         return specializedType.priv.typeArgs[0];
                     }
 
@@ -9916,7 +9922,7 @@ export function createTypeEvaluator(
 
         let returnType = mapSubtypesExpandTypeVars(
             callTypeResult.type,
-            { sortSubtypes: true },
+            { sortSubtypes: false },
             (expandedSubtype, unexpandedSubtype, isLastIteration) => {
                 return useSpeculativeMode(
                     isLastIteration ? undefined : getSpeculativeNodeForCall(errorNode),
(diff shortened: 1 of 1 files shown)


erictraut:
We don't want to use the declared order here. The order of a union should not change type analysis results. This is why union order sorting was added originally. So I don't think this is the correct fix to this issue.

erictraut:
I'm guessing this is an automated bot account, so I'm closing this PR.

Issue #10271, opened 2025-04-08
Type inference bug with self-application of a generic function containing `ParamSpec`


LeeeeT:
**Describe the bug**
In the following example, pyright doesn't correctly "freshen" the ParamSpec `P` resulting in a partially unknown type.

**Code or Screenshots**

```python
from collections.abc import Callable

def id[**P, R](c: Callable[P, R]) -> Callable[P, R]: ...

reveal_type(id(id))
# (c: (...) -> Unknown) -> ((**P(1)@id) -> R(1)@id)
```

PR #10730, opened 2025-07-21
Fix ParamSpec freshening in self-application scenarios

nbom-25i:
Fixes type inference bug where `id(id)` with ParamSpec returns partially unknown types instead of proper generic function type.

#10271 fix


erictraut:
This change doesn't fix the associated bug that it claims to fix. It simply introduces a different bug. It also doesn't include any test.

I'm guessing this is a bot account, so I'm I'm closing the PR.

PR #10762, opened 2025-08-01
Fix literal overload completions in call arguments

el1s7:
Skip the early return in `_tryAddLiterals` when inside a function-call argument so the call-argument literal completion path runs and enumerates all `Literal[...]` overload values. Fixes microsoft/pylance-release#4428.

Before:

image

After:

image


erictraut:
Thanks for the contribution. Since this affects language server functionality, I'll defer to the pylance team. 

It might be worth adding a test case for this new behavior.

@rchiodo, could you or someone else on the pylance team please review this? Thanks!

el1s7:

erictraut:
Also, this is currently failing CI checks.

el1s7:
> Also, this is currently failing CI checks.

It was failing due to a formatter warning, I just fixed it now by running Prettier.

All the tests are also passing locally:
image

image

By the way, you should consider setting a limit for the `jest` max workers. Since running all the tests locally crashed my Linux VM, but after adding the option `-w 1` to the `jest --forceExit` command, it worked fine.

billfelix:
收到，谢谢您的来信。我将及时回复您！ 
祝好！

erictraut:
@rchiodo, are you good with this change?

erictraut:
@el1s7 thanks again for the contribution. Sorry it took so long to get this merged. It will be included in the next release.

Issue #3075, opened 2022-02-18
Lazy analysis for LSP


zsol:
I'm experimenting with `pyright-langserver` in a megarepo with more than 500k Python modules. I realize this is a bit of an unusual use case, so let me try to set 
**some context**:

* The modules are loosely organized into logical groups, and these groups can arbitrarily depend on each other (all facilitated by [buck](https://buck.build/))
* A typical developer only ever needs a small subset of these modules, but the subset can change very dynamically, and is normal to contain around a hundred or so of these logical groups
* The entire repo is not materialized onto disk, only the files needed for a particular task. This is achieved through a virtual filesystem

Our current approach to implementing some parts of the LSP involves [jedi](https://jedi.readthedocs.io/en/latest/), which is designed to do static analysis in a "lazy" way: it only processes source files it definitely needs for a certain operation. This is great for the above environment because it generally doesn't do O(repo) operations unless you ask it to (say, by trying to find all references to a symbol).

**the problem**:
I noticed `pyright-langserver`'s [analyzer](https://github.com/microsoft/pyright/blob/561ec5b093cc09c3b6216c51a161388447b551b4/packages/pyright-internal/src/analyzer/service.ts#L1485) will immediately crawl all files in a workspace when it is registered and then every time the config is changed, which is unfortunate for two reasons:
* As far as I can tell the language server is unresponsive while crawling the directory hierarchy: it doesn't respond to incoming requests/notifications, and not even [this error](https://github.com/microsoft/pyright/blob/561ec5b093cc09c3b6216c51a161388447b551b4/packages/pyright-internal/src/analyzer/service.ts#L1121-L1133) gets sent back to the LSP client
* It generates (in our case unnecessary) load on the system

**potential solution**:
Are there any plans to implement some kind of lazy analysis mode for the language server? There are several very useful LSP requests that could be served without even knowing all files in the workspace: `textDocument/definition`, `textDocument/hover`, `textDocument/documentSymbol`, etc.
This mode would be useful for smaller projects as well, because it could mitigate startup-time latency: go-to-definition could still return results while the entire project is still being processed.

erictraut:
Pyright's type analysis engine is already completely lazy in its analysis. No type file is read, parsed, or analyzed until you open it. All of that work is done "on demand". You can open a large python file and immediately hover over an arbitrary identifier and (typically) get an answer back in under 100ms.

[...]

PR #10786, opened 2025-08-12
Improved startup responsiveness on large monorepos

erictraut:
Improved startup responsiveness on large monorepos by making source file discovery code non-blocking. This addresses #3075.

Without this change, pyright hangs for tens of seconds when used on a large monorepo (>100K files). With this change, pyright's LSP functionality (hover, goto definition, etc.) are responsive almost immediately. Some LSP functions like "find all references" or "workspace symbols" will be inaccurate or incomplete until all project files are enumerated, but these features are not all that usable on a monorepo — at least unless you're using pylance with a populated index.

### DIFF
--- a/packages/pyright-internal/src/analyzer/service.ts
+++ b/packages/pyright-internal/src/analyzer/service.ts
@@ -559,7 +559,11 @@ export class AnalyzerService {
             // the source file enumeration process.
             if (this._sourceEnumerator) {
                 let fileMap: Map<string, Uri>;
-                const maxSourceEnumeratorTime = 1000;
+
+                // Use the "noOpenFilesTimeInMs" limit if it's provided. Otherwise
+                // do all enumeration in one shot. The latter is used for the CLI
+                // and other environments where the user is not blocked on the operation.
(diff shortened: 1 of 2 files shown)


erictraut:
@rchiodo, I'd appreciate a code review for this change. I want to make sure that it doesn't break any pylance functionality.

rchiodo:
Okay thanks. I'll have to pull it into pylance to try it out. I don't think I can really tell just by looking at it.

packages/pyright-internal/src/analyzer/service.ts
```diff
@@ -1267,19 +1320,10 @@ export class AnalyzerService {
...
-
-            for (const file of matchedFiles) {
-                fileMap.set(file.key, file);
-            }
-        });
-
+    // Given a file map returned by the source enumerator, this function
+    // adds any open files that match the include file spec and returns a
+    // final deduped file list.
+    private _getTrackedFileList(fileMap: Map<string, Uri>): Uri[] {
```
bschnurr, line 1326: i wonder if we should change this to return an iterator instead.. for cases where its 100k plus entries.

rchiodo (approved):
Otherwise I tested it with Pylance and it sure seems like things are faster.

PR #10798, opened 2025-08-15
Improved startup performance for large projects by lazily computing t…

erictraut:
…he module name of each source file.


erictraut:
@rchiodo, this optimization may require changes to pylance — in particular, if it provides a custom `ISourceFileFactory`.

PR #10816, opened 2025-08-18
Implemented a number of performance and memory optimizations

erictraut:
* Made implicitImports and filteredImplicitImports fields of ImportResult optional to eliminate the need to create empty maps.
* Avoided creating a new diagnostic rule set for every source file unless it overrides the default rule set.
* Limited memoization of file and web URIs; we were previously allocating and caching an unlimited number of URIs which consumed 100's of Mb of space on large project.
* Changed tokenizer to forcibly clone the value of a string literal so we don't retain a reference to the entire file's text contents.

### DIFF
--- a/packages/pyright-internal/src/analyzer/importResolver.ts
+++ b/packages/pyright-internal/src/analyzer/importResolver.ts
@@ -813,9 +813,7 @@ export class ImportResolver {
         }
 
         const newImportResult = Object.assign({}, importResult);
-        if (filteredImplicitImports.size > 0) {
-            newImportResult.filteredImplicitImports = filteredImplicitImports;
-        }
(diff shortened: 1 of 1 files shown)


erictraut:
@rchiodo, this change should significantly reduce the likelihood of heap exhaustion on large projects (e.g. monorepos). After doing a bunch of memory profiling, I found that `createFileUri` was memoizing every Uri — effectively creating a cache with no upper limit. On large projects with lots of import resolution paths, this resulted in over 1GB of memoized Uri's! I fixed this by limiting the number of memoized Uris to a fixed number (currently 256). Since most cache hits occur in rapid succession, this gives us most of the performance benefits of memoization without a big memory overhead.

PR #10854, opened 2025-08-29
Document editable installs with uv

nickmyatt:


erictraut:
Thanks for the contribution! This is a good tip for users.

@rchiodo, you might want to update the [troubleshooting section](https://github.com/microsoft/pylance-release/blob/main/TROUBLESHOOTING.md#editable-install-modules-not-found) on the pylance-release site to incorporate this update as well.

PR #10888, opened 2025-09-03
Clarify editable install behavior with setuptools and uv

konstin:
Following a user question in uv (https://github.com/astral-sh/uv/issues/15652), clarify the documentation around editables.

From my uv experience, import hooks breaking static type checkers is almost exclusively a setuptools problem (see e.g. https://github.com/astral-sh/uv/issues/1708#issuecomment-2174463476 and https://github.com/astral-sh/uv/issues/3898), so I clarified the docs that this is update using uv with setuptools.

### DIFF
--- a/docs/import-resolution.md
+++ b/docs/import-resolution.md
@@ -41,24 +41,24 @@ Pyright uses the following mechanisms (in priority order) to determine which Pyt
 3. As a fallback, use the default Python environment (i.e. the one that is invoked when typing `python` in the shell).
 
 ### Editable installs
-
-If you want to use static analysis tools with an editable install, you should configure the editable install to use `.pth` files that contain file paths rather than executable lines (prefixed with `import`) that install import hooks. See your package manager’s documentation for details on how to do this. We have provided some basic information for common package managers below.
+If you want to use static analysis tools with an editable install, you should configure the editable install to use `.pth` files that contain file paths rather than executable lines (prefixed with `import`) that install import hooks. Only setuptools is known to use import hooks by default, other build backend generally default to path-based `.pth` files.
 
 Import hooks can provide an editable installation that is a more accurate representation of your real installation. However, because resolving module locations using an import hook requires executing Python code, they are not usable by Pyright and other static analysis tools. Therefore, if your editable install is configured to use import hooks, Pyright will be unable to find the corresponding source files.
 
-#### pip / setuptools
-`pip` (`setuptools`) supports two ways to avoid import hooks:
+#### pip with setuptools
+`pip` with `setuptools` supports two ways to avoid import hooks:
 - [compat mode](https://setuptools.pypa.io/en/latest/userguide/development_mode.html#legacy-behavior)
 - [strict mode](https://setuptools.pypa.io/en/latest/userguide/development_mode.html#strict-editable-installs)
 
-#### uv
-[uv](https://docs.astral.sh/uv/reference/settings/#config-settings) can be
-configured to avoid import hooks with
+#### uv with setuptools
+When using uv with setuptools, uv can be [configured](https://docs.astral.sh/uv/reference/settings/#config-settings) to avoid import hooks:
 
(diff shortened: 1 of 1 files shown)


docs/import-resolution.md
```diff
@@ -41,25 +41,25 @@ Pyright uses the following mechanisms (in priority order) to determine which Pyt
 3. As a fallback, use the default Python environment (i.e. the one that is invoked when typing `python` in the shell).
 
 ### Editable installs
-
-If you want to use static analysis tools with an editable install, you should configure the editable install to use `.pth` files that contain file paths rather than executable lines (prefixed with `import`) that install import hooks. See your package manager’s documentation for details on how to do this. We have provided some basic information for common package managers below.
+If you want to use static analysis tools with an editable install, you should configure the editable install to use `.pth` files that contain file paths rather than executable lines (prefixed with `import`) that install import hooks. Only setuptools is known to use import hooks by default, other build backends generally default to path-based `.pth` files.
```
erictraut, line 44: Nit: The phrase "other build backends..." should be a separate sentence.

Also, I recommend removing "generally" because it's ambiguous in this context. Does it mean that they sometimes don't default to path-based files? Does it mean that a small subset of them don't? If we're confident that setuptools is the only one, then we should say that. If we're not confident of this, then we should probably not call out setuptools specifically.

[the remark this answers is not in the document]
rchiodo: AFAIK other tools do use import hooks. 

Hatch and uv don't, but a bunch of others do.

Here's what copilot gave me when asked for other tools that generate import hooks:

# 🔍 Tools That Create or Use Python Import Hooks

This table summarizes tools and libraries that implement or rely on Python import hooks, typically by modifying `sys.meta_path` or using custom loaders. These hooks enable dynamic behavior during module import, such as loading non-Python files, transforming code, or injecting runtime logic.

| Tool / Library       | Purpose / Use Case                                                                 | Mechanism                     | Link |
|----------------------|-------------------------------------------------------------------------------------|-------------------------------|------|
| **custom-imports**   | Import config files (`.ini`, `.json`, `.csv`) as Python modules                    | Registers custom importers via `sys.meta_path` | [PyPI](https://pypi.org/project/custom-imports/) |
| **imphook**          | Lightweight framework for per-file-type import hooks                               | Custom `MetaPathFinder` and loaders | [GitHub](https://github.com/pfalcon/python-imphook) |
| **importlib**        | Standard library foundation for building custom import hooks                       | Implement `MetaPathFinder`, `PathEntryFinder`, etc. | [Stack Overflow guide](https://stackoverflow.com/questions/43571737/how-to-implement-an-import-hook-that-can-modify-the-source-code-on-the-fly-using) |
| **pyximport**        | Automatically compiles and imports `.pyx` Cython files                             | Adds hook to `sys.meta_path` | [Cython docs](https://cython.readthedocs.io/en/latest/src/tutorial/appendix.html#pyximport) |
| **zipimport / pkgutil / runpy** | Built-in modules that use import hooks internally for zipped packages and dynamic execution | Custom loaders and finders | [zipimport docs](https://docs.python.org/3/library/zipimport.html) |
| **pluggy / stevedore** | Dynamically discover and load plugins                                             | Entry point scanning, sometimes with import hooks | [Pluggy docs](https://pluggy.readthedocs.io/en/latest/) |

---

## 🧠 Notes

- Most tools use `sys.meta_path` to register custom finders/loaders.
- Import hooks are powerful for plugin systems, config injection, and non-standard module formats.
- Setuptools used to embed Python code in `.pth` files for import-time behavior, but modern tools like Hatch and uv avoid this.

konstin: In my user support experience in uv, the problems I saw were always setuptools. That doesn't mean other build backends can't do this, but so far as a gotcha that needs to be specifically documented I've only seen it from setuptools.

erictraut:
Thanks for the update. Looks good.

PR #10891, opened 2025-09-04
Removed "reportShadowedImports" check 

erictraut:
Removed "reportShadowedImports" check and associated "renameShadowedFile" action. This feature is off by default in pyright although enabled by default in pylance.

It's problematic for a number of reasons:
1. It's not a type checking feature. It's a linting feature, so it really don't belong in a type checker.
2. The implementation is very problematic because it introduces an inverted dependency. Namely, it causes the Checker to depend on a language server provider and source mapper.
3. This dependency results in significant performance problems when the check is enabled.

I'd prefer to leave this feature out permanently and ask users to rely on a linter if they want to enable a check like this. If the pylance team feels that this is really important to retain, then it should be reimplemented in a way that doesn't introduce bad dependencies and performance problems.


rchiodo:
@erictraut, we'd like to reimplement this. Are you okay with Pylance having different diagnostics than Pyright? 

I could reimplement it in Pyright but make it much simpler (just have the checker verify the file name isn't one of a known set of system modules). Or we can do the same in Pylance instead but then our errors would diverge.

erictraut:
> Are you okay with Pylance having different diagnostics than Pyright?

That's up to you and the pylance team to decide. It's going to create usability issues, so it's not the approach I would personally recommend.

Some questions I think the pylance team should consider before going this route:
1. Do you agree with my assertion that this is a linter-style check, as opposed to check that belongs in a type checker?
2. Do you want the pylance language server to offer linter-style checks? Historically, it has avoided this and left such checks up to dedicated linters like pylint and ruff.
3. If you want pylance to be in the business of reporting linter-style issues, what principle will you use for deciding which linter checks to include and which are out of scope? Is there something special about this particular check that makes it different from other linter-style checks?

If you end up implementing this in pylance, please consider how developers who use pylance in VS Code and pyright + ruff in their CI can get consistent diagnostics in both places.

rchiodo:
I agree that this is a linter style check and should be part of a linter, but I'm not sure others will know or care what the difference is. I believe we just want it there for newbies and newbies aren't going to have pylint/ruff setup. We have a hard enough time explaining to people just starting out that they need to install a 'Python' extension.

@Luabud do you have any thoughts?

Maybe an alternative here is to finish TSP and get 'Ty' shipping with Pylance. It would show this error because it runs all the diagnostics that ruff does I believe.

rchiodo:
Or another idea is we just ship ruff itself with Pylance.

luabud:
@rchiodo agreed it's a linter style check, but it's a different one since it's not meant to maintain consistency on the code base, it's to prevent users from making unintentional mistakes.

The example we saw happen was a user who named some module after a standard library module name accidentally (I believe it was `copy.py`), which was indirectly getting imported. Unfortunately the module was written as a script to delete a bunch of files, so when they ran their project, it accidentally called that script and wiped out some files. It's an easy mistake to make when you don't understand how the Python import system works.  

I also agree with you that newbies aren't going to have linters set up. I guess the other option would be to implement this on the Python extension side? Maybe @karthiknadig  would disagree but from my perspective it makes sense to have a "set up" warning like this coming from the Python extension, rather than treating this as a linter rule.

Issue #10671, opened 2025-07-02
workspaceEdit missing changeAnnotations


latifboubyan:
Hello,

I'm facing an issue where the `rename` capability is broken because the issue reported here:
https://github.com/neovim/neovim/issues/34731

It seems problem has to do with the fact that when `rename` generates a workspaceEdit, it returns AnnotatedTextEdits but without a changeAnnotations field. So a client trying to use the annotation IDs will break.

[This is a good description of the issue.
](https://github.com/neovim/neovim/issues/34731#issuecomment-3029046310)

latifboubyan:
This was fixed in basedpyright 
https://github.com/DetachHead/basedpyright/pull/1352

erictraut:
I'd accept a PR to fix this issue if you're interested in making a contribution.

stvsmth:
I have been running the suggested change committed in `basedpyright` for a few weeks now. I am in the process of testing/linting and will open a PR presently.

PR #10902, opened 2025-09-06
Fixed bug that causes workspace edits to include annotation IDs without changeAnnotations field.

stvsmth:
Addresses #10671.

Tests/Checks all pass locally.


erictraut:
Thanks for the contribution. The change looks good to me. I'd like to tag someone from the pylance team to review this since this is code they originally contributed to pyright. cc @rchiodo

rchiodo:
Shouldn't the same change be applied to `CreateFile.create` call above? I assume it would cause the same problem.

erictraut:
@stvsmth, please take a look at the question from rchiodo

stvsmth:
Sorry I missed the earlier question, I disabled GitHub email notifications in my last job and never re-enabled them after starting new gig that uses GitLab. Fixed and I should be more responsive.

I'm a Python backend developer; I'm new to hacking on language servers but willing to dig in. My Typescript is passable, more so when the existing code is good.

@rchiodo: Updating `CreateFile.create` seems reasonable. What do we think of the other similar invocations that follow: `RenameFile.create` and `DeleteFile.create`. These seem reasonable to me; however, I haven't taken the time to trigger these errors in Neovim. 

@erictraut ... Do you have a preference of an additional commit? or rebased, one commit (and, while I'm at it, bringing in latest main). I will add in some basic tests.

erictraut:
@stvsmth, I just updated your branch to merge the latest changes from main. No need to rebase. Just submit additional commits to the same branch.

stvsmth:

erictraut:
LGTM. Thanks for the contribution.

Issue #10910, opened 2025-09-09
Pyright fails on files that are too large


alkasm:
Pyright started failing on our repo a few weeks ago, providing incorrect hints in VS Code, though mypy was successfully passing. We noticed that the issue was triggered when we bumped one of the dependencies, which boils down to one very large (~90k line), autogenerated Python file. I then turned this into a min repro with the `pyright` CLI.

**Describe the bug**

With `nominal-api==0.802.0` installed from PyPI, the following script passes a `pyright` type check. With `nominal-api==0.803.0` (and above), the following script fails to type check. However, the code should pass in both cases.

**Code or Screenshots**

```python asdf.py
from nominal_api.module import ModuleApplication, ModuleRef

def f(s0: str, s1: str, s2: str) -> str:
    return s0 + s1 + s2

mod = ModuleApplication("a", "b", "c", ModuleRef("d"), "e")
f(mod.rid, mod.asset_rid, mod.module.rid)
```

To repro, name the above script `asdf.py`, then:

```sh
% uv add pyright nominal-api==0.802.0
% uv run pyright asdf.py --verbose
% uv add nominal-api==0.803.0
% uv run pyright asdf.py --verbose
```

Note that the diff from the `nominal-api` versions 0.802.0 to 0.803.0 is pretty small compared to the file size, and ultimately inconsequential:

<details>

<summary>

```diff 802_impl.py 803_impl.py```

</summary>

```
14800a14801,14805
> 
> 
> class module_LatestVersionStrategy(ConjureBeanType):
>     """This strategy refers to the latest version of the module.
>     """
14801a14807,14810
>     @builtins.classmethod
>     def _fields(cls) -> Dict[str, ConjureFieldDefinition]:
>         return {
>         }
14802a14812,14820
>     __slots__: List[str] = []
> 
> 
> 
> module_LatestVersionStrategy.__name__ = "LatestVersionStrategy"
> module_LatestVersionStrategy.__qualname__ = "LatestVersionStrategy"
> module_LatestVersionStrategy.__module__ = "nominal_api.module"
> 
> 
15492a15511,15624
> 
> 
> class module_PinnedVersionStrategy(ConjureBeanType):
>     """This strategy refers to a specific version of the module.
>     """
> 
>     @builtins.classmethod
>     def _fields(cls) -> Dict[str, ConjureFieldDefinition]:
>         return {
>             'version': ConjureFieldDefinition('version', module_ModuleVersion)
>         }
> 
[...]

PR #10913, opened 2025-09-09
feat: add max complexity configuration option

alkasm:
See #10910


alkasm:
@microsoft-github-policy-service agree company="Nominal, Inc"

erictraut:
This isn't something that should be exposed as a configuration option. Closing.

PR #10945, opened 2025-09-18
Fix mypy_primer to run again

rchiodo:
Mypy_primer suddenly seems to need 3.11. This moves the actions to use 3.11


hauntsaninja:
Thanks for fixing! The context here is:
- https://github.com/hauntsaninja/mypy_primer/pull/209

Issue #129, opened 2019-05-20
pyright could have option to consider .gitignore files


mikest:
**Is your feature request related to a problem? Please describe.**
We often have files excluded in our codebase via .gitignore, this includes intermediary build files and such. It doesn't appear as if pyright considers these filters when doing its scan. This means we have to maintain two instances of the filter list, or a pyright scan will start picking up code files it should not.

**Describe the solution you'd like**
An option to include the .gitignore rules when selecting files. Preferably defaulting to "on"

**Additional context**

return42:
This issue was closed but it is not clear to me if there is an option to include the .gitignore rules when selecting files. The patch 5ec260f that refers this issue number (#129) seems not related to this issue.  Could you please clarify the status of the .gitignore option / thanks a lot!

erictraut:
The ".gitignore" file is specific to git and code management. It has little if any overlap with type checking.

If there are files or directories that you want to exclude from type checking, you can configure that in the "pyrightconfig.json" file. We don't have any plans for pyright to read the ".gitignore" file.

return42:
> We don't have any plans for pyright to read the ".gitignore" file.

Thanks for clarifying.

PR #10959, opened 2025-09-26
Idea for automatically excluding .gitignore entries

rchiodo:
I know this might end up adding a lot of entries to the exclude list, but this discussion item has been on the back of my mind for some time:
https://github.com/microsoft/pylance-release/discussions/7236#discussioncomment-13638782

It makes sense to me that you'd never want .gitignore files to be included in analysis. This PR makes that automatic.


erictraut:
I think this is problematic and would prefer not to include this. The `.gitignore` file is a config file for a different tool, and I don't think we should presume overlap between the two. In general, I think it's a bad idea for one tool to read the config file of another tool. Configuration should be independent. This idea was already discussed [here](https://github.com/microsoft/pyright/issues/129) and rejected.

rchiodo:
Okay thought it was worth a discussion. Didn't realize it had been discussed before.

Issue #10996, opened 2025-10-06
Missing kwargs in signature help when using Unpack (PEP 692) with extra items (PEP 728)


Viicos:
Code sample in [pyright playground](https://pyright-play.net/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMoCCKcANFACqICmAJgCJIDGMZAqiggIaMDWAUAMYAbTgGdRFOgApKCWg2ZkqADxghOAfSQwqEUQF4icAJQAuPlEtROpzChgC%2BNKsCjAArikZSAVD54A7pwgaKK27Fy8ANrkdAC6ZlBc4o58Hl5SxnxAA)

```python
from typing import Any, TypedDict, Unpack


class TD(TypedDict, extra_items=Any):
    a: int


def func(**kwargs: Unpack[TD]): pass



func()

```

<img width="460" height="276" alt="Image" src="https://github.com/user-attachments/assets/0eb2117f-20f7-435b-924c-d615915b896e" />

**VS Code extension or command-line**
pyright

erictraut:
The signature help logic is owned and maintained by the pylance team.

@rchiodo, could you please transfer this to pylance-release? Thanks!

rchiodo:
Do you still want to transfer? @Viicos has submitted a PR in Pyright.

erictraut:
Just noticed that. Leave it for now. Thanks.

erictraut:
This is addressed in pyright 1.1.407

Viicos:
Hum, seems like I can still reproduce the issue.. ([playground](https://pyright-play.net/?pythonVersion=3.14&code=GYJw9gtgBALgngBwJYDsDmUkQWEMoAqiApgCYAiSAxjADRQCqKCAhlQNb0ByYMASsQCOAVyQgyAKClUANiwDO8qAFkwANyTEAFEQRlKNesQAeMECwD6SGMQjyAvKhgBKAFwSonqO2JwAjK5Q8mYeXj5wAEyBPPxCouKkANpOALrScopQAILuXlCkxMCwJKSk1DBa8sQywPQsgU70AEYNKHRQAFQd7ADuLCBo8oFMrByJqhrEKc5QALQAfFA8KMSBAHQbUhLGUPbZWs5SxmvweqXlB0A)). Perhaps my fix wasn't targeting the right logic?

erictraut:
Seems to be working.

[...]

PR #10997, opened 2025-10-06
Fixed bug that omit `**kwargs` parameter with an unpacked TypedDict with extra items.

Viicos:
This addresses #10996. I assume this should fix it?


Viicos:

erictraut:
@Viicos, thanks for the contribution. The logic looks good to me. Once you fix the style issue, it should be good to merge.

PR #11027, opened 2025-10-13
Warn when `self` or `cls` parameter has a default value

decorator-factory:
Pyright currently does not raise any diagnostics on this:

```py
class Foo:
    def bar(self=42) -> None:  # allowed
        pass

foo = Foo()
foo.bar()  # also allowed
```

Having a default value for the first parameter of a method or classmethod is generally dubious, and is probably a typo from hastily editing the code:
```diff
- def bird(self, ex_parrot=42):
+ def bird(self=42):
```

### DIFF
--- a/packages/pyright-internal/src/analyzer/checker.ts
+++ b/packages/pyright-internal/src/analyzer/checker.ts
@@ -7156,6 +7156,7 @@ export class Checker extends ParseTreeWalker {
             }
 
             this._validateClsSelfParamType(node, functionType, classType, /* isCls */ true);
+            this._disallowDefaultInClsSelfParam(node);
             return;
         }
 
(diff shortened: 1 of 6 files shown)


erictraut:
Is this targeting an existing issue in the issue tracker? I don't see one. We generally want to see an issue filed and triaged before a fix is considered.

I don't think the proposed fix is correct. Disallowing all defaults for the `self` or `cls` parameter is more of a linting rule, not a type checker rule. Implementing this as a type checking rule will be more involved. I'm not (yet) convinced that this problem is prevalent to justify such a check. This is unlikely to be a common source of bugs.

jorenham:
`42` isn't assignable to `Foo`. But in `def bar(self=42) -> None:`, the type of `self` is `Foo` (via `self: Self`). So why is `self=42` allowed right now?

PR #11055, opened 2025-10-23
task scheduling queue

khresth:
Implement priority based task scheduling for background analysis via PriorityTaskScheduler with min-heap implementation. Replaced the direct worker.post Message with priority queue system.


rchiodo:
I don't think this actually works. 

For it to have worked you'd have to enqueue more than one thing at a time and that's impossible given the entire queue drains as soon as it adds something. 

We implemented something similar to this in Pylance and it uses another worker thread as the priority scheduler. This allows us enqueue requests independent of scheduling them.

PR #11177, opened 2025-12-17
Fix completions for unpacking TypeDicts with missing fields

rchiodo:
Pylance has a bug that I thought I'd look into: https://github.com/microsoft/pylance-release/issues/6732

In error cases, completions don't come up. 

```
from typing import Unpack, TypedDict

class InnerDict(TypedDict):
    a: int
    b: str

class OuterDict(TypedDict):
    inner: InnerDict
    field_1: str

def test(**kwargs: Unpack[OuterDict]):
    pass

#  Completions for "a", "b" inside {} don't show up because field_1 is not present
test(inner={})

# Completions do work when field_1 is present
test(field_1="x", inner={})
```

The fix here was to include the inference context when validating arguments.


erictraut (approved):
I didn't have a chance to test it, but the code looks good to me.

PR #11180, opened 2025-12-19
Fix Type Narrowing for Asymmetric Attributes After Assignment

mfish33:
This PR fixes the issue where an asymmetric attribute can not be narrowed after it has been assigned in a particular scope.

Take the following for example:
```python

class MyAttr:
   def __set__(self, instance: object, value: float) -> None:
      ...

   def __get__(self, instance: object, owner: type) -> int | None:
      ...

class MyClass:
   attr: MyAttr

MyClass.attr = 1.5

# Should be able to narrow types after initial assignment
assert MyClass.attr is not None
assert_type(MyClass.attr, int)
```

The current version of Pyright fails since it refuses to narrow `MyClass.attr`. The solution to this is to save the original type of `MyClass.attr` when an asymmetric assignment is found. This type can then be retrieved later during flow analysis.


erictraut:
Pyright is working as intended here. This PR modifies its behavior in a way that is incorrect. For an explanation of the current behavior, refer to [this documentation](https://microsoft.github.io/pyright/#/mypy-comparison?id=type-narrowing-for-asymmetric-descriptors).

mfish33:
This PR does not implement narrowing based on assignment. The current implementation prevents narrowing the type after assignment. This is a bug.

The following currently works fine with Pyright:
```python

class MyAttr:
   def __set__(self, instance: object, value: float) -> None:
      ...

   def __get__(self, instance: object, owner: type) -> int | None:
      ...

class MyClass:
   attr: MyAttr

# When the assignment is commented out the following type narrowing works as expected
# MyClass.attr = 1.5

# Should be able to narrow types after initial assignment
assert MyClass.attr is not None
assert_type(MyClass.attr, int)
```

erictraut:
I still don't consider the current behavior a bug, but I think there is an argument to be made for supporting type narrowing based on type guards for asymmetric attributes. 

I'd appreciate it you'd file a feature request so we can discuss it there before posting a PR.

I've reopened the PR for now.

mfish33:
Hi Eric, happy to open a bug. However, I am unsure if it should be considered a feature request. As I showed above, type guard narrowing is currently supported by Pyright except for when you try to do it after an assignment.

mfish33:
Hi Eric, I am gently bumping this. Surprisingly, this is quite a typical code pattern here at Arista and is blocking wider Pyright adoption.  I have even run mypy-primer on the change, and it found one instance of this issue in the wild.

mypy_primer run: https://github.com/mfish33/pyright/pull/2
Issue in Colour: https://github.com/colour-science/colour/blob/develop/colour/io/tm2714.py#L1014. `self.path` is an asymmetric attribute that gets set in the same scope and thus can not be later narrowed.

mfish33:
@rchiodo  Is there a chance you can take a look at this?

rchiodo:
@mfish33, it looks good to me, but Eric would be the one that needs to approve. He's the owner/expert in this area.

erictraut:
As I mentioned before, I'd like to discuss this in a feature request before considering a PR.

mfish33:
As I also mentioned above, I am happy to open a bug/ft request, but I feel pretty strongly that this is indeed a bug. As demonstrated in my previous comments and with the mypy-primer run, this PR is not introducing a new feature, but rather fixing a bug in a corner case in an existing feature that is not being handled properly. @erictraut, I strongly encourage you to look at the diff as it is self-evident of these facts.

mfish33:
Created #11233

erictraut:
Please refer to my comments in the feature request.

PR #11209, opened 2026-01-07
Auto-detect src paths for execution environments with autoSearchPaths enabled

heejaechang:
## Description

Fixes an issue where `python.analysis.autoSearchPaths` only auto-detected and added `<project-root>/src` to extraPaths, but did not apply the same logic to execution environment roots. When a project has multiple execution environments with their own `src` directories (e.g., `env1/src`, `env2/src`), these were not being automatically added to the search paths.

This change extends the auto-detection behavior to also check for and add `<executionEnvironment.root>/src` directories when:
- `autoSearchPaths` is enabled
- Neither the top-level config nor the execution environment explicitly specifies `extraPaths`
- The `src` directory exists and is not a package (no `__init__.py`)

This mirrors the existing behavior for project-root `src` directories and helps maintain consistent import resolution across multi-environment projects.

The investigation focused on two key areas:
1. **`configOptions.ts`**: Located the existing auto-detection logic in `initializeFromJson()` that handles project-root `src` paths. The logic checks for `autoSearchPaths`, verifies the `src` directory exists, and ensures it's not a package before adding it.
2. **`setupExecutionEnvironments()`**: Extended this method to apply similar auto-detection logic to each execution environment's root directory.

The implementation needed to respect the suppression mechanism where explicitly setting `extraPaths` to an empty array (either at config or environment level) prevents auto-detection.

## Implementation

Key changes:

1. **Refactored auto-detection logic** into a reusable static method `_tryGetAutoDetectedSrcExtraPath()` that:
   - Takes a root URI and checks for a `src` subdirectory
   - Returns the path only if `src` exists and is not a package
   - Returns `undefined` otherwise

2. **Extended `setupExecutionEnvironments()`** to:
   - Accept `fs` (FileSystem) and `autoSearchPaths` parameters
   - Track whether `extraPaths` was specified at the config level
   - For each execution environment with a root, attempt auto-detection if neither config-level nor env-level `extraPaths` were specified
   - Deduplicate to avoid adding the same path twice

3. **Updated call sites** in `service.ts` and `testState.ts` to pass the required parameters

## Testing

Added comprehensive test coverage in `config.test.ts` with the following scenarios:

1. **`AutoSearchPathsOnWithExecutionEnvironmentRoots`**: Verifies auto-detection adds `src` paths for multiple execution environments
2. **`AutoSearchPathsOnWithExecutionEnvironmentRootsConfigExtraPathsEmptySuppresses`**: Ensures empty `extraPaths` at config level suppresses auto-detection
3. **`AutoSearchPathsOnWithExecutionEnvironmentRootsEnvExtraPathsEmptySuppresses`**: Ensures empty `extraPaths` at environment level suppresses auto-detection
4. **`AutoSearchPathsOnWithExecutionEnvironmentRootsSrcIsPkgSuppresses`**: Verifies `src` directories containing `__init__.py` are not added
5. **`AutoSearchPathsOnWithExecutionEnvironmentRootDotDedupes`**: Ensures no duplicates when execution environment root is "." (project root)

Test fixtures include sample projects with various configurations to cover all edge cases.

## Addresses

Addresses microsoft/pylance-release#7378


erictraut:
This change doesn't make sense to me. The `autoSearchPaths` mechanism was added to support pylance scenarios for users with no interest in type checking and a low threshold for adding custom settings. Execution environments are a power user feature for users who are using type checking and want to highly customize their configuration. If someone has a `src` subdirectory in a monorepo, they should specify that as the execution environment root, not its parent directory.

heejaechang:
alright, I will close the issue then. thank you!

flying-sheep:
@erictraut I disagree, if someone has a `src` directory, there’s almost no case when they don’t want that, so making users configure this manually is kind of a needless chore.

> specify that as the execution environment root

what does that even mean?

Pylance still does weird things when I add `src/**/*` to `python.analysis.include` as documented in https://github.com/microsoft/pylance-release/blob/main/docs/settings/python_analysis_include.md

It adds this when I have it add an import, which almost never makes sense:

```py
from ...src.anndata.utils import asarray
```

PR #11211, opened 2026-01-07
Update typeshed to latest

rchiodo:
Also add script to do this automatically


erictraut:
Historically, I did this by hand because:
1. It's important to stay on top of changes in key stubs like `typing.pyi` and `builtins.pyi`.
2. We don't want to ship the `tensorflow` stubs from typeshed because they are very incomplete and are worse in many respects than having no stubs. So I always excluded that.

PR #11231, opened 2026-01-14
Refactor import resolver infrastructure and enhance file system interactions

heejaechang:
This PR is `to gather feedback` before we make these changes in Pylance and push them to Pyright. `It is not intended to be checked into Pyright directly`.

The main purpose of this PR is to improve Jest test performance and provide knobs to control performance in tests.

This change mainly includes two updates.

Improvements to `TestFileSystem` around mounting real folders to our test virtual file system. It now provides related data at once rather than gathering information one by one through separate calls, and mounts more lazily by supporting mounting a file rather than an entire folder.

Extracting caches in the import resolver as services so they can be mocked or use different implementations in tests. Usually, we create a new import resolver for each test, and most of those tests share the same typeshed, site-packages, etc., but we rebuild them (stdlib version map, typestub root maps, etc.) repeatedly. Each build is cheap (10–100 milliseconds), but doing this a few thousand times is expensive. `The actual implementation of the services is a 1:1 match with the previous implementation`, and in Pyright it always uses the default implementation, which is the same as before.

- Here is the list of changes:

Updated `FileSystemResolver` to return detailed directory entries, including kind and size.
Improved `TestFileSystem` to lazily populate links for mounted directories, optimizing path traversal.
Introduced `ImportResolverFileSystem` to cache directory enumeration and file existence checks, reducing IO operations.
Implemented `TypeshedInfoProvider` to manage typeshed paths and version information, with caching for efficiency.

Added comprehensive unit tests for the new import resolver infrastructure, ensuring correct behavior and cache management.

...

`TestFileSystem` changes made each jest test to be up to a few seconds faster (depends on how many tests are in the jest file)
`ImportResolver` changes made the target language server tests that enabled the mocked service a few seconds faster. combined with TestFileSystem improvement, ~10 seconds faster.

* it also has a test knob for `partial stubs` but pyright doesn't have bundled `partial stub` so it shouldn't matter for pyright's test.

### DIFF
--- a/packages/pyright-internal/src/analyzer/importResolver.ts
+++ b/packages/pyright-internal/src/analyzer/importResolver.ts
@@ -8,25 +8,27 @@
  * runtime rules of Python.
  */
 
-import type { Dirent } from 'fs';
-
 import { appendArray, flatten, getMapValues, getOrAdd } from '../common/collectionUtils';
 import { ConfigOptions, ExecutionEnvironment, matchFileSpecs } from '../common/configOptions';
 import { Host } from '../common/host';
 import { stubsSuffix } from '../common/pathConsts';
 import { getFileExtension, stripFileExtension } from '../common/pathUtils';
-import { PythonVersion, pythonVersion3_0 } from '../common/pythonVersion';
+import { PythonVersion } from '../common/pythonVersion';
 import { ServiceProvider } from '../common/serviceProvider';
+import { ServiceKeys } from '../common/serviceKeys';
 import * as StringUtils from '../common/stringUtils';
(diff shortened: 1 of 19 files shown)


heejaechang:
@erictraut can you take a look?

erictraut:
Sorry, I won't have time to review this in the near future.

heejaechang:
Although the PR appears to include many changes, most are related to testing rather than the product itself. The modification in ImportResolver affects only testing and has no impact on the actual product. It's primarily a refactoring that moves existing code from one location to another.

Given that, the change should be safe and have no effect on the product. We're considering moving forward with it.

erictraut:
Sounds good to me.

PR #11239, opened 2026-01-19
Fix workspace initialization when workspaceFolders capability is enabled

jacks0n:
## Summary
- Fix `handleInitialized()` to always call `updateSettingsForAllWorkspaces()`, not just when `hasWorkspaceFoldersCapability=false`
- Add test for workspace initialization with `workspaceFolders` capability enabled

## Problem
When `hasWorkspaceFoldersCapability=true`, `handleInitialized()` never called `updateSettingsForAllWorkspaces()`, leaving workspaces created during `initialize()` with unresolved `isInitialized.promise`. Any code awaiting this promise would hang forever.

## Solution
Restructure the method to:
1. Conditionally register the workspace folders change handler (when capability is enabled)
2. Always call `dynamicFeatures.register()` and `updateSettingsForAllWorkspaces()`

## Test plan
- [x] New test `Initialize with workspace folder support` verifies hover requests work after initialization with `workspaceFolders=true`
- [x] Existing test `Initialize without workspace folder support` still passes
- [x] Test hangs/times out before the fix, passes after

### DIFF
--- a/packages/pyright-internal/src/languageServerBase.ts
+++ b/packages/pyright-internal/src/languageServerBase.ts
@@ -702,16 +702,13 @@ export abstract class LanguageServerBase implements LanguageServerInterface, Dis
         // not send config updates before this point.
         this._initialized = true;
 
-        if (!this.client.hasWorkspaceFoldersCapability) {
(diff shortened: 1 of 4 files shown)


jacks0n:

PR #11260, opened 2026-01-31
Add pyrighht caching

jdave-zip:


jdave-zip:
@microsoft-github-policy-service agree [company="ZipHQ"]

jdave-zip:
@microsoft-github-policy-service agree company="ZipHQ"

erictraut:
For a feature of this size and magnitude, please start with a discussion in a feature request rather than submitting a bunch of code.

We've explored persistent caches in the past and were not satisfied with the tradeoffs. I'll need a lot of evidence to be convinced that this is something that should be added. There are many downsides to persistent caches, so there would need to be commensurate benefits to justify such a feature.

PR #11262, opened 2026-02-02
Microoptimize memoization without args

Zaczero:


erictraut:
If it doesn't have a measurable impact on performance, then it's not worth the regression risk.

PR #11263, opened 2026-02-02
Optimize findClosestMatch complexity

Zaczero:
This change improved the performance on openstreetmap-ng codebase by about 6%.

14.5s → 13.6s (-6.6% ± 0.6%)

- Old algorithm: O(n) per query (n = map.size), even though only a handful of ancestors could possibly match.
- New algorithm: O(depth) map lookups (depth = number of path segments, usually small), and returns as soon as the best match is found.

The correctness is preserved because “best match” is defined purely in terms of ancestor relationship + maximum path length, and walking ancestors from deepest to root returns the same maximum-length ancestor when present.


Zaczero:

rchiodo (approved):

Zaczero:
I was checking for possible regressions and I noticed that packages/pyright-internal/src/common/uri/constantUri.ts has isRoot that always returns false. So I updated the stop condition to work universally and abort when the URI is no longer changing.

PR #11264, opened 2026-02-02
Microoptimize time measurements

Zaczero:
Removes Duration object allocations in TimingStat by switching to Date.now() deltas; same semantics, less GC churn.


erictraut:
Not worth regression risk.

PR #11266, opened 2026-02-02
Avoid double fileContents hashing and fix lastFileContentHash=0 bug

Zaczero:
Avoid SourceFile double file hashing on read by passing down the computed hash.

Additionally fix lastFileContentHash=0 comparison bug by replacing || with ??.


erictraut:
It's not clear from your commit notes whether you're fixing a bug. For bug fixes, please include a link to an active bug report. If this isn't a bug fix and it doesn't improve performance, then I think we'll pass on it.

Zaczero:
The bug is that || recalculates hash unnecessarily when it's set to zero. So it's all performance oriented, no behavioral change. It's a small bugfix that integrates nicely with this PR (same scope, same idea: avoid double hashing).

erictraut:
Thanks, I think we'll pass on this.

Issue #11253, opened 2026-01-28
Running pyright with --threads will exit with success when it OOMs


tjni:
**Describe the bug**

This is related to https://github.com/microsoft/pyright/issues/11252. In my code base, we had some code trigger an OOM. That code is really very weird and I'm not convinced we should keep it. But the bigger issue we ran into was that we were running `pyright` with `--threads 4` and that returned a successful error code.

**Code or Screenshots**

In the reproducer in https://github.com/microsoft/pyright/issues/11252, run as-is and check the exit code. It will be non-zero. Now modify the script to add `--threads 4` to the pyright invocation and run it. It will report an exit code of zero.

PR #11274, opened 2026-02-07
Convert worker OOM to a fatal error.

tjni:
We can attach an exit handler to the child process and detect when an exit was unexpected. In that situation, we report a fatal error exit code.

Closes https://github.com/microsoft/pyright/issues/11253.


erictraut:
Thanks for the contribution.

Issue #9190, opened 2024-10-10
Pyright does not infer generic class parameter when specified as argument to class method


superlopuh:
Here's a simple example, T is completely specified, yet Pyright complains:
``` python
from typing import Generic, TypeVar

T = TypeVar("T")

class GenericClass(Generic[T]):

    @classmethod
    def method(cls, param: T) -> T:
        return param

val = GenericClass.method(1)
```

Pyright errors:
```
pyright xdsl/dialects/bla.py
12:1 - error: Type of "val" is unknown (reportUnknownVariableType)
12:7 - error: Type of "method" is partially unknown
    Type of "method" is "(param: Unknown) -> Unknown" (reportUnknownMemberType)
2 errors, 0 warnings, 0 informations 
```

With `pyright 1.1.384`

erictraut:
Pyright's behavior here is correct, so this isn't a bug.

You've defined a class-scoped type variable here, so the type argument is associated with the class (`GenericClass`), not with the method (`method`). Any class-scoped type variables that appear within a method's signature are specialized at the time you access the method through the class. In this case, the class is `GenericClass[Unknown]` because you haven't provided an explicit type argument. If you had provided a default value for the type variable (as per [PEP 696](https://peps.python.org/pep-0696/)), then `GenericClass` would be specialized with that default value. Since there was no default specified, the default is implicitly `Unknown`. That means the type of `GenericClass.method` is `(param: Unknown) -> Unknown`. There is no type variable to solve in this case, so it doesn't matter what arguments you pass.

There's only one case specified in the typing spec where class-scoped type variables can be inferred when calling a method, and that's in the case of a constructor call. This special case doesn't apply to class methods.

Issue #7721, opened 2024-04-17
 Pyright unable to infer the type of Generic classmethod


sunfkny:
**Describe the bug**
Since version 1.1.354, Pyright unable to infer the type of a generic class method in certain scenarios.

**Code or Screenshots**
```python
from typing import Generic, TypeVar, reveal_type

T = TypeVar("T")

class Option(Generic[T]):
    def __init__(self, value: T):
        self._val = value

    @classmethod
    def Some(cls, val: T) -> "Option[T]":
        return cls(val)

reveal_type(Option(1))           # Pyright correctly infers Option[int]
reveal_type(Option.Some(2))      # Pyright incorrectly infers Option[Unknown]

```

[...]

Issue #11187, opened 2025-12-23
Problem inferring nested generic classes with defaults


cmp0xff:
## Describe the bug
`pyright` cannot infer nested generic classes with default values properly.

## Code or Screenshots

```python
from typing import Any, Generic
from collections.abc import Sequence
from typing_extensions import TypeVar, Self, reveal_type

_OrderableT = TypeVar("_OrderableT", default=Any)

class Interval(Generic[_OrderableT]):
    def __new__(cls, left: _OrderableT, right: _OrderableT) -> Self: ...

IntervalT = TypeVar("IntervalT", bound=Interval, default=Interval)

S2 = TypeVar("S2")

class IntervalIndex(Generic[S2]):
    def __new__(cls, data: Sequence[IntervalT]) -> IntervalIndex[IntervalT]: ...

reveal_type(Interval(0, 1))  # type checkers all give Interval[int]
reveal_type(IntervalIndex([Interval(0, 1)]))  # mypy gives IntervalIndex[Interval[int]], pyright gives IntervalIndex[Interval[Any]]
```

## VS Code extension or command-line

Command line, `pyright==1.1.407`

just-maiyak:
Hello !

I think I have a concrete example of this bug in the library `psycopg3`:

```py
import psycopg
from psycopg.rows import dict_row

def test():
    with psycopg.connect(
        row_factory=dict_row # This fails to typecheck
    ) as conn:
        pass
```

I have reproduced a minimal setup for this (mimicking what is done in the library):

```py
from typing import Any, Generic, TypeAlias, TypeVar, Protocol


TupleRow: TypeAlias = tuple[Any, ...]
DictRow: TypeAlias = dict[str, Any]
Row = TypeVar("Row", covariant=True, default="TupleRow")


class RowMaker(Protocol[Row]):
    def __call__(self, __test: Any) -> Row: ...


class RowFactory(Protocol[Row]):
    def __call__(self, __test: Any) -> RowMaker[Row]: ...


class Connection(Generic[Row]):
    @classmethod
    def connect(
        cls,
        conninfo: str = "",
        *,
        row_func: RowFactory[Row] | None = None
    ) -> None:
        pass


def dict_row(__x: Any) -> RowMaker[DictRow]: ...

Connection.connect(row_func=dict_row) # Does not typecheck under pyright but does with mypy
```

I get the following result upon inspection with pyright :

```
/Users/just.maiyak/Projects/rnd/submarine/test_pyright.py
[...]

PR #11280, opened 2026-02-13
Fix classmethod inference for TypeVars with PEP 696 defaults

jacks0n:
When calling a classmethod on an unparameterized generic class where a TypeVar has a PEP 696 default, `specializeWithDefaultTypeArgs` eagerly resolves the TypeVar before argument matching, preventing inference from classmethod arguments.

For example:
```python
T = TypeVar("T", default=int)

class ClassA(Generic[T]):
    @classmethod
    def create(cls, *, factory: Factory[T] | None = None) -> Self: ...

# Expected: ClassA[str], got: ClassA[int]
v = ClassA.create(factory=str_factory)
```

This fix defers default specialization for classmethods on classes with explicit TypeVar defaults, allowing inference from arguments. Unsolved TypeVars still receive their defaults after argument matching via `constructorTypeVarScopeId`, following the same pattern used for `__init__` / `__new__`.

Only affects classes with PEP 696 explicit defaults (`isDefaultExplicit`). Unrelated to #9190 / #7721 which involve classmethods without TypeVar defaults.

## Changes

- **typeEvaluator.ts**: Skip `specializeWithDefaultTypeArgs` for classmethods when the class has TypeVars with PEP 696 explicit defaults, deferring specialization to argument matching
- **types.ts**: Extend `constructorTypeVarScopeId` propagation to classmethods (previously only `__init__` / `__new__`), so unsolved TypeVars receive their defaults after inference
- **typeVarDefaultClass5.py**: Covers factory classmethod, direct parameter, mixed TypeVars, subclass inheritance, and instance method non-regression

### DIFF
--- a/packages/pyright-internal/src/analyzer/typeEvaluator.ts
+++ b/packages/pyright-internal/src/analyzer/typeEvaluator.ts
@@ -2340,16 +2340,40 @@ export function createTypeEvaluator(
         }
 
         // If this is an unspecialized generic class, specialize it using the default
-        // values for its type parameters.
+        // values for its type parameters. Skip this if we're suppressing the use
+        // of attribute access override, such as with dundered methods (like __call__).
         if (
             isInstantiableClass(objectType) &&
             !objectType.priv.includeSubclasses &&
             objectType.shared.typeParams.length > 0
         ) {
-            // Skip this if we're suppressing the use of attribute access override,
-            // such as with dundered methods (like __call__).
             if ((flags & MemberAccessFlags.SkipAttributeAccessOverride) === 0) {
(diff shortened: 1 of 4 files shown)


erictraut:
I don't think this change is correct. Pyright's current behavior is in compliance with the typing spec.

erictraut:
We recommend opening a bug report or feature request and discuss changes with maintainers before creating a PR.

just-maiyak:
This has been flagged as a bug in #11187 but had no recent activity from maintainers, Mypy and Pyright are not consistent on this typing spec from my understanding.

PR #11285, opened 2026-02-17
Preserve diagnostic rule overrides when positional args override include

ptarjan:
# Summary
 
When pyright is invoked with positional directory arguments, diagnostic rule overrides from `pyrightconfig.json` are silently ignored.
 
**Works correctly** (config overrides applied):
```bash
pyright --project pyrightconfig.json
```
 
**Broken** (config overrides ignored):
```bash
pyright --project pyrightconfig.json src/
```
 
# Reproduction steps
 
1. Create `pyrightconfig.json`:
```json
{
  "include": ["src"],
  "reportPrivateImportUsage": false
}
```
2. Create `src/mymodule/__init__.py` that re-exports from `src/mymodule/_internal.py`
3. Create `src/consumer.py` that imports from `mymodule._internal`
4. `pyright --project pyrightconfig.json` — no errors (correct)
5. `pyright --project pyrightconfig.json src/` — reports `reportPrivateImportUsage` error (incorrect)
 
# Root cause
 
Positional args flow through `pyright.ts:253-272` → `commandLineOptions.includeFileSpecsOverride` → `service.ts:1152-1158` which replaces `configOptions.include`. The config file is still parsed and `configOptions.diagnosticRuleSet` has the correct overrides.
 
The problem is in `Program.addTrackedFile()`: the new `SourceFile._diagnosticRuleSet` is initialized to `getBasicDiagnosticRuleSet()`. The correct rule set from the execution environment is only applied during `parse()`. For tracked files, diagnostics can be consulted before `parse()` applies the execution environment rules, so basic defaults (e.g. `reportPrivateImportUsage: "error"`) take effect instead of the config's overrides (`"none"`).
 
# Fix
 
Add `SourceFile.setInitialDiagnosticRuleSet()` and call it from `Program.addTrackedFile()` immediately after construction. Only `addTrackedFile` is changed — other `createSourceFile` call sites don't need this because imported/interim files always go through `parse()` before diagnostics are checked. The setter avoids changing `ISourceFileFactory`, keeping it non-invasive for third-party API consumers.
 
# Changes
- `sourceFile.ts`: Add `setInitialDiagnosticRuleSet()` public method
- `program.ts`: Call the setter in `addTrackedFile()` after construction
- `privateImportUsage.test.ts`: End-to-end test reproducing the bug
- `sourceFile.test.ts`: Unit test for the setter

### DIFF
--- a/packages/pyright-internal/src/analyzer/sourceFile.ts
+++ b/packages/pyright-internal/src/analyzer/sourceFile.ts
@@ -11,7 +11,12 @@ import { isMainThread } from 'worker_threads';
 
 import { OperationCanceledException } from '../common/cancellationUtils';
 import { appendArray } from '../common/collectionUtils';
-import { ConfigOptions, DiagnosticRuleSet, ExecutionEnvironment, getBasicDiagnosticRuleSet } from '../common/configOptions';
+import {
+    ConfigOptions,
+    DiagnosticRuleSet,
+    ExecutionEnvironment,
+    getBasicDiagnosticRuleSet,
+} from '../common/configOptions';
 import { ConsoleInterface, StandardConsole } from '../common/console';
 import { assert } from '../common/debug';
 import { Diagnostic, DiagnosticCategory, TaskListToken, convertLevelToCategory } from '../common/diagnostic';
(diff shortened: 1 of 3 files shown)


ptarjan:
@microsoft-github-policy-service agree [company="Anthropic"]

ptarjan:

PR #11302, opened 2026-03-02
Made PEP 747 (Annotating Type Forms) no longer experimental

Viicos:
as it was recently accepted: https://discuss.python.org/t/55984/103.

### DIFF
--- a/docs/features.md
+++ b/docs/features.md
@@ -38,7 +38,7 @@ Pyright supports [configuration files](configuration.md) that provide granular c
 * [PEP 728](https://www.python.org/dev/peps/pep-0728/) TypedDict with typed extra items
 * [PEP 742](https://www.python.org/dev/peps/pep-0742/) narrowing types with TypeIs
 * [PEP 746](https://www.python.org/dev/peps/pep-0746/) (experimental) type checking annotated metadata
-* [PEP 747](https://www.python.org/dev/peps/pep-0747/) (experimental) annotating type forms
+* [PEP 747](https://www.python.org/dev/peps/pep-0747/)  annotating type forms
 * [PEP 764](https://www.python.org/dev/peps/pep-0764/) (experimental) inline typed dictionaries
 * Type inference for function return values, instance variables, class variables, and globals
 * Type guards that understand conditional code flow constructs like if/else statements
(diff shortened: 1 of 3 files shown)


rchiodo:
@Viicos, it looks like the changes are causing a single test failure.

Viicos:
Looks like a bug, that can be reproduce with the following MRE:

```python
from typing import NamedTuple

# X is necessary to reproduce. Doesn't reproduce with only "a" as a field
X = NamedTuple("X", [("a", int), ("b", str)])
X(1, "1")


class Z(NamedTuple):
    c: list[str]
    d: list[int]


q6 = Z(["1"], [3])  # Error: Class definition for "Z" depends on itself

# This is necessary to reproduce
for a, b in zip(*q6):  # needs to be named a and b to reproduce
    pass
```

<details>

<summary>LLM analysis</summary>

  The `isTypeFormSupported` guard on `getTypeOfStringList` (line 1715) was the critical one. By removing it, the commit causes every string literal to be    
  speculatively evaluated as a type expression to extract `TypeForm` info.                                                                                 
                                                                                                                                                         
  The chain of events:                                                                                                                                   
                                                                                                                                                         
  1. `X = NamedTuple("X", [("a", int), ("b", str)])` is being processed                                                                                    
  2. The string "a" is now speculatively parsed as a type expression (line 1739: `getTypeOfStringListAsType`)                                              
  3. "a" is parsed as the identifier a, which resolves to the for-loop variable a from for a, b in `zip(*q6)`:                                             
  4. To infer a's type, pyright evaluates the for-loop expression `zip(*q6)
`  5. This requires q6's type → code flow follows to `q6 = Z(["1"], [3])
`  6. Evaluating `Z(["1"], [3])` triggers `Z`'s class evaluation → `Z` is marked `PartiallyEvaluated`
  7. `validateConstructorArgs` calls `validateMetaclassCall` first (constructors.ts:132), which calls `getBoundCallMethod` → getTypeOfBoundMember(Z,
  "__call__")
  8. Z is still `PartiallyEvaluated` → cycle error

  The reason both X and the zip call are needed:
  - X is needed because its "a" string literal is what triggers the speculative type evaluation
  - *zip(q6) is needed because the for-loop variable a is what the identifier a resolves to, and evaluating that loop requires Z's constructor

  The fix should restore the `isTypeFormSupported` guard specifically for the string-as-TypeForm path, or find a way to avoid the speculative type
  evaluation causing this cascade.

</details>

rchiodo:
Closing as superseded by microsoft/pyright#11412 (Enable TypeForm support by default), merged on June 25, 2026. That change removed the TypeForm experimental gates and addressed the string-literal evaluation issue discussed here; current main also already includes the sentinel change. The remaining difference is the stale experimental label for PEP 747 in docs/features.md, which will be handled in a separate documentation-only PR. Thanks for the original contribution!

PR #11323, opened 2026-03-10
Fallback to root `pyproject.toml` if project `pyproject.toml` doesnt contain `[tool.pyright]`

### LINKED ISSUE #11320: `--project` as a directory should ignore `pyproject.toml` without `tool.pyright`
# Feature Request

We have a monorepo setup (root `pyproject.toml` with `[tool.pyright]` with the "default" configuration all projects _should_ inherit (unless they choose to override).

For typechecking, we run `pyright` "per-project-directory" using `--project` (to ensure the correct `pyproject.toml` gets loaded for that project).

This all works if the project either:
- Doesn't have a `pyproject.toml` (uses root config ✅ )
- Has a `pyproject.toml` and has a `[tool.pyright]` (ideally using `extends = "../../pyproject.toml"`, but that's on us to enforce) ✅ 

However, some projects have a `pyproject.toml` for non-pyright reasons (specifying project metadata, configuring other tools like `ruff`). And for these projects, the mere existence of `pyproject.toml` forces `pyright` to load config from it, of which there is no config, and therefore we get the `pyright` defaults.

## Example bad behavior

Given

```
.
├── pyproject.toml                // [tool.pyright].reportUnnecessaryTypeIgnoreComment = true
├── no-pyprojecttoml
│   └── file.py                   // x = 1  # type: ignore
├── pyprojecttoml-with-config
│   ├── file.py                   // x = 1  # type: ignore
│   └── pyproject.toml            // [tool.pyright].reportUnnecessaryTypeIgnoreComment = false
└── pyprojecttoml-without-config
    ├── file.py                   // x = 1  # type: ignore
    └── pyproject.toml            // (empty)
```

Then:
- `uv tool run pyright --verbose --project pyprojecttoml-without-config pyprojecttoml-without-config/file.py
  - Loads root `pyproject.toml` and therefore reports unnecessary type ignore ✅ 
- `uv tool run pyright --verbose --project pyprojecttoml-with-config pyprojecttoml-with-config/file.py`
  - Loads project `pyproject.toml` and therefore doesn't report unnecessary type ignore ✅ 
- `uv tool run pyright --verbose --project pyprojecttoml-without-config pyprojecttoml-without-config/file.py `
  - Loads project `pyproject.toml` (warns that it `has no "[tool.pyright]" section`) and doesn't report unnecessary type ignore :x:

(`uv tool run pyright --version` gives me `1.1.408` )

# Solution

After the warning about missing `"[tool.pyright]"` section, try and load config from the CWD (as if the project `pyproject.toml` didn't exist)

(FWIW `ruff` actually does this, so there's some precedence)

thejcannon:
Fixes #11320 by having `_getExtendedConfigurations` take in a "search directory" to fallback to searching if we load the `pyproject.toml` and it doesn't contain a `[tool.pyright]` (which it then loads by calling `_getExtendedConfigurations`).

This makes it so we still only read/parse the config once, while being able to "fallback" correctly.


packages/pyright/package-lock.json
```diff
@@ -856,6 +856,7 @@
             "integrity": "sha512-NZyJarBfL7nWwIq+FDL6Zp/yHEhePMNnnJ0y3qfieCrmNvYct8uvtiV41UvlSe6apAfk0fY1FbWx+NwfmpvtTg==",
             "dev": true,
             "license": "MIT",
+            "peer": true,
```
rchiodo: This is because your node version is too new. Should be Node v22

thejcannon:
I absolutely hate to be a bother, but I'd love to know if there's a release cadence or some ETA on the next Pyright release, so I can work on the upgrade locally. Thanks!

rchiodo:
It usually depends upon there being enough fixes to bother. Right now it seems there's no bugs marked as addressed in next version? Oh this PR actually closed the bug. We usually just mark them as 'addressed in next version'. Let me check to see if there are others.

rchiodo:
I think we want to fix a number of issues and after that release. Typeshed updated too. Hopefully in the next week

thejcannon:
Y'all's time and energy is very much appreciated ♥️

PR #11331, opened 2026-03-14
feat: add `pyright.clearDiagnosticsOnChange` opt-in setting

### LINKED ISSUE #11308: Feature request: opt-in setting to clear diagnostics immediately on didChange
**Is your feature request related to a problem? Please describe.**

When pyright is used as a language server by AI coding tools, stale diagnostics cause problems. The tool edits a file via `textDocument/didChange`, then reads diagnostics to decide what to do next. Because reanalysis is async, the diagnostics it reads are from the previous state. This causes the tool to re-fix already-fixed errors, miss newly introduced errors, or loop on diagnostics that no longer apply.

This isn't a problem for interactive editors where a human is still typing and the debounce is desirable.

**Describe the solution you'd like**

An opt-in setting (`pyright.eagerDiagnosticInvalidation`, default `false`) that publishes empty diagnostics for a file immediately on `didChange`, before reanalysis completes. No behavior change for anyone who doesn't enable it.

The setting would be configured by tool vendors in their LSP initialization config, not by end users.

I have a working implementation with tests if there's interest.

WolffM:
## Summary

Fixes microsoft/pyright#11308: feat: add `pyright.clearDiagnosticsOnChange` opt-in setting

## Changes

This PR addresses the issue described above. Changes were developed on the `fix/11308-feat-add-pyright-cleardiagnosticsonchange-opt-in-s` branch.

## Related Issue

- Closes #11308

### DIFF
--- a/packages/pyright-internal/src/common/languageServerInterface.ts
+++ b/packages/pyright-internal/src/common/languageServerInterface.ts
@@ -46,6 +46,7 @@ export interface ServerSettings {
     ignoreFileSpecs?: string[];
     taskListTokens?: TaskListToken[];
     functionSignatureDisplay?: SignatureDisplayType | undefined;
+    clearDiagnosticsOnChange?: boolean | undefined;
 }
 
 export interface MessageAction {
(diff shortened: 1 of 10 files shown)


erictraut:
This feature doesn't address the comment I made in #11308. I don't think this is the right way to add this feature.

It also introduces workflows and other files that shouldn't have been included in the PR.

WolffM:
hey @erictraut thank you for reviewing this.

my bad for adding those workflow files, i use them for local testing in my fork.
let me try again with a numeric debounce value instead.

Issue #10845, opened 2025-08-25
Methods of a variadic generic class become uncallable when specified with TypeVarTuple and a suffix


96-LB:
When a variadic generic class is indexed with an unpacked TypeVarTuple followed by some suffix, any methods of that class which unpack the variadic generic fail to typecheck regardless of what arguments you pass in. It seems that the variadic args somehow swallow up all following arguments, making it impossible to pass in arguments corresponding to the suffix types.

In the following example, the call to `foo.foo` is expected to typecheck. However, it instead throws a cryptic error which is unable to be resolved with any set of arguments.
```python
class Foo[*Ts]:
    def foo(self, *args: *Ts):
        ...

def bar[*Ts, X](foo: Foo[*Ts, X], x: X, *args: *Ts):
    foo.foo(*args, x) # error: Argument missing for parameter "__p2" (reportCallIssue)
```

Here are some similar constructs which do *not* trigger this error:
```python
class Foo[*Ts]:
    def foo(self, *args: *Ts):
        ...

# swapping order of generic parameters to Foo
def bar[*Ts, X](foo: Foo[X, *Ts], x: X, *args: *Ts):
    foo.foo(x, *args) # ok
```
```python
from typing import Callable

# using a callable directly
def bar[*Ts, X](foo: Callable[[*Ts, X], None], x: X, *args: *Ts):
    foo(*args, x) # ok
```
```python
from typing import Callable

# using an attribute which returns a function instead of a method definition
class Foo[*Ts]:
    def __init__(self, foo: Callable[[*Ts], None]):
        self.foo = foo

def bar[*Ts, X](foo: Foo[*Ts, X], x: X, *args: *Ts):
    foo.foo(*args, x) # ok
```

Despite the last non-example, it seems the mere presence of a method declaraton causes the error to reappear, even though the revealed type of `foo.foo` doesn't change.
```python
from typing import Callable

class Foo[*Ts]:
    def __init__(self, foo: Callable[[*Ts], None]):
        self.foo = foo # ok
    
    def foo(self, *args: *Ts):
        ...

def bar[*Ts, X](foo: Foo[*Ts, X], x: X, *args: *Ts):
    reveal_type(foo.foo) # Type of "foo.foo" is "(*Ts@bar, X@bar) -> None"
    foo.foo(*args, x) # Argument missing for parameter "__p2"
```

This issue occurs with:
- Command-line version 1.1.404
- VS Code Pylance extension release version 2025.7.1
[...]

PR #11363, opened 2026-04-07
Fix match statement issue

gramster:
Fix pattern matching regression for variadic tuple types
When a tuple with an unbounded entry (e.g., tuple[int, *tuple[str, ...], int])
is matched against a pattern shorter than the tuple, the unbounded entry is
removed to make lengths match. However, this loses the information that the
tuple was originally variadic and could have different lengths.

The fix tracks when an unbounded entry is removed (removedIndeterminate flag)
and marks such matches as isPotentialNoMatch=true. This prevents the entire
variadic tuple from being eliminated in negative narrowing, preserving
reachability of subsequent match cases.

Fixes https://github.com/microsoft/pyright/issues/10845


.gitignore
```diff
@@ -125,4 +125,10 @@ junit.xml
 Thumbs.db
 
 # Potential venv
-.venv
\ No newline at end of file
+.venv
```
rchiodo: Copilot generated:
-134

**Remove this entire hunk from the PR.** These additions (`serena/`, `.beads/`, `AGENTS.md`) are unrelated to the pattern matching fix and appear to be local dev environment artifacts. Notably, `AGENTS.md` exists at the repo root and is already tracked — adding it to `.gitignore` signals wrong intent and could confuse contributors. Keep the PR focused on the bug fix.

packages/pyright-internal/src/analyzer/patternMatching.ts
```diff
@@ -1435,6 +1441,7 @@ function getSequencePatternInfo(
```
rchiodo: Copilot generated:
-1442

After `tupleIndeterminateIndex = -1`, the resulting `SequencePatternInfo` will have `isUnboundedTuple: false` even though the original tuple IS unbounded. This works today because `isPotentialNoMatch = true` prevents elimination before `isUnboundedTuple` is consulted, but any future consumer of `isUnboundedTuple` that trusts it as ground truth will get incorrect information. Consider: `isUnboundedTuple: removedIndeterminate || tupleIndeterminateIndex >= 0` at the push site (around line 1510) to preserve semantic accuracy.

PR #11369, opened 2026-04-08
Apply type narrowing in ternary expressions

gramster:


packages/pyright-internal/src/analyzer/operations.ts
```diff
@@ -790,7 +819,19 @@ export function getTypeOfTernaryOperation(
         fileInfo.definedConstants
     );
 
-    if (constExprValue !== false && evaluator.isNodeReachable(node.d.ifExpr)) {
+    // Check if we should apply flow-sensitive narrowing. We avoid narrowing for
+    // simple name references with literal bool types because the variable could
+    // be reassigned, even though the type is a literal. This also applies to
+    // `not <Name>` expressions to maintain consistency.
+    // Note: This guard is specific to ternary expressions. The and/or operators
```
rchiodo: Copilot generated:
**[Medium]** The comment explains that and/or operators "don't need this guard because they operate on already-evaluated types from their operands." Both the Skeptic and Architect note this distinction is debatable — `and`/`or`'s `leftType` is also flow-narrowed via `getTypeOfExpression`. The observable inconsistency: `maybe = True; maybe and 1 or "no"` narrows (no guard), but `1 if maybe else "no"` does not (guard fires). Consider acknowledging this inconsistency in the comment rather than asserting a structural distinction.

packages/pyright-internal/src/tests/samples/conditionalExpr1.py
```diff
@@ -0,0 +1,106 @@
...
+
+def func6(val: list[int] | None):
+    # After this guard, val is known to be not None.
+    if val is None:
+        return
+    
+    # However, val could still be an empty list (falsy), so the else branch
+    # is still reachable. Both branches should contribute to the type.
+    ts1: list[int] | str = val if val else "empty"
+    assert_type(val if val else "empty", list[int] | str)
```
rchiodo, line 90: Copilot generated:
**[High]** `func_bool_literal` and `func_bool_literal_not` are vacuous — they pass identically with the `isBoolLiteralName` guard deleted. Pyright widens `maybe = True` to `bool` (not `Literal[True]`), so `isBoolLiteralName` returns `false` and the guard never fires. `canBeTruthy(bool)` and `canBeFalsy(bool)` both return `true`, producing `int | str` regardless.

Replace these with tests that actually produce `Literal[True]` via flow narrowing:
```python
def func_guard_flow_narrowed(x: bool):
    if x:
        # x is Literal[True] here — guard should fire
        ts: int | str = 1 if x else "no"
        assert_type(1 if x else "no", int | str)

def func_guard_flow_narrowed_not(x: bool):
    if x:
        ts: int | str = 1 if not x else "yes"
        assert_type(1 if not x else "yes", int | str)
```

packages/pyright-internal/src/tests/samples/conditionalExpr1.py
```diff
@@ -0,0 +1,106 @@
+# This sample tests type narrowing for conditional expressions (ternary operator)
```
rchiodo, line 1: Copilot generated:
**[Medium]** (Unresolved prior feedback) Missing test cases that would strengthen coverage: (1) `def f(x: Literal[True]): return 1 if x else "no"` — documents the guard's behavior for explicitly-annotated parameters where no reassignment occurred. (2) A nested ternary `a if x else (b if y else c)` for regression coverage. (3) A custom class without `__bool__`/`__len__` to verify `canBeFalsy` is the deciding factor (not `isNodeReachable`).

packages/pyright-internal/src/analyzer/operations.ts
```diff
@@ -765,6 +765,33 @@ export function getTypeOfUnaryOperation(
     return { type, isIncomplete, magicMethodDeprecationInfo: deprecatedInfo };
 }
```
rchiodo, line 767: Copilot generated:
**[Medium]** (Unresolved prior feedback) `isBoolLiteralName` only guards `bool` literals, not other literal types like `Literal[42]`. The Advocate's defense that bool is uniquely common as a mutable flag pattern is reasonable, but this design choice should be documented. Add a comment explaining why only `bool` is guarded — e.g., "Only bool is guarded because bool literals are the overwhelmingly common case for mutable flag variables used as ternary conditions; non-bool literals like `Literal[42]` are rarely used as ternary conditions where reassignment would cause over-narrowing."

packages/pyright-internal/src/analyzer/operations.ts
rchiodo, line 826: Copilot generated:
**[Medium]** (Unresolved prior feedback) The comment "and/or operators don't need this guard because they operate on already-evaluated types from their operands" asserts a structural distinction that all four reviewers agree is inaccurate — `and`/`or`'s `leftType` is also flow-narrowed via `getTypeOfExpression`. Revise to something like: "The and/or operators historically apply narrowing without this guard. The guard is added here as conservative behavior for the new ternary narrowing, since ternary conditions are more commonly simple name references that may have been reassigned."

Issue #11359, opened 2026-04-02
Significant Performance Regression in 1.1.405+


arranf:
**Describe the bug**
I'm seeing a reproducible significant (~20s -> ~9 minutes!) performance regression on our large codebase that I've narrowed down to beginning with 1.1.405. I have also tested 1.1.406 and 1.1.408 and can confirm it reproduces there.

Looking at the stats/verbose output, the only difference I can seem to find is that one particular file takes `524587ms` in 1.1.405+ up from `159ms` in `1.1.404`. When excluding that file, another (similar) file spikes in time taken for checking (`387ms` on the non-excluded run, `493372ms` on the excluded run).

**20 second run**
`uv run --with pyright==1.1.404 pyright --stats --verbose`

```
Completed in 21.828sec

Analysis stats
Total files parsed and bound: 4964
Total files checked: 1454

Timing stats
Find Source Files:    0.03sec
Read Source Files:    0.88sec
Tokenize:             0.94sec
Parse:                0.77sec
Resolve Imports:      1.2sec
Bind:                 1.01sec
Check:                16.66sec
Detect Cycles:        0sec
```

**9 minute run**
`uv run --with pyright==1.1.405 pyright --stats --verbose`

```
Completed in 545.704sec

Analysis stats
Total files parsed and bound: 4965
Total files checked: 1454

Timing stats
Find Source Files:    0.03sec
Read Source Files:    0.86sec
Tokenize:             0.95sec
Parse:                0.71sec
Resolve Imports:      0.58sec
Bind:                 0.98sec
Check:                541.27sec
Detect Cycles:        0sec
```

**Specs/Configuration**
Running using Python 31.4 on a M5 Mac with Node 24

Default configuration besides `pythonVersion = "3.14"` and a `includes` array.

I've omitted the full verbose logs, and haven't included the file, to avoid leaking any information about the codebase publicly but am happy to share full logs privately with contributors and I'm happy to experiment with that file to help identify what the time is being taken on.

rchiodo:
Thanks for the issue but I think we'd need some piece of code that reproduces the problem.

arranf:
> Thanks for the issue but I think we'd need some piece of code that reproduces the problem.

Thanks for the quick response. I'm happy to produce a reproducible case but I can't share the current code directly. Can you provide some guidance in how I can narrow this down to produce a reproducible case?

rchiodo:
I guess first you see if you can reproduce the check problem with a single file on the command line. 

[...]

Issue #11076, opened 2025-10-30
Process never finishes in `1.1.405`, `1.1.406`, `1.1.407`


gothicVI:
**Describe the bug**
We have a large code base that uses a custom library which has a large, auto-generated file of pydantic models. Since `1.1.405` including all following versions we're seeing the issue that `pyright` does not finish. It will consume all of the cores it gets but simply run ad infinitum - our CI pipelines kill the job after an hour. Going back to `1.1.404` finishes in ~5 minutes.

I am willing to debug this but I have no idea where to even start looking and I can't share code.

erictraut:
Thanks for reporting. I won't be able to investigate this without additional details. We'll need to figure out the exact file and piece of code within that file that's causing the problem.

Here are two techniques that may be useful.

1. Run with the `--verbose` option. This will print additional details and help you narrow it down. 

2. Bisect the code base and/or the code within a file. That is, delete half of the code and run the type checker on the remaining half. Depending on whether it hangs, either delete the other half or delete half of the remaining.

gothicVI:
I'm not sure this is related to https://github.com/microsoft/pyright/issues/10881 but there might be something off with memory:
In `1.1.404` we get a ~24sec run with output: https://gist.github.com/gothicVI/0ae6ad94ae5aa5e178ccbdb13b19d8d5
In `1.1.407` I aborted the run after ~15min with output: https://gist.github.com/gothicVI/f6b78bcaa646eeab30023a27078dd915
There are additional `Long operation` messages and more memory related output.

@erictraut are the provided logs of any help?

gothicVI:
@erictraut any chance this can get looked at? I'd be willing to help debugging as best I can but I'd need some input on what to provide.

gothicVI:
@erictraut this remains an issue in `1.1.408`

gothicVI:
@erictraut I'd really like to help debug this but you'd need to let me know how I could be of help...

rchiodo:
If you can try the bisect that he suggested, that would narrow it down. Additionally, if you can share the code that causes the problem, it would make it possible to debug the issue.

gothicVI:
@erictraut @rchiodo I did bisect between releases `1.1.404` and `1.1.405`. Each step was done via
```
npm run install:all
cd packages/pyright
npm run build
cd -
# in a different terminal in the code repository
time node ./path/to/pyright/packages/pyright/index.js .
# back to the terminal in the pyright repository
git bisect ...
```
with result
[...]

Issue #10850, opened 2025-08-26
bug: spurious "partially unknown"


bukzor:
**Describe the bug**

I don't understand this bug at all, but I can tell you how to reproduce it, and I think it's clearly a bug.

Setup:

```console
$ tail -vn99 demo*
==> demo_typed.py <==
from __future__ import annotations

from typing import TypeVar

J = TypeVar("J", bound="JSON")
JSONObjectOf = dict[str, J]
JSON = str | JSONObjectOf["JSON"]
JSONObject = JSONObjectOf[JSON]

==> demo.py <==
from demo_typed import JSON
from demo_typed import JSONObjectOf


def identity(json: JSONObjectOf[JSON]):
    return json

==> demo0.py <==
import demo_typed


def identity0(json0: demo_typed.JSONObjectOf[demo_typed.JSON]):
    return json0

==> demo1.py <==
from demo_typed import JSON
from demo_typed import JSONObjectOf


def identity1(json1: JSONObjectOf[JSON]):
    return json1

==> demo2.py <==
from demo_typed import JSONObject


def identity2(json2: JSONObject):
    return json2

==> demo3.py <==
import demo_typed


def identity3(json3: demo_typed.JSONObject):
    return json3
```

**expected behavior**:

```
$ pyright demo0.py demo2.py | sed 's#'"$PWD/"'#./#g'
0 errors, 0 warnings, 0 informations
```

**actual behavior**:

```
$ pyright demo2.py demo0.py | sed 's#'"$PWD/"'#./#g'
./demo0.py
  ./demo0.py:4:5 - error: Return type, "dict[str, str | Unknown]", is partially unknown (reportUnknownParameterType)
  ./demo0.py:4:15 - error: Type of parameter "json0" is partially unknown
    Parameter type is "dict[str, str | Unknown]" (reportUnknownParameterType)
  ./demo0.py:4:46 - error: Type of "JSON" is partially unknown
    Type of "JSON" is "type[str] | Unknown" (reportUnknownMemberType)
  ./demo0.py:5:12 - error: Return type, "dict[str, str | Unknown]", is partially unknown (reportUnknownVariableType)
./demo2.py
  ./demo2.py:4:5 - error: Return type, "dict[str, str | Unknown]", is partially unknown (reportUnknownParameterType)
  ./demo2.py:4:15 - error: Type of parameter "json2" is partially unknown
    Parameter type is "dict[str, str | Unknown]" (reportUnknownParameterType)
[...]

PR #11372, opened 2026-04-09
Fix 1.1.405 regressions

### LINKED ISSUE #11321: Regression in 1.1.405 involving loops
**Describe the bug**
A clear and concise description of the behavior you are seeing and the expected behavior along with steps to reproduce it.
Pyright seems to be confused about variable type when modifying said variable inside the loop.
In this example, `lines` should infer as `list[str]`.

**Code or Screenshots**

```python
lines = ["ab", "bc", "ac", "ca"]  # lines: list[str]
while lines:  # lines: list[str] | list[str | Unknown]
    print(lines)  # lines: list[str] | list[Unknown]
    lines = [x for x in lines[1:]]  # x: str | Unknown
```
Working fine in [1.1.404](https://pyright-play.net/?pyrightVersion=1.1.404&strict=true&code=DYSwdgpgzgBAvDA2gIgIYCNkBobPQY211UJ2X1WQF0AoAdwAsRgIZRIoAuGmXmABwBO4AC4AKdtACUPPpNgJEADxgAzAPaCYK8G3DREARk5VaQA)
Broken in [1.1.405](https://pyright-play.net/?pyrightVersion=1.1.405&strict=true&code=DYSwdgpgzgBAvDA2gIgIYCNkBobPQY211UJ2X1WQF0AoAdwAsRgIZRIoAuGmXmABwBO4AC4AKdtACUPPpNgJEADxgAzAPaCYK8G3DREARk5VaQA)
Still broken in [1.1.408](https://pyright-play.net/?pyrightVersion=1.1.408&strict=true&code=DYSwdgpgzgBAvDA2gIgIYCNkBobPQY211UJ2X1WQF0AoAdwAsRgIZRIoAuGmXmABwBO4AC4AKdtACUPPpNgJEADxgAzAPaCYK8G3DREARk5VaQA) (latest version as of today)

gramster:
Fix 1.1.405 regressions: use position check instead of isFlowPathBetweenNodes in inferTypeOfSymbolForUsage

Fix: replace the isFlowPathBetweenNodes call with a simple source-position check. Skip declarations that textually appear after the usage (these can only be reached via loop back-edges). Retain declarations that appear before the usage (preserving the #10850 fix for type aliases like JSON = ... used in JSONObjectOf[JSON]).

Should fix https://github.com/microsoft/pyright/issues/11321, https://github.com/microsoft/pyright/issues/11076 and https://github.com/microsoft/pyright/issues/11359


packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -23304,7 +23304,15 @@ export function createTypeEvaluator(
                     const usageScope = ParseTreeUtils.getExecutionScopeNode(usageNode);
                     const declScope = ParseTreeUtils.getExecutionScopeNode(decl.node);
                     if (usageScope === declScope) {
-                        if (!isFlowPathBetweenNodes(decl.node, usageNode)) {
+                        // Skip declarations that appear after the usage in the source.
```
rchiodo, line 23307: @copilot
**[Medium]** The comment claims "such declarations can only be reached via loop back-edges," but this isn't strictly true for all same-scope constructs. An untyped declaration in an `if` branch textually before usage in the `else` branch is flow-unreachable yet would now be included. While the Skeptic confirmed the flow engine at the usage site overrides `effectiveType` for local variables (mitigating the impact on narrowed types), the wider `effectiveType` feeds into metadata flags like `includesVariableDecl` and `addTypeFormForSymbol`. Consider softening the comment to "typically only reached via loop back-edges" and adding a regression test for branching patterns (e.g. `if`/`else` with untyped assignment in only one branch).

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -23304,7 +23304,15 @@ export function createTypeEvaluator(
...
+                        // Such declarations can only be reached via loop back-edges, and
+                        // including them causes circular evaluation (producing Unknown).
+                        // Declarations that textually precede the usage are retained so
+                        // that order-independent type aliases resolve correctly.
+                        // Note: for the cases filtered out here (decl after usage in same
+                        // scope), the code-flow engine handles loop-carried narrowing when
+                        // it evaluates the type at the actual usage site.
+                        if (decl.node.start >= usageNode.start) {
```
rchiodo, line 23319: @copilot
**[Medium]** The position heuristic replaces a semantic reachability check, which is the correct fix for the circular-evaluation regressions. However, no test exercises non-loop branching within the same execution scope (e.g. `try`/`except`, `if`/`else`). Add at least one test where an untyped declaration textually precedes usage but is flow-unreachable, to confirm the flow engine's safety net produces the correct narrowed type at the usage site:
```python
def f(flag: bool):
    if flag:
        x = "hello"
        return
    reveal_type(x)  # should be str | Unbound, not str
```

rchiodo (approved):
Approved via Review Center.

rchiodo (approved):
Approved via Review Center.

PR #11375, opened 2026-04-10
reportPrivateUsage: honor __all__ for module-level _name in non-py.typed source

pt-anthropic:
`reportPrivateUsage` consults `__all__` for module-level single-underscore names only when the declaring file is bound as a stub or a py.typed package. For plain source files, `_bindNameValueToScope` calls `setIsPrivateMember()` directly, so an explicit `__all__` entry does not suppress the diagnostic.

This means the same symbol is reported or not depending on whether its file is in `include`:

```python
# pkg/__init__.py
__all__ = ["_exported"]
def _exported() -> int: return 1
```
```python
# app.py
from pkg import _exported  # reportPrivateUsage fires iff pkg is in include
```

This PR routes the non-stub, non-py.typed path through `_potentialPrivateSymbols` as well, so `__all__` is honored consistently.

The new `private3.py`/`private4.py` sample test fails before the change (2 errors) and passes after (1 error - only the name not in `__all__`).

### DIFF
--- a/packages/pyright-internal/src/analyzer/binder.ts
+++ b/packages/pyright-internal/src/analyzer/binder.ts
@@ -3550,12 +3550,8 @@ export class Binder extends ParseTreeWalker {
                             // Don't include private-named symbols in the builtin scope.
                             symbol.setIsExternallyHidden();
                         } else {
-                            // Defer the private/protected decision until __all__ has
-                            // been processed so an explicit __all__ entry can mark
-                            // the symbol public. Previously this path called
-                            // setIsPrivateMember() directly for non-stub,
-                            // non-py.typed source files, which meant __all__ was
-                            // honored only for stubs and py.typed packages.
+                            // Defer the private/protected decision until __all__ is processed
+                            // so an explicit __all__ entry can promote the symbol to public.
                             this._potentialPrivateSymbols.set(name, symbol);
                         }
                     }
--- a/packages/pyright-internal/src/analyzer/binder.ts
+++ b/packages/pyright-internal/src/analyzer/binder.ts
@@ -3546,15 +3546,17 @@ export class Binder extends ParseTreeWalker {
                             } else {
                                 this._potentialPrivateSymbols.set(name, symbol);
                             }
-                        } else if (this._fileInfo.isStubFile || this._fileInfo.isInPyTypedPackage) {
-                            if (this._currentScope.type === ScopeType.Builtin) {
-                                // Don't include private-named symbols in the builtin scope.
-                                symbol.setIsExternallyHidden();
-                            } else {
-                                this._potentialPrivateSymbols.set(name, symbol);
-                            }
+                        } else if (this._currentScope.type === ScopeType.Builtin) {
+                            // Don't include private-named symbols in the builtin scope.
+                            symbol.setIsExternallyHidden();
                         } else {
-                            symbol.setIsPrivateMember();
+                            // Defer the private/protected decision until __all__ has
+                            // been processed so an explicit __all__ entry can mark
+                            // the symbol public. Previously this path called
+                            // setIsPrivateMember() directly for non-stub,
+                            // non-py.typed source files, which meant __all__ was
+                            // honored only for stubs and py.typed packages.
+                            this._potentialPrivateSymbols.set(name, symbol);
                         }
                     }
                 }
--- a/packages/pyright-internal/src/tests/checker.test.ts
+++ b/packages/pyright-internal/src/tests/checker.test.ts
@@ -47,6 +47,13 @@ test('Private1', () => {
     TestUtils.validateResults(analysisResults, 4);
 });
 
+test('Private2', () => {
+    const configOptions = new ConfigOptions(Uri.empty());
+    configOptions.diagnosticRuleSet.reportPrivateUsage = 'error';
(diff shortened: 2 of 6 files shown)


packages/pyright-internal/src/tests/checker.test.ts
```diff
@@ -47,6 +47,13 @@ test('Private1', () => {
     TestUtils.validateResults(analysisResults, 4);
 });
```
rchiodo, line 48: Copilot generated:
-53

**Missing test for `_usesUnsupportedDunderAllForm` interaction** (High)

The test only covers the simple `__all__ = ["..."]` form. Add a test with a computed `__all__` (e.g., `__all__ = _base + ["public"]`) and a `_name` import to validate the fallback behavior. This is the exact path where the regression identified above occurs.

packages/pyright-internal/src/analyzer/binder.ts
```diff
@@ -3546,15 +3546,17 @@ export class Binder extends ParseTreeWalker {
...
+                            symbol.setIsExternallyHidden();
                         } else {
-                            symbol.setIsPrivateMember();
+                            // Defer the private/protected decision until __all__ has
+                            // been processed so an explicit __all__ entry can mark
+                            // the symbol public. Previously this path called
+                            // setIsPrivateMember() directly for non-stub,
+                            // non-py.typed source files, which meant __all__ was
+                            // honored only for stubs and py.typed packages.
+                            this._potentialPrivateSymbols.set(name, symbol);
```
rchiodo, line 3555: Copilot generated:

Regression when _usesUnsupportedDunderAllForm is true (High)

The old code at this location called setIsPrivateMember() directly for plain-source _name symbols. The new code defers them into _potentialPrivateSymbols, but the resolution loop at line 366 is guarded by !this._usesUnsupportedDunderAllForm. When a module uses a computed __all__ form (e.g., __all__ = _list + ["foo"]), the flag is true, the loop is skipped, and setIsPrivateMember() is never called — silently suppressing reportPrivateUsage for all _name imports from that module.

The Advocate acknowledged this as a genuine weakness. Consider falling back to setIsPrivateMember() for symbols in _potentialPrivateSymbols when _usesUnsupportedDunderAllForm is true, matching the old unconditional behavior. Add a test with a computed __all__ form to cover this path.

packages/pyright-internal/src/analyzer/binder.ts
```diff
@@ -3546,15 +3546,17 @@ export class Binder extends ParseTreeWalker {
...
-                            }
+                        } else if (this._currentScope.type === ScopeType.Builtin) {
+                            // Don't include private-named symbols in the builtin scope.
+                            symbol.setIsExternallyHidden();
                         } else {
-                            symbol.setIsPrivateMember();
+                            // Defer the private/protected decision until __all__ has
+                            // been processed so an explicit __all__ entry can mark
+                            // the symbol public. Previously this path called
+                            // setIsPrivateMember() directly for non-stub,
```
rchiodo: info · L3590–3595
-3595

The comment is well-intentioned but includes 4 lines of historical narration ("Previously this path called setIsPrivateMember() directly..."). Per repo coding standards, prefer comments that explain why, not changelog-style what-was. The first two lines are good; trim to:

// Defer the private/protected decision until __all__ is processed
// so an explicit __all__ entry can promote the symbol to public.
this._potentialPrivateSymbols.set(name, symbol);
The historical context belongs in the commit message / PR description (where it already exists).

packages/pyright-internal/src/tests/checker.test.ts
rchiodo, line 48: Copilot generated:
-53

**Missing test for `_usesUnsupportedDunderAllForm` interaction** (High)

The test only covers the simple `__all__ = ["..."]` form. Add a test with a computed `__all__` (e.g., `__all__ = _base + ["public"]`) and a `_name` import to validate the fallback behavior. This is the exact path where the regression identified above occurs.

packages/pyright-internal/src/analyzer/binder.ts
pt-anthropic: Thanks for flagging — I traced this and don't believe there's a regression here: the `_potentialPrivateSymbols` resolution loop (binder.ts ~L344) isn't gated on `_usesUnsupportedDunderAllForm`. For `__all__ = _base + [...]` the assignment handler sets `_dunderAllNames = []`, so `!this._dunderAllNames?.some(...)` is true and `setIsPrivateMember()` is called as before. This is also the pre-existing behavior for stub/py.typed packages, which already routed through this same path. Added `private5.py`/`private6.py` in bc747f4 covering exactly this case to lock it in — happy to add an explicit fallback if I'm misreading the loop.

packages/pyright-internal/src/analyzer/binder.ts
pt-anthropic: Done in bc747f4 — trimmed to the two-line version.

packages/pyright-internal/src/tests/checker.test.ts
pt-anthropic: Added in bc747f4 — `private5.py` defines `_name` with `__all__ = _base + ["public"]` (computed form), `private6.py` imports it. Test expects 1 `reportPrivateUsage` error.

PR #11383, opened 2026-04-14
tokenizer: accept ':' in type: ignore code names for namespaced codes

### LINKED ISSUE #11345: Type ignore comments with ty-namespaced codes are not recognized as blanket type ignores
**Describe the bug**

ty 0.0.25 added support for type ignore comments that use namespaced codes and look like this: `type: ignore[ty:unresolved-reference]` (https://github.com/astral-sh/ruff/pull/24096).

When these namespaced codes are used, Pyright fails to treat the comment as a blanket type ignore.

**Code or Screenshots**

```python
a[1]  # type: ignore[name-defined, ty:unresolved-reference]
```

```
error: "a" is not defined (reportUndefinedVariable)
```

1.1.408 (command-line version)

alliasgher:
## Summary

Fixes #11345.

The regexes for `type: ignore` and `pyright: ignore` comments used `[\s\w-,]` as the character class for codes inside `[...]`. The character `:` was missing, so tool-namespaced codes such as `ty:unresolved-reference` (introduced by [ty/ruff](https://github.com/astral-sh/ruff/pull/24096)) failed to match — pyright treated the comment as if it didn't exist and still emitted the diagnostic.

### Fix

Add `:` to the character class: `[\s\w-,]` → `[\s\w:,-]`. The change is identical in both `typeIgnoreCommentRegEx` and `pyrightIgnoreCommentRegEx`.

### Tests

Added `TypeIgnoreNamespacedCode` and `TypeIgnoreMixedNamespacedCodes` to `tokenizer.test.ts`. Full `tokenizer.test` suite (88 tests) passes.

### DIFF
--- a/packages/pyright-internal/src/parser/tokenizer.ts
+++ b/packages/pyright-internal/src/parser/tokenizer.ts
@@ -142,7 +142,11 @@ const _byteOrderMarker = 0xfeff;
 
 const defaultTabSize = 8;
 const magicsRegEx = /\\\s*$/;
-const typeIgnoreCommentRegEx = /((^|#)\s*)type:\s*ignore(\s*\[([\s\w-,]*)\]|\s|$)/;
+// The character class for type: ignore rule codes includes ':' so that
+// tool-namespaced codes such as "ty:unresolved-reference" are accepted.
+// pyright: ignore uses the original class since tool-namespaced codes
+// are not expected there.
+const typeIgnoreCommentRegEx = /((^|#)\s*)type:\s*ignore(\s*\[([\s\w:,-]*)\]|\s|$)/;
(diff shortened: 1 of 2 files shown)


alliasgher:

correctmost:
> The change is identical in both `typeIgnoreCommentRegEx` and `pyrightIgnoreCommentRegEx`

I don't think the Pyright regex should be modified.  I wouldn't expect tool-specific ignores to be used with `pyright: ignore`.

alliasgher:
Good catch — reverted the change to `pyrightIgnoreCommentRegEx`, keeping only the `typeIgnoreCommentRegEx` update.

rchiodo (approved):

rchiodo:
Thanks for the PR

Issue #11008, opened 2025-10-08
Slowdown when enabling enableExperimentalFeatures on string-heavy files


correctmost:
**Describe the bug**

I ran Pyright on the [psf/black](https://github.com/psf/black) codebase and noticed that it took much longer when `enableExperimentalFeatures` was set to true.

**Code or Screenshots**

Most of the time is spent checking one file that has a dictionary with 40,000+ string entries in it:
https://github.com/psf/black/blob/94ebcb50851e40cba88c6ad2c14f9dfe45e08921/profiling/dict_huge.py

```
[tool.pyright]
enableExperimentalFeatures = false

$ time npx pyright ./profiling/dict_huge.py 
real	1m22.881s
user	1m22.398s
sys	0m0.646s
```

```
[tool.pyright]
enableExperimentalFeatures = true

$ time npx pyright ./profiling/dict_huge.py 
real	3m21.485s
user	2m54.655s
sys	0m25.637s
```

I looked at some profiler data and it seems like `getTypeOfStringListAsType` is a big culprit.

When I apply this dummy patch to avoid `getTypeOfStringListAsType` calls, I see similar performance with `enableExperimentalFeatures = true` and `enableExperimentalFeatures = false`:

```diff
diff --git a/packages/pyright-internal/src/analyzer/typeEvaluator.ts b/packages/pyright-internal/src/analyzer/typeEvaluator.ts
index fa5750def..698198e32 100644
--- a/packages/pyright-internal/src/analyzer/typeEvaluator.ts
+++ b/packages/pyright-internal/src/analyzer/typeEvaluator.ts
@@ -1718,7 +1718,8 @@ export function createTypeEvaluator(
         if (
             node.d.strings.length !== 1 ||
             node.d.strings[0].nodeType !== ParseNodeType.String ||
-            !isTypeFormSupported(node)
+            !isTypeFormSupported(node) ||
+            1 == 1
         ) {
             return typeResult;
         }
```

**VS Code extension or command-line**
9ed33cf7be6 (command-line version)

bschnurr:
so I think we need to set a 'maxInferredListLength' like tuples. https://github.com/microsoft/pyright/blob/9ed33cf7be6150dc53244e8d4009c06e4daedeae/packages/pyright-internal/src/analyzer/tuples.ts#L243-L264

correctmost:
The slowdown has been fixed by #11412.

PR #11412, opened 2026-05-01
Enable TypeForm support by default

davidfstr:
...rather than requiring the --enableExperimentalFeatures bit to be set

---

CONTEXT:
* PEP 747 (TypeForm) is accepted for Python 3.15 which is being finalized (feature-wise), so it seems like a good time to turn on TypeForm recognition in pyright by default.

OPEN QUESTIONS:
* I am curious to know what the performance implications (if any) are for enabling recognition of TypeForms everywhere by default. **Does your CI infrastructure have statistics that could be used to detect performance changes?** Maybe mypy_primer could be run in a mode that checks performance deltas?


davidfstr:

packages/pyright-internal/src/analyzer/typeEvaluator.ts
davidfstr, line 1: Nit: Misspelling in commit message: "current interence context" -> "current inference context"

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -1743,10 +1744,18 @@ export function createTypeEvaluator(
...
 
+        // Only attempt to interpret the string as a TypeForm forward reference when
+        // there's a signal that a TypeForm value is wanted in this context. Doing it
+        // unconditionally can trigger expensive (and recursion-prone) type lookups
+        // for plain string literals in non-type contexts.
+        const wantsTypeForm =
+            (flags & EvalFlags.TypeFormArg) !== 0 ||
+            (inferenceContext !== undefined && expectedTypeWantsTypeForm(inferenceContext.expectedType));
+
         if (
```
Viicos: Do I assume that this fixes the test failure I was seeing in https://github.com/microsoft/pyright/pull/11302/?

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -28701,13 +28694,6 @@ export function createTypeEvaluator(
         return { sourceType: simpleSrcType, destType: simpleDestType };
     }
 
-    function isTypeFormSupported(node: ParseNode) {
```
rchiodo, line 28704: Is TypeForm only supported in 3.15 or is it backwards compatible? I'm wondering if this should stick around and be checking the environment for 3.15 or not

plan/test-output.txt
```diff
@@ -0,0 +1,529 @@
+$ (cd packages/pyright-internal && npm test)
```
rchiodo: I'm guessing you didn't mean to submit this.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
rchiodo: Oh no it seems it's in typing_extensions.
https://typing.python.org/en/latest/spec/type-forms.html#type-forms

rchiodo:
There's a prettier error too:



> check:prettier
> prettier -c .

Checking formatting...
[warn] packages/pyright-internal/src/analyzer/typeEvaluator.ts
[warn] Code style issues found in the above file. Forgot to run Prettier?



You should be able to run `npm run fix:prettier` from the command line and submit the results.

plan/test-output.txt
davidfstr: Heh, indeed not. I’m accustomed to the plan/ directory being gitignored. Will fix.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
davidfstr: Ah, a sibling PR. I will take a look at that PR and its test failure too.

davidfstr:
* Will fix prettier issues, and anything else reported by CI
* Will investigate mypy_primer output.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
davidfstr: Yes. The 2nd commit in this PR fixes the exact test from that PR.

correctmost:
> * I am curious to know what the performance implications (if any) are for enabling recognition of TypeForms everywhere by default

I previously reported a performance issue with TypeForm support: #11008.  This PR mostly fixes that large slowdown, but it seems like Pyright is still a little slower with TypeForm support enabled for that testcase.

davidfstr:
Analysis of mypy_primer output:
* The only project tracked by mypy-primer that had any change in error counts was `sympy (https://github.com/sympy/sympy)`
* sympy had *preexisting errors* when run under pyright, as indicated by mypy-primer showing *removed* lines in the diff
* Many of the preexisting errors changed to a slightly different form after the TypeForm enablement change
    * Specifically, 121 locations have both removed and added errors at the same location, indicating modified error messages rather than new errors.
    * ✅ I'm not too concerned about preexisting errors changing. Probably they changed to be a more narrow/precise finding.
* Some preexisting errors were removed
    * Specifically, 33 locations had errors that disappeared entirely
    * ✅ I'm not too concerned about errors being removed
* There *do* exist new errors:
    * ⚠️ Specifically, 171 locations have errors that are completely new (no prior error at that
  location)
    * The new errors are concentrated in:
      * 👉 polys/euclidtools.py: Lines 1974 (type parameter variance issues)
      * 👉 solvers/diophantine/diophantine.py: Heavy concentration (504-912+) with operator type errors like Operator "*" not supported for types "Unknown | Basic"

### Analysis of new errors in diophantine.py

Errors at diophantine.py:175-183 are in the `DiophantineEquationType.__init__` function, and involve the unannotated `equation` parameter variable:

```python
class DiophantineEquationType:
    """..."""
    name: str

    def __init__(self, equation, free_symbols=None):  # LINE 175
        self.equation = _sympify(equation).expand(force=True)

        if free_symbols is not None:
            self.free_symbols = free_symbols
        else:
            self.free_symbols = list(self.equation.free_symbols)
            self.free_symbols.sort(key=default_sort_key)  # LINE 182
```

* With TypeForm enabled, `_sympify(equation)` (where `equation` is unannotated) resolves to the first matching overload of sympify (returning `Integer`), not the last fallback (returning `Basic`). This single difference cascades through the rest of the function. For reference, `_sympify` has the following overloads:

```python
@overload
def sympify(a: int, *, strict: bool = False) -> Integer: ... # type: ignore
@overload
def sympify(a: float, *, strict: bool = False) -> Float: ...
@overload
def sympify(a: Expr | complex, *, strict: bool = False) -> Expr: ...
@overload
def sympify(a: Tbasic, *, strict: bool = False) -> Tbasic: ...
@overload
def sympify(a: Any, *, strict: bool = False) -> Basic: ...

def sympify(a, locals=None, convert_xor=True, strict=False, rational=False,
        evaluate=None): ...
```

* It is not clear to me why enabling TypeForm would alter the chosen overload in this situation. If an unannotated `equation` was passed to `sympify(equation)`, I'd expect pyright to presume that `equation: Unknown` (which is equivalent to `equation: Any` AFAIK) and thus select the `def sympify(a: Any, *, strict: bool = False) -> Basic: ...` overload, which is the behavior that existed before. **I plan to investigate this change in overload section further.**

* All of the new errors at diophantine.py:504+ are cascading consequences of the differing overload selected.

For reference, the revealed types within diophantine.py:175-183 changed in the following ways:

```
  ┌────────────────────────────┬─────────────────────┬──────────────────────┐
  │                            │ Baseline (TypeForm  │   TypeForm enabled   │
  │                            │        off)         │                      │
  ├────────────────────────────┼─────────────────────┼──────────────────────┤
  │                            │ Basic (last         │ Integer (first       │
  │ _sympify(equation)         │ overload, a: Any →  │ overload, a: int →   │
  │                            │ Basic)              │ Integer)             │
  ├────────────────────────────┼─────────────────────┼──────────────────────┤
  │ .expand(force=True)        │ error: Basic has no │ Expr (inherited)     │
  │                            │  .expand → Unknown  │                      │
  ├────────────────────────────┼─────────────────────┼──────────────────────┤
  │ self.equation              │ Unknown             │ Expr                 │
  ├────────────────────────────┼─────────────────────┼──────────────────────┤
  │ self.equation.free_symbols │ Unknown             │ set[Basic]           │
  │                            │                     │ (annotated)          │
  ├────────────────────────────┼─────────────────────┼──────────────────────┤
  │ self.free_symbols          │ Unknown |           │ Unknown |            │
  │                            │ list[Unknown]       │ list[Basic]          │
  ├────────────────────────────┼─────────────────────┼──────────────────────┤
  │ x, y = var                 │ x: Unknown          │ x: Unknown | Basic   │
  ├────────────────────────────┼─────────────────────┼──────────────────────┤
  │ x**2 at line 504           │ OK (Unknown)        │ error — Basic has no │
  │                            │                     │  __pow__             │
  └────────────────────────────┴─────────────────────┴──────────────────────┘
```

### Analysis of new errors in euclidtools.py

Affected code:
```python
def dmp_cancel(...) -> tuple[dmp[Er], dmp[Er]] | tuple[Er, Er, dmp[Er], dmp[Er]]:
    ...

    p = dmp_mul_ground(p, cp, u, K)  # LINE 1973
    q = dmp_mul_ground(q, cq, u, K)  # LINE 1974

    return p, q  # LINE 1976
```

One new error at line 1974 (`q = dmp_mul_ground(q, cq, u, K)`):
```
+   .../projects/sympy/sympy/polys/euclidtools.py:1974:24 - error: Argument of type "dmp[Er@dmp_cancel] | Unknown | dup[Unknown]" cannot be assigned to parameter "f" of type "dmp[Er@dmp_mul_ground]" in function "dmp_mul_ground"
+     Type "dmp[Er@dmp_cancel] | Unknown | dup[Unknown]" is not assignable to type "dmp[Er@dmp_mul_ground]"
+       "builtins.list" is not assignable to "builtins.list"
+         Type parameter "_T@list" is invariant, but "dmp" is not the same as "dmp"
+         Consider switching from "list" to "Sequence" which is covariant (reportArgumentType)
+   .../projects/sympy/sympy/polys/euclidtools.py:1974:34 - error: Argument of type "Domain[Er@dmp_cancel] | Ring[Unknown]" cannot be assigned to parameter "K" of type "Domain[Er@dmp_mul_ground]" in function "dmp_mul_ground"
+     Type "Domain[Er@dmp_cancel] | Ring[Unknown]" is not assignable to type "Domain[Er@dmp_mul_ground]"
+       "Domain[Er@dmp_cancel]" is not assignable to "Domain[Er@dmp_mul_ground]"
+         Type parameter "Er@Domain" is invariant, but "Er@dmp_cancel" is not the same as "Er@dmp_mul_ground" (reportArgumentType)
```

The two new errors at line 1974 are word-for-word identical to the two errors already at line 1973 (with q/cq substituted for p/cp).

On main branch (Baseline) vs. this feature branch (TypeForm), the inputs to this section of code are identical:
```
  ┌─────────────────┬───────────────────────────────────────────────────────┐
  │    Variable     │                   Type (both modes)                   │
  ├─────────────────┼───────────────────────────────────────────────────────┤
  │ p (before line  │ list[dmp] | Unknown | list[Unknown]                   │
  │ 1973)           │                                                       │
  ├─────────────────┼───────────────────────────────────────────────────────┤
  │ q (before line  │ list[dmp] | Unknown | list[Unknown] ← same as p       │
  │ 1974)           │                                                       │
  ├─────────────────┼───────────────────────────────────────────────────────┤
  │ cp              │ Er@dmp_cancel | Unknown | RingElement* (3-way,        │
  │                 │ includes TypeVar bound)                               │
  ├─────────────────┼───────────────────────────────────────────────────────┤
  │ cq              │ Er@dmp_cancel | Unknown (2-way, no bound)             │
  ├─────────────────┼───────────────────────────────────────────────────────┤
  │ K               │ Domain[Er@dmp_cancel] | Ring[Unknown]                 │
  └─────────────────┴───────────────────────────────────────────────────────┘
```

So p and q have identical types going in, yet:
- Baseline: flags dmp_mul_ground(p, cp, u, K) but silently accepts
  dmp_mul_ground(q, cq, u, K).
- TypeForm: flags both consistently.

<details><summary>Probable cause: A constraint-solver asymmetry in baseline.</summary>
<p>
The only difference between the two calls is the second argument's type. cp is
a 3-way union that includes RingElement* (the Er TypeVar's bound); cq is a
2-way union without it. Baseline pyright treats these very differently when
solving for Er@dmp_mul_ground:

- With cp (containing the bound): widens the solution, fails the dup[Unknown]
  invariance check on f → error.
- With cq (no bound): solves cleanly, lets f's dup[Unknown] slide through
  Unknown-permissiveness → no error, and q ends up narrowed to
  dmp[Er@dmp_cancel] (the "clean" type seen at line 1976 in baseline).

TypeForm enabling makes the solver flag both cases uniformly. The q value at
line 1976 then carries dmp[Er@dmp_cancel | Unknown] — the Unknown that
baseline silently absorbed.
</p>
</details> 

TypeForm enabling makes the solver flag both cases uniformly.

Thus the "new" error at line 1974 is not a regression. ✅ 

### Summary of next steps

* [ ] Investigate why the chosen overload in diophantine.py changed, to a choice that makes less sense to me that than the original choice

davidfstr:
Holy moly. diophantine.py is complex. **I think I'm going to disregard the remaining error deltas in sympy.** sympy was already producing very many errors under pyright before this branch and now its producing very many but different errors after this branch. Not worth the investigation time: sympy was already broken at the start.

I've just pushed up commits integrating the outstanding feedback.

## Next

So the next item (and possibly last) item I want to investigate is any **performance regressions** that this branch might introduce. I'll look at the psf/black codebase as issue #11008 did.

rchiodo:
GitHub cannot anchor PR review comments to unchanged lines in the diff. Falling back to a general PR comment for [packages/pyright-internal/src/analyzer/typeEvaluator.ts:L5574](https://github.com/microsoft/pyright/blob/b87b75fc9798102bc882a9f0ba887bdfff41c693/packages/pyright-internal/src/analyzer/typeEvaluator.ts#L5574).

Copilot generated:
`expectedTypeWantsTypeForm` uses `doForEachSubtype` which iterates all subtypes unconditionally (returns `void`, no early exit). The existing `someSubtypes` helper at `typeUtils.ts:786` uses `.some()` with short-circuit semantics and returns `boolean` — exactly what this function needs.

**Suggested fix:**
```typescript
function expectedTypeWantsTypeForm(expectedType: Type): boolean {
    return someSubtypes(expectedType, (subtype) =>
        isClassInstance(subtype) && ClassType.isBuiltIn(subtype, 'TypeForm')
    );
}
```
🔍 Structurally confirmed: `doForEachSubtype` signature at typeUtils.ts:771 returns `void`; `someSubtypes` at typeUtils.ts:786 returns `boolean` with `.some()`.

[verified]

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -1743,11 +1744,15 @@ export function createTypeEvaluator(
             };
```
rchiodo, line 1745: Copilot generated:
The `wantsTypeForm` guard is well-designed — it prevents expensive string-as-TypeForm interpretation for plain string literals. However, `[unverified]` — it's unclear whether `inferenceContext` propagates deeply enough through container literal evaluation. For example, in `d: dict[str, TypeForm] = {"key": "int"}`, does the `TypeForm` expected type reach the string `"int"` via inference context? A targeted test case for container scenarios (`list[TypeForm]`, `dict[str, TypeForm]`) would confirm or refute this.

[verified]

rchiodo:
GitHub cannot anchor PR review comments to unchanged lines in the diff. Falling back to a general PR comment for [packages/pyright-internal/src/analyzer/typeEvaluator.ts:L7659](https://github.com/microsoft/pyright/blob/b87b75fc9798102bc882a9f0ba887bdfff41c693/packages/pyright-internal/src/analyzer/typeEvaluator.ts#L7659).

Copilot generated:
📍 packages/pyright-internal/src/analyzer/typeEvaluator.ts (all `cloneWithTypeForm` sites)

`cloneWithTypeForm` always performs two object spreads (`{...type}` + `{...type.props}`) even when the TypeForm value being set is identical to what's already there. Compare with `cloneForCondition` (types.ts:366) which short-circuits when both old and new values are `undefined`. Consider adding a similar short-circuit: `if (type.props?.typeForm === typeForm) return type;` — this would avoid unnecessary allocations on the ~19 call sites that are now unconditional. [unverified — impact depends on how often the no-op case occurs]

[verified]

rchiodo:
GitHub cannot anchor PR review comments to unchanged lines in the diff. Falling back to a general PR comment for [packages/pyright-internal/src/analyzer/typeEvaluator.ts:L990](https://github.com/microsoft/pyright/blob/b87b75fc9798102bc882a9f0ba887bdfff41c693/packages/pyright-internal/src/analyzer/typeEvaluator.ts#L990).

Copilot generated:
-1044 (not in diff, but related)

All 7 TypeForm test files (typeForm1–7) still explicitly set `enableExperimentalFeatures = true`. The stated purpose of this PR is that TypeForm works by default, but no test validates this. Add at least one test that exercises TypeForm **without** `enableExperimentalFeatures` to prove the gate removal is effective.

[verified]

davidfstr:
Update: The following item is still my next item for this PR:

> So the next item (and possibly last) item I want to investigate is any performance regressions that this branch might introduce. I'll look at the psf/black codebase as issue https://github.com/microsoft/pyright/issues/11008 did.

Before I start the item, I will investigate/fix [a performance regression on a similar PR in mypy](https://github.com/python/mypy/pull/21262#issuecomment-4364872582). I expect I may be able to take learnings from that future fix and apply them to this PR.

davidfstr:
Update: I'm [continuing to fix performance regressions in mypy](https://github.com/python/mypy/pull/21262#issuecomment-4554325597).

davidfstr:
I profiled this branch's changes against two targets ("corpuses"):

1. the entire `psf/black` repo and 
2. `psf/black`'s string-heavy `profiling/dict_huge.py` file (the testcase that motivated #11008)

and found **no significant performance regression** in either.

## Performance

`build/perfCompare.py`, single-process (pyright's default), paired per-round deltas. CPU time (lowest-variance), n=50.

Setup: Clone the corpus next to the pyright repo:

```
git clone --depth 1 --branch 26.5.1 https://github.com/psf/black ../black
```

### psf/black's entire repo - branch base vs branch tip

```
python build/perfCompare.py --metric cpu --workers1 --warmup-runs 3 --num-runs 50 \
    --corpus ../black \
    3619ecd f05ea58
```

- n=50: **+19.4 ms ±29.1 (+0.31%)**, CI [−9.7, +48.5].

Not significant (CI includes 0).

### psf/black's `profiling/dict_huge.py` - branch base vs branch tip

```
python build/perfCompare.py --metric cpu --workers1 --warmup-runs 3 --num-runs 50 \
    --corpus-file ../black/profiling/dict_huge.py \
    3619ecd f05ea58
```

- n=50: **+13.4 ms ±23.1 (+0.72%)**, CI [−9.7, +36.5].

Not significant (CI includes 0).

No significant regression on either corpus, including the string-heavy file that motivated #11008.

## Next Step

I will review/integrate feedback from @rchiodo 's 4 comments above.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
davidfstr: Added targeted tests.

davidfstr:
@rchiodo I've added commits addressing all your outstanding feedback.

The PR branch is now pretty messy, with a lot of commits that should ultimately be squashed. Would you mind if I reordered/squashed the commits to get a clean history?

rchiodo:
If you want to squash that's fine. When we merge the PR, we'll also squash everything so it doesn't really matter which.

rchiodo (approved):

rchiodo:
@davidfstr, thanks for the PR

PR #11415, opened 2026-05-04
Fix hang in constraint solver for cyclic TypeVar lower bounds (#11413)

### LINKED ISSUE #11413: Pyright hangs indefinitely when using wrapt > 2.0.0

**Describe the bug**
  Pyright hangs indefinitely when analyzing code that uses wrapt==2.1.2 with
  generic type parameters and wrap_function_wrapper(). The same code works
  correctly with wrapt<2.0.0.

**Code or Screenshots**
See [minimum reproducible example](https://github.com/Gatedd/pyright-minimal-repo)

command line

rchiodo:
## Summary

Fixes #11413.

Pyright would hang indefinitely while analyzing code that combined wrapt 2.1.x's `wrap_function_wrapper` with a generic `make_wrapper` factory whose return type was a union containing both `R` and `Awaitable[R]`. The user-reported repro is at https://github.com/Gatedd/pyright-minimal-repo.

## Root cause

In `assignUnconstrainedTypeVar` (`packages/pyright-internal/src/analyzer/constraintSolver.ts`), when a TypeVar's first lower bound was being established, the source type was recorded unconditionally — even when that source type contained the destination TypeVar nested inside another generic constructor.

For the wrapt case the constraint solver received:

```
R := R | Awaitable[R]
```

There is no finite type that satisfies this. Each subsequent round of `solveAndApplyConstraints` substituted `R` with its own ever-growing bound, producing types of the form:

```
R | Awaitable[R] | Awaitable[R | Awaitable[R]] | Awaitable[R | Awaitable[R] | Awaitable[R | Awaitable[R]]] | ...
```

The recursion-depth cap (20) was hit per call, but the per-call combinatorial work grew exponentially, producing a hang rather than a clean infinite loop.

## Fix

Add a classical [occurs check](https://en.wikipedia.org/wiki/Occurs_check) before recording the lower bound. If `adjSrcType` references `destType` at a strictly nested position (e.g. `R := F[R]` or `R := R | Awaitable[R]`), refuse the assignment and return `false`.

The check is carefully scoped to allow two benign shapes that the existing logic already handles correctly:

1. **Bare top-level identity** (`T := T`) — fine, it's the identity constraint.
2. **Top-level union member equal to the TypeVar** (`T := T | int`, which arises naturally from protocol matching against a method returning `T | int`) — also fine; the existing solver resolves it. This is exercised by the existing `Protocol44` test.

Only strictly nested references trigger the failure path.

## Test

Adds `paramSpec56` regression sample + test. Without the fix the test hangs the jest worker (SIGTERM after timeout); with the fix analysis completes in ~640ms.

## Verification

- All 1,233 `typeEvaluator*` and `checker` tests pass locally.
- Existing related tests verified unaffected: `Tuple19` (similar cyclic shape, surfaces an error via Eric's earlier tuple-depth cap), `Protocol44` (legitimate `T | int` shape), all 56 other `ParamSpec*` tests.

## Caveat

For the wrapt case the constraint solver detects the cycle and refuses to bind the TypeVar, but the failure isn't surfaced as a user-visible diagnostic in this particular path — it occurs deep inside bidirectional inference for an argument expression where the diag isn't threaded through. The user's original code is type-incorrect, but pyright won't currently flag it; the important behavioral change here is that **analysis completes** instead of hanging. Threading a diagnostic through that call site would be a separate improvement.

### DIFF
--- a/packages/pyright-internal/src/analyzer/constraintSolver.ts
+++ b/packages/pyright-internal/src/analyzer/constraintSolver.ts
@@ -778,9 +778,7 @@ function assignUnconstrainedTypeVar(
             // recorded as the lower bound and existing logic resolves it.
             if (typeVarOccursIn(destType, adjSrcType)) {
                 diag?.addMessage(
-                    LocAddendum.typeAssignmentMismatch().format(
-                        evaluator.printSrcDestTypes(adjSrcType, destType)
-                    )
+                    LocAddendum.typeAssignmentMismatch().format(evaluator.printSrcDestTypes(adjSrcType, destType))
                 );
                 return false;
             }
@@ -1319,11 +1317,7 @@ function assignParamSpec(
 function typeVarOccursIn(typeVar: TypeVarType, type: Type): boolean {
     // A bare top-level TypeVar reference is not a cycle. Compare by name +
     // scope id rather than identity since pyright sometimes clones TypeVars.
-    if (
-        isTypeVar(type) &&
-        type.shared.name === typeVar.shared.name &&
-        type.priv.scopeId === typeVar.priv.scopeId
-    ) {
+    if (isTypeVar(type) && type.shared.name === typeVar.shared.name && type.priv.scopeId === typeVar.priv.scopeId) {
         return false;
     }
 
@@ -1340,10 +1334,7 @@ function typeVarOccursIn(typeVar: TypeVarType, type: Type): boolean {
 
     const tvars = getTypeVarArgsRecursive(type);
     for (const tv of tvars) {
-        if (
-            tv.shared.name === typeVar.shared.name &&
-            tv.priv.scopeId === typeVar.priv.scopeId
-        ) {
+        if (tv.shared.name === typeVar.shared.name && tv.priv.scopeId === typeVar.priv.scopeId) {
             return true;
         }
     }
--- a/packages/pyright-internal/src/analyzer/constraintSolver.ts
+++ b/packages/pyright-internal/src/analyzer/constraintSolver.ts
@@ -763,6 +763,27 @@ function assignUnconstrainedTypeVar(
     } else {
         if (!curLowerBound || isTypeSame(destType, curLowerBound)) {
             // There was previously no lower bound. We've now established one.
+            // Apply an occurs check: if `adjSrcType` references `destType`
+            // (the TypeVar being solved) at a strictly nested position
+            // (e.g. `R := F[R]` or `R := R | Awaitable[R]`), recording it as
+            // the lower bound creates a cyclic constraint that subsequent
+            // widening / substitution rounds expand into an exponentially
+            // growing recursive type. Such a constraint has no finite
+            // solution, so report the assignment as a failure rather than
+            // letting the analyzer hang. See microsoft/pyright#11413.
+            //
+            // Top-level union members that are exactly `destType` (e.g.
+            // `T := T | int` arising from protocol matching against `T | int`)
+            // are *not* considered cyclic - the original `adjSrcType` is
+            // recorded as the lower bound and existing logic resolves it.
+            if (typeVarOccursIn(destType, adjSrcType)) {
+                diag?.addMessage(
+                    LocAddendum.typeAssignmentMismatch().format(
+                        evaluator.printSrcDestTypes(adjSrcType, destType)
+                    )
+                );
+                return false;
+            }
             newLowerBound = adjSrcType;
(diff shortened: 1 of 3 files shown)


rchiodo:
## mypy_primer analysis

I went through every diff. Summary: **all the changes are either correct improvements or genuine bugs that the cycle was previously masking.** Net +8 errors across 5 projects, but those errors are real, and many of the "removed" errors were false positives. Several remaining errors also became dramatically more readable.

### home-assistant/core — improvement (no count change)

All `-` lines have absurdly nested types like:
```
Future[Coroutine[Any, Any, Coroutine[Any, Any, Coroutine[Any, Any, Coroutine[Any, Any, _R]]]]]
```
This is exactly the cyclic-explosion this PR is fixing — the constraint solver was repeatedly substituting `_R := Coroutine[Any, Any, _R] | _R` into itself. With the fix the diagnostics show the original `Coroutine[Any, Any, _R] | _R` cleanly.

### sympy — mostly correct

**Removed (false positives):**
- `crv_types.py:2544` / `:2723` — `Method "_cdf" overrides ... returns type "ComplexInfinity | NaN | Rational | Unknown | Zero | Expr"` not assignable to `None`. The base `_cdf` returns `None`; the overrides return `Expr`. The cyclic widening was previously fooling pyright into believing the base also returned `Expr`-like values, so the override "matched". Removing this error is correct (the real fix is unrelated — these overrides are arguably valid in sympy's design, but the error wasn't from this code path).
- `test_solvers.py:1214`, `test_solveset.py:445` (×2) — same family.

**New (genuine):**
- `polys/rings.py:852-853` — `PolyElement[Er]` doesn't satisfy `RingElement` protocol because its `__add__` overload signature has a **wider return type** (`PolyElement[Er] | PolyElement[PolyElement[Er]]`) than `RingElement.__add__` requires (`PolyElement[Er]`). I confirmed this in source: `PolyElement.__add__` is genuinely overloaded with two return shapes, but the implementation signature widens. This is a real protocol mismatch.
- `integrals.py:107` — nested `tuple([tuple(xab) for xab in self.limits])` — same `_T_co` flowing into itself; previously masked by cyclic widening.
- Several `test_solveset.py` errors about `Set` / `Basic` not satisfying `Sized` / `Iterable` / `__getitem__` — these are real protocol mismatches that the cycle was hiding.

**Improved messages (no count change):**
- ~10 lines under `plot/...`: old message just said `"FunctionType" is not assignable to "PlotInterval"`; new message includes the actual function signature (`Type "(self: Self@PlotInterval) -> ..."`). More useful diagnostic.

### prefect — genuine bugs uncovered (+5 errors)

All five new errors are in `_internal/concurrency/api.py` and have the shape:
```
override returns type "Awaitable[T]" / "CoroutineType[Any, Any, T]"
"Awaitable[T]" is not assignable to type "T"
Function return type "T | Awaitable[T]" is incompatible with type "T"
```
These are genuine override mismatches: the base class declares `-> T` but the subclass returns `Awaitable[T]`. Previously the cyclic constraint `T := T | Awaitable[T]` silently widened, making the subclass appear to return `T`. The fix correctly rejects the cycle, surfacing the real `reportIncompatibleMethodOverride` error.

### jax — genuine bugs uncovered (+2 errors)

`checkify.py:133` and `:155` — `tree_flatten` overrides return `tuple[list[Unknown], tuple[Unknown, Unknown]]` while base returns `tuple[list[Unknown], Unknown]`. The cycle was masking a real override-type-mismatch on the second tuple element.

### xarray — minor, arguably correct (+1 error)

`netCDF4_.py:439` — `NetCDF4DataStore.__init__` has unannotated `manager`. After an `isinstance` branch that reassigns to `DummyFileManager`, `self._manager` ends up typed `DummyFileManager[Unknown] | Unknown`. In `get_child_store(self) -> Self`, `type(self)(self._manager, ...)` now flags that the union argument can't satisfy the constructor's inferred parameter type. The error itself is awkwardly worded (anything *should* be assignable to `Unknown`), but the underlying issue is that the constructor is unannotated and the `Unknown | Unknown` union is being matched against itself with different scopes — same family of cycle. One spurious-looking error against the rest of the wins seems fine.

### Bottom line

| Project | Error count change | Verdict |
|---|---|---|
| home-assistant/core | 0 | improved messages |
| sympy | mixed | net improvement |
| prefect | +5 | real bugs uncovered |
| jax | +2 | real bugs uncovered |
| xarray | +1 | minor regression in error wording |

I don't think any of these warrant a fix on this PR — the new errors are correct, the removed errors were false positives, and the improved messages are a nice bonus.

heejaechang (approved):

PR #11419, opened 2026-05-06
Short-circuit NoReturn analysis for Any or Unknown calls

bschnurr:
## Summary

- Stops NoReturn call analysis as soon as a callable union subtype is `Any` or `Unknown`.
- Uses a single-pass subtype loop with a true early exit.
- Adds coverage for both `NoReturn | Unknown` and `NoReturn | Any` call targets.

## Why the early exit is correct

A call is considered `NoReturn` only when every possible call-target subtype is known not to return normally. `Any` and `Unknown` can represent a `NoReturn` callable, but they can also represent a callable that returns normally. They therefore provide insufficient information to prove that the call does not return.

This is also equivalent to the previous counting implementation. Each subtype increments `subtypeCount`, but an `Any` or `Unknown` subtype matches none of the cases that increment `noReturnTypeCount`. Once one is encountered, `noReturnTypeCount === subtypeCount` cannot be true, even if every remaining subtype is `NoReturn`. Returning `false` immediately preserves the existing reachability result while avoiding unnecessary analysis of later union members.

Any reproducible mypy_primer diagnostic differences should be reviewed for correctness rather than treated as regressions solely because diagnostics changed.

The separate `typeUtils` helper optimization remains in #11681.

## Validation

- `pnpm --dir packages/pyright-internal exec jest typeEvaluator1.test -t "Unreachable1" --forceExit`
- `pnpm --dir packages/pyright-internal run build`
- `pnpm exec prettier --check packages/pyright-internal/src/analyzer/codeFlowEngine.ts`

### DIFF
--- a/packages/pyright-internal/src/analyzer/codeFlowEngine.ts
+++ b/packages/pyright-internal/src/analyzer/codeFlowEngine.ts
@@ -48,6 +48,7 @@ import {
     ClassType,
     combineTypes,
     FunctionType,
+    isAnyOrUnknown,
     isClass,
     isClassInstance,
     isFunction,
@@ -59,6 +60,7 @@ import {
     isTypeSame,
     isTypeVar,
     isTypeVarTuple,
+    isUnion,
     maxTypeRecursionCount,
     NeverType,
     OverloadedType,
@@ -1772,9 +1774,16 @@ export function getCodeFlowEngine(
             const callTypeResult = evaluator.getTypeOfExpression(node.d.leftExpr, EvalFlags.CallBaseDefaults);
             const callType = callTypeResult.type;
 
-            doForEachSubtype(callType, (callSubtype) => {
+            const callSubtypes = isUnion(callType) ? callType.priv.subtypes : [callType];
+            for (let callSubtype of callSubtypes) {
                 // Track the number of subtypes we've examined.
                 subtypeCount++;
+
+                // Any or Unknown can never establish a guaranteed NoReturn result, so any
+                // union containing either is immediately a negative answer.
+                if (isAnyOrUnknown(callSubtype)) {
+                    return false;
+                }
 
                 if (isInstantiableClass(callSubtype)) {
                     // Does the class have a custom metaclass that implements a `__call__` method?
@@ -1782,7 +1791,7 @@ export function getCodeFlowEngine(
                     // in this case that the __call__ method is not a NoReturn type.
                     const metaclassCallResult = getBoundCallMethod(evaluator, node, callSubtype);
                     if (metaclassCallResult) {
-                        return;
+                        continue;
                     }
 
                     const newMethodResult = getBoundNewMethod(evaluator, node, callSubtype);
@@ -1839,7 +1848,7 @@ export function getCodeFlowEngine(
                         }
                     }
                 }
-            });
+            }
 
             // The call is considered NoReturn if all subtypes evaluate to NoReturn.
             const callIsNoReturn = subtypeCount > 0 && noReturnTypeCount === subtypeCount;
--- a/packages/pyright-internal/src/analyzer/codeFlowEngine.ts
+++ b/packages/pyright-internal/src/analyzer/codeFlowEngine.ts
@@ -48,6 +48,7 @@ import {
     ClassType,
     combineTypes,
     FunctionType,
+    isAnyOrUnknown,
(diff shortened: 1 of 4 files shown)


packages/pyright-internal/src/analyzer/codeFlowEngine.ts
```diff
@@ -1772,17 +1774,24 @@ export function getCodeFlowEngine(
             const callTypeResult = evaluator.getTypeOfExpression(node.d.leftExpr, EvalFlags.CallBaseDefaults);
             const callType = callTypeResult.type;
 
-            doForEachSubtype(callType, (callSubtype) => {
+            const callSubtypes = isUnion(callType) ? callType.priv.subtypes : [callType];
```
rchiodo, line 1789: Copilot generated:
**Consider using `someSubtypes` as a guard clause instead of restructuring to an explicit loop.** The Architect identified that the codebase already has `someSubtypes(type, predicate)` (in `typeUtils.ts:786`) which handles `isUnion` dispatch internally and supports early-exit semantics. This would achieve identical behavior:

```ts
if (someSubtypes(callType, isAnyOrUnknown)) {
    return false;
}
doForEachSubtype(callType, (callSubtype) => { /* unchanged */ });
```

This avoids direct `.priv.subtypes` access and keeps the iteration pattern consistent with the rest of the analyzer. The explicit loop works correctly, but introduces an ad-hoc pattern where utilities already exist.

[verified]

packages/pyright-internal/src/tests/samples/unreachable1.py
```diff
@@ -92,6 +92,14 @@ def func9():
     return 3
 
 
+def func9_1(noreturn_func: Callable[[], NoReturn], unknown_func, flag: bool):
```
rchiodo, line 95: Copilot generated:
**Consider adding a companion test with an explicit `Any`-typed parameter.** The current test covers `Unknown` (via the unannotated `unknown_func`), but not `Any` directly. While `isAnyOrUnknown` handles both uniformly, an explicit `Any` test case would provide direct coverage of both branches. Example:

```python
def func9_2(noreturn_func: Callable[[], NoReturn], any_func: Any, flag: bool):
    callback = noreturn_func if flag else any_func
    callback()
    return 3  # Should not be unreachable
```

Low priority since the utility function is shared, but strengthens regression coverage.

[verified]

packages/pyright-internal/src/analyzer/codeFlowEngine.ts
bschnurr: Ya the problem is that `doForEachSubtype` doesn't support early exit

packages/pyright-internal/src/analyzer/typeUtils.ts
```diff
@@ -796,11 +791,11 @@ export function someSubtypes(type: Type, callback: (type: Type) => boolean): boo
 export function allSubtypes(type: Type, callback: (type: Type) => boolean): boolean {
     if (isUnion(type)) {
         return type.priv.subtypes.every((subtype) => {
-            callback(subtype);
+            return callback(subtype);
```
bschnurr: `every` expects the return result from callback

packages/pyright-internal/src/analyzer/typeUtils.ts
```diff
@@ -1301,18 +1290,19 @@ export function getLiteralTypeClassName(type: Type): string | undefined {
...
     if (isUnion(type)) {
         let className: string | undefined;
-        let foundMismatch = false;
 
-        doForEachSubtype(type, (subtype) => {
+        for (const subtype of type.priv.subtypes) {
             const subtypeLiteralTypeName = getLiteralTypeClassName(subtype);
             if (!subtypeLiteralTypeName) {
-                foundMismatch = true;
+                return undefined;
```
bschnurr: use early exit

rchiodo:

rchiodo:

rchiodo (approved):
Approved via Review Center.

rchiodo:

rchiodo (approved):

rchiodo:

rchiodo (approved):

heejaechang:
### 🟢 Pyright CLI QA — `a12becea` — verdict: green

12 focused check(s) completed.

PR #11431, opened 2026-05-09
Make PEP 661 (sentinel) support non-experimental

JelleZijlstra:
PEP 661 was accepted and is going to be in Python 3.15, so pyright
support no longer needs to be experimental.


rchiodo (approved):
Thanks for the submission :)

JelleZijlstra:
Thanks! From a quick look the CI failure looks like a flake and the mypy-primer output looks unrelated (pre-existing nondeterminism?), but let me know if you'd like me to look further.

PR #11444, opened 2026-05-14
Fix wrong symbol range in `callHierarchy/incomingCalls` response

d34db3ff:
### Context
This fixes #11443 

Return the range that encloses the function node instead of just the function name.

PR #11455, opened 2026-05-20
Fix incoming call hierarchy ranges

bschnurr:
## Summary

Redo of #11444 on current upstream `main`, with regression coverage added.

This fixes incorrect `callHierarchy/incomingCalls` item ranges for function callers and ensures incoming call hierarchy also resolves when the request position lands on a `Function` node rather than only a `Name` node.

### Production fixes

- In `callHierarchyProvider.ts`, incoming call items for function callers now use:
  - `range`: the full function range
  - `selectionRange`: the function name range

- In `referencesProvider.ts`, `getDeclarationForPosition` now handles `ParseNodeType.Function` by resolving through `node.d.name`.

### Tests

Added regression coverage for incoming call hierarchy:

- `showcallhierarchy.incomingCalls.function.fourslash.ts`
  - verifies full caller `range`
  - verifies name-only `selectionRange`
  - verifies lookup when the request starts on a `Function` node

Updated adjacent incoming-call expectation:

- `showcallhierarchy.incomingCalls.aliasedFunction.fourslash.ts`
  - now expects the full caller function range

Extended the shared fourslash incoming-call assertion to support optional `selectionRange` checks.

## Validation

Ran the call hierarchy fourslash slice:

- `npx jest src/tests/fourSlashRunner.test.ts -t "showcallhierarchy" --runInBand --forceExit`

Result:
- 6 passed

Ran the full fourslash runner:

- `npx jest src/tests/fourSlashRunner.test.ts --runInBand --forceExit`

Result:
- 251 passed

### DIFF
--- a/packages/pyright-internal/src/languageService/referencesProvider.ts
+++ b/packages/pyright-internal/src/languageService/referencesProvider.ts
@@ -403,6 +403,10 @@ export class ReferencesProvider {
             return this.getDeclarationForNode(program, fileUri, node, reporter, useCase, token);
         }
 
+        if (node.nodeType === ParseNodeType.Function) {
+            return this.getDeclarationForNode(program, fileUri, node.d.name, reporter, useCase, token);
+        }
+
         // For other node types, there are no references to be found.
         return undefined;
     }
--- a/packages/pyright-internal/src/tests/fourslash/showcallhierarchy.incomingCalls.function.fourslash.ts
+++ b/packages/pyright-internal/src/tests/fourslash/showcallhierarchy.incomingCalls.function.fourslash.ts
@@ -1,32 +1,35 @@
 /// <reference path="typings/fourslash.d.ts" />
 
 // @filename: declare.py
-//// def /*marker1*/func():
+//// /*marker3*/def /*marker1*/func():
 ////    return 1
 
 // @filename: consume.py
 //// from declare import func
-//// from declare import /*marker2*/func as foobar
 ////
-//// def [|callByName|]():
-////    /*marker3*/func()
-//// def callByAlias():
-////    foobar()
+//// def [|/*callByNameSelection*/callByName|](): /*marker2*/func()
 
 // @filename: consume2.py
 //// from declare import func
 ////
-//// def [|callByName2|]():
-////    func()
+//// def [|/*callByName2Selection*/callByName2|](): func()
 
 {
-    const ranges = helper.getRanges();
-    const references = ranges.map((range) => {
-        return { path: range.fileName, range: helper.convertPositionRange(range) };
-    });
+    const callByNameSelectionRange = helper.getPositionRange('callByNameSelection');
+    const callByName2SelectionRange = helper.getPositionRange('callByName2Selection');
     const itemList = [
-        { filePath: references[0].path, range: references[0].range, name: 'callByName' },
-        { filePath: references[1].path, range: references[1].range, name: 'callByName2' },
+        {
+            filePath: helper.getMappedFilePath('consume.py'),
+            range: helper.expandPositionRange(callByNameSelectionRange, 4, 10),
+            selectionRange: callByNameSelectionRange,
+            name: 'callByName',
+        },
+        {
+            filePath: helper.getMappedFilePath('consume2.py'),
+            range: helper.expandPositionRange(callByName2SelectionRange, 4, 10),
+            selectionRange: callByName2SelectionRange,
+            name: 'callByName2',
+        },
     ];
 
     helper.verifyShowCallHierarchyGetIncomingCalls({
--- a/packages/pyright-internal/src/tests/harness/fourslash/testState.ts
+++ b/packages/pyright-internal/src/tests/harness/fourslash/testState.ts
@@ -1272,6 +1272,9 @@ export class TestState {
             const expectedFilePath = map[name].items.map((x) => x.filePath);
             const expectedRange = map[name].items.map((x) => x.range);
             const expectedName = map[name].items.map((x) => x.name);
+            const expectedSelectionRange = map[name].items
+                .map((x) => x.selectionRange)
+                .filter((x): x is PositionRange => x !== undefined);
 
             const position = this.convertOffsetToPosition(fileName, marker.position);
             const actual = new CallHierarchyProvider(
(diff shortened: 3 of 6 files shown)


packages/pyright-internal/src/languageService/referencesProvider.ts
```diff
@@ -403,6 +403,10 @@ export class ReferencesProvider {
             return this.getDeclarationForNode(program, fileUri, node, reporter, useCase, token);
         }
 
+        if (node.nodeType === ParseNodeType.Function) {
```
rchiodo, line 406: Copilot generated:
**Broader behavioral scope than documented.** `getDeclarationForPosition` is also called by `renameProvider.ts` and `reportReferences`. This change means Rename and Find All References now resolve from the `def` keyword position (previously returned `undefined`). This is likely a net improvement, but worth noting in the PR description and considering additional test coverage for rename-from-def scenarios.

[verified]

packages/pyright-internal/src/tests/fourslash/showcallhierarchy.incomingCalls.function.fourslash.ts
```diff
@@ -1,32 +1,35 @@
...
-    const references = ranges.map((range) => {
-        return { path: range.fileName, range: helper.convertPositionRange(range) };
-    });
+    const callByNameSelectionRange = helper.getPositionRange('callByNameSelection');
+    const callByName2SelectionRange = helper.getPositionRange('callByName2Selection');
     const itemList = [
-        { filePath: references[0].path, range: references[0].range, name: 'callByName' },
-        { filePath: references[1].path, range: references[1].range, name: 'callByName2' },
+        {
+            filePath: helper.getMappedFilePath('consume.py'),
```
rchiodo, line 22: Copilot generated:
**Magic numbers in `expandPositionRange` are fragile.** The offsets `(4, 10)` encode `"def "` (4 chars) and `"(): func()"` (10 chars). Any future edit to the test Python (renaming, adding type annotations) silently breaks these without a compile error. Consider using a named range `[|...|]` wrapping the full function definition for the `range` assertion. This matches the robustness guidance in the test instructions.

[verified]

rchiodo:

packages/pyright-internal/src/tests/harness/fourslash/testState.ts
```diff
@@ -1289,6 +1292,12 @@ export class TestState {
                 for (const a of actual) {
                     assert.strictEqual(expectedRange?.filter((e) => this._deepEqual(a.from.range, e)).length, 1);
                     assert.strictEqual(expectedName?.filter((e) => this._deepEqual(a.from.name, e)).length, 1);
+                    if (expectedSelectionRange.length > 0) {
+                        assert.strictEqual(
+                            expectedSelectionRange.filter((e) => this._deepEqual(a.from.selectionRange, e)).length,
+                            1
+                        );
+                    }
```
rchiodo, line 1300: > **Warning** · Non-blocking recommendation

`selectionRange` is optional per expected item, but once any item supplies it this loop requires every actual item to match one of the supplied ranges. Match each actual call against one complete expected item and check `selectionRange` only when that matched expectation provides it; a mixed specified/unspecified test would confirm the failure.

[verified]

rchiodo:

rchiodo (approved):
Approved via Review Center.

rchiodo:
The production fix looks right: `range` now covers the whole function, `selectionRange` covers its name, and using `selectionRange.start` for follow-up requests avoids resolving at `def` or a decorator.

I'd suggest three follow-ups:

1. **Match selection ranges per caller in the assertion helper.** Pooling expected selection ranges independently loses their association with callers. It rejects valid results when one expected item omits its optional `selectionRange`, but accepts incorrect results with selection ranges swapped between callers. Match each actual item against its expected URI/name/range, then check `selectionRange` only when that specific expectation supplies one.

2. **Preserve the existing import-site coverage.** The function fourslash test removes the marker on `func` in `from declare import func as foobar`, along with the alias-calling function. Querying the original name at that import site is distinct from querying the alias itself. Please retain that scenario while updating the expectations. Explicit start/end markers would also make the expected ranges clearer and less brittle than the hard-coded `4`, `10`, and `12` offsets.

3. **Refresh the PR description.** It still describes a `referencesProvider.ts` change to handle `Function` nodes, which isn't in the current diff. It should instead describe the `selectionRange.start` changes in `languageServerBase.ts` and the incoming/outgoing round-trip coverage for decorated and undecorated callers.

packages/pyright-internal/src/languageService/referencesProvider.ts
rchiodo: Removed the `Function`-node fallback in 56f60d0. Both incoming and outgoing call-hierarchy handlers now resolve returned items using `selectionRange.start`, so the full function/decorator range no longer broadens Rename or Find All References behavior. The regression coverage exercises real returned-item round trips for decorated and undecorated callers.

packages/pyright-internal/src/tests/fourslash/showcallhierarchy.incomingCalls.function.fourslash.ts
rchiodo: Fixed in 91d216c. Both incoming-call fixtures now use explicit nested multiline ranges for the whole function and its name; the hard-coded 4/10/12 offsets are gone. The function fixture also restores the original-name marker in `from declare import func as foobar`, its alias-calling function, and the separate call-site query. All three related suites pass on both the PR-head source and the GitHub merge snapshot.

packages/pyright-internal/src/tests/harness/fourslash/testState.ts
rchiodo: Fixed in 91d216c. The helper matches each actual caller to one expected URI/name/full-range tuple, consumes that expectation to enforce one-to-one matching, and checks `selectionRange` only when that specific expectation supplies it. Added coverage for an omitted selection range on either caller, swapped URI/name/range/selectionRange fields, reordered results, cardinality, and duplicates. All seven new cases fail against the old helper on the GitHub merge snapshot and pass with the fix.

rchiodo:
Addressed all three points from https://github.com/microsoft/pyright/pull/11455#issuecomment-5800066644 in 91d216c and the updated PR description:

1. Incoming-call expectations now match one-to-one by URI/name/full range, with selectionRange checked only for that matched caller. Seven harness regression cases cover mixed optional fields, swapped fields, ordering, cardinality, and duplicates.
2. Restored the original-name query at `from declare import func as foobar` and its alias-calling function. Both affected incoming-call fixtures now use explicit multiline/name ranges instead of arithmetic offsets.
3. Rewrote the description to document the full incoming function range, name-only selectionRange, both `languageServerBase.ts` follow-up handlers using `selectionRange.start`, and actual decorated/undecorated incoming/outgoing round trips. It no longer claims a ReferencesProvider Function fallback.

The complete testState, languageServer, and fourslash suites pass on PR-head source (312 tests) and the GitHub merge snapshot with these fixes (327 tests), with each snapshot's server rebuilt. All seven new harness cases fail against the old helper on the merge snapshot. TypeScript, formatting, applicable lint, and whitespace checks also pass. Replied to and resolved the related inline threads.

PR #11458, opened 2026-05-20
Update rspack dependencies to version 2.0.4

bschnurr:
## Summary

Updates Pyright's rspack toolchain to `@rspack/cli` and `@rspack/core` `^2.0.4` across the repo packages.

This refreshes the package-local lockfiles and updates the build/test configuration needed for rspack 2 compatibility.

## Changes

- Updates rspack dependencies and lockfiles in the Pyright CLI, internal, and VS Code extension packages.
- Updates rspack config files for rspack 2, including async `@rspack/core` loading and `output.library.type`.
- Fixes `lerna exec` typecheck command forwarding so `tsc --noEmit` is passed as the child command.
- Adds Node setup to the mypy_primer workflows so rspack 2 runs on a supported Node version in CI.
- Keeps npm legacy peer resolution for mypy_primer setup of historical checkouts.
- Updates Jest/ts-jest config so tests resolve VS Code language package export maps correctly.
- Imports worker port message transports from `vscode-jsonrpc/node`, where `PortMessageReader` and `PortMessageWriter` are exported.

## mypy_primer Workflow

The PR mypy_primer workflow compares the PR merge commit against the base-side parent. The base commit currently cannot install/build with its original dependency metadata after the rspack 2 update, so the workflow creates two local refs before running mypy_primer:

- `new_commit`: the GitHub pull request merge commit.
- `base_commit`: the merge commit's first parent, patched only with generated build metadata from the PR merge commit and a temporary `.npmrc` for legacy peer resolution.

The build metadata list is generated from the PR merge commit and includes root package manifests, package manifests and lockfiles, and package-level rspack configs. This keeps analyzer/source behavior for the old checker at the base revision while allowing the old checker to install and build with the updated rspack-compatible build tooling.

The PR workflow checkout is also faster now:

- uses `fetch-depth: 2`
- uses `fetch-tags: false`
- derives `base_commit` from `new_commit`'s first parent instead of fetching full branch/tag history
- passes `--new new_commit --old base_commit` to mypy_primer so both refs remain reachable from the shallow local clone

The push mypy_primer workflow is left with full history because its implicit old revision can depend on tags.

## Validation

- `npm run install:all`
- `npm run build:cli:dev`
- `npm run build:extension:dev`
- `cd packages/pyright-internal && npm run webpack:testserver`
- `npx lerna exec --stream --no-bail -- "tsc --noEmit"`
- `npm run typecheck`
- `cd packages/pyright-internal && npx jest src/tests/languageServer.test.ts --runInBand --no-cache`
- `cd packages/pyright-internal && npm run test:norebuild`
- Reproduced the mypy_primer base checkout failure at `4243a1b36c91204720c4cf80e5bbcb4a0737a759` with `npm run install:all`.
- Confirmed `NPM_CONFIG_LEGACY_PEER_DEPS=true npm run install:all` succeeds for that base checkout and for the PR merge checkout.
- Confirmed the synthetic mypy_primer base checkout installs locally when npm legacy peer resolution and the updated build metadata are present.

Note: rspack 2 warns on local Node `22.11.0`; CI uses Node `22`, and local validation passed despite the warning.

### DIFF
--- a/.github/workflows/mypy_primer_pr.yaml
+++ b/.github/workflows/mypy_primer_pr.yaml
@@ -72,11 +72,33 @@ jobs:
           git checkout -b base_commit $MERGE_BASE
           echo "base commit"
           git rev-list --format=%s --max-count=1 base_commit
+          # Keep the old checker source at the merge base, but use this PR's
+          # dependency and bundler metadata so the old checkout can install and build.
+          git checkout $GITHUB_SHA -- \
+            package.json \
+            package-lock.json \
+            packages/pyright/package.json \
+            packages/pyright/package-lock.json \
+            packages/pyright/rspack.config.js \
+            packages/pyright-internal/package.json \
+            packages/pyright-internal/package-lock.json \
+            packages/vscode-pyright/package.json \
+            packages/vscode-pyright/package-lock.json
           printf 'legacy-peer-deps=true\n' > .npmrc
-          git add .npmrc
+          git add \
+            .npmrc \
+            package.json \
+            package-lock.json \
+            packages/pyright/package.json \
+            packages/pyright/package-lock.json \
+            packages/pyright/rspack.config.js \
+            packages/pyright-internal/package.json \
+            packages/pyright-internal/package-lock.json \
+            packages/vscode-pyright/package.json \
+            packages/vscode-pyright/package-lock.json
           git config user.name github-actions[bot]
           git config user.email 41898282+github-actions[bot]@users.noreply.github.com
-          git commit -m "Configure npm for mypy_primer base"
+          git commit -m "Configure build dependencies for mypy_primer base"
 
           echo ''
           cd ..
--- a/.github/workflows/mypy_primer_pr.yaml
+++ b/.github/workflows/mypy_primer_pr.yaml
@@ -68,6 +68,7 @@ jobs:
           cd pyright_to_test
           echo "new commit"
           git rev-list --format=%s --max-count=1 $GITHUB_SHA
+          git branch new_commit $GITHUB_SHA
 
           MERGE_BASE=$(git rev-parse "$GITHUB_SHA^1")
           git checkout -b base_commit $MERGE_BASE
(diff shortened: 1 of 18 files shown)


rchiodo (approved):

rchiodo (approved):
Approved via Review Center.

rchiodo (approved):
Approved via Review Center.

bschnurr:
Follow-up on the mypy_primer workflow changes:

- The PR primer checkout now uses `fetch-depth: 2` and `fetch-tags: false` instead of full history. The workflow only needs the GitHub PR merge commit and its first parent; tag history is not needed for the PR comparison.
- Because the checkout is shallow, the workflow now creates a local `new_commit` branch before switching to `base_commit`. mypy_primer clones the local checkout again, so passing `--new new_commit` keeps the PR merge commit reachable in that clone.
- The old baseline branch is still created from the merge commit's first parent and then patched only with this PR's dependency/build metadata so the base analyzer source is preserved while the old checker can install/build with the updated rspack toolchain.

This keeps the checkout faster while preserving the old/new comparison shape that mypy_primer expects.

.github/workflows/mypy_primer_pr.yaml
```diff
@@ -41,34 +44,74 @@ jobs:
...
         with:
           path: pyright_to_test
-          fetch-depth: 0
+          fetch-depth: 2
+          fetch-tags: false
       - uses: actions/setup-python@v5
         with:
           python-version: '3.11'
+      - uses: actions/setup-node@v6
+        with:
```
rchiodo, line 53: Copilot generated:
-62

The hardcoded file list for `git checkout $GITHUB_SHA --` creates a brittle implicit contract: if a future PR adds a new package or build config file (e.g. a new rspack plugin config or a new monorepo package), this list silently becomes stale and the workflow fails with a confusing install/build error rather than a clear signal.

Consider adding a sync-comment like `# SYNC: update this list when adding new packages or build configs` to draw attention. Alternatively, a small shell check that verifies the cherry-pick produced a buildable state (`npm run install:all || echo 'ERROR: mypy_primer base-commit file list may be stale'`) would fail more clearly.

[verified]

rchiodo (approved):
Approved via Review Center.

rchiodo (approved):
Approved via Review Center.

PR #11480, opened 2026-06-06
Avoid unnecessary workspace configuration request

rchl:
The code requested the python section first and, right after that, the `python.analysis` section. Each triggered a `workspace/configuration` request (for clients that support the "configuration" capability). The `python.analysis` request is unnecessary because it's a subset of the already-requested `python` section, so avoiding it prevents a redundant request.

### DIFF
--- a/packages/pyright-internal/src/tests/lsp/languageServerTestUtils.ts
+++ b/packages/pyright-internal/src/tests/lsp/languageServerTestUtils.ts
@@ -712,7 +712,6 @@ export async function runPyrightServer(
                         finalConfiguration = configuration;
                     }
                 }
-
                 result.push(finalConfiguration);
             }
 
--- a/packages/pyright-internal/src/tests/lsp/languageServerTestUtils.ts
+++ b/packages/pyright-internal/src/tests/lsp/languageServerTestUtils.ts
@@ -268,7 +268,7 @@ function isPlainObject(value: unknown): boolean {
  * Returns true if the update was applied, false otherwise.
  */
 function updateConfigurationSection(
-    configuration: any,
+    configuration: Record<string, any>,
     requestedSection: string | undefined,
     targetSection: string | undefined,
     targetValue: any,
@@ -283,7 +283,11 @@ function updateConfigurationSection(
     // Path relative to the object we actually hold.
     const relativeParts = targetSectionParts.slice(requestedSectionParts.length);
     if (relativeParts.length === 0) {
-        return false; // target == the section itself; nothing to set
+        // target == the section itself
+        for (const [key, value] of Object.entries(targetValue)) {
+            configuration[key] = value;
+        }
+        return true;
     }
 
     // Walk to the parent, creating missing intermediate objects along the way.
--- a/packages/pyright-internal/src/tests/lsp/languageServerTestUtils.ts
+++ b/packages/pyright-internal/src/tests/lsp/languageServerTestUtils.ts
@@ -701,17 +701,18 @@ export async function runPyrightServer(
             const result = [];
             const mappedSettings = settingsMap.get(info) || [];
             for (const item of p.items) {
-                const setting = mappedSettings.find(
+                const matchingSettings = mappedSettings.filter(
                     (s) =>
                         (s.item.scopeUri === item.scopeUri || s.item.scopeUri === undefined) &&
                         (isRequestedConfigurationSectionIncludesTargetSection(item.section, s.item.section))
                 );
-                let finalConfiguration = undefined;
-                if (setting) {
+                if (matchingSettings.length) {
                     // Indicate we queried at least one setting.
                     info.queriedConfigSettings.resolve();
-
-                    const configuration = {};
+                }
+                let finalConfiguration = undefined;
+                for (const setting of matchingSettings) {
+                    const configuration: Record<string, any> = finalConfiguration ?? {};
                     if (updateConfigurationSection(configuration, item.section, setting.item.section, setting.value)) {
                         finalConfiguration = configuration;
                     }
--- a/packages/pyright-internal/src/tests/lsp/languageServerTestUtils.ts
+++ b/packages/pyright-internal/src/tests/lsp/languageServerTestUtils.ts
@@ -281,9 +281,7 @@ function updateConfigurationSection(
     const relativeParts = targetSectionParts.slice(requestedSectionParts.length);
     if (relativeParts.length === 0) {
         // target == the section itself
-        for (const [key, value] of Object.entries(targetValue)) {
-            configuration[key] = value;
-        }
+        Object.assign(configuration, targetValue);
         return true;
     }
 
--- a/packages/pyright-internal/src/tests/lsp/languageServerTestUtils.ts
+++ b/packages/pyright-internal/src/tests/lsp/languageServerTestUtils.ts
@@ -243,14 +243,11 @@ export function createFileSystem(projectRoot: string, testData: FourSlashData, o
 
 const settingsMap = new Map<PyrightServerInfo, { item: ConfigurationItem; value: any }[]>();
 
-function isRequestedConfigurationSectionIncludesTargetSection(
-    requestedSection: string | undefined,
-    targetSection: string | undefined
-) {
-    const requestedSectionParts = requestedSection ? requestedSection.split('.') : [];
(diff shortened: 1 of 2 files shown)


rchl:
I think the way the tests are currently written, those depends on this extra request. Tests should be adapted.

packages/pyright-internal/src/tests/lsp/languageServerTestUtils.ts
```diff
@@ -243,17 +243,67 @@ export function createFileSystem(projectRoot: string, testData: FourSlashData, o
 
 const settingsMap = new Map<PyrightServerInfo, { item: ConfigurationItem; value: any }[]>();
 
-export function updateSettingsMap(info: PyrightServerInfo, settings: { item: ConfigurationItem; value: any }[]) {
-    const ignoreCase = toBoolean(info.testData.globalOptions[GlobalMetadataOptionNames.ignoreCase]);
-    // Normalize the URIs for all of the settings.
-    settings.forEach((s) => {
-        if (s.item.scopeUri) {
-            s.item.scopeUri = UriEx.parse(s.item.scopeUri, !ignoreCase).toString();
```
rchl, line 251: `updateSettingsMap` was unused so removed.

Added `updateConfigurationSection` which is used to apply extra settings like [this one](https://github.com/rchl/pyright/blob/4cc1392284c66116b8bb13a6e2965faa152c77a6/packages/pyright-internal/src/tests/languageServer.test.ts#L276-L288) regardless of which configuration section is requested (`python.analysis` or just `python`) as long as the requested section contains the section specified in extra setting.

It's quite a bit more complex than previous logic but previous wasn't technically correct as it relied on pyright only expecting the overrides to be present when requesting `python.analysis` section but did not care about the overrides when requesting `python` section. Now the overrides are present regardless of which section is requested as long as it contains the section specified in overrides.

And this extra complexity is of course only in the testsuite so normal shipped code is not affected.

packages/pyright-internal/src/tests/lsp/languageServerTestUtils.ts
```diff
@@ -650,13 +700,19 @@ export async function runPyrightServer(
                 const setting = mappedSettings.find(
```
rchiodo: Copilot generated:
Exact-section requests now silently return `undefined`. When a test registers a section *equal* to the requested one (e.g. `{ section: 'python' }` or `{ section: 'pyright' }`), `updateConfigurationSection` computes `relativeParts.length === 0` and returns `false`, so the mock pushes `undefined` instead of the value — inverting the old harness behavior (which returned `setting.value` on exact match). The Skeptic and Architect both confirmed this by tracing; it is test-only and not hit today (only `python.analysis` is registered), but it's a latent regression in a shared utility. Fix: when `relativeParts.length === 0`, return the stored value directly (`finalConfiguration = setting.value`) so the mock remains a faithful superset of real client behavior.

[verified]

rchiodo: Copilot generated:
`.find()` returns only the first matching registration, so sibling sub-sections are silently dropped. If a future test registers both `python.analysis` and another `python.*` key, a single `python` request matches both via the prefix check but only the first is returned — unlike a real client, which merges all `python.*` keys into one tree. Raised by all three of Skeptic/Architect/Advocate. Consider `filter` + folding all matches into one `configuration` object before pushing.

[verified]

packages/pyright-internal/src/tests/lsp/languageServerTestUtils.ts
```diff
@@ -243,17 +243,67 @@ export function createFileSystem(projectRoot: string, testData: FourSlashData, o
...
+) {
+    const requestedSectionParts = requestedSection ? requestedSection.split('.') : [];
+    const targetSectionParts = targetSection ? targetSection.split('.') : [];
+    for (const [i, requestedSectionPart] of requestedSectionParts.entries()) {
+        if (targetSectionParts[i] !== requestedSectionPart) {
+            return false;
         }
-    });
+    }
+    return true;
```
rchiodo, line 254: Copilot generated:
Prettier violations will fail the required `npm run check` gate: `isPlainObject` uses `typeof value === "object"` and `updateConfigurationSection` uses `split(".")` plus a trailing comma after `targetValue: any,`. The repo enforces `singleQuote: true`, and the sibling helpers in this same file use single quotes. Flagged by all three of Advocate/Skeptic/Architect. Reformat to single quotes / run Prettier before merge.

[verified]

packages/pyright-internal/src/tests/lsp/languageServerTestUtils.ts
```diff
@@ -243,17 +243,67 @@ export function createFileSystem(projectRoot: string, testData: FourSlashData, o
 
 const settingsMap = new Map<PyrightServerInfo, { item: ConfigurationItem; value: any }[]>();
 
-export function updateSettingsMap(info: PyrightServerInfo, settings: { item: ConfigurationItem; value: any }[]) {
-    const ignoreCase = toBoolean(info.testData.globalOptions[GlobalMetadataOptionNames.ignoreCase]);
-    // Normalize the URIs for all of the settings.
-    settings.forEach((s) => {
-        if (s.item.scopeUri) {
-            s.item.scopeUri = UriEx.parse(s.item.scopeUri, !ignoreCase).toString();
+function isRequestedConfigurationSectionIncludesTargetSection(
```
rchiodo: Copilot generated:
The name `isRequestedConfigurationSectionIncludesTargetSection` buries a generic dot-path prefix check inside domain vocabulary and reads grammatically awkwardly. Both the Rules reviewer (helpers-generic.md) and the Architect flagged this. Rename to something like `isDotPathPrefix(prefix, path)` or `isSectionAtOrUnder(target, requested)` and update the two call sites — clearer intent and correct argument-order signaling.

[verified]

packages/pyright-internal/src/tests/lsp/languageServerTestUtils.ts
rchl: fixed

rchl: Using filter now to apply all overrides.

That said, it's a bit unclear how multiple setting overrides should be handled. Whether
- later overrides should overwrite earlier?
- should all overrides be merged together (if so, what strategy to use)?

I suppose it's fine as is for now as normally only a single override is used per test.

packages/pyright-internal/src/tests/lsp/languageServerTestUtils.ts
rchl: fixed

packages/pyright-internal/src/tests/lsp/languageServerTestUtils.ts
rchl: renamed

packages/pyright-internal/src/tests/lsp/languageServerTestUtils.ts
```diff
@@ -243,17 +243,68 @@ export function createFileSystem(projectRoot: string, testData: FourSlashData, o
...
+    if (!isDotPathPrefix(requestedSection, targetSection)) {
+        return false;
+    }
+
+    const requestedSectionParts = requestedSection ? requestedSection.split('.') : [];
+    const targetSectionParts = targetSection ? targetSection.split('.') : [];
+
+    // Path relative to the object we actually hold.
+    const relativeParts = targetSectionParts.slice(requestedSectionParts.length);
+    if (relativeParts.length === 0) {
```
rchiodo, line 282: Copilot generated:
New regression introduced by the exact-section fix (flagged independently by the Skeptic, Architect, and Rules reviewer): `Object.entries(targetValue)` is unguarded for non-object values. A primitive (`{ section: 'pyright.typeCheckingMode', value: 'standard' }`) silently becomes a char-indexed object, and a `null`/`undefined` value throws `TypeError`, taking down the `ConfigurationRequest` handler. The descend branch guards with `isPlainObject` but this one doesn't. Dormant today (all overrides are object-valued) but a latent footgun in a shared mock. Use `Object.assign(configuration, targetValue)` (a no-op for primitive sources) or guard with `isPlainObject(targetValue)`.

[verified]

rchiodo (approved):

PR #11482, opened 2026-06-11
Fix callable positional count diagnostics

### LINKED ISSUE #11464: Function assignability error message for positional parameters swaps functions
**Describe the bug**
In the function assignability error message for parameters, the sub-message has the functions swapped
```
Type "() -> int" is not assignable to type "(int) -> int"
  Function accepts too many positional parameters; expected 0 but received 1 
```
As `() -> int` is the scrutinized type being checked for assignability, the message should say
```
  Function accepts too few positional parameters; expected 1 but received 0
```

**Code or Screenshots**
Code sample in [pyright playground](https://pyright-play.net/?strict=true&code=GYJw9gtgBALgngBwJYDsDmUkQWEMoDCAhgDYlEBGJApgLABQDAJtcFCwMa5Ey4AUwAK4oOALkKlyVagG0ZqGAF0ANJhRKAlFAC0APgllKNOQpVqlohlGtQQ1GIJAooQkQwYABTt14hmrFzAwPi09c0t6G1t7R2cARgYgA)

```python
from typing import Callable

def decorator(func: Callable[[int], int]) -> Callable[[int], int]:
    return func

@decorator
def foo() -> int:
    return 1

```

Tested on online playground (version 1.1.409)

popsiclelmlm:
## Summary
Fixes microsoft/pyright#11464 by reporting callable positional parameter count mismatches from the source function's perspective.

## Reproduction
```python
from typing import Callable

def decorator(func: Callable[[int], int]) -> Callable[[int], int]:
    return func

@decorator
def foo() -> int:
    return 1
```

Before this change, Pyright reported:

```text
Function accepts too many positional parameters; expected 0 but received 1
```

For `foo`, the clearer diagnostic is:

```text
Function accepts too few positional parameters; expected 1 but received 0
```

## Root cause
When callable parameter counts differed, `typeEvaluator.ts` selected the `tooFew`/`tooMany` addendum and populated the expected/received values from the destination callable perspective. For diagnostics shown on the source callable being assigned, this inverted the user-facing message.

## Changes
- Swap the positional-count diagnostic addendum direction for callable assignability failures.
- Report expected/received counts from the callable being assigned.
- Add a fourslash diagnostic regression test that checks both too-few and too-many positional count messages.

## Tests
- `npm install`
- `npm run build` from `packages/pyright-internal`
- `node --max-old-space-size=8192 --expose-gc ./node_modules/jest/bin/jest --runTestsByPath src/tests/diagnostics.test.ts --runInBand --forceExit` from `packages/pyright-internal`
- `npx prettier -c packages/pyright-internal/src/analyzer/typeEvaluator.ts packages/pyright-internal/src/tests/diagnostics.test.ts`
- `git diff --check`

## Screenshots/Logs
Not applicable; diagnostic text is covered by the regression test.

### DIFF
--- a/packages/pyright-internal/src/analyzer/typeEvaluator.ts
+++ b/packages/pyright-internal/src/analyzer/typeEvaluator.ts
@@ -26924,9 +26924,9 @@ export function createTypeEvaluator(
                 ).length;
 
                 diag?.createAddendum().addMessage(
-                    LocAddendum.functionTooFewParams().format({
-                        expected: nonDefaultSrcParamCount,
-                        received: destPositionalCount,
+                    LocAddendum.functionTooManyParams().format({
+                        expected: destPositionalCount,
+                        received: nonDefaultSrcParamCount,
                     })
                 );
                 canAssign = false;
@@ -26997,9 +26997,9 @@ export function createTypeEvaluator(
 
                 if (srcPositionalCount < adjDestPositionalCount) {
                     diag?.addMessage(
-                        LocAddendum.functionTooManyParams().format({
-                            expected: srcPositionalCount,
-                            received: destPositionalCount,
+                        LocAddendum.functionTooFewParams().format({
+                            expected: adjDestPositionalCount,
+                            received: srcPositionalCount,
                         })
                     );
                     canAssign = false;
--- a/packages/pyright-internal/src/tests/diagnostics.test.ts
+++ b/packages/pyright-internal/src/tests/diagnostics.test.ts
@@ -39,3 +39,44 @@ test('pyright ignore unused import', async () => {
         marker: { category: 'none', message: '' },
     });
 });
+
+test('callable assignment positional parameter count', async () => {
+    const code = `
+// @filename: test.py
+//// from typing import Callable
+////
+//// def decorator(func: Callable[[int], int]) -> Callable[[int], int]:
+////     return func
+////
+//// def decorator0(func: Callable[[], int]) -> Callable[[], int]:
+////     return func
+////
+//// @[|/*marker*/decorator|]
+//// def foo() -> int:
+////     return 1
+////
+//// @[|/*marker2*/decorator0|]
+//// def bar(value: int, /) -> int:
+////     return value
+    `;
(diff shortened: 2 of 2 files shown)


rchiodo:
GitHub cannot anchor PR review comments to unchanged lines in the diff. Falling back to a general PR comment for [packages/pyright-internal/src/analyzer/typeEvaluator.ts:L26959](https://github.com/microsoft/pyright/blob/8cd09f7f9f0a39f69038a89c563cc44cfee1de10/packages/pyright-internal/src/analyzer/typeEvaluator.ts#L26959).

Copilot generated:
The Skeptic identified that this `functionTooManyParams` line and the sibling `argsPositionOnly` addendum now use opposite expected/received conventions. In the `marker2` case the user sees `Position-only parameter mismatch; expected 1 but received 0` followed by `Function accepts too many positional parameters; expected 0 but received 1` — mirror-image numbers in one error. Before this PR the two lines agreed; the swap desynced them. Worth a deliberate decision: either accept the position-only case reads as contradictory, or align the conventions. [unverified — message rendering traced by hand, not executed]

[verified]

rchiodo:
GitHub cannot anchor PR review comments to unchanged lines in the diff. Falling back to a general PR comment for [packages/pyright-internal/src/analyzer/typeEvaluator.ts:L26959](https://github.com/microsoft/pyright/blob/8cd09f7f9f0a39f69038a89c563cc44cfee1de10/packages/pyright-internal/src/analyzer/typeEvaluator.ts#L26959).

Copilot generated:
Asymmetric count basis between the two branches (flagged by Skeptic and Advocate): block 2 reports `adjDestPositionalCount` (the `*args`/overload-adjusted threshold the guard actually compares), but this block reports the unadjusted `destPositionalCount`. For a dest signature with `*args` or trailing defaulted positionals taking the 'too many' path, the displayed `expected` may not match the count the user reasons about. The asymmetry is pre-existing, but this PR is the natural moment to make the reciprocal pair consistent. [unverified — no `*args` fixture exercises this path]

[verified]

PR #11485, opened 2026-06-11
Allowing using external typeshed

mochaaP:
Adds an environment variable `PYRIGHT_TYPESHED_PATH` for overriding builtin typeshed with an external one.

### DIFF
--- a/packages/pyright-internal/src/analyzer/pythonPathUtils.ts
+++ b/packages/pyright-internal/src/analyzer/pythonPathUtils.ts
@@ -14,7 +14,7 @@ import { Host } from '../common/host';
 import * as pathConsts from '../common/pathConsts';
 import { PythonVersion } from '../common/pythonVersion';
 import { Uri } from '../common/uri/uri';
-import { getFileSystemEntries, isDirectory, tryStat } from '../common/uri/uriUtils';
+import { getFileSystemEntries, isDirectory, tryStat, UriEx } from '../common/uri/uriUtils';
 import { ImportLogger } from './importLogger';
 
 export interface PythonPathResult {
@@ -25,7 +25,18 @@ export interface PythonPathResult {
 export const stdLibFolderName = 'stdlib';
 export const thirdPartyFolderName = 'stubs';
 
-export function getTypeShedFallbackPath(fs: Pick<FileSystem, 'getModulePath' | 'existsSync' | 'realCasePath'>) {
+export function getTypeShedFallbackPath(
+    fs: Pick<FileSystem, 'getModulePath' | 'existsSync' | 'realCasePath' | 'statSync'>
+) {
+    const envTypeshedPath = process.env.PYRIGHT_TYPESHED_PATH;
+    if (envTypeshedPath) {
+        const envTypeshedUri = UriEx.file(envTypeshedPath);
+        const stats = fs.statSync(envTypeshedUri);
+        if (stats.isDirectory()) {
+            return fs.realCasePath(envTypeshedUri);
+        }
+    }
+
     const moduleDirectory = fs.getModulePath();
     if (!moduleDirectory || moduleDirectory.isEmpty()) {
         return undefined;


rchiodo:
Might be nice to add a test to verify this works. Like what if the env variable doesn't end with a '/'. Does it still work?

packages/pyright-internal/src/analyzer/pythonPathUtils.ts
```diff
@@ -26,6 +26,11 @@ export const stdLibFolderName = 'stdlib';
 export const thirdPartyFolderName = 'stubs';
 
 export function getTypeShedFallbackPath(fs: Pick<FileSystem, 'getModulePath' | 'existsSync' | 'realCasePath'>) {
+    const envTypeshedPath = process.env.PYRIGHT_TYPESHED_PATH
+    if (fs.existsSync(envTypeshedPath)) {
```
rchiodo: existsSync requires a URI, not a string, so this would not actually work. You'd need something like 

```typescript
const envTypeshedPath = UriEx.file(process.env.PYRIGHT_TYPESHED_PATH);
```

rchiodo (changes requested):

mochaaP:
> Might be nice to add a test to verify this works. Like what if the env variable doesn't end with a '/'. Does it still work?

Updated, catching this with `fs.statSync` now. Errors are intentionally not caught to indicate a configuration error

packages/pyright-internal/src/analyzer/pythonPathUtils.ts
```diff
@@ -25,7 +25,18 @@ export interface PythonPathResult {
...
-export function getTypeShedFallbackPath(fs: Pick<FileSystem, 'getModulePath' | 'existsSync' | 'realCasePath'>) {
+export function getTypeShedFallbackPath(
+    fs: Pick<FileSystem, 'getModulePath' | 'existsSync' | 'realCasePath' | 'statSync'>
+) {
+    const envTypeshedPath = process.env.PYRIGHT_TYPESHED_PATH;
+    if (envTypeshedPath) {
+        const envTypeshedUri = UriEx.file(envTypeshedPath);
+        const stats = fs.statSync(envTypeshedUri);
+        if (stats.isDirectory()) {
+            return fs.realCasePath(envTypeshedUri);
```
rchiodo, line 36: 📍 packages/pyright-internal/src/analyzer/pythonPathUtils.ts:32

`fs.statSync(envTypeshedUri)` has no existence guard and no try/catch, so a missing/typo'd/stale `PYRIGHT_TYPESHED_PATH` throws `ENOENT` out of `getTypeShedFallbackPath`. Nothing on the call path wraps this, so instead of falling back to bundled typeshed the analyzer's stub resolution breaks. This file already imports `tryStat`/`isDirectory` precisely because `statSync` throws. Replace with the safe idiom: `if (isDirectory(fs, envTypeshedUri)) { return fs.realCasePath(envTypeshedUri); }`.

packages/pyright-internal/src/analyzer/pythonPathUtils.ts
```diff
@@ -25,7 +25,18 @@ export interface PythonPathResult {
 export const stdLibFolderName = 'stdlib';
 export const thirdPartyFolderName = 'stubs';
 
-export function getTypeShedFallbackPath(fs: Pick<FileSystem, 'getModulePath' | 'existsSync' | 'realCasePath'>) {
+export function getTypeShedFallbackPath(
+    fs: Pick<FileSystem, 'getModulePath' | 'existsSync' | 'realCasePath' | 'statSync'>
+) {
+    const envTypeshedPath = process.env.PYRIGHT_TYPESHED_PATH;
```
rchiodo, line 31: 📍 packages/pyright-internal/src/analyzer/pythonPathUtils.ts:30

The precedent `PYRIGHT_TMPDIR` reads `process.env.PYRIGHT_TMPDIR?.trim()`, but this code omits `.trim()`. A trailing newline/space (common in shell/CI env exports) becomes part of the path and is then fed into the stat. Read the value as `process.env.PYRIGHT_TYPESHED_PATH?.trim()` to match the existing convention and avoid surprising failures.

packages/pyright-internal/src/analyzer/pythonPathUtils.ts
```diff
@@ -25,7 +25,18 @@ export interface PythonPathResult {
...
+    if (envTypeshedPath) {
+        const envTypeshedUri = UriEx.file(envTypeshedPath);
+        const stats = fs.statSync(envTypeshedUri);
+        if (stats.isDirectory()) {
+            return fs.realCasePath(envTypeshedUri);
+        }
+    }
+
```
rchiodo, line 39: 📍 packages/pyright-internal/src/analyzer/pythonPathUtils.ts:30

No tests cover the new behavior. Add a focused regression test for the three cases: valid dir → override used; path is a file → silently ignored; non-existent path → no throw, graceful fallback to bundled typeshed. The missing-path test would catch the crash above directly. Restore/delete `process.env.PYRIGHT_TYPESHED_PATH` in `afterEach` to avoid leaking global state into other suites.

rchiodo:
Core idea is fine, but the env-var path must use the safe stat idiom already present in this file (isDirectory/tryStat) so a bad PYRIGHT_TYPESHED_PATH falls back gracefully instead of throwing. Please also add regression tests for the valid-dir / file / missing-path cases.

rchiodo:
Core concern: the env-var override uses an unguarded `fs.statSync`, which can throw and break stub resolution instead of gracefully falling back. Please use the existing `isDirectory`/`tryStat` idiom, trim the env value, and add regression tests for the valid-dir / file / missing-path cases.

rchiodo:
Main blocker: guard the env-path lookup using the existing `isDirectory`/`tryStat` idiom already imported in this file, since `statSync` throws on missing paths. Adding a small regression test and trimming the env value would round it out.

rchiodo:

packages/pyright-internal/src/analyzer/pythonPathUtils.ts
rchiodo, line 39: The raw `fs.statSync(envTypeshedUri)` throws `ENOENT` when `PYRIGHT_TYPESHED_PATH` points to a non-existent path — a very likely misconfiguration (typo, stale export, moved checkout). That throw propagates through `getTypeShedFallbackPath` into typeshed/import resolution with no try/catch in the chain, turning the intended graceful fallback into broken analysis. This file already imports `tryStat` and `isDirectory` for exactly this stat-with-guard pattern (and the rest of the function uses `existsSync` guards). Fix: `const stats = tryStat(fs, envTypeshedUri); if (stats?.isDirectory()) { return fs.realCasePath(envTypeshedUri); }`. This also lets you drop `'statSync'` from the widened `Pick<FileSystem, …>` signature.

rchiodo, line 39: No tests were added. The analogous `PYRIGHT_TMPDIR` override ships with `realTempFile.test.ts` covering set/unset cases. Add coverage for: (1) valid external directory → returned via `realCasePath`; (2) non-existent path → does NOT throw and falls back to bundled typeshed (this case would have caught the `statSync` throw above); (3) file-instead-of-directory → ignored. Since `process.env` is global mutable state, save/restore it in `afterEach` to avoid leaking into other suites.

rchiodo, line 39: The new user-facing `PYRIGHT_TYPESHED_PATH` override is undocumented, making it undiscoverable. Add a one-line mention in `docs/` alongside the existing `typeshedPath` config. Optionally, log via `importLogger` when the override is applied so users can confirm it took effect when debugging a wrong external typeshed.

rchiodo:
Core concern: the `PYRIGHT_TYPESHED_PATH` override should fail gracefully. Use the already-imported `tryStat`/`isDirectory` guard instead of a raw `statSync`, and add basic coverage for the valid/missing/not-a-directory cases.

rchiodo:
Closing for now

PR #11486, opened 2026-06-12
fix(dataClasses): recognize `KW_ONLY` separator for any variable name

### LINKED ISSUE #11451: KW_ONLY is only recognized when assigned to _
### **Description**
When creating a `KW_ONLY` field in a dataclass, the typical convention is to attach it to `_`. However, as Python documentation states, _[this is simply a convention](https://docs.python.org/3/library/dataclasses.html#dataclasses.KW_ONLY)_, and the name is ignored regardless. At runtime, the field effectively doesn't exist.

However, Pyright will not recognize it under anything else, as the `KW_ONLY` check is locked behind a field specifically being `_`. This goes against official Python behavior and documentation.

https://github.com/microsoft/pyright/blob/63998f4d0720a86447f4c4a04716e34f3e703660/packages/pyright-internal/src/analyzer/dataClasses.ts#L378-L388

As such, it will not recognize which fields should be keyword-only. In some cases, Pyright will not report anything, making the issue invisible. However, it will raise an false positive if `KW_ONLY` is after fields with defaults, as seen below.

### **Code example**

The real world example that inspired this issue. Another common convention with `_` is to assign [`gettext`](https://docs.python.org/3/library/gettext.html#gettext.gettext) to it. However, when assigning translatable text into a dataclass field's default, it will conflict with `_: KW_ONLY`, as now `_` has been reassigned (to nothing) in the class scope.

The simplest solution, which seems to generally satisfy linters, is to assign `KW_ONLY` to `__` instead. However, Pyright will complain about this too, stating "Fields without default values cannot appear after fields with default values". This error is reported as simply "`reportGeneralTypeIssues`", meaning it cannot be selectively silenced. Any call to the class constructor will also now be flagged, as Pyright believes there is a required (ironically positional) argument that is missing.

The issue can currently be silenced by assigning `field(default_factory=KW_ONLY)` to the `KW_ONLY` typed field, but this is rather abnormal and not very Pythonic.

```python
from dataclasses import KW_ONLY, dataclass, field
from gettext import gettext

_ = gettext

@dataclass
class PartiallyKWONLY:
    im_positional: int
    __: KW_ONLY  # no error
    im_kwonly: str

@dataclass
class PartiallyKWONLYandDefaults:
    im_positional: int = 67
    __: KW_ONLY  # error: reportGeneralTypeIssues
    im_kwonly: str = _("with a default")

@dataclass
class PartiallyKWONLYbutFullyDefaults:
    im_positional: int = 67
    __: KW_ONLY = field(default_factory=KW_ONLY)  # no error
    im_kwonly: str = _("with a default")
```

yen0304:
Closes #11451

---

`dataclasses.KW_ONLY` marks a separator field: every parameter declared after it in the class body becomes keyword-only. The [Python docs](https://docs.python.org/3/library/dataclasses.html#dataclasses.KW_ONLY) explicitly state that *the variable name is ignored* — `_` is merely a convention, and any identifier works equally well at runtime.

Before this change, Pyright checked for `=== '_'` before evaluating the annotation type, so non-`_` names like `__` or `kw_sep` were treated as ordinary fields instead of separators. This caused two classes of false-positive diagnostics:

1. **`reportGeneralTypeIssues`** — "Fields without default values cannot appear after fields with default values" — when a `KW_ONLY` separator with no default appeared after a field that did have one.
2. Downstream constructor-call errors, because the keyword-only fields were incorrectly treated as positional.

**Fix:** Remove the `=== '_'` guard so any `TypeAnnotation` whose RHS resolves to `KW_ONLY` is treated as a separator.

A new test sample (`dataclassKwOnly2.py`) exercises the three common patterns: `__: KW_ONLY`, `__: KW_ONLY` after a field with a default, and an arbitrary name (`kw_sep: KW_ONLY`).

> Note: This fix was developed with AI-agent assistance.

### DIFF
--- a/packages/pyright-internal/src/analyzer/dataClasses.ts
+++ b/packages/pyright-internal/src/analyzer/dataClasses.ts
@@ -376,7 +376,8 @@ export function synthesizeDataClassMethods(
                         });
 
                     // Is this a KW_ONLY separator introduced in Python 3.10?
-                    if (!isNamedTuple && statement.d.valueExpr.d.value === '_') {
+                    // Per the Python docs, the variable name is ignored — any name works.
+                    if (!isNamedTuple) {
                         const annotatedType = variableTypeEvaluator();
 
                         if (isClassInstance(annotatedType) && ClassType.isBuiltIn(annotatedType, 'KW_ONLY')) {
--- a/packages/pyright-internal/src/tests/samples/dataclassKwOnly2.py
+++ b/packages/pyright-internal/src/tests/samples/dataclassKwOnly2.py
@@ -1,0 +1,47 @@
+# This sample tests that KW_ONLY can be assigned to any variable name,
+# not just `_`. Python docs state the name is ignored at runtime.
+# See: https://docs.python.org/3/library/dataclasses.html#dataclasses.KW_ONLY
+
+from dataclasses import KW_ONLY, dataclass
+
+
+@dataclass
+class DC1:
+    a: int
+    __: KW_ONLY
+    b: str
+
+
+DC1(1, b="hi")
+DC1(a=1, b="hi")
+
+# This should generate an error because "b" is keyword-only.
+DC1(1, "hi")
+
+
+@dataclass
+class DC2:
+    a: int = 0
+    __: KW_ONLY
+    b: str = "hi"
+
+
+DC2(a=0, b="hi")
+DC2(b="hi")
+
+# This should generate an error because "b" is keyword-only.
+DC2(0, "hi")
+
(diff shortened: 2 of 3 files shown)


rchiodo (approved):
Approved via Review Center.

rchiodo:
Looks good — correct, well-scoped fix that aligns with the documented behavior that the KW_ONLY separator name is ignored, with appropriate regression coverage. Approving.

rchiodo (approved):
Approved via Review Center.

PR #11488, opened 2026-06-14
Fix reachability for generic context managers with bool __exit__

### LINKED ISSUE #11483: Incorrect type inference from context manager __exit__ return type with Generic
**Describe the bug**
When a context manager's `__exit__()` function is annotated (or inferred) as returning `None` or `Literal[False]`, pyright (correctly) infers that the entire body of the `with` block must have run for any code after the block, since the context manager (if the typing is correct) won't suppress any exceptions. If the `__exit__()` function is annotated (or inferred) as returning `bool` or `Literal[True]`, then it (correctly) infers that some or none of the body of the `with` block may have run if there was an exception that was suppressed.

[Pyright playground link with all code samples](https://pyright-play.net/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMoDiApikSEgMYA0UAykQDbA0AqiRAagIYg0hEA3IlwYB9eAiIAoKSygBeKG0ncQACgDkLDTQBGYAK4oAJvI36wDAD4A5MKQ0BKGRQZcAzu6gBhezCIAHjAAslwoXGhkdqQAXFJQCVDGRMBQoqIk-iDpau6MwI5QALQAfHT5cYlVUAB0dfGJyanpgUgwOXlMNIEU4uzdAb0CIgO9MLqFpVDRRJXVCXU1Lm6ePn6BIWERZABCYJZzCU1pGShZHfmTZfRMh9WLDUcpJ63torn5o32SX8MMX%2BMrlALAw7lUHlJXB4vL4zhtQuFIupiKRyBQANosAC6jjuxxacOy706BWK1wqj3u9Sq%2BIyATaFy6UB63yIvxGzMG4gmZKUYMSEIAxFAAKKDIgIGBIexQUhCEBQdwGBAIfieIjGGjuMBQCAGdz4EBGKAAA2A%2BwUUAAjCapOadYoNAaQBopAB3NoACzWcKCCO2IBmakKHigFBgATu-CEIlZanDARqLXpbzUMxo6em9iIjkKwoASoJhAwNVBQyaZraqvbLVapNHi3H7XmoIWYyXjGWvCaADJtMgidFWrG2qTCsUUCVSmU4fxnJAiBhwRXK1VEdWaxU6iBcZcoMCG41mi2KG12k9QJ0wF3ur0%2B-x%2BrZIvaWYNdsMRqNF2MSIjxiNJnSDLvJmoHZrmUAFt%2BHbviaIJVokNanvW0FNvsLZtsWpbln2WSDleLo0MOo7juKkrSigspFgqSoqmqeSbtqur6oelHHg61q2khl7Oq6HowN6sIPpsiJkOiMxYm%2BoYJl%2B7ZxgmgGvDkYGkBm4EYdB2HdpWlLcXWDY-uwajNpBraaZ2OH9iAg7ETIpGTuRM4HpkC4MEuK50euDFatuu6ygeUBGmxelcReBF8XeQnwk%2BYkgpJIZeDJlIGWIv7-omybAWm2ZqapWakBBUHtlpprwbpF76ahaUmcKACEACSNjeAA8vm%2BYit4LC1WZxUWd2uEDgwQ4jlIQA)

```python
from typing import Literal, Self, reveal_type

class ContextManagerNone:
    def __enter__(self) -> Self:
        ...
    def __exit__(self, exc_type, exc_val, exc_tb) -> None:
        ...

class ContextManagerBool:
    def __enter__(self) -> Self:
        ...
    def __exit__(self, exc_type, exc_val, exc_tb) -> bool:
        ...

# Exception never suppressed, so must run `foo = 1`
foo = 'str'
with ContextManagerNone() as ctx:
    reveal_type(ctx.__exit__(None, None, None)) # Revealed as `None`
    foo = 1
reveal_type(foo) # Revealed as `Literal[1]`

# Exception potentially suppressed, so may not run `foo = 1`
foo = 'str'
with ContextManagerBool() as ctx:
    reveal_type(ctx.__exit__(None, None, None)) # Revealed as `bool`
    foo = 1
reveal_type(foo) # Revealed as `Literal['str', 1]`
```

So far so good. The problem comes when the context manager is a Generic and the return type of the `__exit__()` function is annotated with a TypeVar:

```python
from typing import Generic, Self, TypeVar, reveal_type

T = TypeVar('T', bound='bool|None')

class ContextManager(Generic[T]):
    def __enter__(self) -> Self:
        ...
    def __exit__(self, exc_type, exc_val, exc_tb) -> T:
        ...

# Exception never suppressed, so must run `foo = 1`
foo = 'str'
with ContextManager[None]() as ctx:
    reveal_type(ctx.__exit__(None, None, None)) # Revealed as `None`
    foo = 1
reveal_type(foo) # Revealed as `Literal[1]`

# Exception potentially suppressed, so may not run `foo = 1`
foo = 'str'
with ContextManager[bool]() as ctx:
    reveal_type(ctx.__exit__(None, None, None)) # Revealed as `bool`
    foo = 1
reveal_type(foo) # !!!INCORRECT!!! Revealed as `Literal[1]`
```

On the last line, despite revealing the expected type for the return value of `__exit__()`, pyright seems to behave as if the context manager won't suppress any exceptions.

First spotted while running pylance as a VS Code extension (version 2026.2.1). Confirmed using the pyright playground.

KirtiRamchandani:
Fixes #11483

`isExceptionContextManager` only treated `__exit__` as swallowing exceptions when the declared return type was literally `bool`. For `ContextManager[bool]`, the declared return is the TypeVar `T`.

Specialize the return type from the context manager instance type args before the bool check. Added `with7.py` regression sample.

### DIFF
--- a/packages/pyright-internal/src/analyzer/codeFlowEngine.ts
+++ b/packages/pyright-internal/src/analyzer/codeFlowEngine.ts
@@ -1979,6 +1979,12 @@ export function getCodeFlowEngine(
                         }
                     }
 
+                    // Generic context managers (e.g. ContextManager[bool]) declare __exit__ as returning
+                    // a TypeVar. Specialize using the instance type args before checking for bool.
+                    if (isTypeVar(returnType) && cmType.priv.typeArgs && cmType.priv.typeArgs.length >= 1) {
+                        returnType = cmType.priv.typeArgs[0];
+                    }
+
                     cmSwallowsExceptions = false;
                     if (isClassInstance(returnType) && ClassType.isBuiltIn(returnType, 'bool')) {
                         if (returnType.priv.literalValue === undefined || returnType.priv.literalValue === true) {
--- a/packages/pyright-internal/src/analyzer/codeFlowEngine.ts
+++ b/packages/pyright-internal/src/analyzer/codeFlowEngine.ts
@@ -68,6 +68,8 @@ import {
     UnknownType,
 } from './types';
 import {
+    applySolvedTypeVars,
+    buildSolution,
     cleanIncompleteUnknown,
     derivesFromStdlibClass,
     doForEachSubtype,
@@ -1979,10 +1981,14 @@ export function getCodeFlowEngine(
                         }
                     }
 
-                    // Generic context managers (e.g. ContextManager[bool]) declare __exit__ as returning
-                    // a TypeVar. Specialize using the instance type args before checking for bool.
-                    if (isTypeVar(returnType) && cmType.priv.typeArgs && cmType.priv.typeArgs.length >= 1) {
-                        returnType = cmType.priv.typeArgs[0];
+                    // Generic context managers can declare __exit__ as returning a TypeVar
+                    // that isn't necessarily the first type parameter. Specialize the declared
+                    // return type using the context manager instance's type arguments.
+                    if (cmType.shared.typeParams.length > 0 && cmType.priv.typeArgs) {
+                        returnType = applySolvedTypeVars(
+                            returnType,
+                            buildSolution(cmType.shared.typeParams, cmType.priv.typeArgs)
+                        );
                     }
 
                     cmSwallowsExceptions = false;
--- a/packages/pyright-internal/src/tests/samples/with7.py
+++ b/packages/pyright-internal/src/tests/samples/with7.py
@@ -1,0 +1,20 @@
+# Regression test for https://github.com/microsoft/pyright/issues/11483
+# Generic ContextManager[T] should treat __exit__ return type like typing.ContextManager[bool].
+
+from typing import Generic, Self, TypeVar, reveal_type
+
+T = TypeVar("T", bound="bool|None")
+
+
+class ContextManager(Generic[T]):
+    def __enter__(self) -> Self: ...
+
+    def __exit__(self, exc_type, exc_val, exc_tb) -> T: ...
+
+
+def test_generic_bool_context_manager() -> None:
+    foo = "str"
+    with ContextManager[bool]() as ctx:
+        reveal_type(ctx.__exit__(None, None, None), expected_text="bool")
+        foo = 1
+    reveal_type(foo, expected_text="Literal['str', 1]")
--- a/packages/pyright-internal/src/tests/samples/with7.py
+++ b/packages/pyright-internal/src/tests/samples/with7.py
@@ -18,3 +18,21 @@ def test_generic_bool_context_manager() -> None:
         reveal_type(ctx.__exit__(None, None, None), expected_text="bool")
         foo = 1
     reveal_type(foo, expected_text="Literal['str', 1]")
+
+
+TEnter = TypeVar("TEnter")
+TExit = TypeVar("TExit", bound="bool|None")
+
+
+class SplitContextManager(Generic[TEnter, TExit]):
+    def __enter__(self) -> TEnter: ...
+
+    def __exit__(self, exc_type, exc_val, exc_tb) -> TExit: ...
+
+
+def test_multi_typevar_context_manager() -> None:
+    bar = "str"
+    with SplitContextManager[int, bool]() as ctx:
+        reveal_type(ctx, expected_text="int")
+        bar = 1
+    reveal_type(bar, expected_text="Literal['str', 1]")
--- a/packages/pyright-internal/src/tests/checker.test.ts
+++ b/packages/pyright-internal/src/tests/checker.test.ts
@@ -212,6 +212,12 @@ test('With6', () => {
     TestUtils.validateResults(analysisResults, 0);
 });
 
+test('With7', () => {
+    const analysisResults = TestUtils.typeAnalyzeSampleFiles(['with7.py']);
+
+    TestUtils.validateResults(analysisResults, 0);
+});
+
 test('Mro1', () => {
     const analysisResults = TestUtils.typeAnalyzeSampleFiles(['mro1.py']);
 


packages/pyright-internal/src/analyzer/codeFlowEngine.ts
```diff
@@ -1979,6 +1979,12 @@ export function getCodeFlowEngine(
                         }
                     }
 
+                    // Generic context managers (e.g. ContextManager[bool]) declare __exit__ as returning
```
rchiodo: Copilot generated:
Positional `typeArgs[0]` is the wrong type argument for any context manager whose `__exit__` TypeVar isn't the class's first type parameter. All three reviewers tool-verified this: `AbstractContextManager(Protocol[_T_co, _ExitT_co])` puts the `__enter__` type at slot 0 and the `__exit__` type at slot 1. Concretely, `CM[bool, int]` with `def __exit__(...) -> T` (T at slot 1, actual return `int`) is read as `typeArgs[0]` = `bool` → `cmSwallowsExceptions = true`, a regression: pre-PR this correctly returned `false`. Conversely `CM[int, bool]` reads slot 0 = `int` and fails to fix the bug. Replace with identity-aware specialization: `returnType = partiallySpecializeType(returnType, cmType, /* typeClassType */ undefined);` (typeUtils.ts:1496). This resolves the TypeVar by identity regardless of slot, handles multi-param, inherited-base, and async cases uniformly, and lets you drop the `isTypeVar` / `typeArgs.length` guards entirely.

[verified]

rchiodo: Copilot generated:
The PR's titled scenario `typing.ContextManager[bool]` only passes by coincidence. Tool-verified: `typing.ContextManager[_T_co]` subclasses `AbstractContextManager[_T_co, bool | None]`, so its real `__exit__` return is the union `bool | None`, while `typeArgs[0]` reads the *enter* type `bool`. `typing.ContextManager[int]` (enter `int`, exit still `bool | None`) would regress to not-swallowing. Fixing the slot via `partiallySpecializeType` addresses this, but note the downstream `isClassInstance(returnType) && ClassType.isBuiltIn(returnType, 'bool')` check (line ~1985) still won't recognize `bool` inside a `bool | None` union — consider widening it to treat a union as swallowing if any subtype is a non-`False`-literal `bool`.

[verified]

rchiodo: Copilot generated:
The new block executes after the `__aexit__` Coroutine unwrap, so async generic managers (`async with CM[bool, int]()` where `__aexit__ -> Coroutine[Any, Any, T]`) hit the same positional mismapping and the same reachability corruption. The PR neither addresses the async case correctly nor tests it. The `partiallySpecializeType` fix above covers sync and async uniformly.

[verified]

rchiodo: Copilot generated:
Root-cause note (all reviewers): the fix targets the right layer — specializing the `__exit__` return before the bool check rather than guarding reachability downstream — but uses the wrong mechanism. The underlying cause is reading the *unspecialized* `exitType.shared.declaredReturnType` and then re-deriving the substitution by hand with an index, reaching around the type system's existing identity-based specialization. This is 'right place, wrong tool,' which the recommended `partiallySpecializeType` call resolves at the origin.

[verified]

packages/pyright-internal/src/tests/samples/with7.py
```diff
@@ -0,0 +1,20 @@
+# Regression test for https://github.com/microsoft/pyright/issues/11483
+# Generic ContextManager[T] should treat __exit__ return type like typing.ContextManager[bool].
+
+from typing import Generic, Self, TypeVar, reveal_type
+
+T = TypeVar("T", bound="bool|None")
+
+
```
rchiodo, line 8: Copilot generated:
The sample only exercises the single-type-parameter shape (`class ContextManager(Generic[T])` with `__exit__ -> T` at slot 0) — the one case the positional shortcut happens to get right, so the suite stays green even though the implementation is wrong. Add cases that actually constrain the fix: (1) a two-typevar manager where `__exit__` returns the *second* TypeVar (e.g. `CM[bool, int]` returning the slot-1 type — the regression direction), (2) the named `typing.ContextManager[...]` case, (3) an inherited generic `__exit__` (TypeVar scoped to a base class in the MRO), and (4) an `async with` variant.

[verified]

packages/pyright-internal/src/analyzer/codeFlowEngine.ts
```diff
@@ -1979,6 +1979,12 @@ export function getCodeFlowEngine(
...
+                    // a TypeVar. Specialize using the instance type args before checking for bool.
+                    if (isTypeVar(returnType) && cmType.priv.typeArgs && cmType.priv.typeArgs.length >= 1) {
+                        returnType = cmType.priv.typeArgs[0];
+                    }
```
rchiodo, line 1992: Using `cmType.priv.typeArgs[0]` assumes the `__exit__` return TypeVar is always the class's *first* type parameter. That isn't generally true. The stdlib type this PR is modeled on, `contextlib.AbstractContextManager(Protocol[_T_co, _ExitT_co])`, declares `__exit__ -> _ExitT_co` (the **second** parameter), so this picks the wrong arg. It also creates a new false positive: for `class CM(Generic[A, B]): def __exit__(...) -> B`, `CM[bool, None]()` actually returns `None` from `__exit__`, but this grabs `typeArgs[0] == bool` and wrongly marks the manager as swallowing exceptions, corrupting reachability/narrowing. Prefer resolving the declared TypeVar to its specialized return type (e.g. via the class's type-var context / effective return type) rather than assuming index 0.

rchiodo, line 1992: Note that `typing.ContextManager[bool]` expands to `AbstractContextManager[bool, bool | None]`, where `_ExitT_co` defaults to `bool | None`. So `__exit__`'s return resolves to the *union* `bool | None`, not a bare TypeVar. `isTypeVar(returnType)` is then false, this block is skipped, and the downstream `isClassInstance(returnType, 'bool')` check fails on the union, meaning the actual stdlib case named in the sample may remain unfixed. Consider adding a test against `contextlib.AbstractContextManager` and handling a `bool | None` union in the bool check.

packages/pyright-internal/src/tests/samples/with7.py
```diff
@@ -0,0 +1,20 @@
...
+    def __exit__(self, exc_type, exc_val, exc_tb) -> T: ...
+
+
+def test_generic_bool_context_manager() -> None:
+    foo = "str"
+    with ContextManager[bool]() as ctx:
+        reveal_type(ctx.__exit__(None, None, None), expected_text="bool")
+        foo = 1
+    reveal_type(foo, expected_text="Literal['str', 1]")
```
rchiodo, line 20: This sample only covers a single-TypeVar generic where `typeArgs[0]` coincides with the exit TypeVar, which is the one shape the current fix gets right. Add a multi-TypeVar regression case (e.g. `class CM(Generic[A, B])` with `__exit__ -> B`, instantiated as `CM[None, bool]` and as `CM[bool, None]`) asserting the correct swallow / no-swallow reachability behavior. This documents intent and guards against the positional-index pitfall.

rchiodo:
The core fix idea is right, but resolving the exit return type by grabbing `typeArgs[0]` is positionally fragile. Resolve the TypeVar to its actual specialized return type instead, and add a multi-TypeVar regression test to lock the behavior in.

rchiodo (approved):
Approved via Review Center.

rchiodo (approved):
Approved via Review Center.

rchiodo (approved):
Approved via Review Center.

PR #11489, opened 2026-06-14
Fix generic NamedTuple match narrowing for bounded type vars

### LINKED ISSUE #11487: Generic NamedTuple narrows type in `match` case to `Unknown`
**Describe the bug**

Pyright should be able to tell that `value.foo` is always a `bool`.

**Code or Screenshots**

Code sample in [pyright playground](https://pyright-play.net/?strict=true&code=GYJw9gtgBALgngBwJYDsDmUkQWEMoByAhhAKYAmAKgK4IA2pAUMwMZ1EDOHUlAFqmgDalAFxQARmDB0AugApiZKrQYBKEYyhaowKWMrNG5UsFikOMOQDcidaqTFhxAK1IsYqqAFoAfFAsgGtpQEEQwLLxQNnYOmsHaLJykPPzocp5IptH2AHS6YEHxRSCkMNQgKFAARDCpaFVxRYkcyQD6hUXBJWUV1Q1AA)

```python
from typing import NamedTuple


class Thing[T: bool](NamedTuple):
    foo: T


def test(value: object) -> str:
    match value:
        case Thing() if value.foo:
            return "thing"
        case _:
            return ""

```

```
Type of "foo" is unknown  (reportUnknownMemberType)
```

KirtiRamchandani:
## Summary

Fixes pyright incorrectly reporting `Unknown` for fields on generic `NamedTuple` classes after a successful class pattern match when the type parameter has a bound (e.g. `Thing[T: bool]`).

## Problem

```python
class Thing[T: bool](NamedTuple):
    foo: T

def test(value: object) -> str:
    match value:
        case Thing():
            reveal_type(value.foo)  # was Unknown
```

## Fix

- When preparing class patterns, prefer bound type args over `Unknown` for bounded type params
- When narrowing after a match against `object`, specialize bounded type params to their bounds
- When replacing unsolved type vars during constraint application, fall back to bound types before `Unknown`

## Testing

- Added `matchNamedTupleGeneric1.py` + `NamedTuple12` test
- `npm run test:norebuild -- --testNamePattern=NamedTuple12` passes
- `MatchClass1` regression test passes

Fixes #11487

### DIFF
--- a/packages/pyright-internal/src/analyzer/patternMatching.ts
+++ b/packages/pyright-internal/src/analyzer/patternMatching.ts
@@ -754,12 +754,17 @@ function narrowTypeBasedOnLiteralPattern(
 function specializeBoundedMatchTypeParams(
     evaluator: TypeEvaluator,
     matchType: ClassType,
-    subjectTypeArgs: Type[] | undefined,
     solvedTypeArgs: Type[] | undefined,
     condition: ReturnType<typeof getTypeCondition>
 ): Type {
+    // `solvedTypeArgs` is indexed by `matchType`'s own type parameters, so the
+    // caller must align it to the pattern class (e.g. pass `resultType.priv.typeArgs`
+    // where `resultType` is the same generic class). The subject's own type arguments
+    // must not be used here: the subject may be a different generic class (e.g. a
+    // generic supertype), and indexing it by the pattern class's parameters would
+    // misread unrelated arguments.
     const typeArgs = matchType.shared.typeParams.map((param, index) => {
-        const specializedArg = subjectTypeArgs?.[index] ?? solvedTypeArgs?.[index];
+        const specializedArg = solvedTypeArgs?.[index];
 
         // Keep an argument unless it is genuinely Unknown. A bare in-scope TypeVar
         // (e.g. a subject `Thing[S]` from a generic function `def f[S: bool]`) is a
@@ -1075,13 +1080,9 @@ function narrowTypeBasedOnClassPattern(
                                 (param) => isTypeVar(param) && param.shared.boundType
                             )
                         ) {
-                            const subjectTypeArgs = isClassInstance(subjectSubtypeExpanded)
-                                ? subjectSubtypeExpanded.priv.typeArgs
-                                : undefined;
                             resultType = specializeBoundedMatchTypeParams(
                                 evaluator,
                                 unexpandedSubtype,
-                                subjectTypeArgs,
                                 resultType.priv.typeArgs,
                                 getTypeCondition(subjectSubtypeExpanded)
                             );
--- a/packages/pyright-internal/src/analyzer/patternMatching.ts
+++ b/packages/pyright-internal/src/analyzer/patternMatching.ts
@@ -68,7 +68,6 @@ import {
     addConditionToType,
     containsAnyOrUnknown,
     convertToInstance,
-    convertToInstantiable,
     doForEachSubtype,
     getTypeCondition,
     getTypeVarScopeIds,
@@ -86,7 +85,6 @@ import {
     partiallySpecializeType,
     preserveUnknown,
     requiresSpecialization,
-    selfSpecializeClass,
     specializeTupleClass,
     specializeWithUnknownTypeArgs,
     transformPossibleRecursiveTypeAlias,
@@ -748,6 +746,41 @@ function narrowTypeBasedOnLiteralPattern(
     });
 }
 
+// When a class pattern matches a generic class whose type parameters have an
+// upper bound (e.g. `class Thing[T: bool]`), the constraint solver may leave the
+// parameters unsolved (Unknown) because the subject carries no type arguments.
+// Rather than surfacing Unknown, fall back to each parameter's bound while
+// preserving any argument that was concretely solved. The subject's condition is
+// reapplied to the resulting instance.
+function specializeBoundedMatchTypeParams(
+    evaluator: TypeEvaluator,
+    matchType: ClassType,
+    subjectTypeArgs: Type[] | undefined,
+    solvedTypeArgs: Type[] | undefined,
+    condition: ReturnType<typeof getTypeCondition>
+): Type {
+    const typeArgs = matchType.shared.typeParams.map((param, index) => {
+        const specializedArg = subjectTypeArgs?.[index] ?? solvedTypeArgs?.[index];
+
+        // Keep an argument that was solved to a concrete (non-bound, non-Unknown) type.
+        if (
+            specializedArg &&
+            !requiresSpecialization(specializedArg) &&
+            !containsAnyOrUnknown(specializedArg, /* recurse */ true)
+        ) {
+            return specializedArg;
+        }
+
+        if (isTypeVar(param) && param.shared.boundType) {
+            return convertToInstance(param.shared.boundType);
+        }
+
+        return specializedArg ?? getUnknownForTypeVar(param, evaluator.getTupleClassType());
+    });
+
+    return addConditionToType(convertToInstance(ClassType.specialize(matchType, typeArgs)), condition);
+}
+
 function narrowTypeBasedOnClassPattern(
     evaluator: TypeEvaluator,
     type: Type,
@@ -920,10 +953,7 @@ function narrowTypeBasedOnClassPattern(
                 const expandedSubtypeInstance = convertToInstance(expandedSubtype);
                 const isPatternMetaclass = isMetaclassInstance(expandedSubtypeInstance);
 
-                return evaluator.mapSubtypesExpandTypeVars(
-                    type,
-                    /* options */ undefined,
-                    (subjectSubtypeExpanded, subjectSubtypeUnexpanded) => {
+                return evaluator.mapSubtypesExpandTypeVars(type, /* options */ undefined, (subjectSubtypeExpanded) => {
                     if (isAnyOrUnknown(subjectSubtypeExpanded)) {
                         if (isInstantiableClass(expandedSubtype) && ClassType.isBuiltIn(expandedSubtype, 'Callable')) {
                             // Convert to an unknown callable type.
@@ -1027,73 +1057,35 @@ function narrowTypeBasedOnClassPattern(
                                                 },
                                             }
                                         ) as ClassType;
-
-                                        if (
-                                            unexpandedSubtype.shared.typeParams.some(
-                                                (param) => isTypeVar(param) && param.shared.boundType
-                                            )
-                                        ) {
-                                            const subjectTypeArgs = isClassInstance(subjectSubtypeExpanded)
-                                                ? subjectSubtypeExpanded.priv.typeArgs
-                                                : undefined;
-                                            const solvedTypeArgs = isClassInstance(resultType)
-                                                ? resultType.priv.typeArgs
-                                                : undefined;
-                                            const typeArgs = unexpandedSubtype.shared.typeParams.map((param, index) => {
-                                                const specializedArg = subjectTypeArgs?.[index] ?? solvedTypeArgs?.[index];
-                                                if (specializedArg && !requiresSpecialization(specializedArg)) {
-                                                    return specializedArg;
-                                                }
-
-                                                if (isTypeVar(param) && param.shared.boundType) {
-                                                    return convertToInstance(param.shared.boundType);
-                                                }
-
-                                                return specializedArg ?? getUnknownForTypeVar(param, evaluator.getTupleClassType());
-                                            });
-
-                                            resultType = addConditionToType(
-                                                convertToInstance(ClassType.specialize(unexpandedSubtype, typeArgs)),
-                                                getTypeCondition(subjectSubtypeExpanded)
-                                            );
-                                        }
                                     }
-
                                 }
                             }
                         } else {
                             return undefined;
                         }
 
+                        // For a generic class pattern whose type parameters are bounded
+                        // (e.g. `class Thing[T: bool]`), the subject may carry no type arguments,
+                        // leaving the parameters unsolved. Fall back to each parameter's bound -
+                        // but only for the pattern class itself. When the subject narrowed to a
+                        // proper subclass of the pattern class, its own type arguments must be
+                        // preserved rather than rebuilt from the (widened) base.
                         if (
                             isClassInstance(resultType) &&
                             isInstantiableClass(unexpandedSubtype) &&
-                            unexpandedSubtype.shared.typeParams.some((param) => isTypeVar(param) && param.shared.boundType)
+                            ClassType.isSameGenericClass(resultType, ClassType.cloneAsInstance(unexpandedSubtype)) &&
+                            unexpandedSubtype.shared.typeParams.some(
+                                (param) => isTypeVar(param) && param.shared.boundType
+                            )
                         ) {
-                            const genericMatchType = unexpandedSubtype;
-                            const subjectTypeArgs = isClassInstance(subjectSubtypeUnexpanded)
-                                ? subjectSubtypeUnexpanded.priv.typeArgs
+                            const subjectTypeArgs = isClassInstance(subjectSubtypeExpanded)
+                                ? subjectSubtypeExpanded.priv.typeArgs
                                 : undefined;
-                            const solvedTypeArgs = resultType.priv.typeArgs;
-                            const typeArgs = genericMatchType.shared.typeParams.map((param, index) => {
-                                const specializedArg = subjectTypeArgs?.[index] ?? solvedTypeArgs?.[index];
-                                if (
-                                    specializedArg &&
-                                    !requiresSpecialization(specializedArg) &&
-                                    !containsAnyOrUnknown(specializedArg, /* recurse */ true)
-                                ) {
-                                    return specializedArg;
-                                }
-
-                                if (isTypeVar(param) && param.shared.boundType) {
-                                    return convertToInstance(param.shared.boundType);
-                                }
-
-                                return specializedArg ?? getUnknownForTypeVar(param, evaluator.getTupleClassType());
-                            });
-
-                            resultType = addConditionToType(
-                                convertToInstance(ClassType.specialize(genericMatchType, typeArgs)),
+                            resultType = specializeBoundedMatchTypeParams(
+                                evaluator,
+                                unexpandedSubtype,
+                                subjectTypeArgs,
+                                resultType.priv.typeArgs,
                                 getTypeCondition(subjectSubtypeExpanded)
                             );
                         }
@@ -1130,8 +1122,7 @@ function narrowTypeBasedOnClassPattern(
                     }
 
                     return undefined;
-                    }
-                );
+                });
             }
 
             return undefined;
--- a/packages/pyright-internal/src/analyzer/patternMatching.ts
+++ b/packages/pyright-internal/src/analyzer/patternMatching.ts
@@ -58,6 +58,7 @@ import {
     isNever,
     isSameWithoutLiteralValue,
     isTypeSame,
+    isTypeVar,
     isTypeVarTuple,
     isUnknown,
     isUnpackedTypeVar,
@@ -67,9 +68,11 @@ import {
     addConditionToType,
     containsAnyOrUnknown,
     convertToInstance,
+    convertToInstantiable,
     doForEachSubtype,
     getTypeCondition,
     getTypeVarScopeIds,
+    getUnknownForTypeVar,
     getUnknownTypeForCallable,
     isLiteralType,
     isLiteralTypeOrUnion,
@@ -82,6 +85,7 @@ import {
     mapSubtypes,
     partiallySpecializeType,
     preserveUnknown,
+    selfSpecializeClass,
     specializeTupleClass,
     specializeWithUnknownTypeArgs,
     transformPossibleRecursiveTypeAlias,
@@ -755,7 +759,14 @@ function narrowTypeBasedOnClassPattern(
     // specialize it with Unknown type arguments.
     if (isClass(exprType) && !exprType.props?.typeAliasInfo) {
         exprType = ClassType.cloneRemoveTypePromotions(exprType);
-        exprType = specializeWithUnknownTypeArgs(exprType, evaluator.getTupleClassType());
+        if (
+            isInstantiableClass(exprType) &&
+            exprType.shared.typeParams.some((param) => isTypeVar(param) && param.shared.boundType)
+        ) {
+            exprType = selfSpecializeClass(exprType, { useBoundTypeVars: true });
+        } else {
+            exprType = specializeWithUnknownTypeArgs(exprType, evaluator.getTupleClassType());
+        }
     }
 
     // Are there any positional arguments? If so, try to get the mappings for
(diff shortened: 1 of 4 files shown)


packages/pyright-internal/src/analyzer/typeUtils.ts
```diff
@@ -4281,6 +4290,16 @@ class ApplySolvedTypeVarsTransformer extends TypeVarTransformer {
             return this._solveDefaultType(typeVar, recursionCount);
```
rchiodo, line 4281: Copilot generated:
Global blast radius (High). All three non-rule reviewers independently flagged this: `ApplySolvedTypeVarsTransformer.transformTypeVar` is the central specialization engine, called from ~31 sites across 9 files. This unconditional unsolved-bounded-TypeVar→bound block fires for *every* `replaceUnsolved` caller, not just pattern matching — the Skeptic verified that `first([])` (with `T: bool`) infers `Unknown` today and would now become its bound engine-wide. The comment even admits the motive is pattern-matching-specific. Localize this: add an opt-in flag on `replaceUnsolved` (e.g. `useBoundForUnsolved`) and thread it from the pattern-matching path (`patternMatching.ts:1012`) rather than changing the default for all callers. This is an upstream-owned subrepo requiring surgical diffs.

[verified]

rchiodo, line 4281: Copilot generated:
Missing `useUnknown` guard (High). The Advocate and Skeptic both note the existing default-type branch at line 4291 is guarded by `&& !this._options.replaceUnsolved?.useUnknown`, but this new bound block has no such guard and sits right before the final return. A caller that explicitly passes `replaceUnsolved.useUnknown = true` (the "force Unknown" path, e.g. `specializeWithUnknownTypeArgs` consumers) would now silently receive bounds instead — overriding documented caller intent. Gate this block with `&& !this._options.replaceUnsolved?.useUnknown`.

[verified]

packages/pyright-internal/src/analyzer/typeUtils.ts
```diff
@@ -4192,6 +4192,15 @@ class ApplySolvedTypeVarsTransformer extends TypeVarTransformer {
```
rchiodo, line 4192: Copilot generated:
Solved-to-Unknown→bound changes error-recovery inference globally (Medium). The Skeptic verified this keys on `isUnknown` (not `isAny`), so `Any` propagation is correctly preserved — good. But a bounded TypeVar genuinely *solved* to `Unknown` (e.g. an argument sourced from an unresolved import / error type) will now snap to its bound across the whole evaluator, changing gradual/error-recovery behavior. Same fix as above: gate behind the pattern-match-specific option rather than the global transformer.

[verified]

packages/pyright-internal/src/analyzer/patternMatching.ts
```diff
@@ -755,7 +759,14 @@ function narrowTypeBasedOnClassPattern(
     // specialize it with Unknown type arguments.
```
rchiodo, line 795: Copilot generated:
Untested edge case: positional generic NamedTuple match (Medium). The Skeptic verified `selfSpecializeClass` (typeUtils.ts:1166-1184) does NOT receive `getTupleClassType()` and uses `cloneAsBound`, whereas the replaced `specializeWithUnknownTypeArgs(exprType, evaluator.getTupleClassType())` specifically threaded the tuple class type. For a NamedTuple, positional matching (`case Thing(x):`) relies on tuple structure via `__match_args__`. The sample only tests `case Thing()` with attribute access, so a positional-match regression here would be silent. Add a positional-match sample to confirm this path still works.

[verified]

packages/pyright-internal/src/analyzer/typeUtils.ts
rchiodo, line 4192: Copilot generated:
Introduced duplication (Architect, High). The bound-conversion ternary `TypeBase.isInstantiable(x) ? convertToInstantiable(x.shared.boundType, /* includeSubclasses */ false) : convertToInstance(x.shared.boundType)` appears 3× (two here, one in `patternMatching.ts`). Extract a single `getBoundTypeForTypeVar(typeVar)` helper. The predicate `typeParams.some(p => isTypeVar(p) && p.shared.boundType)` is also duplicated 2× in patternMatching.ts.

[verified]

packages/pyright-internal/src/analyzer/patternMatching.ts
```diff
@@ -1020,6 +1031,32 @@ function narrowTypeBasedOnClassPattern(
                                             }
```
rchiodo, line 1060: Copilot generated:
This block clobbers correctly-solved type args (Medium). The Skeptic showed `resultType` is first computed by `solveAndApplyConstraints` (lines 1012-1022), then **unconditionally overwritten** whenever any param has a bound. Solved args are discarded — e.g. a subject `Thing[Literal[True]]` matched by `case Thing():` widens to the bound `bool`; and unbounded params in the same class get substituted with `getUnknownForTypeVar` even when they were solved. This is an untested narrowing regression and is likely **redundant** with the typeUtils change (constraint solving already routes through `replaceUnsolved`). Either drop this block, or only fill the bound for params that remained unsolved while preserving solved args.

[verified]

rchiodo, line 1060: Copilot generated:
Shotgun-fix consolidation (High). All three reviewers note there are three overlapping mechanisms implementing the same "use the bound" idea: (1) the early `selfSpecializeClass`, (2) this post-match override, and (3) the global typeUtils transformer. The Advocate recommends a concrete experiment: revert both typeUtils hunks, keep only the patternMatching changes, and run `NamedTuple12` + `MatchClass1`; if green, the global edits are unnecessary. Prove the single minimal load-bearing site and delete the others before merge.

[verified]

packages/pyright-internal/src/analyzer/patternMatching.ts
```diff
@@ -755,7 +759,14 @@ function narrowTypeBasedOnClassPattern(
     // specialize it with Unknown type arguments.
     if (isClass(exprType) && !exprType.props?.typeAliasInfo) {
         exprType = ClassType.cloneRemoveTypePromotions(exprType);
-        exprType = specializeWithUnknownTypeArgs(exprType, evaluator.getTupleClassType());
+        if (
```
rchiodo: Copilot generated:
The Architect flags `useBoundTypeVars` as a naming trap: verified at typeUtils.ts:1166-1185/:210-217, it does NOT mean the TypeVar's `boundType` — it calls `TypeVarType.cloneAsBound` (the scope-bound-vs-free distinction), filling args with `T@Thing`, not `bool`. The actual `T→bool` substitution happens later in the typeUtils transformer. Two different meanings of "bound" collide in one fix. Add a short comment clarifying this so a future reader doesn't misread the intent.

[verified]

packages/pyright-internal/src/analyzer/patternMatching.ts
rchiodo, line 795: Copilot generated:
Implicit coupling (Architect, Low). The early `selfSpecializeClass(... useBoundTypeVars ...)` only produces the correct result because the typeUtils transformer later rewrites unsolved bounded vars to their bound. Neither side documents the dependency — altering one silently regresses the other (action-at-a-distance). If both edits are kept, add a comment noting this contract.

[verified]

packages/pyright-internal/src/tests/typeEvaluator8.test.ts
```diff
@@ -629,6 +629,14 @@ test('NamedTuple11', () => {
     TestUtils.validateResults(analysisResults, 3);
 });
 
+test('NamedTuple12', () => {
```
rchiodo, line 632: Copilot generated:
Validation depth below blast radius (High, raised by all). The PR ran only `NamedTuple12` + `MatchClass1`, but the typeUtils edits touch the hottest substitution routine in the evaluator. Run the full `typeEvaluator1..8` suites and review sample-snapshot deltas before merge. If the global change is kept, add a test that pins an unsolved bounded TypeVar *outside* pattern matching (e.g. `first([])`) to its intended result, plus coverage for constrained TypeVars (`T: (int, str)`, which use `shared.constraints` not `shared.boundType` and currently fall through to `Unknown`) and instantiable bounded params.

[verified]

rchiodo:
The core fix is reasonable, but it spreads across three overlapping mechanisms and pushes pattern-matching-specific policy into the shared, upstream-owned ApplySolvedTypeVarsTransformer. Please scope the bound-fallback to the pattern-matching path, avoid unconditionally discarding the constraint-solved resultType, and run the full typeEvaluator* suite to confirm no regressions before merge.

rchiodo:
The core fix is right to target generic NamedTuple match narrowing, but it spreads across four overlapping mechanisms — two of which are global edits to the shared `ApplySolvedTypeVarsTransformer` substitution primitive. Please scope the change to pattern matching, avoid overriding the explicit `useUnknown` contract, stop unconditionally overwriting the constraint-solved `resultType`, and bisect which mechanisms are actually load-bearing for `NamedTuple12` so the redundant ones can be removed. Consolidating the duplicated bound-conversion idiom and broadening test coverage (constrained/mixed/positional/concrete-arg cases) would also harden this.

rchiodo:
The targeted `patternMatching.ts` self-specialization is reasonable, but the fix currently spreads across shared type-substitution infrastructure and adds an unconditional result overwrite. Please scope the bound-fallback to pattern matching (avoid changing the global unsolved-TypeVar behavior in `typeUtils.ts`), preserve concretely-solved type args instead of widening back to the bound, and run the full `typeEvaluator*` suite before merge.

rchiodo:
The core fix is well-targeted at the pattern-matching path, but it leaks feature-specific policy into shared infrastructure (typeUtils.ts) and layers multiple overlapping mechanisms. Please scope the bound-fallback to the pattern-matching caller (or revert the typeUtils.ts changes if patternMatching.ts alone fixes the sample), avoid widening concretely-solved type args, and consolidate the duplicated bound-projection idiom. Given packages/pyright is an upstream-owned subrepo, keeping the fix localized to pattern matching is strongly preferred.

packages/pyright-internal/src/analyzer/patternMatching.ts
```diff
@@ -82,6 +85,8 @@ import {
     mapSubtypes,
     partiallySpecializeType,
     preserveUnknown,
+    requiresSpecialization,
+    selfSpecializeClass,
```
rchiodo: 📍 packages/pyright-internal/src/analyzer/patternMatching.ts:71
`convertToInstantiable` and `selfSpecializeClass` are added to the import list but referenced nowhere in the change. The repo enforces `@typescript-eslint/no-unused-vars: "error"`, so `npm run check` will fail and block merge. Remove both imports. Their presence suggests an abandoned `selfSpecializeClass`-based approach — worth confirming that path wasn't the intended cleaner fix.

packages/pyright-internal/src/analyzer/patternMatching.ts
```diff
@@ -915,7 +920,10 @@ function narrowTypeBasedOnClassPattern(
                 const expandedSubtypeInstance = convertToInstance(expandedSubtype);
                 const isPatternMetaclass = isMetaclassInstance(expandedSubtypeInstance);
 
-                return evaluator.mapSubtypesExpandTypeVars(type, /* options */ undefined, (subjectSubtypeExpanded) => {
+                return evaluator.mapSubtypesExpandTypeVars(
+                    type,
+                    /* options */ undefined,
+                    (subjectSubtypeExpanded, subjectSubtypeUnexpanded) => {
```
rchiodo: 📍 packages/pyright-internal/src/analyzer/patternMatching.ts:923
The `mapSubtypesExpandTypeVars(...)` call was rewrapped across multiple lines and the arrow `(subjectSubtypeExpanded, subjectSubtypeUnexpanded) => {` moved to a deeper nesting level, but the large callback body keeps its prior indentation rather than being re-indented one level inside the arrow. As written this will fail `npm run check:prettier`. Run `npx prettier --write` on the file before merge.

packages/pyright-internal/src/analyzer/patternMatching.ts
```diff
@@ -1019,13 +1027,77 @@ function narrowTypeBasedOnClassPattern(
...
+                                return specializedArg ?? getUnknownForTypeVar(param, evaluator.getTupleClassType());
+                            });
+
+                            resultType = addConditionToType(
+                                convertToInstance(ClassType.specialize(genericMatchType, typeArgs)),
+                                getTypeCondition(subjectSubtypeExpanded)
+                            );
+                        }
+
                         // Are there any positional arguments? If so, try to get the mappings for
```
rchiodo, line 1091: 📍 packages/pyright-internal/src/analyzer/patternMatching.ts:1030
Block A (inserted right after `resultType = solveAndApplyConstraints(...)`) appears to be dead code. When it runs, control falls through to Block B (added after the if/else-if/else), whose guard `isClassInstance(resultType) && isInstantiableClass(unexpandedSubtype) && <bounded>` is a superset of A's post-state — so B always recomputes `resultType` and discards A's result. The two blocks have also already drifted: B adds a `containsAnyOrUnknown(arg, /* recurse */ true)` gate that A lacks. Per AGENTS.md "avoid copy/paste variants", delete Block A and extract Block B into a single helper, e.g. `specializeBoundedMatchTypeParams(matchType, subjectTypeArgs, solvedTypeArgs, condition)`. Confirm `NamedTuple12` still passes after deleting A.

packages/pyright-internal/src/analyzer/patternMatching.ts
```diff
@@ -1058,7 +1130,8 @@ function narrowTypeBasedOnClassPattern(
                     }
```
rchiodo, line 1120: 📍 packages/pyright-internal/src/analyzer/patternMatching.ts:1102
Potential subclass-narrowing regression. This block runs for ALL class patterns with a bounded-generic pattern class, not just NamedTuple, and there is no `return` before it on the exact-match branch. So for a subject that is a subclass of a bounded-generic pattern class (e.g. `class Sub(Base[bool])` matched against `case Base()` where `class Base[T: bool]`), `resultType` (the subclass) is rebuilt from `unexpandedSubtype` (the base), potentially clobbering the narrowed subclass type. Add a subclass test case and either gate this block to the synthesized-pattern (`else if`) path, or skip it when `resultType` is a proper subtype of the pattern class.

packages/pyright-internal/src/tests/samples/matchNamedTupleGeneric1.py
```diff
@@ -0,0 +1,25 @@
...
+            return ""
+
+
+def test_literal(value: Thing[bool] | object) -> str:
+    match value:
+        case Thing(foo=True):
+            reveal_type(value.foo, expected_text="bool")
+            return "thing"
+        case _:
+            return ""
```
rchiodo, line 25: 📍 packages/pyright-internal/src/tests/samples/matchNamedTupleGeneric1.py:6
Test coverage is limited to a final bound (`bool`, which cannot be subclassed so `T ≡ bool`) — the degenerate case. Add coverage for: (1) a non-final bound (`T: SomeBase`) where narrowing-to-bound is an actual widening — the more interesting correctness case; (2) a subclass subject (see the clobbering concern in patternMatching.ts); and (3) optionally a constrained TypeVar (`T: (int, str)`), which currently still resolves to `Unknown`. As written, the suite would stay green even if the subclass regression ships.

rchiodo:
The core narrowing fix looks correct for the `Thing[T: bool]` repro, but please address the CI blockers first (remove the unused `convertToInstantiable`/`selfSpecializeClass` imports and run Prettier on the rewrapped callback). Beyond CI, the two near-identical bounded-typeparam specialization blocks should be de-duplicated into one helper (the inner block is overwritten by the outer one), and the outer block runs for all bounded-generic class patterns — add a subclass-subject and non-final-bound test to prove it doesn't clobber subclass narrowing.

packages/pyright-internal/src/analyzer/patternMatching.ts
```diff
@@ -743,6 +746,41 @@ function narrowTypeBasedOnLiteralPattern(
...
+    condition: ReturnType<typeof getTypeCondition>
+): Type {
+    const typeArgs = matchType.shared.typeParams.map((param, index) => {
+        const specializedArg = subjectTypeArgs?.[index] ?? solvedTypeArgs?.[index];
+
+        // Keep an argument that was solved to a concrete (non-bound, non-Unknown) type.
+        if (
+            specializedArg &&
+            !requiresSpecialization(specializedArg) &&
+            !containsAnyOrUnknown(specializedArg, /* recurse */ true)
```
rchiodo: New narrowing regression. The keep-guard `!requiresSpecialization(specializedArg)` over-widens a solved-to-TypeVar arg. `requiresSpecialization` returns `true` for any bare in-scope TypeVar, so for a subject from a generic context — e.g. `def f[S: bool](value: Thing[S])` matched by `case Thing()` — `specializedArg = S` fails the keep-branch and the helper returns `convertToInstance(param.shared.boundType)` = `bool`, silently replacing the legitimately-narrowed `S` with its bound. This applies to every bounded-generic class (not just NamedTuple) and has zero test coverage. Trigger replacement on genuine `Unknown` only — e.g. `if (specializedArg && !containsAnyOrUnknown(specializedArg, /* recurse */ true)) return specializedArg;` (drop the `requiresSpecialization` term) — and add a generic-function sample asserting `expected_text="S@f"`.

packages/pyright-internal/src/tests/samples/matchNamedTupleGeneric1.py
```diff
@@ -0,0 +1,59 @@
...
+    foo: T
+
+
+def test(value: object) -> str:
+    match value:
+        case Thing():
+            reveal_type(value.foo, expected_text="bool")
+            return "thing" if value.foo else ""
+        case _:
+            return ""
```
rchiodo, line 16: Positional-match coverage gap. Every sample uses only `case Thing()` / keyword `case Thing(foo=True)` / `case Container()`. Add a positional `case Thing(x)` / `case Container(x)` sample to confirm `__match_args__` binding still works for generic NamedTuples — a positional regression would otherwise ship silently.

rchiodo:
Core fix is sound, but the `!requiresSpecialization(specializedArg)` keep-guard will silently replace a narrowed in-scope TypeVar subject arg (e.g. `Thing[S]` where `S: bool`) with its bound. Please tighten that guard and add a generic-function regression sample before merge. Note: several first-pass comments referenced an earlier version (Block A/B, unused `convertToInstantiable`/`selfSpecializeClass` imports, missing non-final/subclass tests) that don't match this diff, so they were dropped.

rchiodo:
The new `specializeBoundedMatchTypeParams` helper over-widens a legitimately-narrowed type-var argument: because `requiresSpecialization` is true for any in-scope TypeVar, a bounded generic subject (e.g. `def f[S: bool](value: Thing[S])` matched by `case Thing()`) has its precise `S` replaced by its bound `bool`, a regression with no test coverage. Most other candidates are false positives against the actual diff (claimed unused imports, a duplicate 'Block A/B', and an unhandled subclass case are not present — the diff has a single block already guarded by `isSameGenericClass` with a passing `test_subclass`).

packages/pyright-internal/src/analyzer/patternMatching.ts
```diff
@@ -1026,6 +1064,32 @@ function narrowTypeBasedOnClassPattern(
...
+                        // leaving the parameters unsolved. Fall back to each parameter's bound -
+                        // but only for the pattern class itself. When the subject narrowed to a
+                        // proper subclass of the pattern class, its own type arguments must be
+                        // preserved rather than rebuilt from the (widened) base.
+                        if (
+                            isClassInstance(resultType) &&
+                            isInstantiableClass(unexpandedSubtype) &&
+                            ClassType.isSameGenericClass(resultType, ClassType.cloneAsInstance(unexpandedSubtype)) &&
+                            unexpandedSubtype.shared.typeParams.some(
+                                (param) => isTypeVar(param) && param.shared.boundType
```
rchiodo, line 1080: Sync↔async parity follow-up. The Pylance async copy of `narrowTypeBasedOnClassPattern` (packages/pylance-internal/src/analyzer/patternMatching.ts) is untouched, so the Pylance async product path (`enableAsyncProgram=true`) won't get this fix. Deferral is acceptable for an upstream-pyright-targeted PR, but track it as an explicit follow-up to mirror the fix and add Pylance-harness coverage, per the repo's sync↔async parity policy.

rchiodo:
Main blocker: the bounded-fallback helper should only replace genuine Unknown, not any arg that `requiresSpecialization` flags (which includes bare in-scope TypeVars), otherwise a subject like `Thing[S]` from a generic context gets widened to its bound. Worth a positional `case Thing(x)` test and a follow-up to mirror this into the Pylance async path.

packages/pyright-internal/src/analyzer/patternMatching.ts
```diff
@@ -743,6 +745,39 @@ function narrowTypeBasedOnLiteralPattern(
...
+// reapplied to the resulting instance.
+function specializeBoundedMatchTypeParams(
+    evaluator: TypeEvaluator,
+    matchType: ClassType,
+    subjectTypeArgs: Type[] | undefined,
+    solvedTypeArgs: Type[] | undefined,
+    condition: ReturnType<typeof getTypeCondition>
+): Type {
+    const typeArgs = matchType.shared.typeParams.map((param, index) => {
+        const specializedArg = subjectTypeArgs?.[index] ?? solvedTypeArgs?.[index];
```
rchiodo: 📍 packages/pyright-internal/src/analyzer/patternMatching.ts:759
New narrowing regression (High). `specializedArg = subjectTypeArgs?.[index] ?? solvedTypeArgs?.[index]` indexes `subjectTypeArgs` by the *pattern* class's `typeParams`, but `subjectTypeArgs` is sourced at the call site from `subjectSubtypeExpanded.priv.typeArgs` — which may be a *different* generic class. The Skeptic traced this and I confirmed it against the Branch-2 path (L982–1024): for subject `Base[int]` matched by `case Thing()` where `class Thing[T: bool](Base[int])`, the block fires (guard only relates `resultType` to `unexpandedSubtype`, never to the subject), and `Base.U=int` is misread as `Thing.T`, narrowing `value.foo` to `int` instead of the bound `bool`. This affects every bounded-generic class with a generic supertype and has zero coverage. Drop the `subjectTypeArgs` source and rely solely on `resultType.priv.typeArgs` (which the guard already aligns to the pattern class — it is redundant in Branch 1 and undefined for `object` subjects, so nothing covered is lost), and add a generic-supertype sample pinning the bound.

[verified]

packages/pyright-internal/src/analyzer/patternMatching.ts
```diff
@@ -743,6 +745,44 @@ function narrowTypeBasedOnLiteralPattern(
...
+    // must not be used here: the subject may be a different generic class (e.g. a
+    // generic supertype), and indexing it by the pattern class's parameters would
+    // misread unrelated arguments.
+    const typeArgs = matchType.shared.typeParams.map((param, index) => {
+        const specializedArg = solvedTypeArgs?.[index];
+
+        // Keep an argument unless it is genuinely Unknown. A bare in-scope TypeVar
+        // (e.g. a subject `Thing[S]` from a generic function `def f[S: bool]`) is a
+        // legitimately narrowed argument and must not be widened to the bound.
+        if (specializedArg && !containsAnyOrUnknown(specializedArg, /* recurse */ true)) {
```
rchiodo, line 772: The keep-guard `!containsAnyOrUnknown(specializedArg, /* recurse */ true)` can over-widen a structurally-known arg that merely contains a nested `Unknown`. For a subject `Box[list]` where `class Box[T: Sequence[int]]` (`list == list[Unknown]`), `resultType` is `Box[list[Unknown]]`; with `recurse: true` the walker descends into `list`'s arg, finds `Unknown`, discards the arg, and widens it to the bound `Sequence[int]` — silently losing the user-written `list` structure (pre-PR this returned the subject as-is). All current samples use `T: bool` (no nested structure), so this is uncovered. Consider `recurse: false` (top-level and union-member `Unknown` are still detected, so the `object`-subject and union cases still widen correctly), or gate the block to fire only when at least one arg is genuinely unsolved/Unknown — and add a container-bounded sample pinning the behavior.

rchiodo:
Looks solid overall and the generic-supertype case is already guarded and tested. One non-blocking note inline about the keep-guard potentially over-widening container-bounded type params with nested Unknown.

rchiodo (approved):
Approved via Review Center.

rchiodo:
Thanks for the PR

PR #11490, opened 2026-06-14
Allow Callable class attributes to be overridden by instance methods

### LINKED ISSUE #11475: Cannot override a class's callable variable with a method
**Describe the bug**

When a base class has a Callable variable without a `Self` parameter, overriding this variable in a subclass causes an error, which is inconsistent with the behavior of mypy and ty.

**Code or Screenshots**

```python
from typing import override


class A:
    hello: Callable[[], None] = lambda: print("hello")


class B(A):
    @override
    def hello(self) -> None:
        print("hi")


b = B()
b.hello()
# Output: hi
```

```ps
(report-callback-exception) PS G:\programming\python\lab\report_callback_exception> pyright main4.py
g:\programming\python\lab\report_callback_exception\main4.py
  g:\programming\python\lab\report_callback_exception\main4.py:11:9 - error: Method "hello" overrides class "A" in an incompatible manner
    Positional parameter count mismatch; base method has 0, but override has 1 (reportIncompatibleMethodOverride)
(report-callback-exception) PS G:\programming\python\lab\report_callback_exception> mypy main4.py        
(report-callback-exception) PS G:\programming\python\lab\report_callback_exception> ty check main4.py
All checks passed!
```

- mypy=2.1.0
- pyright=1.1.409
- ty=0.0.40

KirtiRamchandani:
## Summary

Fixes incompatible method override diagnostics when a base class declares a `Callable`-typed attribute and a subclass overrides it with an instance method.

## Problem

```python
class A:
    hello: Callable[[], None] = lambda: print("hello")

class B(A):
    @override
    def hello(self) -> None:
        print("hi")
```

Pyright reported a positional parameter count mismatch (0 vs 1). mypy and ty accept this.

## Fix

Treat a zero-parameter Callable class attribute overridden by a one-parameter instance method as compatible.

## Testing

- Added `methodOverrideCallable1.py` + `MethodOverrideCallable1` test (0 errors expected)

Fixes #11475

### DIFF
--- a/packages/pyright-internal/src/analyzer/typeEvaluator.ts
+++ b/packages/pyright-internal/src/analyzer/typeEvaluator.ts
@@ -27819,19 +27819,32 @@ export function createTypeEvaluator(
                     }
                 }
             } else if (overrideParamDetails.positionParamCount > baseParamDetails.positionParamCount) {
-                // Verify that all of the override parameters that extend the
-                // signature are either *args, **kwargs or parameters with
-                // default values.
+                const isCallableAttrOverriddenByInstanceMethod =
+                    baseParamDetails.positionParamCount === 0 &&
+                    overrideParamDetails.positionParamCount === 1 &&
+                    FunctionType.isInstanceMethod(overrideMethod) &&
+                    !FunctionType.isStaticMethod(baseMethod) &&
+                    !FunctionType.isClassMethod(baseMethod);
 
-                for (let i = baseParamDetails.positionParamCount; i < overrideParamDetails.positionParamCount; i++) {
-                    const overrideParam = overrideParamDetails.params[i].param;
+                if (!isCallableAttrOverriddenByInstanceMethod) {
+                    // Verify that all of the override parameters that extend the
+                    // signature are either *args, **kwargs or parameters with
+                    // default values.
 
-                    if (
-                        overrideParam.category === ParamCategory.Simple &&
-                        overrideParam.name &&
-                        !overrideParamDetails.params[i].defaultType
+                    for (
+                        let i = baseParamDetails.positionParamCount;
+                        i < overrideParamDetails.positionParamCount;
+                        i++
                     ) {
-                        foundParamCountMismatch = true;
+                        const overrideParam = overrideParamDetails.params[i].param;
+
+                        if (
+                            overrideParam.category === ParamCategory.Simple &&
+                            overrideParam.name &&
+                            !overrideParamDetails.params[i].defaultType
+                        ) {
+                            foundParamCountMismatch = true;
+                        }
                     }
                 }
             }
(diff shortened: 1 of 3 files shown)


packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -27819,19 +27819,32 @@ export function createTypeEvaluator(
...
-                // default values.
-
-                for (let i = baseParamDetails.positionParamCount; i < overrideParamDetails.positionParamCount; i++) {
-                    const overrideParam = overrideParamDetails.params[i].param;
+                const isCallableAttrOverriddenByInstanceMethod =
+                    baseParamDetails.positionParamCount === 0 &&
+                    overrideParamDetails.positionParamCount === 1 &&
+                    FunctionType.isInstanceMethod(overrideMethod) &&
+                    !FunctionType.isStaticMethod(baseMethod) &&
+                    !FunctionType.isClassMethod(baseMethod);
```
rchiodo, line 27827: 📍 packages/pyright-internal/src/analyzer/typeEvaluator.ts:27822
The guard hardcodes `baseParamDetails.positionParamCount === 0 && overrideParamDetails.positionParamCount === 1`, which only covers the zero-arg case. The semantically-identical `Callable[[int], None]` overridden by `def hello(self, x: int)` still errors (base count is 1, so the new branch is skipped), and the defaulted-param variant (`def zero(self, x: int = 5)` over `Callable[[], None]`) also still errors — both accepted by mypy/ty. The general rule is "non-method callable base with N positionals; instance-method override with N+1 whose extra leading param is `self`." Consider generalizing (bind the override, strip `self`, compare the bound signature against the base callable) so a near-identical issue doesn't resurface. If the scoped fix is preferred for upstreaming, at minimum document the narrowness.

packages/pyright-internal/src/tests/samples/methodOverrideCallable1.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class B(A):
+    @override
+    def hello(self) -> None:
+        print("hi")
+
+
+b = B()
+b.hello()
```
rchiodo, line 18: 📍 packages/pyright-internal/src/tests/samples/methodOverrideCallable1.py:1
The sample only exercises the happy path (0 errors). Consider adding a negative/lock-in case proving a genuinely incompatible callable-attr override still errors — e.g. `Callable[[], None]` overridden by `def hello(self, y) -> None` (a required extra non-`self` positional) or a wrong return type. Without it, nothing guards against a future broadening of this branch silently suppressing real `reportIncompatibleMethodOverride` errors.

rchiodo:
Looks good and correctly targets the reported case. Two non-blocking suggestions: the suppression is hardcoded to the zero-arg Callable case (the equivalent N-positional case still errors), and adding a negative lock-in test would guard against future broadening silently suppressing real override errors.

rchiodo (approved):
Approved via Review Center.

PR #11491, opened 2026-06-14
Fix NewType identity comparisons with base literals

### LINKED ISSUE #11463: Incorrect `is` behaviour with NewType
## Bug
Pyright considers `left is right` to always be false in cases where `left` is a `NewType` instance with a base of `T` and `right` is a `T` literal.

## Code
```python
from types import NoneType
from typing import NewType, reveal_type

Apple = NewType("Apple", NoneType)
Banana = NewType("Banana", bool)
Cherry = NewType("Cherry", int)

def f(a: Apple, b: Banana, c: Cherry) -> None:
    if a is None:
        reveal_type(a)  # pyright reports unreachable

    if b is True:
        reveal_type(b)  # pyright reports unreachable

    if c is False:
        reveal_type(c)  # pyright reports unreachable

f(Apple(None), Banana(True), Cherry(False))
```

Runtime output:
```
Runtime type is 'NoneType'
Runtime type is 'bool'
Runtime type is 'bool'
```

Interestingly, pyright also doesn't produce `reportUnnecessaryComparison` diagnostics for any of the conditions.

(this also works with `int` and an integer literal, but `is` with an integer literal is bad so I didn't include it)

## Pyright version
pyright 1.1.409, python 3.14

KirtiRamchandani:
## Summary

Fixes pyright treating valid `is` comparisons between `NewType` instances and their underlying base literals as unreachable.

## Problem

```python
Banana = NewType("Banana", bool)

def f(b: Banana) -> None:
    if b is True:  # incorrectly marked unreachable
        reveal_type(b)
```

## Fix

Compare `NewType` instances against their underlying base types when evaluating `is` narrowing for literals and `None`.

## Testing

- Added `newTypeIsLiteral1.py` + `NewType8` test (0 errors expected)

Fixes #11463

### DIFF
--- a/packages/pyright-internal/src/analyzer/typeGuards.ts
+++ b/packages/pyright-internal/src/analyzer/typeGuards.ts
@@ -1081,10 +1081,21 @@ function narrowTupleTypeForIsNone(evaluator: TypeEvaluator, type: Type, isPositi
     });
 }
 
+// Walks a chain of NewType instances down to the innermost non-NewType base
+// instance (e.g. NewType("Y", NewType("X", bool)) -> an instance of bool).
+// Returns undefined if `subtype` is not a NewType instance or its base cannot
+// be resolved to a class.
 function getInnermostNewTypeBaseInstance(subtype: Type): Type | undefined {
-    let currentType = subtype;
+    if (!isClassInstance(subtype) || !ClassType.isNewTypeClass(subtype)) {
+        return undefined;
+    }
 
-    while (isClassInstance(currentType) && ClassType.isNewTypeClass(currentType) && currentType.shared.baseClasses.length > 0) {
+    let currentType: Type = subtype;
+    while (isClassInstance(currentType) && ClassType.isNewTypeClass(currentType)) {
+        if (currentType.shared.baseClasses.length === 0) {
+            return undefined;
+        }
+
         const baseClass = currentType.shared.baseClasses[0];
         if (!isClass(baseClass)) {
             return undefined;
@@ -1094,20 +1105,6 @@ function getInnermostNewTypeBaseInstance(subtype: Type): Type | undefined {
     }
 
     return currentType;
-}
-
-function isNewTypeWithSingletonBase(
-    evaluator: TypeEvaluator,
-    subtype: Type,
-    targetType: Type,
-    isExactMatch: (type: Type) => boolean
-) {
-    const baseInstance = getInnermostNewTypeBaseInstance(subtype);
-    if (!baseInstance || baseInstance === subtype) {
-        return false;
-    }
-
-    return isExactMatch(baseInstance) || evaluator.assignType(baseInstance, targetType);
 }
 
 // Handle type narrowing for expressions of the form "x is None" and "x is not None".
@@ -1157,8 +1154,14 @@ function narrowTypeForIsNone(evaluator: TypeEvaluator, type: Type, isPositiveTes
 
             const adjustedSubtype = useExpandedSubtype ? subtype : unexpandedSubtype;
 
-            // NewType instances can be identity-compared with their underlying base type.
-            if (isNewTypeWithSingletonBase(evaluator, adjustedSubtype, evaluator.getNoneType(), isNoneInstance)) {
+            // A NewType whose innermost base is exactly None (e.g. NewType("Apple", NoneType))
+            // can be identity-compared with None: keep the NewType identity on the positive
+            // branch and eliminate it on the negative branch. A NewType whose base is merely
+            // None-compatible (e.g. NewType("Obj", object)) is intentionally not handled here;
+            // it falls through to the generic checks below so that "is not None" does not
+            // incorrectly collapse to Never.
+            const newTypeBaseInstance = getInnermostNewTypeBaseInstance(adjustedSubtype);
+            if (newTypeBaseInstance && isNoneInstance(newTypeBaseInstance)) {
                 resultIncludesNoneSubtype = true;
                 return isPositiveTest ? adjustedSubtype : undefined;
             }
@@ -1186,10 +1189,8 @@ function narrowTypeForIsNone(evaluator: TypeEvaluator, type: Type, isPositiveTes
     // of the subtypes are None types with conditions, retain those.
     if (isPositiveTest && resultIncludesNoneSubtype) {
         return mapSubtypes(result, (subtype) => {
-            return isNoneInstance(subtype) ||
-                isNewTypeWithSingletonBase(evaluator, subtype, evaluator.getNoneType(), isNoneInstance)
-                ? subtype
-                : undefined;
+            const baseInstance = getInnermostNewTypeBaseInstance(subtype);
+            return isNoneInstance(subtype) || (baseInstance && isNoneInstance(baseInstance)) ? subtype : undefined;
         });
     }
 
@@ -1231,7 +1232,11 @@ function narrowTypeForIsEllipsis(evaluator: TypeEvaluator, node: ExpressionNode,
                     ? unexpandedSubtype
                     : subtype;
 
-            if (isNewTypeWithSingletonBase(evaluator, adjustedSubtype, ellipsisType, isEllipsisInstance)) {
+            // Only a NewType whose innermost base is exactly the ellipsis type is treated
+            // as the singleton (see narrowTypeForIsNone for the rationale); a merely
+            // ellipsis-compatible base falls through so "is not ..." does not collapse to Never.
+            const newTypeBaseInstance = getInnermostNewTypeBaseInstance(adjustedSubtype);
+            if (newTypeBaseInstance && isEllipsisInstance(newTypeBaseInstance)) {
                 resultIncludesEllipsisSubtype = true;
                 return isPositiveTest ? adjustedSubtype : undefined;
             }
@@ -1257,8 +1262,8 @@ function narrowTypeForIsEllipsis(evaluator: TypeEvaluator, node: ExpressionNode,
     // of the subtypes are ellipsis types with conditions, retain those.
     if (isPositiveTest && resultIncludesEllipsisSubtype) {
         return mapSubtypes(result, (subtype) => {
-            return isEllipsisInstance(subtype) ||
-                isNewTypeWithSingletonBase(evaluator, subtype, ellipsisType, isEllipsisInstance)
+            const baseInstance = getInnermostNewTypeBaseInstance(subtype);
+            return isEllipsisInstance(subtype) || (baseInstance && isEllipsisInstance(baseInstance))
                 ? subtype
                 : undefined;
         });
--- a/packages/pyright-internal/src/analyzer/typeGuards.ts
+++ b/packages/pyright-internal/src/analyzer/typeGuards.ts
@@ -1081,6 +1081,25 @@ function narrowTupleTypeForIsNone(evaluator: TypeEvaluator, type: Type, isPositi
     });
 }
 
+function isNewTypeWithSingletonBase(
+    evaluator: TypeEvaluator,
+    subtype: Type,
+    targetType: Type,
+    isExactMatch: (type: Type) => boolean
+) {
+    if (!isClassInstance(subtype) || !ClassType.isNewTypeClass(subtype) || subtype.shared.baseClasses.length === 0) {
+        return false;
+    }
+
+    const baseClass = subtype.shared.baseClasses[0];
+    if (!isClass(baseClass)) {
+        return false;
+    }
+
+    const baseInstance = ClassType.cloneAsInstance(baseClass);
+    return isExactMatch(baseInstance) || evaluator.assignType(baseInstance, targetType);
+}
+
 // Handle type narrowing for expressions of the form "x is None" and "x is not None".
 function narrowTypeForIsNone(evaluator: TypeEvaluator, type: Type, isPositiveTest: boolean) {
     const expandedType = mapSubtypes(type, (subtype) => {
@@ -1129,22 +1148,9 @@ function narrowTypeForIsNone(evaluator: TypeEvaluator, type: Type, isPositiveTes
             const adjustedSubtype = useExpandedSubtype ? subtype : unexpandedSubtype;
 
             // NewType instances can be identity-compared with their underlying base type.
-            if (isClassInstance(adjustedSubtype) && ClassType.isNewTypeClass(adjustedSubtype) && adjustedSubtype.shared.baseClasses.length > 0) {
-                const baseClass = adjustedSubtype.shared.baseClasses[0];
-                if (isClass(baseClass)) {
-                    const baseInstance = ClassType.cloneAsInstance(baseClass);
-                    if (isNoneInstance(baseInstance)) {
-                        resultIncludesNoneSubtype = true;
-                        return isPositiveTest ? adjustedSubtype : undefined;
-                    }
-
-                    if (evaluator.assignType(baseInstance, evaluator.getNoneType())) {
-                        resultIncludesNoneSubtype = true;
-                        return isPositiveTest
-                            ? addConditionToType(evaluator.getNoneType(), subtype.props?.condition)
-                            : adjustedSubtype;
-                    }
-                }
+            if (isNewTypeWithSingletonBase(evaluator, adjustedSubtype, evaluator.getNoneType(), isNoneInstance)) {
+                resultIncludesNoneSubtype = true;
+                return isPositiveTest ? adjustedSubtype : undefined;
             }
 
             // Is it an exact match for None?
@@ -1170,7 +1176,10 @@ function narrowTypeForIsNone(evaluator: TypeEvaluator, type: Type, isPositiveTes
     // of the subtypes are None types with conditions, retain those.
     if (isPositiveTest && resultIncludesNoneSubtype) {
         return mapSubtypes(result, (subtype) => {
-            return isNoneInstance(subtype) ? subtype : undefined;
+            return isNoneInstance(subtype) ||
+                isNewTypeWithSingletonBase(evaluator, subtype, evaluator.getNoneType(), isNoneInstance)
+                ? subtype
+                : undefined;
         });
     }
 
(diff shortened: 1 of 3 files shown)


packages/pyright-internal/src/analyzer/typeGuards.ts
```diff
@@ -2741,8 +2760,19 @@ function narrowTypeForLiteralComparison(
             }
 
             if (isIsOperator || isNoneInstance(subtype)) {
-                const isSubtype = evaluator.assignType(subtype, literalType);
-                return isSubtype ? literalType : undefined;
+                let compareType: Type = subtype;
```
rchiodo: Copilot generated:
The Skeptic execution-verified that the fix only unwraps **one** level of `NewType`. For `X = NewType("X", bool)` then `Y = NewType("Y", X)`, `y is True` is still marked unreachable: `compareType` becomes the inner NewType `X`, so `assignType(X, Literal[True])` is false → returns `undefined` → dead branch. This is the exact bug class this PR fixes, one level deeper. Either walk the NewType base chain to the ultimate non-NewType base (a loop / shared helper) and add a nested-NewType row to `newTypeIsLiteral1.py`, or document the single-level limitation explicitly.

[verified]

packages/pyright-internal/src/analyzer/typeGuards.ts
```diff
@@ -1128,6 +1128,25 @@ function narrowTypeForIsNone(evaluator: TypeEvaluator, type: Type, isPositiveTes
 
             const adjustedSubtype = useExpandedSubtype ? subtype : unexpandedSubtype;
 
+            // NewType instances can be identity-compared with their underlying base type.
```
rchiodo: Copilot generated:
The Architect notes this new block re-implements None-detection (`isNoneInstance` + `assignType(..., getNoneType())`) that already exists in the original checks ~10 lines below. Preferred direction: unwrap the NewType to its base instance once and let the existing exact-None / potentially-None checks run against it, returning the original NewType `subtype` on the positive branch — mirroring the cleaner structure the literal-comparison hunk already uses. This collapses two parallel None paths into one.

[verified]

packages/pyright-internal/src/analyzer/typeGuards.ts
```diff
@@ -1128,6 +1128,25 @@ function narrowTypeForIsNone(evaluator: TypeEvaluator, type: Type, isPositiveTes
...
+            if (isClassInstance(adjustedSubtype) && ClassType.isNewTypeClass(adjustedSubtype) && adjustedSubtype.shared.baseClasses.length > 0) {
+                const baseClass = adjustedSubtype.shared.baseClasses[0];
+                if (isClass(baseClass)) {
+                    const baseInstance = ClassType.cloneAsInstance(baseClass);
+                    if (isNoneInstance(baseInstance)) {
+                        resultIncludesNoneSubtype = true;
+                        return isPositiveTest ? adjustedSubtype : undefined;
```
rchiodo: Copilot generated:
[unverified] The Skeptic flagged an identity inconsistency: when the base is `NoneType` (Apple case), this returns the NewType (`adjustedSubtype`), but when the base merely *contains* None (e.g. `NewType("Obj", object)`), the positive branch returns bare `None` via `addConditionToType(getNoneType(), …)` instead of the NewType. This diverges from the literal-comparison path, which always returns the NewType `subtype` on a positive match. Make this a deliberate decision and add a test either way.

[verified]

packages/pyright-internal/src/analyzer/typeGuards.ts
```diff
@@ -2741,8 +2760,19 @@ function narrowTypeForLiteralComparison(
             }
 
             if (isIsOperator || isNoneInstance(subtype)) {
-                const isSubtype = evaluator.assignType(subtype, literalType);
-                return isSubtype ? literalType : undefined;
+                let compareType: Type = subtype;
+                if (isClassInstance(subtype) && ClassType.isNewTypeClass(subtype) && subtype.shared.baseClasses.length > 0) {
```
rchiodo: Copilot generated:
The Architect, Skeptic, and Advocate all note the NewType-unwrap pattern (`isClassInstance(x) && isNewTypeClass(x) && baseClasses.length > 0` → `cloneAsInstance(baseClasses[0])`) is open-coded in both hunks, with `isNewTypeClass` checked a third time in this return ternary. Extracting a single `getNewTypeBaseInstance(t): ClassType | undefined` helper localizes the new dependency on the `baseClasses[0]` synthesis invariant (`createNewType`, typeEvaluator.ts:13788) to one place and makes the two hunks obviously symmetric.

[verified]

packages/pyright-internal/src/analyzer/typeGuards.ts
rchiodo: Copilot generated:
Nit (Advocate, structurally confirmed): the new guard line is ~145 columns including indentation, past the repo's `printWidth: 120`. Prettier will reflow it, so `npm run check` may bounce until `fix:prettier` is run. Mechanical — resolve before CI.

[verified]

packages/pyright-internal/src/tests/samples/newTypeIsLiteral1.py
```diff
@@ -0,0 +1,22 @@
...
+
+from types import NoneType
+from typing import NewType, reveal_type
+
+Apple = NewType("Apple", NoneType)
+Banana = NewType("Banana", bool)
+Cherry = NewType("Cherry", int)
+
+
+def f(a: Apple, b: Banana, c: Cherry) -> None:
```
rchiodo: Copilot generated:
The Advocate flags positive-only coverage. The sample exercises only positive `is` branches; there is no coverage for the negative branch (`b is not True`), for a `NewType` inside a union, for a nested `NewType` (`NewType("Y", X)` — see the verified gap above), or the `NewType("Obj", object) is None` identity case. Adding these would lock in the intended behavior and guard against the residual gaps the reviewers identified.

[verified]

packages/pyright-internal/src/analyzer/typeGuards.ts
```diff
@@ -1128,6 +1128,25 @@ function narrowTypeForIsNone(evaluator: TypeEvaluator, type: Type, isPositiveTes
...
+                    }
+
+                    if (evaluator.assignType(baseInstance, evaluator.getNoneType())) {
+                        resultIncludesNoneSubtype = true;
+                        return isPositiveTest
+                            ? addConditionToType(evaluator.getNoneType(), subtype.props?.condition)
+                            : adjustedSubtype;
+                    }
+                }
+            }
```
rchiodo, line 1167: 📍 packages/pyright-internal/src/analyzer/typeGuards.ts:1131
The `is None` fix does not fix its own primary scenario and produces the opposite of correct narrowing. For `Apple = NewType("Apple", NoneType)`, `if a is None:` should be the reachable branch (`Apple`), but the positive branch collapses to `Never`. The first sub-branch returns `adjustedSubtype` (the NewType instance, a non-`None` type) while setting `resultIncludesNoneSubtype = true`; the unchanged post-filter at lines 1152-1156 (`isNoneInstance(st) ? st : undefined`) then strips that subtype — `isNoneInstance` is `false` for a NewType because it lacks the `BuiltIn` flag — yielding `Never`. Fix: for the `isNoneInstance(baseInstance)` case the positive result must be a value the post-filter keeps (a `None` instance carrying the NewType identity), or skip setting `resultIncludesNoneSubtype` and return the NewType directly.

[verified]

packages/pyright-internal/src/tests/samples/newTypeIsLiteral1.py
```diff
@@ -0,0 +1,22 @@
...
+    if a is None:
+        reveal_type(a, expected_text="Apple")
```
rchiodo, line 23: 📍 packages/pyright-internal/src/tests/samples/newTypeIsLiteral1.py:12
This assertion is vacuous and hides the functional bug. Because `if a is None:` is unreachable (see typeGuards.ts:1131), pyright never evaluates `reveal_type(a, expected_text="Apple")`, so the `expected_text` check never fires — flipping the literal to `expected_text="WRONG_XYZ"` still passes with 0 errors. A `reveal_type` in dead code is not an assertion. Add negative coverage that fails when the branch is unreachable (e.g. a statement after the block that depends on the narrowed binding, or assert the `else` branch type) so a `Never` regression can't pass silently.

[verified]

packages/pyright-internal/src/analyzer/typeGuards.ts
```diff
@@ -1128,6 +1128,25 @@ function narrowTypeForIsNone(evaluator: TypeEvaluator, type: Type, isPositiveTes
...
+                if (isClass(baseClass)) {
+                    const baseInstance = ClassType.cloneAsInstance(baseClass);
+                    if (isNoneInstance(baseInstance)) {
+                        resultIncludesNoneSubtype = true;
+                        return isPositiveTest ? adjustedSubtype : undefined;
+                    }
+
+                    if (evaluator.assignType(baseInstance, evaluator.getNoneType())) {
+                        resultIncludesNoneSubtype = true;
+                        return isPositiveTest
```
rchiodo: 📍 packages/pyright-internal/src/analyzer/typeGuards.ts:1137
For a NewType over a None-compatible non-None base (e.g. `ObjNT = NewType("ObjNT", object)`), `if o is None:` narrows to bare `None`, dropping the `ObjNT` brand. This is inconsistent with the literal path (line 2773), which deliberately preserves the NewType identity. The behavior mirrors existing base-`None` narrowing so it may be defensible, but the two newly added paths answer "preserve NewType identity?" differently. Decide the intended behavior, add a test for this path (currently untested), and add a one-line comment explaining the divergence so a future reader doesn't mistake it for a bug.

[verified]

packages/pyright-internal/src/analyzer/typeGuards.ts
rchiodo, line 1167: 📍 packages/pyright-internal/src/analyzer/typeGuards.ts:1131
The NewType-base unwrap is copy-pasted into two functions (here and lines 2762-2769): identical `isClassInstance && isNewTypeClass && baseClasses.length > 0` → `baseClasses[0]` → `isClass()` guard → `cloneAsInstance`. AGENTS.md explicitly calls out this copy/paste-variant pattern. Extract a single `getNewTypeBaseInstance(subtype): ClassType | undefined` helper so both call sites collapse to one line and there's a single place to evolve depth/Any-base handling.

[verified]

packages/pyright-internal/src/analyzer/typeGuards.ts
```diff
@@ -2741,8 +2760,19 @@ function narrowTypeForLiteralComparison(
...
             if (isIsOperator || isNoneInstance(subtype)) {
-                const isSubtype = evaluator.assignType(subtype, literalType);
-                return isSubtype ? literalType : undefined;
+                let compareType: Type = subtype;
+                if (isClassInstance(subtype) && ClassType.isNewTypeClass(subtype) && subtype.shared.baseClasses.length > 0) {
+                    const baseClass = subtype.shared.baseClasses[0];
+                    if (isClass(baseClass)) {
+                        compareType = ClassType.cloneAsInstance(baseClass);
+                    }
+                }
```
rchiodo: 📍 packages/pyright-internal/src/analyzer/typeGuards.ts:2760
The unwrap is single-level only. `NewType`'s base may itself be a NewType (`Y = NewType("Y", X)` where `X = NewType("X", bool)` is legal and instantiable). The patch unwraps one level, so `compareType` stays a NewType and `assignType(X_instance, True)` still fails, leaving nested NewType comparisons marked unreachable. Either loop/recurse until the first non-NewType base (natural fit for the suggested shared helper) or explicitly scope nested NewType out with a comment.

[verified]

rchiodo:
GitHub cannot anchor PR review comments to unchanged lines in the diff. Falling back to a general PR comment for [packages/pyright-internal/src/analyzer/typeGuards.ts:L1162](https://github.com/microsoft/pyright/blob/a1fec8b4c974c014a93a0065fca5d68fe8a9c017/packages/pyright-internal/src/analyzer/typeGuards.ts#L1162).

📍 packages/pyright-internal/src/analyzer/typeGuards.ts:1162
`narrowTypeForIsEllipsis` is the structural twin of `narrowTypeForIsNone` (same `isXInstance`/`assignType` shape) but received no NewType handling. `X = NewType("X", EllipsisType); x is ...` would still be flagged unreachable — the same class of bug this PR fixes, left inconsistent in the adjacent context. Rare, but at minimum note why ellipsis was intentionally excluded, or handle it via the shared helper.

[verified]

rchiodo:
The literal-comparison path (bool/int cases) is fixed correctly, but the `is None` path does not actually fix its own scenario. For `NewType("Apple", NoneType)`, the positive `if a is None:` branch narrows to `Never` (the post-filter at lines 1152-1156 strips the NewType instance via `isNoneInstance`), so the added `reveal_type` assertion is dead code and `NewType8` passes vacuously. Please make the `is None` positive result survive the post-filter (return a None-typed value carrying the NewType identity, or skip the post-filter for this case) and add non-vacuous coverage that fails on a `Never` regression.

packages/pyright-internal/src/analyzer/typeGuards.ts
```diff
@@ -1128,6 +1157,12 @@ function narrowTypeForIsNone(evaluator: TypeEvaluator, type: Type, isPositiveTes
 
             const adjustedSubtype = useExpandedSubtype ? subtype : unexpandedSubtype;
 
+            // NewType instances can be identity-compared with their underlying base type.
+            if (isNewTypeWithSingletonBase(evaluator, adjustedSubtype, evaluator.getNoneType(), isNoneInstance)) {
+                resultIncludesNoneSubtype = true;
+                return isPositiveTest ? adjustedSubtype : undefined;
+            }
```
rchiodo, line 1167: Negative-branch regression (verified empirically). `isNewTypeWithSingletonBase` returns true when `isExactMatch(base) || assignType(base, target)` — i.e. "could the value be the singleton?". That's correct for the positive branch's `adjustedSubtype` return, but wrong for the negative branch's `undefined` (eliminate) return. For `ObjNT = NewType("ObjNT", object)`, the base unwraps to `object` and `assignType(object, None)` is `true`, so this block fires and `o is not None` returns `undefined` → `Never`, marking the true branch unreachable (reproduced: `reportUnreachable='error'` flags the `if o is not None:` body). In the pre-PR code this case falls through to the `assignType(subtype, None)` check and correctly keeps the type. Fix: gate the negative-branch elimination on `isExactMatch(base)` only, so non-singleton bases fall through to the existing exact-/potentially-None checks.

packages/pyright-internal/src/analyzer/typeGuards.ts
```diff
@@ -1193,6 +1231,11 @@ function narrowTypeForIsEllipsis(evaluator: TypeEvaluator, node: ExpressionNode,
                     ? unexpandedSubtype
                     : subtype;
 
+            if (isNewTypeWithSingletonBase(evaluator, adjustedSubtype, ellipsisType, isEllipsisInstance)) {
+                resultIncludesEllipsisSubtype = true;
```
rchiodo, line 1240: Same negative-branch defect in the ellipsis twin. For an `object`-based NewType, `assignType(object, ellipsisType)` is `true`, so `e is not ...` returns `undefined` → `Never` here too, whereas the existing fall-through at the `assignType(subtype, ellipsisType)` check would correctly keep the type. Apply the same fix as the `is None` path (single root cause).

packages/pyright-internal/src/analyzer/typeGuards.ts
```diff
@@ -1081,6 +1081,35 @@ function narrowTupleTypeForIsNone(evaluator: TypeEvaluator, type: Type, isPositi
...
+    if (!baseInstance || baseInstance === subtype) {
+        return false;
+    }
+
+    return isExactMatch(baseInstance) || evaluator.assignType(baseInstance, targetType);
+}
+
 // Handle type narrowing for expressions of the form "x is None" and "x is not None".
 function narrowTypeForIsNone(evaluator: TypeEvaluator, type: Type, isPositiveTest: boolean) {
     const expandedType = mapSubtypes(type, (subtype) => {
```
rchiodo, line 1112: The predicate name `isNewTypeWithSingletonBase` overstates what it checks: for an `object` base the unwrapped base is not a singleton, yet `assignType(object, None)` makes it return `true`. This name is what makes the negative-branch elimination look correct. Consider renaming (e.g. `newTypeBaseCouldMatch`) and documenting that a `true` result only justifies *keeping* a subtype on the positive branch, not *eliminating* it on the negative branch.

packages/pyright-internal/src/tests/samples/newTypeIsLiteral1.py
```diff
@@ -0,0 +1,45 @@
...
+
+    if bb is True:
+        reveal_type(bb, expected_text="Plantain")
+
+    if c is False:
+        reveal_type(c, expected_text="Cherry")
+
+    if d is ...:
+        reveal_type(d, expected_text="Dragonfruit")
+
```
rchiodo, line 39: Coverage gap that let the negative-branch regression pass. The sample exercises only positive `is` branches; there's no `is not None` / `is not ...` coverage and no `object`-based NewType. Add a NewType over a None-compatible non-singleton base (e.g. `ObjNT = NewType("ObjNT", object)`) and assert that the `else` of `if o is None:` (and the body of `if o is not None:`) stays reachable as `ObjNT` under `reportUnreachable='error'`, so this class of bug is caught.

rchiodo:
Core fix for the positive `is`/`is None` cases works and nested NewTypes are handled correctly (verified: `a is None` narrows to the NewType, not Never). The blocking problem is on the negative branch: `isNewTypeWithSingletonBase` answers "could the base be the singleton?" via `assignType(base, target)`, which is right for keeping a subtype on the positive branch but wrong for eliminating it on the negative branch. For a NewType over a non-singleton base that's merely compatible with None/ellipsis (e.g. `object`), this turns `is not None` / `is not ...` into Never. Please gate the negative-branch elimination on the exact-match case only, and add negative-branch + object-base test coverage.

rchiodo:
The core fix works and nested NewTypes are handled correctly. However, there is one real regression: `isNewTypeWithSingletonBase` uses `assignType(base, target)`, which is correct for keeping a subtype on the positive branch but wrong for eliminating it on the negative branch. For a NewType whose base is merely compatible with (not equal to) None/ellipsis (e.g. an `object`-based NewType), `is not None` / `is not ...` collapse to `Never` and mark reachable code unreachable. I reproduced this under `reportUnreachable='error'`. Restricting the negative-branch elimination to an exact singleton-base match, plus adding negative-branch/object-based test coverage, should resolve it. (The other first-pass notes about the primary branch collapsing to Never, a vacuous test, single-level unwrap, copy-paste duplication, and missing ellipsis handling did not reproduce and were dropped.)

rchiodo (approved):
Approved via Review Center.

rchiodo:
Thanks for the PR :)

KirtiRamchandani:
Yipee! :)

PR #11494, opened 2026-06-15
Fix #11450: [FR]: `struct.unpack` type inference

### LINKED ISSUE #11450: [FR]: `struct.unpack` type inference
Implement `struct.unpack` type inference instead of returning `Any`/`Unknown` to avoid having to use `cast()`. 
See similar PR here:
https://github.com/astral-sh/ruff/pull/23130
https://github.com/astral-sh/ruff/pull/22562

rchiodo:
## Description

Synthesizes a precise return type for `struct.unpack`, `struct.unpack_from`, and `struct.iter_unpack` when the format string is a literal. Instead of the declared `tuple[Any, ...]`, callers now get an exact tuple type (e.g. `tuple[int, int]`) derived from the format codes, so element types and arity are checked and propagated.

`struct.unpack` returns `tuple[Any, ...]` in typeshed, losing all element information even when the format string is a known literal. Pyright already applies function-result transforms for special-cased callables (e.g. `functools.total_ordering`) in `functionTransform.ts`, which is the natural hook for synthesizing a more specific return type without touching the wire protocol or typeshed.

## Implementation

In `applyFunctionTransform`, dispatch the `_struct.unpack` / `unpack_from` / `iter_unpack` callables to a new `applyStructUnpackTransform`:

-   Reads the first positional argument and only proceeds when it is a `str`/`bytes` **literal**.
-   `parseStructFormat` walks the format string, honoring the optional byte-order character, whitespace, repeat counts, and per-code element kinds (`int`/`float`/`bool`/`bytes`), with `x` producing no element and `s`/`p` collapsing their count into a single `bytes`. The native-only codes `n`/`N`/`P` are only treated as `int` in native mode (no prefix or `@`); under an explicit byte-order prefix they fall back to the declared type, matching the runtime `struct.error`.
-   Builds an exact tuple via `makeTupleObject`; `iter_unpack` wraps it in `Iterator[...]`.
-   Falls back to the declared type for non-literals, unknown codes, or counts above `maxStructUnpackElementCount` (256) to bound performance.

## Testing

Added `structUnpack1.py` with `assert_type` checks covering numeric codes, repeat counts, mixed kinds, `s`/`c`/`x`, bytes-literal formats, empty/byte-order-only formats, whitespace, `unpack_from`, `iter_unpack`, the zero repeat count (`"0i"` → `tuple[()]`), the element-count cap fallback (`"257i"` → `tuple[Any, ...]`), native-mode `n`/`N`/`P` vs. their invalid prefixed forms, and fallback for unknown codes and non-literal formats. Wired up via `StructUnpack1` in `typeEvaluator8.test.ts` (expects 0 errors).

## Addresses

Addresses https://github.com/microsoft/pylance-release/issues/11450

---

### DIFF
--- a/packages/pyright-internal/src/analyzer/functionTransform.ts
+++ b/packages/pyright-internal/src/analyzer/functionTransform.ts
@@ -250,8 +250,12 @@ function parseStructFormat(format: string): StructElementType[] | undefined {
     const elements: StructElementType[] = [];
     let index = 0;
 
-    // An optional leading byte-order/size/alignment character.
+    // An optional leading byte-order/size/alignment character. The 'n', 'N',
+    // and 'P' codes are only valid in native mode (no prefix or '@'); under an
+    // explicit byte-order prefix ('=', '<', '>', '!') they raise struct.error.
+    let isNativeMode = true;
     if (index < format.length && '@=<>!'.includes(format[index])) {
+        isNativeMode = format[index] === '@';
         index++;
     }
 
@@ -290,7 +294,7 @@ function parseStructFormat(format: string): StructElementType[] | undefined {
         const code = format[index];
         index++;
 
-        const elementKind = getStructElementType(code);
+        const elementKind = getStructElementType(code, isNativeMode);
         if (elementKind === undefined) {
             return undefined;
         }
@@ -318,7 +322,7 @@ function parseStructFormat(format: string): StructElementType[] | undefined {
     return elements;
 }
 
-function getStructElementType(code: string): StructElementType | 'pad' | undefined {
+function getStructElementType(code: string, isNativeMode: boolean): StructElementType | 'pad' | undefined {
     switch (code) {
         case 'x':
             return 'pad';
@@ -338,10 +342,15 @@ function getStructElementType(code: string): StructElementType | 'pad' | undefin
         case 'L':
         case 'q':
         case 'Q':
+            return 'int';
+
         case 'n':
         case 'N':
         case 'P':
-            return 'int';
+            // These codes are only available in native mode (no prefix or '@').
+            // Under an explicit byte-order prefix they are invalid, so fall back
+            // to the declared return type rather than synthesizing 'int'.
+            return isNativeMode ? 'int' : undefined;
 
         case '?':
             return 'bool';
--- a/packages/pyright-internal/src/analyzer/functionTransform.ts
+++ b/packages/pyright-internal/src/analyzer/functionTransform.ts
@@ -148,8 +148,12 @@ function applyTotalOrderingTransform(
 // Distinguishes between the `struct` functions whose return type can be
 // synthesized from a literal format string. `unpack` and `unpack_from`
 // return a tuple; `iter_unpack` returns an iterator of tuples.
-type StructUnpackKind = 'tuple' | 'iterator';
-
+//
+// Only the module-level `_struct.*` free functions are handled here. The
+// reused-format API (`struct.Struct(fmt).unpack()` / `.unpack_from()` /
+// `.iter_unpack()`) is out of scope: it would require threading the
+// constructor's literal format through to the method calls, so those still
+// infer the declared `tuple[Any, ...]`.
 function getStructUnpackKind(fullName: string): StructUnpackKind | undefined {
     switch (fullName) {
         case '_struct.unpack':
@@ -163,9 +167,6 @@ function getStructUnpackKind(fullName: string): StructUnpackKind | undefined {
             return undefined;
     }
 }
-
-// The synthesized element type produced by a single struct format code.
-type StructElementType = 'int' | 'float' | 'bool' | 'bytes';
 
 // To avoid performance issues with very large repeat counts, fall back to
 // the declared return type (a homogeneous `Any` tuple) above this many elements.
@@ -364,3 +365,10 @@ function getStructElementType(code: string, isNativeMode: boolean): StructElemen
             return undefined;
     }
 }
+
+// The kind of return type synthesized for a dispatched `struct` function:
+// a tuple (`unpack`/`unpack_from`) or an iterator of tuples (`iter_unpack`).
+type StructUnpackKind = 'tuple' | 'iterator';
+
+// The synthesized element type produced by a single struct format code.
+type StructElementType = 'int' | 'float' | 'bool' | 'bytes';
--- a/packages/pyright-internal/src/analyzer/functionTransform.ts
+++ b/packages/pyright-internal/src/analyzer/functionTransform.ts
@@ -208,8 +208,7 @@ function applyStructUnpackTransform(
     const getElementType = (elementKind: StructElementType): Type | undefined => {
         let elementType = elementTypeCache.get(elementKind);
         if (!elementType) {
-            const builtInName = elementKind === 'bytes' ? 'bytes' : elementKind;
-            const builtInType = evaluator.getBuiltInObject(errorNode, builtInName);
+            const builtInType = evaluator.getBuiltInObject(errorNode, elementKind);
             if (!isClassInstance(builtInType)) {
                 return undefined;
             }
@@ -273,8 +272,11 @@ function parseStructFormat(format: string): StructElementType[] | undefined {
                 count = count * 10 + (format.charCodeAt(index) - 0x30);
                 index++;
 
-                // Guard against pathologically large counts.
-                if (count > maxStructUnpackElementCount) {
+                // Guard against pathologically large counts. This bounds the
+                // accumulator itself; the produced element count is bounded
+                // separately below so byte-length codes ('s'/'p') aren't
+                // rejected for large counts.
+                if (count > Number.MAX_SAFE_INTEGER) {
                     return undefined;
                 }
             }
@@ -295,7 +297,7 @@ function parseStructFormat(format: string): StructElementType[] | undefined {
 
         if (code === 's' || code === 'p') {
             // For 's' and 'p', the count is the byte length of a single value,
-            // so it always produces exactly one element.
+            // so it always produces exactly one element regardless of count.
             elements.push('bytes');
         } else if (elementKind === 'pad') {
             // Pad bytes ('x') produce no elements.
@@ -303,6 +305,12 @@ function parseStructFormat(format: string): StructElementType[] | undefined {
             const repeat = count < 0 ? 1 : count;
             for (let i = 0; i < repeat; i++) {
                 elements.push(elementKind);
+
+                // Bound the produced element count to avoid performance issues
+                // with very large repeat counts.
+                if (elements.length > maxStructUnpackElementCount) {
+                    return undefined;
+                }
             }
         }
     }
--- a/packages/pyright-internal/src/analyzer/functionTransform.ts
+++ b/packages/pyright-internal/src/analyzer/functionTransform.ts
@@ -12,6 +12,7 @@ import { DiagnosticRule } from '../common/diagnosticRules';
 import { LocMessage } from '../localization/localize';
 import { ExpressionNode, ParamCategory } from '../parser/parseNodes';
 import { Symbol, SymbolFlags } from './symbol';
+import { makeTupleObject } from './tuples';
 import { Arg, FunctionResult, TypeEvaluator } from './typeEvaluatorTypes';
 import {
     ClassType,
@@ -22,6 +23,7 @@ import {
     isFunction,
     isInstantiableClass,
     OverloadedType,
+    TupleTypeArg,
     Type,
 } from './types';
 import { ClassMember, lookUpObjectMember, MemberAccessFlags, synthesizeTypeVarForSelfCls } from './typeUtils';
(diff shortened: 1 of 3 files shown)


packages/pyright-internal/src/analyzer/functionTransform.ts
```diff
@@ -137,3 +144,206 @@ function applyTotalOrderingTransform(
...
+
+            // A count must be followed by a format code.
+            if (index >= format.length) {
+                return undefined;
+            }
+        }
+
+        const code = format[index];
+        index++;
+
```
rchiodo, line 297: 📍 packages/pyright-internal/src/analyzer/functionTransform.ts:285
This `if (count > maxStructUnpackElementCount) return undefined` guard fires during digit accumulation, treating every count as an element multiplier. For `s`/`p` the count is a *byte length* that collapses to exactly one `bytes`, so common fixed-width formats like `"1024s"`, `"512s"`, or `"300p"` are wrongly rejected and fall back to `tuple[Any, ...]` — defeating the feature for a case the PR claims to handle. Bound `elements.length` after deciding a code's actual element contribution (or rely on the caller's length check with an in-loop bound) instead of capping the raw `count` during digit parsing. The test sample only covers `"3s"`, so this is untested.

packages/pyright-internal/src/analyzer/functionTransform.ts
```diff
@@ -137,3 +144,206 @@ function applyTotalOrderingTransform(
...
+        return result;
+    }
+
+    const formatType = evaluator.getTypeOfArg(argList[0], /* inferenceContext */ undefined).type;
+    if (!isClassInstance(formatType)) {
+        return result;
+    }
+
+    // The format must be a `str` or `bytes` literal. Both store their value
+    // as a string in `literalValue`.
```
rchiodo, line 193: 📍 packages/pyright-internal/src/analyzer/functionTransform.ts:191
`const builtInName = elementKind === 'bytes' ? 'bytes' : elementKind;` is a no-op — all four kinds (`'int'|'float'|'bool'|'bytes'`) are already valid `getBuiltInObject` names, so the ternary returns the same value either way. It implies a mapping that doesn't exist and will mislead future readers; replace with `const builtInName = elementKind;`.

rchiodo:
Solid, well-tested transform. One real (non-blocking) gap worth addressing: the per-code count guard rejects `s`/`p` formats whose byte-length count exceeds 256 (e.g. `"1024s"`), even though those collapse to a single `bytes`. Otherwise good to merge.

packages/pyright-internal/src/analyzer/functionTransform.ts
rchiodo: Fixed in ec79d696b. The performance guard now bounds the *produced* element count inside the repeat loop rather than capping the raw count during digit parsing. Byte-length codes ('s'/'p') push exactly one `bytes` element regardless of count, so `1024s`, `512s`, and `300p` now correctly infer `tuple[bytes]` instead of falling back. Added test coverage for these cases in structUnpack1.py.

packages/pyright-internal/src/analyzer/functionTransform.ts
rchiodo: Fixed in ec79d696b. Replaced the no-op ternary with a direct `getBuiltInObject(errorNode, elementKind)` call, since all four element kinds are already valid builtin object names.

packages/pyright-internal/src/tests/samples/structUnpack1.py
```diff
@@ -0,0 +1,63 @@
...
+assert_type(struct.unpack("@", buffer), tuple[()])
+
+
+# An unrecognized format code falls back to the declared return type.
+assert_type(struct.unpack("<2z", buffer), tuple[Any, ...])
+
+
+# A non-literal format string falls back to the declared return type.
+def func1(fmt: str) -> None:
+    assert_type(struct.unpack(fmt, buffer), tuple[Any, ...])
```
heejaechang, line 99: 📍 packages/pyright-internal/src/tests/samples/structUnpack1.py:1

The sample doesn't directly assert the element-count cap fallback (e.g. `"257i"`, just over `maxStructUnpackElementCount`, should yield `tuple[Any, ...]`) nor the `0`-count edge (`"0i"` → `tuple[()]`). Both are correct by inspection, but explicit `assert_type` cases would lock in these boundaries cheaply.

packages/pyright-internal/src/analyzer/functionTransform.ts
```diff
@@ -137,3 +144,214 @@ function applyTotalOrderingTransform(
...
+        case 'x':
+            return 'pad';
+
+        case 'c':
+        case 's':
+        case 'p':
+            return 'bytes';
+
+        case 'b':
+        case 'B':
```
heejaechang, line 337: 📍 packages/pyright-internal/src/analyzer/functionTransform.ts:320

Native-only codes are over-accepted: `struct.unpack("<n", ...)`, `"<N"`, and `"<P"` raise `struct.error: bad char in struct format` (native-only codes are invalid under an explicit byte-order prefix), yet `getStructElementType` maps `n`/`N`/`P` to `int` unconditionally, so the PR synthesizes `tuple[int]` for a call that fails at runtime. This doesn't regress vs the declared fallback, but a tighter version would only treat `n`/`N`/`P` as `int` when the effective mode is native (`@` or no prefix). Worth a follow-up.

packages/pyright-internal/src/tests/samples/structUnpack1.py
```diff
@@ -0,0 +1,63 @@
+# This sample tests the synthesized return type of `struct.unpack`,
+# `struct.unpack_from` and `struct.iter_unpack` when the format string
+# is a literal.
+
+import struct
```
heejaechang, line 5: 📍 packages/pyright-internal/src/tests/samples/structUnpack1.py:1

Issue-linkage convention: the PR description uses `Fixes https://github.com/microsoft/pylance-release/issues/11450`, but repo convention is `Addresses` for `microsoft/pylance-release` issues (`Fixes` is reserved for `microsoft/pyrx` issues). Please update the description.

heejaechang:
Main blocker: the struct.unpack transform was added only to the pyright-internal (sync) copy of functionTransform.ts. Pylance's async path uses a separate pylance-internal copy that wasn't updated, so inference diverges between sync and async modes. Please mirror the change (and add Pylance-harness coverage for both modes) per the repo's sync↔async parity policy.

packages/pyright-internal/src/tests/samples/structUnpack1.py
rchiodo: Added explicit `assert_type` cases for both boundaries in `structUnpack1.py`:

- `assert_type(struct.unpack("257i", buffer), tuple[Any, ...])` — just over `maxStructUnpackElementCount`, exercises the cap fallback.
- `assert_type(struct.unpack("0i", buffer), tuple[()])` — the zero repeat-count edge.

Cheap to lock in, thanks for the nudge.

packages/pyright-internal/src/analyzer/functionTransform.ts
rchiodo: Good catch — tightened this. `parseStructFormat` now tracks the effective mode from the optional prefix and passes it to `getStructElementType`, which only maps `n`/`N`/`P` to `int` in native mode (no prefix or `@`). Under an explicit byte-order prefix they return `undefined`, so we fall back to the declared `tuple[Any, ...]` instead of synthesizing a tuple for a call that fails at runtime with `struct.error`.

Added coverage in `structUnpack1.py`: `nNP` / `@nNP` → `tuple[int, int, int]`, and `<n` / `=N` / `>P` → `tuple[Any, ...]`.

packages/pyright-internal/src/tests/samples/structUnpack1.py
rchiodo: Updated the PR description to use `Addresses https://github.com/microsoft/pylance-release/issues/11450` instead of `Fixes`, per the repo convention.

rchiodo:
Thanks for flagging the sync↔async parity concern. I've broken that out into a separate follow-up issue so it can be tracked and reviewed independently — this PR stays focused on the pyright change.

heejaechang:
Core change looks solid, but it needs the async counterpart updated for parity before merge: the new struct.unpack transform lives only in the pyright (sync) functionTransform.ts, while packages/pylance-internal/src/analyzer/functionTransform.ts still handles only functools.total_ordering. Please mirror the transform (and add a Pylance-harness regression test covering both modes). Two smaller follow-ups noted inline.

heejaechang:
Solid, well-tested fix on the Pyright side. The one blocker is sync↔async parity: the async copy in packages/pylance-internal/src/analyzer/functionTransform.ts still only transforms functools.total_ordering, so struct.unpack results differ when enableAsyncProgram=true. Please mirror the transform there and add Pylance-harness coverage for both modes.

packages/pyright-internal/src/analyzer/functionTransform.ts
```diff
@@ -137,3 +144,223 @@ function applyTotalOrderingTransform(
...
+type StructUnpackKind = 'tuple' | 'iterator';
+
+function getStructUnpackKind(fullName: string): StructUnpackKind | undefined {
+    switch (fullName) {
+        case '_struct.unpack':
+        case '_struct.unpack_from':
+            return 'tuple';
+
+        case '_struct.iter_unpack':
+            return 'iterator';
```
heejaechang, line 164: Review-rule violation (`ts-named-types-at-bottom`): `StructUnpackKind` and `StructElementType` are declared inline, interleaved with the function implementations. Extraction is correct (both appear in multiple signatures, so the tiny-one-off carve-out doesn't apply), but placement is wrong — move both `type` declarations to the bottom of the file, after `getStructElementType`.

heejaechang:
Solid feature with thorough tests. Two blocking items before merge: (1) port the transform into the async functionTransform copy in pylance-internal so sync/async modes agree, and (2) move the two type aliases to the bottom of the file per the repo convention. The native-mode handling for n/N/P is already correct and well-tested.

heejaechang:
The struct.unpack synthesis itself looks solid and well-tested, but it was only applied to the pyright/sync copy of applyFunctionTransform. The async copy (packages/pylance-internal/src/analyzer/functionTransform.ts) still handles only functools.total_ordering, which violates the sync↔async parity requirement and yields divergent results across modes. Please port the transform there and add a Pylance-harness regression test covering both modes.

heejaechang:
Main blocker: sync↔async parity. The new struct.unpack transform exists only in packages/pyright-internal/src/analyzer/functionTransform.ts but not in the pylance-internal async copy, so results diverge by mode. Please port the transform to the async copy and add a Pylance-harness regression test covering both modes. The implementation itself (literal parsing, native-mode n/N/P handling, element-count cap) looks solid and well-tested.

heejaechang:

packages/pyright-internal/src/analyzer/functionTransform.ts
heejaechang, line 164: Only the three module-level `_struct.*` free functions are dispatched, so the reused-format API `struct.Struct(fmt).unpack()` / `.unpack_from()` / `.iter_unpack()` still infers `tuple[Any, ...]`. Threading the constructor's literal format through is reasonably deferred, but the asymmetry may surprise users — consider a brief "out of scope" note or a follow-up issue.

packages/pyright-internal/src/tests/samples/structUnpack1.py
```diff
@@ -0,0 +1,81 @@
...
+assert_type(struct.unpack(">P", buffer), tuple[Any, ...])
+
+
+# An unrecognized format code falls back to the declared return type.
+assert_type(struct.unpack("<2z", buffer), tuple[Any, ...])
+
+
+# A non-literal format string falls back to the declared return type.
+def func1(fmt: str) -> None:
+    assert_type(struct.unpack(fmt, buffer), tuple[Any, ...])
```
heejaechang, line 99: Coverage is representative but not exhaustive: no direct assertion for `d`/`f`/`e`, `q`/`Q`, `l`/`L`, `h`/`H`, `b`/`B`, no-count `s`/`p` (`"s"` → `tuple[bytes]`), or `"0c"` → `tuple[()]`. Since `getStructElementType` is a flat switch, one mixed-code line (e.g. `"bhlqefd"`) would cheaply lock the full mapping and guard against a future typo (e.g. `e`→int).

heejaechang:
Looks solid and well-tested. Approving. Two minor optional notes below; nothing blocking.

heejaechang (approved):
Approved via Review Center.

packages/pyright-internal/src/tests/samples/structUnpack1.py
rchiodo: Added coverage in `structUnpack1.py`: a mixed line `"bhlqefd"` -> `tuple[int, int, int, int, float, float, float]` that locks the full numeric mapping in one place (guards against e.g. an `e`->int typo), `"BHILQ"` for the uppercase codes, no-count `"s"`/`"p"` -> `tuple[bytes]`, and `"0c"` -> `tuple[()]`.

packages/pyright-internal/src/analyzer/functionTransform.ts
rchiodo: Fixed — moved both `StructUnpackKind` and `StructElementType` `type` declarations to the bottom of the file (after `getStructElementType`), per `ts-named-types-at-bottom`. The descriptive text now lives on the functions; the types carry short declarations at the bottom.

rchiodo: Added an explicit out-of-scope note on `getStructUnpackKind`: only the module-level `_struct.*` free functions are dispatched, and the reused-format API (`struct.Struct(fmt).unpack()` / `.unpack_from()` / `.iter_unpack()`) still infers `tuple[Any, ...]` because threading the constructor's literal format through to the method calls is deferred.

PR #11495, opened 2026-06-15
Report duplicate KW_ONLY separators in dataclasses

rchiodo:
## Description

Pyright recognized the dataclass `KW_ONLY` sentinel only when it was assigned to a field named `_`. At runtime, the `dataclasses` module treats **any** field annotated as `KW_ONLY` as the keyword-only separator regardless of name. That core name-guard removal (the original #11451 fix) has since landed on `main` via #11486, so this PR is rebased on top of it and contributes the remaining test coverage plus a related correctness diagnostic.

## Changes

After rebasing onto current `main` (which already contains the #11486 name-guard removal), this PR adds:

1. **New diagnostic — duplicate `KW_ONLY` separators.** CPython raises `TypeError: '<name>' is KW_ONLY, but KW_ONLY has already been specified` when more than one `KW_ONLY` separator appears in a single dataclass. Pyright previously accepted this silently. `dataClasses.ts` now checks `sawKeywordOnlySeparator` before setting it and reports a `reportGeneralTypeIssues` diagnostic on the second separator.
   - New NLS key `Diagnostic.dataClassDuplicateKwOnly` (`localize.ts` + `package.nls.en-us.json`). The key is inserted as its own entry without modifying the adjacent `dataClassFieldInheritedDefault` line, so there is no Prettier/JSON whitespace change.
2. **Test coverage** in `dataclassKwOnly1.py`:
   - `DC5` — `KW_ONLY` separator bound to a non-`_` name (`__`), exercising the #11486 behavior (positive calls plus a keyword-only enforcement error).
   - `DC6` — two `KW_ONLY` separators, expecting the new duplicate diagnostic.
   - Expected error count updated `3 → 5` in `typeEvaluator4.test.ts`.

## Testing

- `npx jest typeEvaluator4 -t "DataClassKwOnly1"` passes.
- `prettier -c` and `eslint` pass on the changed files.

## Addresses

Relates to https://github.com/microsoft/pylance-release/issues/11451 (the core recognition fix landed separately in #11486).

### DIFF
--- a/packages/pyright-internal/src/analyzer/dataClasses.ts
+++ b/packages/pyright-internal/src/analyzer/dataClasses.ts
@@ -381,6 +381,18 @@ export function synthesizeDataClassMethods(
                         const annotatedType = variableTypeEvaluator();
 
                         if (isClassInstance(annotatedType) && ClassType.isBuiltIn(annotatedType, 'KW_ONLY')) {
+                            // CPython raises a TypeError if more than one KW_ONLY
+                            // separator appears within a single dataclass.
+                            if (sawKeywordOnlySeparator) {
+                                evaluator.addDiagnostic(
+                                    DiagnosticRule.reportGeneralTypeIssues,
+                                    LocMessage.dataClassDuplicateKwOnly().format({
+                                        name: variableNameNode.d.value,
+                                    }),
+                                    variableNameNode
+                                );
+                            }
+
                             sawKeywordOnlySeparator = true;
                             variableNameNode = undefined;
                             typeAnnotationNode = undefined;
--- a/packages/pyright-internal/src/localization/package.nls.en-us.json
+++ b/packages/pyright-internal/src/localization/package.nls.en-us.json
@@ -233,6 +233,7 @@
         "dataClassBaseClassNotFrozen": "A frozen class cannot inherit from a class that is not frozen",
         "dataClassConverterFunction": "Argument of type \"{argType}\" is not a valid converter for field \"{fieldName}\" of type \"{fieldType}\"",
         "dataClassConverterOverloads": "No overloads of \"{funcName}\" are valid converters for field \"{fieldName}\" of type \"{fieldType}\"",
+        "dataClassDuplicateKwOnly": "\"{name}\" is a duplicate KW_ONLY separator; only one is allowed within a dataclass",
         "dataClassFieldInheritedDefault": "\"{fieldName}\" overrides a field of the same name but is missing a default value",
         "dataClassFieldWithDefault": "Fields without default values cannot appear after fields with default values",
         "dataClassFieldWithPrivateName": "Dataclass field cannot use private name",
--- a/packages/pyright-internal/src/tests/samples/dataclassKwOnly1.py
+++ b/packages/pyright-internal/src/tests/samples/dataclassKwOnly1.py
@@ -56,3 +56,35 @@ class DC4(DC3):
 
 DC4("", 0.2, b=3)
 DC4(a="", b=3, c=0.2)
+
+
+# This tests that the KW_ONLY separator is recognized even when it
+# is assigned to a name other than "_".
+@dataclass
+class DC5:
+    a: str
+    __: KW_ONLY
+    b: int = 0
+
+
+DC5("hi")
+DC5(a="hi", b=1)
+DC5("hi", b=1)
+
+# This should generate an error because "b" is keyword-only.
+DC5("hi", 1)
+
+
+# This tests that a duplicate KW_ONLY separator is flagged. CPython raises a
+# TypeError at runtime if more than one KW_ONLY separator appears.
+@dataclass
+class DC6:
+    a: str
+    _: KW_ONLY
+    b: int = 0
+    # This should generate an error because only one KW_ONLY separator is allowed.
+    __: KW_ONLY
+    c: int = 0
+
+
+DC6("hi", b=1, c=2)
--- a/packages/pyright-internal/src/localization/localize.ts
+++ b/packages/pyright-internal/src/localization/localize.ts
@@ -375,6 +375,8 @@ export namespace Localizer {
             new ParameterizedString<{ funcName: string; fieldType: string; fieldName: string }>(
                 getRawString('Diagnostic.dataClassConverterOverloads')
             );
+        export const dataClassDuplicateKwOnly = () =>
+            new ParameterizedString<{ name: string }>(getRawString('Diagnostic.dataClassDuplicateKwOnly'));
         export const dataClassFieldInheritedDefault = () =>
             new ParameterizedString<{ fieldName: string }>(getRawString('Diagnostic.dataClassFieldInheritedDefault'));
         export const dataClassFieldWithDefault = () => getRawString('Diagnostic.dataClassFieldWithDefault');
--- a/packages/pyright-internal/src/tests/typeEvaluator4.test.ts
+++ b/packages/pyright-internal/src/tests/typeEvaluator4.test.ts
@@ -426,7 +426,7 @@ test('DataClassKwOnly1', () => {
     configOptions.defaultPythonVersion = pythonVersion3_10;
     const analysisResults = TestUtils.typeAnalyzeSampleFiles(['dataclassKwOnly1.py'], configOptions);
 
-    TestUtils.validateResults(analysisResults, 3);
+    TestUtils.validateResults(analysisResults, 5);
 });
 
 test('DataClassKwOnly2', () => {


packages/pyright-internal/src/tests/samples/dataclassKwOnly1.py
```diff
@@ -56,3 +56,20 @@ class DC4(DC3):
...
+    __: KW_ONLY
+    b: int = 0
+
+
+DC5("hi")
+DC5(a="hi", b=1)
+DC5("hi", b=1)
+
+# This should generate an error because "b" is keyword-only.
+DC5("hi", 1)
```
rchiodo, line 75: 📍 packages/pyright-internal/src/tests/samples/dataclassKwOnly1.py:60

**Missing Pylance-harness coverage for the path that's still broken.** Per `pylance-pyright-split.instructions.md`, a `pylance-release` issue should get a Pylance-harness regression first. Because coverage exists only in the pyright-internal sample, the still-broken Pylance async path has zero coverage — which is exactly why the parity gap above went unnoticed. Add a `packages/pylance-internal/src/tests/` regression exercising the non-`_` separator under both `enableAsyncProgram=false` and `true`. The Advocate argued pyright-internal placement is correct for standalone #11495; that holds only if this never lands in the monorepo, which it does.

[verified]

packages/pyright-internal/src/analyzer/dataClasses.ts
```diff
@@ -375,8 +375,10 @@ export function synthesizeDataClassMethods(
...
 
-                    // Is this a KW_ONLY separator introduced in Python 3.10?
-                    if (!isNamedTuple && statement.d.valueExpr.d.value === '_') {
+                    // Is this a KW_ONLY separator introduced in Python 3.10? The
+                    // separator is conventionally named "_", but the runtime recognizes
+                    // any field annotated as KW_ONLY regardless of the name it's assigned to.
+                    if (!isNamedTuple) {
                         const annotatedType = variableTypeEvaluator();
 
                         if (isClassInstance(annotatedType) && ClassType.isBuiltIn(annotatedType, 'KW_ONLY')) {
```
rchiodo, line 383: 📍 packages/pyright-internal/src/analyzer/dataClasses.ts:380

[unverified] **New false-negative: duplicate `KW_ONLY` separators are silently accepted.** The Skeptic verified via A/B probe that two differently-named separators (e.g. `_: KW_ONLY` and `__: KW_ONLY`) now produce zero diagnostics, whereas CPython raises `TypeError: ... KW_ONLY has already been specified` at class creation. This shape was inexpressible before (both needed the name `_`), so the change unlocks it without a guard. `sawKeywordOnlySeparator` is only ever set, never re-checked. Low severity (missing diagnostic for invalid code, not a false positive), but worth a follow-up check + test.

[verified]

rchiodo:
Core fix is correct, but it landed only in the pyright-internal copy. Pylance keeps a parallel async implementation that still has the old `=== '_'` name guard, so the bug persists in async mode. Mirror the guard removal there and add Pylance-harness coverage for the non-`_` separator under both sync and async modes.

rchiodo:
The core fix is correct, but it was only applied to the pyright-internal copy of dataClasses.ts. Pylance's async copy (packages/pylance-internal/src/analyzer/dataClasses.ts) still has the old `=== '_'` guard, so the bug persists in async mode and the sync/async paths diverge. Please mirror the guard removal there and add Pylance-harness coverage for the non-`_` separator.

packages/pyright-internal/src/analyzer/dataClasses.ts
rchiodo: Fixed in 4792ddca1. Verified against CPython 3.11: declaring two KW_ONLY separators (e.g. `_: KW_ONLY` and `__: KW_ONLY`) raises `TypeError: '__' is KW_ONLY, but KW_ONLY has already been specified`. `dataClasses.ts` now checks `sawKeywordOnlySeparator` before setting it and emits a new `dataClassDuplicateKwOnly` diagnostic on the second separator. Added a `DC6` regression case to `dataclassKwOnly1.py` and bumped the expected error count (4 → 5).

packages/pyright-internal/src/tests/samples/dataclassKwOnly1.py
rchiodo: Not taking this in this PR. This PR targets the standalone `microsoft/pyright` repo, which has no `packages/pylance-internal` (only `pyright`, `pyright-internal`, and `vscode-pyright`), so a Pylance-harness regression cannot be added here. The `pylance-pyright-split` guidance lives in the `microsoft/pyrx` monorepo; when this change is vendored into pyrx via the pyright subrepo, the Pylance-side regression (covering both `enableAsyncProgram=false`/`true`) belongs there. For this repo, the `pyright-internal` sample (`dataclassKwOnly1.py`) is the correct placement, and it now also covers the non-`_` separator (DC5) plus the duplicate-separator case (DC6). The async-path parity concern is likewise a Pylance product path that does not exist in pyright-internal.

packages/pyright-internal/src/localization/package.nls.en-us.json
```diff
@@ -233,7 +233,8 @@
         "dataClassBaseClassNotFrozen": "A frozen class cannot inherit from a class that is not frozen",
         "dataClassConverterFunction": "Argument of type \"{argType}\" is not a valid converter for field \"{fieldName}\" of type \"{fieldType}\"",
         "dataClassConverterOverloads": "No overloads of \"{funcName}\" are valid converters for field \"{fieldName}\" of type \"{fieldType}\"",
-        "dataClassFieldInheritedDefault": "\"{fieldName}\" overrides a field of the same name but is missing a default value",
+        "dataClassDuplicateKwOnly": "\"{name}\" is a duplicate KW_ONLY separator; only one is allowed within a dataclass",
+        "dataClassFieldInheritedDefault":"\"{fieldName}\" overrides a field of the same name but is missing a default value",
```
heejaechang: 📍 packages/pyright-internal/src/localization/package.nls.en-us.json:237
Prettier/JSON regression on the unrelated `dataClassFieldInheritedDefault` line: the space after the colon was dropped (`"dataClassFieldInheritedDefault":"...`) while every other entry uses `": "`. This will fail `npm run check`. Restore the space and avoid touching this pre-existing line.

[verified]

packages/pyright-internal/src/analyzer/dataClasses.ts
```diff
@@ -375,11 +375,25 @@ export function synthesizeDataClassMethods(
...
+                            if (sawKeywordOnlySeparator) {
+                                evaluator.addDiagnostic(
+                                    DiagnosticRule.reportGeneralTypeIssues,
+                                    LocMessage.dataClassDuplicateKwOnly().format({
+                                        name: variableNameNode.d.value,
+                                    }),
+                                    variableNameNode
+                                );
+                            }
+
```
heejaechang, line 395: 📍 packages/pyright-internal/src/analyzer/dataClasses.ts:382
This duplicate-`KW_ONLY` diagnostic is behaviorally correct (CPython raises the matching `TypeError`, and `variableNameNode` is provably non-null here), but it is a separate feature from #11451 — undisclosed in the title/description and added to the upstream-owned `pyright` subrepo, which should keep diffs surgical. Either split it into its own PR with an accurate description, or update this PR's description to document the new diagnostic, the new NLS key, the `DC6` test, and the corrected error count.

[verified]

packages/pyright-internal/src/analyzer/dataClasses.ts
```diff
@@ -375,11 +375,25 @@ export function synthesizeDataClassMethods(
                             allowClassVar: !isNamedTuple,
                         });
 
-                    // Is this a KW_ONLY separator introduced in Python 3.10?
-                    if (!isNamedTuple && statement.d.valueExpr.d.value === '_') {
+                    // Is this a KW_ONLY separator introduced in Python 3.10? The
+                    // separator is conventionally named "_", but the runtime recognizes
```
heejaechang: 📍 packages/pyright-internal/src/analyzer/dataClasses.ts:378
Stale diff base (verified): the core name-guard removal already exists in the pyright-internal working tree, so this hunk's `-` context line no longer exists and the patch will conflict on apply/rebase. Rebase onto current main; the net new contribution over the tree is essentially just the duplicate diagnostic, which makes the PR title's "#11451" framing misleading.

[verified]

heejaechang:
Main blocker: the fix only touches pyright-internal, but the Pylance async counterpart (packages/pylance-internal/src/analyzer/dataClasses.ts) still gates KW_ONLY on the field name being '_', so the bug still reproduces in Pylance's async path. Please port the name-agnostic recognition there with a Pylance-harness regression test. Also note the diff appears rebased on a stale base (the core fix already exists in the pyright-internal tree) and quietly adds an undisclosed duplicate-KW_ONLY diagnostic to the upstream subrepo.

packages/pyright-internal/src/localization/package.nls.en-us.json
rchiodo: Good catch. The regression came from editing the existing `dataClassFieldInheritedDefault` line to squeeze the new key in front of it. The incoming revision leaves that line untouched and adds `dataClassDuplicateKwOnly` as its own entry with the standard `": "` spacing, so `prettier -c` passes on the file again.

packages/pyright-internal/src/analyzer/dataClasses.ts
rchiodo: Agreed. I've updated the PR title and description to document the duplicate-`KW_ONLY` diagnostic, the new `dataClassDuplicateKwOnly` NLS key, the `DC6` regression case, and the error-count bump. After rebasing onto `main` (where the #11451 name-guard fix already landed via #11486), this diagnostic is the PR's net-new contribution, which the description now makes explicit rather than framing the whole PR as the #11451 fix.

packages/pyright-internal/src/analyzer/dataClasses.ts
rchiodo: Rebasing onto current `main`. Since the name-guard removal already merged via #11486, I'm dropping that hunk entirely — the revised diff is just the duplicate-separator diagnostic plus the `DC5` (non-`_` separator name) and `DC6` (duplicate separator) regression cases, so there's no longer a conflicting context line.

rchiodo:
Thanks for the review @heejaechang. Summary of how this revision addresses the feedback:

- **Stale base** — Rebased onto current `main`. The `KW_ONLY` name-guard removal already merged via #11486, so that hunk is dropped; the remaining diff is just the new duplicate-separator diagnostic plus the `DC5`/`DC6` regression cases.
- **Prettier/JSON regression** — The new `dataClassDuplicateKwOnly` key is added as its own entry without touching the `dataClassFieldInheritedDefault` line, so the dropped-space change is gone and `prettier -c` passes.
- **Undisclosed feature** — Title and description now document the duplicate-`KW_ONLY` diagnostic, the new NLS key, the `DC6` test, and the `3 → 5` error-count change.

- **Pylance async path** — This PR targets the standalone `microsoft/pyright` repo, which has no `packages/pylance-internal` (only `pyright`, `pyright-internal`, and `vscode-pyright`), so the async copy and its harness can't be modified here. The `pylance-pyright-split` guidance lives in the `pyrx` monorepo; the async-side guard removal and a regression covering both `enableAsyncProgram=false`/`true` belong there when this change is vendored in via the pyright subrepo. For this repo, `dataclassKwOnly1.py` (DC5 non-`_` separator + DC6 duplicate separator) is the correct placement.

heejaechang:
The core idea is correct, but the fix doesn't reach Pylance's async program path, and there's an unrelated formatting regression plus undisclosed scope (a new duplicate-KW_ONLY diagnostic in the upstream-owned pyright subrepo). Please port the fix to packages/pylance-internal, restore the dropped space in the NLS file, and update the description/scope before merge.

heejaechang:
The pyright-internal change looks correct, but the equivalent fix is missing from Pylance's forked async analyzer (packages/pylance-internal/src/analyzer/dataClasses.ts), where the bug still reproduces. Please port the name-agnostic KW_ONLY recognition and duplicate detection there with a Pylance-harness regression test. The other candidate notes were false positives or already addressed by the PR description.

heejaechang (approved):
Approved via Review Center.

PR #11496, opened 2026-06-15
Fix #11463: Incorrect `is` behaviour with NewType

### LINKED ISSUE #11463: Incorrect `is` behaviour with NewType

rchiodo:
## Description

Fixes incorrect `is` narrowing for `NewType` instances. At runtime a `NewType` value is simply an instance of its base type, but Pyright treated `NewType` instances as never overlapping with `None` or `bool` literals. This caused `x is None` / `x is True` / `x is False` branches to be wrongly marked unreachable (and narrowing to be skipped) when `x` was a `NewType` over `NoneType`, `bool`, etc.

The reported symptom was that comparing a `NewType` instance against `None`/bool literals never narrowed and the guarded branch was flagged unreachable. Tracing the `is` narrowing logic in `typeGuards.ts` showed that both the `None` check (`narrowTypeForIsNone`) and the literal-comparison check compared the *declared* `NewType` class against the target, which never overlaps. The fix is to test the underlying base type instead, matching runtime behavior.

## Implementation

This branch was reconciled with the fix that landed independently on `main`. The narrowing uses the on-disk helper `getInnermostNewTypeBaseInstance` in `typeGuards.ts`, which recursively unwraps `NewType` classes down to the innermost base type instance (or returns `undefined` if the type isn't a `NewType`):

- `narrowTypeForIsNone`: uses the unwrapped innermost base instance when testing for overlap with `None`. The negative branch is eliminated (returns `undefined`) only when the innermost base is *exactly* `None` (`isNoneInstance`), so a `NewType` over `None` correctly makes the `is not None` branch unreachable, while a None-*compatible* base like `NewType("Obj", object)` stays reachable.
- Literal comparison narrowing preserves the `NewType` brand on the positive branch (`return isClassInstance(subtype) && ClassType.isNewTypeClass(subtype) ? subtype : literalType;`), so `b is True` narrows to the `NewType` (e.g. `Banana`) rather than `Literal[True]`, avoiding a spurious `reportArgumentType` on a later construction.

Non-`NewType` types fall back to the original subtype, so existing behavior is unchanged.

## Testing

Regression coverage is provided by sample `newTypeIsLiteral1.py`, wired up as `NewType8` in `typeEvaluator2.test.ts`. The test runs with `reportUnreachable = 'error'` and asserts `validateResults(analysisResults, 0)` — any branch wrongly greyed out would surface as a `reportUnreachable` error and fail. It covers:

- `NewType` over `NoneType` narrowed by `is None` / `is not None` (exactly-None positive and negative directions).
- Literal narrowing (`is True` / `is False`) that preserves the `NewType` brand.
- None-/ellipsis-*compatible* bases (`NewType("Fig", object)`, `NewType("Guava", object)`) that must stay reachable on `is not None` / `is not ...` and not collapse to `Never`.

## Addresses

Fixes https://github.com/microsoft/pyright/issues/11463


packages/pyright-internal/src/tests/typeEvaluator2.test.ts
```diff
@@ -305,6 +305,12 @@ test('NewType7', () => {
     TestUtils.validateResults(analysisResults, 2);
 });
 
+test('NewType8', () => {
+    const analysisResults = TestUtils.typeAnalyzeSampleFiles(['newType8.py']);
+
+    TestUtils.validateResults(analysisResults, 1);
+});
```
rchiodo, line 314: `validateResults(analysisResults, 1)` only checks the error count, but `newType8.py` enables `# pyright: reportUnreachable=true` and the whole point of the sample is reachability. The unreachable block in `g()` emits a separate `UnreachableCode`-category diagnostic (via `checker.ts` `addUnreachableCode`) in addition to the `reportUnreachable` error, and that count currently goes unasserted — so a regression that wrongly greys out an extra branch in `f()` wouldn't be caught. Consider asserting it explicitly, e.g. `TestUtils.validateResults(analysisResults, 1, 0, undefined, undefined, 1);` (mirroring how `Unreachable1` asserts both error and unreachable counts). Non-blocking.

packages/pyright-internal/src/analyzer/typeGuards.ts
```diff
@@ -1134,8 +1159,11 @@ function narrowTypeForIsNone(evaluator: TypeEvaluator, type: Type, isPositiveTes
...
 
-            // Is it potentially None?
-            if (evaluator.assignType(subtype, evaluator.getNoneType())) {
+            // Is it potentially None? For NewType instances, the runtime value
+            // is an instance of the underlying base type, so test the base type
+            // for overlap with None.
+            const subtypeForNoneCheck = getNewTypeBaseInstance(subtype) ?? subtype;
+            if (evaluator.assignType(subtypeForNoneCheck, evaluator.getNoneType())) {
                 resultIncludesNoneSubtype = true;
                 return isPositiveTest
```
rchiodo, line 1178: The fix only harmonizes the positive overlap path. For `Apple = NewType("Apple", NoneType)`, `Apple(None) is not None` is `False` at runtime (the `else` of `if a is None:` is unreachable), but `narrowTypeForIsNone`'s negative path still returns the original `Apple` subtype and treats that branch as reachable — diverging from how a plain `None` is handled. Same one-directional limitation exists for the literal path (`b is True` narrows, but its `else` keeps `Banana` instead of `Literal[False]`). This is an exotic case and not a regression, but consider either mirroring the unwrap on the negative paths or documenting that only positive-direction narrowing is in scope for this fix.

rchiodo:
Fix looks correct and the sample covers the key cases. Two non-blocking notes below: tighten the test to also assert the unreachable-diagnostic count, and clarify whether negative-direction (`is not`/`else`) narrowing for NewType is intentionally out of scope.

packages/pyright-internal/src/analyzer/typeGuards.ts
```diff
@@ -1081,6 +1081,31 @@ function narrowTupleTypeForIsNone(evaluator: TypeEvaluator, type: Type, isPositi
...
+            break;
+        }
+
+        current = baseClass;
+        found = true;
+    }
+
+    return found ? ClassType.cloneAsInstance(current) : undefined;
+}
+
```
heejaechang, line 1109: 📍 packages/pyright-internal/src/analyzer/typeGuards.ts:1084
This PR adds `getNewTypeBaseInstance`, but the working tree already contains a more complete sibling `getInnermostNewTypeBaseInstance` that solves the same issue and is behaviorally divergent. Landing this PR would add a second near-duplicate helper and produce different `reveal_type` results for the same program. Reconcile to the single on-disk helper and narrowing semantic rather than shipping both.

[verified]

packages/pyright-internal/src/analyzer/typeGuards.ts
```diff
@@ -1134,8 +1159,11 @@ function narrowTypeForIsNone(evaluator: TypeEvaluator, type: Type, isPositiveTes
...
-            // Is it potentially None?
-            if (evaluator.assignType(subtype, evaluator.getNoneType())) {
+            // Is it potentially None? For NewType instances, the runtime value
+            // is an instance of the underlying base type, so test the base type
+            // for overlap with None.
+            const subtypeForNoneCheck = getNewTypeBaseInstance(subtype) ?? subtype;
+            if (evaluator.assignType(subtypeForNoneCheck, evaluator.getNoneType())) {
                 resultIncludesNoneSubtype = true;
                 return isPositiveTest
                     ? addConditionToType(evaluator.getNoneType(), subtype.props?.condition)
```
heejaechang, line 1179: 📍 packages/pyright-internal/src/analyzer/typeGuards.ts:1159
The `is not None` (negative) branch is not narrowed. For `Apple = NewType("Apple", NoneType)`, `a is None` correctly narrows to `None`, but the negative branch returns `adjustedSubtype` (`Apple`), so `a is not None` stays reachable even though an `Apple` is always `None`. Also, `assignType(subtypeForNoneCheck, None)` is true for a NewType over `object` too, conflating "None-compatible" with "exactly None". The on-disk version special-cases an *exactly-None* innermost base and returns `undefined` on the negative branch. This gap is untested — `newType8.py` never exercises `is not None`.

[verified]

packages/pyright-internal/src/analyzer/typeGuards.ts
```diff
@@ -2741,7 +2769,11 @@ function narrowTypeForLiteralComparison(
             }
 
             if (isIsOperator || isNoneInstance(subtype)) {
-                const isSubtype = evaluator.assignType(subtype, literalType);
+                // For NewType instances, the runtime value is an instance of the
+                // underlying base type, so test the base type for overlap with
+                // the literal.
+                const subtypeForCheck = getNewTypeBaseInstance(subtype) ?? subtype;
+                const isSubtype = evaluator.assignType(subtypeForCheck, literalType);
```
heejaechang: 📍 packages/pyright-internal/src/analyzer/typeGuards.ts:2769
Literal narrowing discards the NewType brand: `b is True` narrows to `Literal[True]`, dropping `Banana`. Because a `Literal[True]` is not assignable back to the `Banana` NewType (NewType requires explicit construction), a subsequent call like `needs_banana(b)` would now raise a spurious `reportArgumentType`. Suggested fix matching the on-disk version: `return isClassInstance(subtype) && ClassType.isNewTypeClass(subtype) ? subtype : literalType;`.

[verified]

packages/pyright-internal/src/tests/samples/newType8.py
```diff
@@ -0,0 +1,38 @@
...
+Date = NewType("Date", Apple)
+
+
+def f(a: Apple, b: Banana, c: Cherry, d: Date) -> None:
+    if a is None:
+        reveal_type(a, expected_text="None")
+
+    if b is True:
+        reveal_type(b, expected_text="Literal[True]")
+
```
heejaechang: 📍 packages/pyright-internal/src/tests/samples/newType8.py:21
The expectations here encode the brand-dropping semantics that were reversed on disk: `expected_text="Literal[True]"` / `"Literal[False]"` assert the loss of the `Banana`/`Cherry` brand, and would fail against `getInnermostNewTypeBaseInstance` (which preserves the NewType). Once the narrowing semantic is reconciled, update these expectations accordingly.

[verified]

packages/pyright-internal/src/tests/samples/newType8.py
```diff
@@ -0,0 +1,38 @@
...
+    if d is None:
+        reveal_type(d, expected_text="None")
+
+
+def g(c: Cherry) -> None:
+    # A NewType based on int cannot overlap with None, so this branch
+    # remains unreachable.
+    if c is None:
+        # This should generate an error because the code is unreachable.
+        reveal_type(c)
```
heejaechang: 📍 packages/pyright-internal/src/tests/samples/newType8.py:17
Missing negative-branch coverage. The sample only tests positive `is`/`is True`/`is False` branches plus one unreachable case. Add an `is not None` case on a None-based NewType (`if a is not None: reveal_type(a)` should be `Never`/unreachable) and ideally an `is not None` on a None-*compatible* base like `NewType("Obj", object)` (which must NOT collapse to `Never`). These lock in the chosen semantic and would have caught the negative-branch gap above.

[verified]

heejaechang:
This change overlaps with an already-landed fix on `main`: `getInnermostNewTypeBaseInstance` in typeGuards.ts already implements NewType `is` narrowing, and it diverges from this PR on two behaviors (handling of `is not None` for None-based NewTypes, and preserving the NewType brand on literal narrowing). Please reconcile to the single on-disk helper/semantic rather than landing a second near-duplicate, and update `newType8.py` expectations to match.

packages/pyright-internal/src/analyzer/typeGuards.ts
rchiodo: Reconciled — I merged current `upstream/main` into this branch and dropped the duplicate `getNewTypeBaseInstance`. The branch now uses the single on-disk `getInnermostNewTypeBaseInstance` helper and its narrowing semantics. After the merge the branch is identical to `main` for `typeGuards.ts`, so there is no longer a second near-duplicate helper or divergent `reveal_type` behavior.

packages/pyright-internal/src/analyzer/typeGuards.ts
rchiodo: Addressed by adopting main's helper. `getInnermostNewTypeBaseInstance` only treats a NewType as the singleton when its *innermost* base is exactly `None` (`isNoneInstance(newTypeBaseInstance)`), not merely None-compatible. On the negative branch it returns `undefined` (eliminating `is not None` for a None-based NewType), while a None-compatible base like `NewType("Obj", object)` falls through and stays reachable. This is exactly the exactly-None vs. None-compatible distinction you flagged.

packages/pyright-internal/src/analyzer/typeGuards.ts
rchiodo: Fixed via reconciliation. main's literal path preserves the brand exactly as you suggested: `return isClassInstance(subtype) && ClassType.isNewTypeClass(subtype) ? subtype : literalType;`. So `b is True` now narrows to `Banana` (not `Literal[True]`), avoiding the spurious `reportArgumentType` on a later `needs_banana(b)`.

packages/pyright-internal/src/tests/samples/newType8.py
rchiodo: Covered now. main's `newTypeIsLiteral1.py` includes negative-branch coverage in `g()`: `Fig = NewType("Fig", object)` / `Guava = NewType("Guava", object)` exercise `is not None` and `is not ...` on None-/ellipsis-*compatible* bases, asserting they stay reachable (`reveal_type` = `Fig`/`Guava`) and do not collapse to `Never`. The None-based NewTypes (`Apple`, `Apricot`) cover the exactly-None positive case under `reportUnreachable=error`. This locks in the chosen semantic.

packages/pyright-internal/src/tests/typeEvaluator2.test.ts
rchiodo: Addressed by reconciliation. The brittle `newType8.py`/`validateResults(analysisResults, 1)` pairing was replaced with main's `NewType8` test, which runs `newTypeIsLiteral1.py` with `reportUnreachable = 'error'` and asserts `validateResults(analysisResults, 0)`. Reachability is now enforced directly: any branch wrongly greyed out would surface as a `reportUnreachable` error and fail the test.

rchiodo:
I reconciled this branch with the already-landed fix on `main`. I merged current `upstream/main` and dropped the divergent `getNewTypeBaseInstance` helper plus the `newType8.py` sample, so the branch now uses the single on-disk `getInnermostNewTypeBaseInstance` helper and main's complete narrowing semantics (None/ellipsis/literal, brand-preserving, correct negative branches; covered by `newTypeIsLiteral1.py`).

heejaechang:

heejaechang:
The narrowing fix itself looks correct and well-targeted. Two follow-ups before merge: add a Pylance-harness regression test per the pylance-pyright-split rule, and update the PR description to match the actual implementation (`getInnermostNewTypeBaseInstance`, `newTypeIsLiteral1.py`, test expects 0 errors).

rchiodo:
Thanks — addressed the two follow-ups:

1. **PR description** — Updated to match the reconciled implementation: it now references `getInnermostNewTypeBaseInstance` (not the dropped `getNewTypeBaseInstance`), the `newTypeIsLiteral1.py` sample wired as `NewType8`, the `reportUnreachable = 'error'` + `validateResults(analysisResults, 0)` assertion, the brand-preserving literal narrowing, and the exactly-None vs. None-compatible negative-branch semantics. Also corrected the issue link to `microsoft/pyright#11463`.

2. **Pylance-harness regression test** — Not applicable in this repo. This PR targets the standalone `microsoft/pyright` repo, which has no Pylance harness (`packages/` contains only `pyright`, `pyright-internal`, and `vscode-pyright`). The `pylance-pyright-split` rule is a `microsoft/pyrx` monorepo convention and doesn't map here. The correct pyright-side regression test already exists — `newTypeIsLiteral1.py` (`NewType8`) — and enforces the behavior under `reportUnreachable = 'error'` (0 errors), covering both positive and negative `is None`/`is not None`/literal directions.

Note: after merging current `main`, the net diff of this branch vs `main` is empty — the `getInnermostNewTypeBaseInstance` fix and `newTypeIsLiteral1.py` test have fully landed on `main` already, so there are no remaining code changes on this branch.

bschnurr:
This appears fully superseded by #11491, which fixed #11463 and has already merged. This PR now has no remaining diff against `main`, so I think it can be closed.

PR #11497, opened 2026-06-15
Fix #11472: TypeVarTuple escapes method with very nested recursive tuple aliases

### LINKED ISSUE #11472: TypeVarTuple escapes method with very nested recursive tuple aliases
**Describe the bug**
When calling a method generic over a `TypeVarTuple` on a instance that contains deeply nested recursive tuples in type aliases, the generic "escapes" into the code.
```
Type "ForwardRefParser[tuple[Whitespace, Whitespace, tuple[BlockItem], OO@unpack_then], None]" is not assignable to declared type "ForwardRefParser[tuple[Whitespace, Whitespace, tuple[BlockItem], Whitespace], None]"
```
The issue is the `OO@unpack_then`, which escaped from the method.

On the code sample provided there should be no error.

If on the code sample you remove any of the levels of nesting of the type aliases, the issue goes away.
Eg applying this diff makes the code pass
```diff
- type Whitespace = tuple[tuple[tuple[tuple[Whitespace]]]]
+ type Whitespace = tuple[tuple[tuple[Whitespace]]]
```

Applying these two diffs at the same time makes the code pass:
```diff
          tuple[BlockItem],
-         Whitespace,
      ],
```
```diff
-     lambda: ws.then(ws).unpack_then(block).unpack_then(ws)
+     lambda: ws.then(ws).unpack_then(block)
```

**Code or Screenshots**
Code sample in [pyright playground](https://pyright-play.net/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMrCoCGANgDRRgBuApiCEgCY0BQokUAxmCSTZzCRgUAZwB0RAEadM2XPgDCpElL4t1AAUIpSLTipEioAeQDWAbQAqAXQBcLKI6gB9ZxCIxOAC2dEQaEVcoAF4oAAoAImcqUgBXGgiyAEp1JyhmYBdnVCQYVzCRGhJgChiSeNsoSyT7NLTC4rFouJpKyxCoMvj1Fi1iEj0DIwBRenNhuwcnDKycvOcCopLOlsrhmqm6qAbgJq7WqGGO-Z74BBooACUaEViSGCsKCY6zK2soAB9DsYn1M4uAAp%2BQoga63e7mYxPd6hMF3B4wWIIPjmVAwCjGazQlj-KBAkAggBisRQnEh0I6Sl4qho5nMIhgIAoaKxeOBdDhEKhh2s1h6%2BiIhjZBLo5J5tWmNEyrjm%2BR2FGAJM4lXxRKVYomGy2jh2TUVpJV7JAxNJGphBCVqUlmRgXhoKEh3OME0WxQozjAtrohpFIEdGM1UAAtAA%2BYUg8yI5G0p2Yp6fEwTCVbEBEJCFKAAOU9AElsHwIPaYDRGKNwCArY4ZiSEEROKZnF6HQAqSwAZQxTpdmzqOx9EajKNbbdZE3dnrtIH7ouMXb5aSSwbDqtFg9pw87o4Tzsm2qgqfTF2zMDz0cLKGLpfouHUMwQ4EYsU4tJsYWcbQT2ZQF1CX5oi9DKpKgPDNj1PAsixLMsb0GQUjEJXAAHc-EYa5gBXP1uRdDCzS1JwNGoOgGGYHsZhlFBcjlJYFSVSoqRUSQUXMVkcKw3k8K2XU3BoGAiGcfVlSgeiaTpFijTNc0BJ7eokToMIkiabIKPmMJ7zAR9n3klI-kQC4AHUvFyG5a2fDo10jJEUXM8yDKMkQTJoXleRYRCREqBCQGQkBUKlHDbOLey6xoCg-3NNSNJoeT1BQMAwAQac-QZJks2ERyOnCp9Iu04BcBIWL4qgDyvJ89DxJ7czpKcfzjKCshKscarAufOq93MgAhPL6xzYsICxeqoEahyWqcPq0j-OrzSKlC0IwsIexUCBJEYIhKlcsQmzCVyFJrOsGw2xiwHrbaUBMva7RQTaRBSbTcXAmhz14wRhCMUJrMs2kbMMgKHNZO6Ho8IRRGcrAzyLAHnvcpDpt88S-rBp7RBC1KwofTKopxXSoA6w7TG6%2B6zPeizo3MOGL3BoH3i%2BdrOtxnq%2BQO%2BsEqJlFsa6umke-FH1LRlIgA)

<details>
<summary>Collapsed for readability</summary>

```python
from typing import final, override


@final
class Ok[T]:
    __match_args__ = ("_value",)

    def __init__(self, value: T):
        self._value: T = value


@final
class Err[E]:
    def __init__(self, value: E):
        self._value: E = value


type Result[T, E] = Ok[T] | Err[E]

type ParserResult[O, E] = Result[tuple[int, O], E]
type ParserFunc[O, E] = Callable[[str, int], ParserResult[O, E]]


class Parser[O, E]:
    def __init__(self, func: ParserFunc[O, E]):
        self._func: ParserFunc[O, E] = func

    def then[OO, OE](self, _other: Parser[OO, OE]) -> Parser[tuple[O, OO], E | OE]:
        raise NotImplementedError

    def unpack_then[*TS, OO, OE](
        self: Parser[tuple[*TS], E], _other: Parser[OO, OE]
    ) -> Parser[tuple[*TS, OO], E | OE]:
        raise NotImplementedError

def produce[T](_: T | None = None) -> T: raise NotImplementedError

class ForwardRefParser[O, E](Parser[O, E]):
    @override
    def __init__(self, func: Callable[[], Parser[O, E]]):
        self._meta_func: Callable[[], Parser[O, E]] = func
        super().__init__(produce())


type Whitespace = tuple[tuple[tuple[tuple[Whitespace]]]]
ws: ForwardRefParser[Whitespace, None] = produce()

noop: Parser[str, None] = produce()

forloop: ForwardRefParser[
    tuple[
        Whitespace,
        Whitespace,
        tuple[BlockItem],
        Whitespace,
    ],
    None,
] = ForwardRefParser(
    lambda: ws.then(ws).unpack_then(block).unpack_then(ws)
)

type Implementations = tuple[tuple[tuple[Whitespace], Implementations]]
implementations: ForwardRefParser[Implementations, None] = produce()

type BlockItem = tuple[tuple[Implementations]] | tuple[BlockItem]
block: Parser[tuple[BlockItem], None] = produce()
```

</details>

Tested on online playground pyright version `1.1.410`

rchiodo:
## Description

Fixes a bug where a `TypeVarTuple` could "escape" a method's signature when the receiver's type arguments contained deeply nested recursive tuple type aliases. In these cases, an unsolved TypeVar (e.g. `OO` from `unpack_then`) leaked into the inferred type of an assignment, producing a spurious type error.

The repro chained parser combinators (`then` / `unpack_then`) on values whose types involved recursive tuple aliases (`Whitespace`, `Implementations`, `BlockItem`). Tracing the failure showed that `TypeVarTransformer` bailed out early when transforming tuple types whose container depth exceeded a hard-coded limit (`maxTupleTypeArgRecursionDepth = 10`). When that early-return fired mid-transformation, the method's TypeVars were never substituted, so they escaped into the result.

## Implementation

Removed the `maxTupleTypeArgRecursionDepth` guard in `typeUtils.ts`:

- Deleted the `maxTupleTypeArgRecursionDepth` constant.
- Removed the `getContainerDepth(classType) > maxTupleTypeArgRecursionDepth` early-return in `TypeVarTransformer`'s tuple handling, so deeply nested tuples are transformed fully instead of being returned unmodified.

The depth bail-out was originally a guard against effectively-infinite nesting, but the existing recursion-count limit already protects against runaway types, and the early return caused incorrect TypeVar solving in legitimate code.

## Testing

Added `typeVarTuple31.py` sample reproducing the escaping TypeVar with nested recursive tuple aliases, and wired it up as `TypeVarTuple31` in `typeEvaluator6.test.ts` (Python 3.12). The sample asserts zero errors; the final chained assignment previously failed because `OO` escaped. Ran the type-evaluator test suite to confirm the new test passes and no existing `TypeVarTuple` tests regressed.

## Addresses

Fixes https://github.com/microsoft/pylance-release/issues/11472

---

### DIFF
--- a/packages/pyright-internal/src/analyzer/typeUtils.ts
+++ b/packages/pyright-internal/src/analyzer/typeUtils.ts
@@ -232,6 +232,13 @@ export interface AddConditionOptions {
     skipBoundTypeVars?: boolean;
 }
 
+// There are cases where tuple types can be infinitely nested. The
+// recursion count limit will eventually be hit, but this will create
+// deep types that are expensive to construct. As a performance safeguard,
+// we limit the depth of the tuple type arguments. This value is large
+// enough that we should never hit it in legitimate circumstances.
+const maxTupleTypeArgRecursionDepth = 10;
+
 // Tracks whether a function signature has been seen before within
 // an expression. For example, in the expression "foo(foo, foo)", the
 // signature for "foo" will be seen three times at three different
@@ -3716,6 +3723,16 @@ export class TypeVarTransformer {
 
         // Handle tuples specially.
         if (ClassType.isTupleClass(classType)) {
+            // As a performance safeguard, bail out early on very deeply nested
+            // tuples (the recursion count limit would eventually stop us, but
+            // constructing such deep types is expensive). Only do this when there
+            // are no type variables left to substitute; bailing out while type
+            // variables remain would return the unspecialized class and let those
+            // TypeVars "escape" unsolved (see microsoft/pyright#11472).
+            if (getContainerDepth(classType) > maxTupleTypeArgRecursionDepth && !requiresSpecialization(classType)) {
+                return classType;
+            }
+
             if (classType.priv.tupleTypeArgs) {
                 newTupleTypeArgs = [];
 
--- a/packages/pyright-internal/src/analyzer/typeUtils.ts
+++ b/packages/pyright-internal/src/analyzer/typeUtils.ts
@@ -232,13 +232,6 @@ export interface AddConditionOptions {
     skipBoundTypeVars?: boolean;
 }
 
-// There are cases where tuple types can be infinitely nested. The
-// recursion count limit will eventually be hit, but this will create
-// deep types that will effectively hang the analyzer. To prevent this,
-// we'll limit the depth of the tuple type arguments. This value is
-// large enough that we should never hit it in legitimate circumstances.
-const maxTupleTypeArgRecursionDepth = 10;
-
 // Tracks whether a function signature has been seen before within
 // an expression. For example, in the expression "foo(foo, foo)", the
 // signature for "foo" will be seen three times at three different
@@ -3723,10 +3716,6 @@ export class TypeVarTransformer {
 
         // Handle tuples specially.
         if (ClassType.isTupleClass(classType)) {
-            if (getContainerDepth(classType) > maxTupleTypeArgRecursionDepth) {
-                return classType;
-            }
-
             if (classType.priv.tupleTypeArgs) {
                 newTupleTypeArgs = [];
 
--- a/packages/pyright-internal/src/tests/typeEvaluator6.test.ts
+++ b/packages/pyright-internal/src/tests/typeEvaluator6.test.ts
@@ -440,16 +440,18 @@ test('TypeVarTuple31', () => {
     const analysisResults = TestUtils.typeAnalyzeSampleFiles(['typeVarTuple31.py'], configOptions);
     TestUtils.validateResults(analysisResults, 0, 0, 1);
 
-    // The constructor call's inferred type should be fully concrete; in
-    // particular the solved "OO" TypeVar from "unpack_then" must not escape
-    // into the result. Pin it so a future divergent (but non-erroring)
-    // regression is caught.
-    assert.strictEqual(
-        analysisResults[0].infos[0].message,
-        'Type of "inferred" is "ForwardRefParser[tuple[tuple[tuple[tuple[tuple[Whitespace]]]], ' +
-            'tuple[tuple[tuple[tuple[Whitespace]]]], tuple[tuple[tuple[tuple[tuple[tuple[tuple[tuple[tuple[tuple[Whitespace]]]]], ' +
-            'Implementations]]]] | tuple[BlockItem]], tuple[tuple[tuple[tuple[Whitespace]]]]], None]"'
+    // The constructor call's inferred type should be fully concrete. In
+    // particular, the solved "OO" TypeVar from "unpack_then" must not escape
+    // into the result. Rather than pinning the full (brittle) nested-tuple
+    // expansion, assert the concrete outer shape and the absence of the leaked
+    // TypeVar so the fix stays self-evident without coupling to the exact
+    // type-printer output.
+    const revealedType = analysisResults[0].infos[0].message;
+    assert.ok(
+        revealedType.startsWith('Type of "inferred" is "ForwardRefParser['),
+        `Unexpected inferred type: ${revealedType}`
     );
+    assert.ok(!/\bOO\b/.test(revealedType), `"OO" TypeVar escaped into inferred type: ${revealedType}`);
 });
 
 test('Match1', () => {
--- a/packages/pyright-internal/src/tests/typeEvaluator6.test.ts
+++ b/packages/pyright-internal/src/tests/typeEvaluator6.test.ts
@@ -431,6 +431,14 @@ test('TypeVarTuple30', () => {
     TestUtils.validateResults(analysisResults, 0);
 });
 
+test('TypeVarTuple31', () => {
+    const configOptions = new ConfigOptions(Uri.empty());
+
+    configOptions.defaultPythonVersion = pythonVersion3_12;
+    const analysisResults = TestUtils.typeAnalyzeSampleFiles(['typeVarTuple31.py'], configOptions);
+    TestUtils.validateResults(analysisResults, 0);
+});
+
 test('Match1', () => {
     const configOptions = new ConfigOptions(Uri.empty());
 
--- a/packages/pyright-internal/src/tests/typeEvaluator6.test.ts
+++ b/packages/pyright-internal/src/tests/typeEvaluator6.test.ts
@@ -8,6 +8,8 @@
  * arbitrarily among multiple files so they can run in parallel.
  */
 
+import * as assert from 'assert';
+
 import { ConfigOptions } from '../common/configOptions';
 import { pythonVersion3_10, pythonVersion3_11, pythonVersion3_12, pythonVersion3_8 } from '../common/pythonVersion';
 import { Uri } from '../common/uri/uri';
@@ -436,7 +438,18 @@ test('TypeVarTuple31', () => {
 
     configOptions.defaultPythonVersion = pythonVersion3_12;
     const analysisResults = TestUtils.typeAnalyzeSampleFiles(['typeVarTuple31.py'], configOptions);
-    TestUtils.validateResults(analysisResults, 0);
+    TestUtils.validateResults(analysisResults, 0, 0, 1);
+
+    // The constructor call's inferred type should be fully concrete; in
+    // particular the solved "OO" TypeVar from "unpack_then" must not escape
+    // into the result. Pin it so a future divergent (but non-erroring)
+    // regression is caught.
+    assert.strictEqual(
+        analysisResults[0].infos[0].message,
+        'Type of "inferred" is "ForwardRefParser[tuple[tuple[tuple[tuple[tuple[Whitespace]]]], ' +
+            'tuple[tuple[tuple[tuple[Whitespace]]]], tuple[tuple[tuple[tuple[tuple[tuple[tuple[tuple[tuple[tuple[Whitespace]]]]], ' +
+            'Implementations]]]] | tuple[BlockItem]], tuple[tuple[tuple[tuple[Whitespace]]]]], None]"'
+    );
 });
 
 test('Match1', () => {
--- a/packages/pyright-internal/src/tests/samples/typeVarTuple31.py
+++ b/packages/pyright-internal/src/tests/samples/typeVarTuple31.py
@@ -1,0 +1,73 @@
+# This sample tests the case where a TypeVarTuple is used in a method
+# called on an instance whose type arguments contain deeply nested
+# recursive tuple type aliases. The deep nesting previously caused the
+# solved TypeVar to "escape" because type var transformation bailed out
+# early on deeply nested tuples.
+
+from typing import final, override
+from collections.abc import Callable
+
+
+@final
+class Ok[T]:
+    __match_args__ = ("_value",)
+
+    def __init__(self, value: T):
+        self._value: T = value
+
+
+@final
+class Err[E]:
+    def __init__(self, value: E):
+        self._value: E = value
+
+
+type Result[T, E] = Ok[T] | Err[E]
+
+type ParserResult[O, E] = Result[tuple[int, O], E]
+type ParserFunc[O, E] = Callable[[str, int], ParserResult[O, E]]
+
+
+class Parser[O, E]:
+    def __init__(self, func: ParserFunc[O, E]):
+        self._func: ParserFunc[O, E] = func
+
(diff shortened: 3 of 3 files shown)


packages/pyright-internal/src/analyzer/typeUtils.ts
```diff
@@ -3723,10 +3716,6 @@ export class TypeVarTransformer {
 
         // Handle tuples specially.
         if (ClassType.isTupleClass(classType)) {
-            if (getContainerDepth(classType) > maxTupleTypeArgRecursionDepth) {
-                return classType;
-            }
-
             if (classType.priv.tupleTypeArgs) {
```
rchiodo, line 3736: Removing this early-return fixes the escaping TypeVar (1→0 errors), but it also deletes a documented performance safeguard whose comment explicitly claimed the recursion-count limit is insufficient ("will effectively hang the analyzer"). Consider a narrower, safe-bailout variant: keep a depth bound but make the bailout safe (e.g. return `Unknown` or still solve top-level TypeVars) so nothing escapes at any depth while preserving the perf guard. Since this is vendored, upstream-owned Pyright core (keep diffs minimal, coordinate upstream), prefer the safe-bailout variant or coordinate the bound removal upstream against the original motivating repro, and confirm that repro still terminates without the guard.

packages/pyright-internal/src/tests/typeEvaluator6.test.ts
```diff
@@ -431,6 +431,14 @@ test('TypeVarTuple30', () => {
     TestUtils.validateResults(analysisResults, 0);
 });
 
+test('TypeVarTuple31', () => {
+    const configOptions = new ConfigOptions(Uri.empty());
+
+    configOptions.defaultPythonVersion = pythonVersion3_12;
+    const analysisResults = TestUtils.typeAnalyzeSampleFiles(['typeVarTuple31.py'], configOptions);
```
rchiodo, line 440: `validateResults(analysisResults, 0)` asserts only "no errors" — it would still pass if `OO` were replaced by some other wrong-but-non-erroring type. For a TypeVar-escape bug, add a `reveal_type` (or hover) assertion pinning `forloop`'s concrete inferred type so the fix's correctness is self-evident and a future divergent-but-non-error regression is caught.

rchiodo:
Fix looks correct and well-targeted. Two non-blocking suggestions: (1) the removed depth guard was a documented performance safeguard in vendored Pyright core — please confirm the original motivating workload still terminates and consider coordinating upstream; (2) strengthen the test with a reveal_type assertion to pin the inferred type rather than only asserting zero errors.

packages/pyright-internal/src/tests/typeEvaluator6.test.ts
rchiodo: Done in 44a4c50ed. The test now reveals the inferred type of the constructor call via an unannotated `inferred` binding (so the inferred, not declared, type is exercised) and asserts the exact revealed type string, pinning it to the fully concrete `ForwardRefParser[tuple[...], None]` with no `OO` TypeVar present. A future divergent-but-non-erroring regression (e.g. OO leaking back in) now fails the test. validateResults also asserts infoCount=1.

packages/pyright-internal/src/analyzer/typeUtils.ts
rchiodo: Evaluated the safe-bailout variant and chose to keep the full removal, with reasoning:

heejaechang, line 3736: Consider **narrowing** this guard instead of deleting it. The real defect is that the early-return skipped TypeVar *substitution* and returned the bare unspecialized `classType`, letting `OO` escape. Rather than removing the depth bound entirely, you could short-circuit only when the subtree has no in-scope TypeVars (or keep substituting under a cap, returning `Unknown` for truncated args). That fixes the escape while preserving parity with the sibling depth guards (`maxInferredContainerDepth` in `tuples.ts`, and the guards in `typeEvaluator.ts`). The hang-regression concern was empirically disproven, so this is not a blocker — but at minimum note in the PR why full deletion is preferred over a bounded bail-out.

packages/pyright-internal/src/tests/typeEvaluator6.test.ts
```diff
@@ -431,6 +433,25 @@ test('TypeVarTuple30', () => {
...
+    // regression is caught.
+    assert.strictEqual(
+        analysisResults[0].infos[0].message,
+        'Type of "inferred" is "ForwardRefParser[tuple[tuple[tuple[tuple[tuple[Whitespace]]]], ' +
+            'tuple[tuple[tuple[tuple[Whitespace]]]], tuple[tuple[tuple[tuple[tuple[tuple[tuple[tuple[tuple[tuple[Whitespace]]]]], ' +
+            'Implementations]]]] | tuple[BlockItem]], tuple[tuple[tuple[tuple[Whitespace]]]]], None]"'
+    );
+});
+
 test('Match1', () => {
```
heejaechang, line 457: This `assert.strictEqual` pins a ~250-char fully-expanded nested-tuple `reveal_type` string. Since the stated intent is only to catch the `OO` TypeVar leaking back in, this couples the test to exact type-printer output and will break on benign rendering/ordering changes with a near-unreadable diff. Consider asserting the *absence* of the leak (e.g. `!/\bOO\b/.test(message)`) or a concrete-shape substring instead of the full expansion. `validateResults(..., 0, 0, 1)` already guards that the reveal exists.

heejaechang:
Looks fundamentally sound — approving. The central question reviewers converged on is whether deleting this *global* tuple-depth guard is safe; the hang regression was empirically disproven, so this isn't blocking. To strengthen the change, consider narrowing the guard (substitute-but-cap) rather than deleting it outright, and/or adding a timeout-bounded regression sample for the original infinite-nesting scenario the guard protected.

heejaechang:
Looks fundamentally sound — approving. The main thing worth a second look is whether to narrow the depth guard rather than delete it outright (it was originally hang-protection), plus reducing the brittleness of the exact-string reveal_type assertion. Neither is blocking.

packages/pyright-internal/src/analyzer/typeUtils.ts
rchiodo: Good call — switched from full deletion to the narrowed/safe-bailout variant in 2664550d7. The depth guard (`maxTupleTypeArgRecursionDepth`) is back, but the early return now only fires when the deeply nested tuple has no TypeVars left to substitute:

```ts
if (getContainerDepth(classType) > maxTupleTypeArgRecursionDepth && !requiresSpecialization(classType)) {
    return classType;
}
```

Since we never bail while type vars remain, `OO` can no longer escape unsolved at any depth, and the performance safeguard (plus parity with the sibling depth guards) is preserved. The fast path for deep-but-concrete tuples is kept, since `requiresSpecialization` is false there.

packages/pyright-internal/src/tests/typeEvaluator6.test.ts
rchiodo: Agreed — the full-expansion `strictEqual` was too brittle. Replaced it in 2664550d7 with assertions that pin the concrete outer shape and verify the leaked TypeVar is absent:

```ts
const revealedType = analysisResults[0].infos[0].message;
assert.ok(
    revealedType.startsWith('Type of "inferred" is "ForwardRefParser['),
    `Unexpected inferred type: ${revealedType}`
);
assert.ok(!/\bOO\b/.test(revealedType), `"OO" TypeVar escaped into inferred type: ${revealedType}`);
```

`validateResults(..., 0, 0, 1)` still guards that the reveal exists, and this no longer breaks on benign type-printer/ordering changes while still catching the `OO` leak regression.

heejaechang:
Approving. The fix correctly stops the TypeVar from escaping, but the main feedback is that fully deleting the depth guard (vs. narrowing it so substitution still happens) removes a hang-protection bound — worth a brief justification and a hang-shaped regression sample. The exact-string test assertion is also brittle.

heejaechang (approved):
Approved via Review Center.

packages/pyright-internal/src/analyzer/typeUtils.ts
```diff
@@ -3723,7 +3723,13 @@ export class TypeVarTransformer {
 
         // Handle tuples specially.
         if (ClassType.isTupleClass(classType)) {
-            if (getContainerDepth(classType) > maxTupleTypeArgRecursionDepth) {
+            // As a performance safeguard, bail out early on very deeply nested
+            // tuples (the recursion count limit would eventually stop us, but
+            // constructing such deep types is expensive). Only do this when there
+            // are no type variables left to substitute; bailing out while type
+            // variables remain would return the unspecialized class and let those
+            // TypeVars "escape" unsolved (see microsoft/pyright#11472).
```
heejaechang, line 3731: The new comment ends with an inline issue reference `(see microsoft/pyright#11472)`. Issue links (including GitHub shorthand) are prohibited in production source per the repo's issue-reference policy. Remove `(see microsoft/pyright#11472)` and keep that traceability in the PR/commit metadata instead. The rest of the comment—explaining why the bail-out is gated on `!requiresSpecialization`—is good and should stay.

packages/pyright-internal/src/tests/typeEvaluator6.test.ts
```diff
@@ -431,6 +433,27 @@ test('TypeVarTuple30', () => {
...
+    // into the result. Rather than pinning the full (brittle) nested-tuple
+    // expansion, assert the concrete outer shape and the absence of the leaked
+    // TypeVar so the fix stays self-evident without coupling to the exact
+    // type-printer output.
+    const revealedType = analysisResults[0].infos[0].message;
+    assert.ok(
+        revealedType.startsWith('Type of "inferred" is "ForwardRefParser['),
+        `Unexpected inferred type: ${revealedType}`
+    );
+    assert.ok(!/\bOO\b/.test(revealedType), `"OO" TypeVar escaped into inferred type: ${revealedType}`);
```
heejaechang, line 454: `revealedType.startsWith(...)` plus `!/\bOO\b/.test(revealedType)` are partial assertions on deterministic `reveal_type` output, which the repo's no-partial-asserts test rule prohibits. `reveal_type` of a statically-typed expression is fully deterministic, so the non-determinism exception does not apply. Capture the full inferred-type string (e.g. via the Jest inspector) and assert it exactly with `assert.strictEqual(revealedType, '<full exact string>')`. If shape-only assertions are genuinely wanted here, that requires amending the rule, not bypassing it in this test.

heejaechang:
Core fix looks reasonable (narrowing the depth bail-out with `!requiresSpecialization` so unsolved TypeVars can't escape). Two cleanups needed before merge: drop the inline issue link from the source comment, and assert the full reveal_type string exactly rather than using partial checks.

heejaechang:
Code logic looks sound — note that the actual diff narrows the depth guard with `!requiresSpecialization(classType)` rather than deleting it (the PR description's "Removed the guard" wording is stale and should be corrected). Two rule-based fixes are needed before merge: (1) drop the `(see microsoft/pyright#11472)` reference from the source comment, and (2) reconcile the test assertions with `tests-no-partial-asserts.md`. Minor: per repo convention, `pylance-release` issues use "Addresses", not "Fixes".

PR #11498, opened 2026-06-15
Fix #11470: Pyright is incorrectly handling overload functions

### LINKED ISSUE #11470: Pyright is incorrectly handling overload functions
**Describe the bug**

According to the [official Python spec for overload](https://typing.python.org/en/latest/spec/overload.html), the `@overload` function can be followed by a single non-overload function:

> In regular modules, a series of `@overload`-decorated definitions must be followed by exactly one non-`@overload`-decorated definition (for the same function/method).
>
> The `@overload`-decorated definitions are for the benefit of the type checker only, since they will be overwritten by the non-`@overload`-decorated definition, while the latter is used at runtime but ***should be ignored by a type checker***.

The spec also gives an example of an overloaded function, which does not specify types for the non-overloaded function:

```python
@overload
def utf8(value: None) -> None:
    pass
@overload
def utf8(value: bytes) -> bytes:
    pass
@overload
def utf8(value: unicode) -> bytes:
    pass
def utf8(value):
    <actual implementation>
```

The type checker should assume that the `@overload` functions accurately describe the entire type of the function, and the type checker should ignore the type of the non-overloaded function.

But pyright is incorrectly type checking the non-overload function:

```python
from typing import overload

class Foo:
    @overload
    def foo(self, value: int) -> int: ...

    @overload
    def foo(self, value: str) -> str: ...

    def foo(self, value):
        return value
```

When type checking that code in strict mode, pyright incorrectly gives the following errors:

```
__init__.py:10:19 - error: Type of parameter "value" is unknown (reportUnknownParameterType)

__init__.py:10:19 - error: Type annotation is missing for parameter "value" (reportMissingParameterType)
```

In order to suppress those errors, I need to add an `int | str` union type to the final non-overload function.

In that simple example it might seem okay, but with complicated types it can be very tricky indeed. Here is a real example from my project:

```python
@overload
def _make_item[A: JSON](self, id: str, default: A | None, item_class: type[Item[JSON]]) -> Item[A]: ...

@overload
def _make_item(self, id: str, default: dict[str, JSON] | None, item_class: type["ItemDict"]) -> "ItemDict": ...

@overload
def _make_item(self, id: str, default: list[dict[str, JSON]] | None, item_class: type["ItemList"]) -> "ItemList": ...

def _make_item(self, id, default, item_class):
    # Actual function body...
```

Trying to union the 3 different overloads is very annoying, and a union isn't even correct because of the generic.

In order to comply with the spec, pyright must ignore the type of the final non-overload function in the `@overload` chain.

I would even go further and suggest that pyright should error if the user tries to specify a type for the non-overload function, forcing the non-overload function to remain untyped.

This prevents the footgun where a user adds a type to the non-overload function and is then confused why it's not working.

Of course the *body* of the function should still be type-checked as usual, it is only the function arguments and return type that must remain untyped.

I'm running pyright from the command line, [installed from pypi](https://pypi.org/project/pyright/). It is version 1.1.409.

rchiodo:
## Description

Fixes a false positive where Pyright reported `reportUnknownParameterType` and `reportMissingParameterType` on the **implementation** of an overloaded function. Per the typing spec, only the `@overload` signatures define the function's public type; the implementation signature is ignored by the type checker, so its parameters are permitted to remain unannotated. Pyright was incorrectly flagging unannotated parameters on the implementation under strict mode.

The unknown/missing parameter type checks in `checker.ts` ran unconditionally for every function definition. Overload implementations were not exempted, even though their signatures carry no type-checking weight. The fix is to detect overload implementations and skip these two specific diagnostics for them.

## Implementation

In `packages/pyright-internal/src/analyzer/checker.ts`, compute whether the current function is an overload implementation:

```ts
const isOverloadImplementation =
    isOverloaded(functionTypeResult.decoratedType) &&
    !FunctionType.isOverloaded(functionTypeResult.functionType);
```

The decorated type is overloaded (the symbol has `@overload` signatures), but the function type itself is not marked overloaded — which uniquely identifies the implementation function. This guard is then added to the `reportUnknownParameterType` and `reportMissingParameterType` checks so they are skipped for the implementation only. All other functions (including non-overloaded ones and the `@overload` signatures themselves) continue to be checked as before.

## Testing

Added `packages/pyright-internal/src/tests/samples/overloadImpl3.py` under `# pyright: strict`, covering:

- An overloaded **method** (`Foo.foo`) with an unannotated `value` parameter on the implementation — should produce **no** errors.
- An overloaded **free function** (`func1`) with an unannotated implementation parameter — should produce **no** errors.
- A regular non-overloaded function (`func2`) with an unannotated parameter — should still produce both the unknown-type and missing-annotation errors (2 expected errors).

Wired up `OverloadImpl3` in `typeEvaluator6.test.ts`, asserting exactly 2 errors so the regression coverage confirms both that the implementation is exempt and that ordinary functions are unaffected.

Run with:

```
cd packages/pyright/packages/pyright-internal
node node_modules\jest\bin\jest typeEvaluator6 --runInBand --detectOpenHandles --forceExit --testTimeout=180000 -t "OverloadImpl3"
```

## Addresses

Fixes https://github.com/microsoft/pylance-release/issues/11470

---

### DIFF
--- a/packages/pyright-internal/src/analyzer/checker.ts
+++ b/packages/pyright-internal/src/analyzer/checker.ts
@@ -421,6 +421,15 @@ export class Checker extends ParseTreeWalker {
             const keywordNames = new Set<string>();
             const paramDetails = getParamListDetails(functionTypeResult.functionType);
 
+            // If this function is the implementation of an overloaded function,
+            // its signature is ignored by the type checker (only the @overload
+            // signatures define the function's type). Per the typing spec, the
+            // implementation's parameters are allowed to remain unannotated, so
+            // skip the unknown/missing parameter type checks for it.
+            const isOverloadImplementation =
+                isOverloaded(functionTypeResult.decoratedType) &&
+                !FunctionType.isOverloaded(functionTypeResult.functionType);
+
             // Report any unknown or missing parameter types.
             node.d.params.forEach((param, index) => {
                 if (param.d.name) {
@@ -464,7 +473,10 @@ export class Checker extends ParseTreeWalker {
                         const functionTypeParam = functionTypeResult.functionType.shared.parameters[paramIndex];
                         const paramType = FunctionType.getParamType(functionTypeResult.functionType, paramIndex);
 
-                        if (this._fileInfo.diagnosticRuleSet.reportUnknownParameterType !== 'none') {
+                        if (
+                            !isOverloadImplementation &&
+                            this._fileInfo.diagnosticRuleSet.reportUnknownParameterType !== 'none'
+                        ) {
                             if (
                                 isUnknown(paramType) ||
                                 (isTypeVar(paramType) &&
@@ -504,7 +516,11 @@ export class Checker extends ParseTreeWalker {
                             }
                         }
 
-                        if (!hasAnnotation && this._fileInfo.diagnosticRuleSet.reportMissingParameterType !== 'none') {
+                        if (
+                            !isOverloadImplementation &&
+                            !hasAnnotation &&
+                            this._fileInfo.diagnosticRuleSet.reportMissingParameterType !== 'none'
+                        ) {
                             this._evaluator.addDiagnostic(
                                 DiagnosticRule.reportMissingParameterType,
                                 LocMessage.paramAnnotationMissing().format({ name: param.d.name.d.value }),
--- a/packages/pyright-internal/src/tests/samples/overloadImpl3.py
+++ b/packages/pyright-internal/src/tests/samples/overloadImpl3.py
@@ -1,0 +1,41 @@
+# This sample tests that the parameters of an overload implementation are not
+# subject to the reportUnknownParameterType and reportMissingParameterType
+# checks. The implementation signature is ignored by the type checker, so its
+# parameters are allowed to remain unannotated.
+
+# pyright: strict
+
+from typing import overload
+
+
+class Foo:
+    @overload
+    def foo(self, value: int) -> int: ...
+
+    @overload
+    def foo(self, value: str) -> str: ...
+
+    # The unannotated "value" parameter should not generate errors here.
+    def foo(self, value) -> int | str:
+        return 0
+
+
+@overload
+def func1(value: int) -> int: ...
+
+
+@overload
+def func1(value: str) -> str: ...
+
+
+# The unannotated "value" parameter should not generate errors here.
+def func1(value) -> int | str:
+    return 0
+
+
+# A non-overloaded function with an unannotated parameter should still
+# generate errors.
+def func2(value) -> int:
+    # This should generate an error because "value" has an unknown type and
+    # an error because "value" is missing a type annotation.
+    return 1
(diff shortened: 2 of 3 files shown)


packages/pyright-internal/src/analyzer/checker.ts
```diff
@@ -421,6 +421,15 @@ export class Checker extends ParseTreeWalker {
             const keywordNames = new Set<string>();
             const paramDetails = getParamListDetails(functionTypeResult.functionType);
 
+            // If this function is the implementation of an overloaded function,
+            // its signature is ignored by the type checker (only the @overload
+            // signatures define the function's type). Per the typing spec, the
+            // implementation's parameters are allowed to remain unannotated, so
+            // skip the unknown/missing parameter type checks for it.
+            const isOverloadImplementation =
```
rchiodo, line 429: This uses a non-canonical detection idiom. The same file already exposes the explicit, intended API for exactly this question: `OverloadedType.getImplementation(functionTypeResult.decoratedType) === functionTypeResult.functionType` (used at `checker.ts:718-719`, and several times in `typeUtils.ts`). This change instead copies the looser heuristic from `skipUnsolvableTypeVarCheck` (`checker.ts:2531-2533`), creating a third inline copy of overload-implementation detection. Prefer extracting a single private helper (e.g. `isOverloadImplementation(funcTypeResult)`) built on `getImplementation(...)` and reuse it both here and at the existing heuristic site. Functionally equivalent today, but self-documenting and single-source-of-truth.

packages/pyright-internal/src/analyzer/checker.ts
```diff
@@ -421,6 +421,15 @@ export class Checker extends ParseTreeWalker {
...
+                isOverloaded(functionTypeResult.decoratedType) &&
```
rchiodo, line 430: Please confirm the premise before merging. The justification "the implementation signature is ignored by the type checker" applies to *call-site* typing, but the implementation **body is still type-checked** and the param still resolves to `Unknown`. Pyright's documented `reportMissingParameterType` exemptions are only `self`/`cls`, with no mention of overload implementations, so this may be intentional strict-mode behavior rather than a false positive. Since this is an upstream pyright change to documented strict-mode semantics, verify with the Pyright maintainer that this is genuinely considered a false positive to avoid a likely rejection / local divergence.

rchiodo:
Targeted fix with a focused regression test. Two notes worth addressing before merge: (1) prefer the canonical `OverloadedType.getImplementation(...)` API instead of adding a third inline overload-implementation heuristic, and (2) confirm with upstream that strict-mode reporting on overload implementations is genuinely considered a false positive rather than intended behavior.

heejaechang:
Fix looks correct for the reported overload-implementation false positives. Two follow-ups worth considering before merge: (1) the same 'implementation signature is ignored' rationale arguably applies to the impl's return type, which is not exempted here, and (2) per the repo's Pyright↔Pylance split, the primary regression test should live in the Pylance harness with the pyright-internal test as additional coverage.

heejaechang:
Fix looks correct and the predicate uniquely selects the overload implementation. Two non-blocking notes: (1) per the Pyright↔Pylance split policy, a pylance-release fix should have its primary regression test in the Pylance harness; (2) the same 'implementation signature is ignored' rationale arguably applies to the implementation's return type, so it's worth either extending the exemption or documenting that this is params-only.

heejaechang (approved):
Approved via Review Center.

PR #11499, opened 2026-06-15
Fix #11478: Support overloads on property setters

### LINKED ISSUE #11478: Support overloads on property setters
Edit: more broadly overloaded property setters aren't supported:

Code sample in [pyright playground](https://pyright-play.net/?reportDeprecated=true&code=GYJw9gtgBALgngBwJYDsDmUkQWEMpgBuApiADZgCGAJgFC0DGZlAzi1AIIBctUfUAAQTgEpeL37ViwWAEYAFC2JlgASigBaAHyYUMLlAB0x%2Bv0ExZhpTBikJfAUVIUa9qFJkXFy4ABoohJRkAK7EBqgw6tpQAHJgKGFGJm4CFlbENnZmjiTkVHRmHnLeKv6BIYlxCVE6VYnGhqb8qZbWtiBuRV5KpQFBoQYsMCA1sfH1yfSUUAC8nPKqU4YWs1CyfADEUABEADJI7UEA2rIAutuY7Chg%2BKwsSGgolABGZMSwYDtDINtAA)

```python
from typing import overload

class A:
    @property
    def t1(self) -> int: ...

    @t1.setter
    @overload
    def t1(self, value: int) -> None: ...

    @t1.setter
    @overload
    def t1(self, value: None) -> None: ...

    @t1.setter
    def t1(self, value: str) -> None: ...


a = A()

a.t1 = 1  # "Literal[1]" is not assignable to "str"
```

---

Code sample in [pyright playground](https://pyright-play.net/?reportDeprecated=true&code=GYJw9gtgBALgngBwJYDsDmUkQWEMpgBuApiADZgCGAJgFCiRQDulIKqaAzptrvtcQQhiAY0oxidWiLKVO3AIIAuWlDVQAAkLAJS8VeoHBYARgAUnYmWABKKAFoAfJhQwV6j1AB0P2gbUaMCZeljASIP6aAkKi4pJmAEQAysRhHLCY3NHCYhLUCTaRRqYWVsAANFCElGQArsRKLjB2TlAAcmAoDZEePl5%2BHlrgunhwRcTGMABMpdYtzqhuPep9A%2BqBUyGp4ZEaRKQUNLvZsXmJKWnoGTBg7Z3EmVAnuZIF45MzltaV1XUNd115gDup4Vr5dtMtmFSLt9uQqHQPMVprMKlUavVGosgR0uu5Qd5wYNIaEdkiJrBPmUfhj-osoAAfYE4%2B740GrWiUKAAXigCjMhU5XiCPKgJiF01FuOIfiAA)

```python
from typing import overload
from warnings import deprecated

class A:
    @property
    def t1(self) -> int:
        ...

    @t1.setter
    @deprecated("Setting t is deprecated")
    def t1(self, value: int) -> None:
        ...

    @property
    def t2(self) -> int:
        ...

    @t2.setter
    @overload
    @deprecated("Setting t to None is deprecated")
    def t2(self, value: None) -> None:
        ...

    @t2.setter
    @overload
    def t2(self, value: int) -> None:
        ...

    @t2.setter
    def t2(self, value: int | None) -> None:
        ...

a = A()

a.t1 = 1  # reported
a.t2 = None  # not reported
```

See also: https://discuss.python.org/t/107604

rchiodo:
## Description

Adds support for `@overload`-decorated property setters. Previously a property's `fset` accessor could only hold a single `FunctionType`, so multiple `@x.setter @overload` declarations were not combined and the synthesized `__set__` method ignored the overload signatures. This change accumulates setter overloads and synthesizes one `__set__` overload per setter signature, so assignment type-checking, deprecation reporting, and `reveal_type` all respect the overloaded setter.

The `PropertyMethodInfo.methodType` field was typed as a single `FunctionType`, and `addSetMethodToPropertySymbolTable` synthesized exactly one `__set__`. Tracing how `clonePropertyWithSetter` rebuilds the property on each `@x.setter` declaration showed that earlier setter signatures were discarded rather than merged, which is why overloaded setters did not type-check.

## Implementation

- Broadened `PropertyMethodInfo.methodType` to `FunctionType | OverloadedType`, and updated all property-accessor consumers (`checker.ts`, `typeEvaluator.ts`, `typeGuards.ts`, `typeUtils.ts`, `packageTypeVerifier.ts`) to guard with `isFunction(...)` / handle the overloaded case.
- Added `combineSetterOverloads` in `properties.ts` to accumulate setter overloads across declarations, treating a non-overloaded final setter as the implementation when overloads are present.
- Reworked `addSetMethodToPropertySymbolTable` to synthesize one `__set__` per setter overload (excluding the implementation), via the extracted `createSetMethodFromSetter` helper.

## Testing

- Added `property19.py` (overloaded setter accepts matching types, rejects `str`, deleter still works) and `property20.py` (deprecation resolves to the matching setter overload).
- Wired up `Property19` and `Property20` cases in `typeEvaluator8.test.ts`, including a `reportDeprecated = 'error'` run.
- Ran the relevant `typeEvaluator8` property tests to confirm expected diagnostics.

## Addresses

Fixes https://github.com/microsoft/pylance-release/issues/11478

---

### DIFF
--- a/packages/pyright-internal/src/analyzer/checker.ts
+++ b/packages/pyright-internal/src/analyzer/checker.ts
@@ -6067,6 +6067,10 @@ export class Checker extends ParseTreeWalker {
                     this._evaluator.getTypeClassType()
                 );
 
+                // Overloaded accessors (e.g. overloaded property setters) are
+                // represented as an OverloadedType rather than a FunctionType.
+                // Override-compatibility checking for overloaded accessors is not
+                // yet performed; only single-function accessors are validated here.
                 if (isFunction(baseClassMethodType)) {
                     if (!subclassPropMethod) {
                         // The method is missing.
@@ -7007,6 +7011,10 @@ export class Checker extends ParseTreeWalker {
                     this._evaluator.getTypeClassType()
                 );
 
+                // Overloaded accessors (e.g. overloaded property setters) are
+                // represented as an OverloadedType rather than a FunctionType.
+                // Override-compatibility checking for overloaded accessors is not
+                // yet performed; skip them rather than misreporting.
                 if (!isFunction(baseClassMethodType)) {
                     return;
                 }
--- a/packages/pyright-internal/src/analyzer/checker.ts
+++ b/packages/pyright-internal/src/analyzer/checker.ts
@@ -6047,7 +6047,7 @@ export class Checker extends ParseTreeWalker {
         memberName: string,
         errorNode: ParseNode
     ) {
-        const propMethodInfo: [string, (c: ClassType) => FunctionType | undefined][] = [
+        const propMethodInfo: [string, (c: ClassType) => FunctionType | OverloadedType | undefined][] = [
             ['fget', (c) => c.priv.fgetInfo?.methodType],
             ['fset', (c) => c.priv.fsetInfo?.methodType],
             ['fdel', (c) => c.priv.fdelInfo?.methodType],
@@ -6487,7 +6487,7 @@ export class Checker extends ParseTreeWalker {
                 overrideFunction = impl;
             }
         } else if (isClassInstance(overrideType) && ClassType.isPropertyClass(overrideType)) {
-            if (overrideType.priv.fgetInfo) {
+            if (overrideType.priv.fgetInfo && isFunction(overrideType.priv.fgetInfo.methodType)) {
                 overrideFunction = overrideType.priv.fgetInfo.methodType;
             }
         }
@@ -6551,7 +6551,7 @@ export class Checker extends ParseTreeWalker {
                 }
             }
         } else if (isClassInstance(overrideType) && ClassType.isPropertyClass(overrideType)) {
-            if (overrideType.priv.fgetInfo) {
+            if (overrideType.priv.fgetInfo && isFunction(overrideType.priv.fgetInfo.methodType)) {
                 overrideFunction = overrideType.priv.fgetInfo.methodType;
             }
         }
@@ -6987,7 +6987,7 @@ export class Checker extends ParseTreeWalker {
         overrideSymbol: Symbol,
         memberName: string
     ) {
-        const propMethodInfo: [string, (c: ClassType) => FunctionType | undefined][] = [
+        const propMethodInfo: [string, (c: ClassType) => FunctionType | OverloadedType | undefined][] = [
             ['fget', (c) => c.priv.fgetInfo?.methodType],
             ['fset', (c) => c.priv.fsetInfo?.methodType],
             ['fdel', (c) => c.priv.fdelInfo?.methodType],
--- a/packages/pyright-internal/src/analyzer/packageTypeVerifier.ts
+++ b/packages/pyright-internal/src/analyzer/packageTypeVerifier.ts
@@ -876,7 +876,7 @@ export class PackageTypeVerifier {
             case TypeCategory.Class: {
                 // Properties require special handling.
                 if (TypeBase.isInstance(type) && ClassType.isPropertyClass(type)) {
-                    const propMethodInfo: [string, (c: ClassType) => FunctionType | undefined][] = [
+                    const propMethodInfo: [string, (c: ClassType) => FunctionType | OverloadedType | undefined][] = [
                         ['fget', (c) => c.priv.fgetInfo?.methodType],
                         ['fset', (c) => c.priv.fsetInfo?.methodType],
                         ['fdel', (c) => c.priv.fdelInfo?.methodType],
--- a/packages/pyright-internal/src/analyzer/packageTypeVerifier.ts
+++ b/packages/pyright-internal/src/analyzer/packageTypeVerifier.ts
@@ -46,6 +46,7 @@ import {
     isFunction,
     isInstantiableClass,
     isModule,
+    isOverloaded,
     isTypeSame,
     isUnknown,
     ModuleType,
@@ -884,6 +885,9 @@ export class PackageTypeVerifier {
 
                     const propertyClass = type;
 
+                    const stripStaticMethodFlag = (func: FunctionType) =>
+                        FunctionType.cloneWithNewFlags(func, func.shared.flags & ~FunctionTypeFlags.StaticMethod);
+
                     propMethodInfo.forEach((info) => {
                         const methodAccessor = info[1];
                         let accessType = methodAccessor(propertyClass);
@@ -897,9 +901,17 @@ export class PackageTypeVerifier {
                             // work properly when accessed directly from the property object. We need
                             // to remove this flag here so the method is seen as an instance method rather than
                             // static. Otherwise we'll incorrectly report that "self" is not annotated.
-                            accessType = FunctionType.cloneWithNewFlags(
-                                accessType,
-                                accessType.shared.flags & ~FunctionTypeFlags.StaticMethod
+                            accessType = stripStaticMethodFlag(accessType);
+                        } else if (isOverloaded(accessType)) {
+                            // Apply the same StaticMethod-flag removal to each signature of an
+                            // overloaded accessor (e.g. an overloaded property setter).
+                            const overloads = OverloadedType.getOverloads(accessType).map(stripStaticMethodFlag);
+                            const implementation = OverloadedType.getImplementation(accessType);
+                            accessType = OverloadedType.create(
+                                overloads,
+                                implementation && isFunction(implementation)
+                                    ? stripStaticMethodFlag(implementation)
+                                    : implementation
                             );
                         }
 
--- a/packages/pyright-internal/src/analyzer/properties.ts
+++ b/packages/pyright-internal/src/analyzer/properties.ts
@@ -7,6 +7,7 @@
  * Provides type evaluation logic that is specific to properties.
  */
 
+import { appendArray } from '../common/collectionUtils';
 import { DiagnosticAddendum } from '../common/diagnostic';
 import { DiagnosticRule } from '../common/diagnosticRules';
 import { LocAddendum, LocMessage } from '../localization/localize';
@@ -30,6 +31,7 @@ import {
     isClass,
     isFunction,
     isInstantiableClass,
+    isOverloaded,
     isTypeSame,
     isTypeVar,
     ModuleType,
@@ -192,14 +194,18 @@ export function clonePropertyWithSetter(
     // Update the __get__ and __delete__ methods if present.
     updateGetSetDelMethodForClonedProperty(evaluator, propertyObject);
 
+    // Combine this setter with any overloaded setters accumulated from previous
+    // declarations of this property. This supports overloads on property setters.
+    const combinedSetter = combineSetterOverloads(classType.priv.fsetInfo?.methodType, fset);
+
     // Fill in the new fset method.
     propertyObject.priv.fsetInfo = {
-        methodType: fset,
+        methodType: combinedSetter,
         classType: fset.shared.methodClass,
     };
 
     // Fill in the __set__ method.
-    addSetMethodToPropertySymbolTable(evaluator, propertyObject, fset);
+    addSetMethodToPropertySymbolTable(evaluator, propertyObject, combinedSetter);
 
     // Fill in the getter, setter and deleter methods.
     addDecoratorMethodsToPropertySymbolTable(propertyObject);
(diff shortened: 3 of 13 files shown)


packages/pyright-internal/src/analyzer/properties.ts
```diff
@@ -434,7 +499,7 @@ function updateGetSetDelMethodForClonedProperty(evaluator: TypeEvaluator, proper
     }
 
     const fsetInfo = propertyObject.priv.fsetInfo;
-    if (fsetInfo && isFunction(fsetInfo.methodType)) {
+    if (fsetInfo && (isFunction(fsetInfo.methodType) || isOverloaded(fsetInfo.methodType))) {
         addSetMethodToPropertySymbolTable(evaluator, propertyObject, fsetInfo.methodType);
     }
```
rchiodo, line 523: Copilot generated:
The Skeptic, Architect, and Advocate all flagged this: `assignProperty`'s `getFunction` return type was widened to `FunctionType | OverloadedType | undefined`, but the body still gates on `isFunction(...)`. Two consequences for the *new* overloaded-setter feature: (1) when an overloaded-setter property is the assignment **source**, `!isFunction(srcAccessType)` is true → the `missingSetter` branch fires and reports a **false-positive** on valid code (e.g. a class with an overloaded setter assigned to a `Protocol` with a plain setter); (2) when an overloaded setter is the **dest**, `isFunction(destAccessType)` is false → setter compatibility is silently skipped (soundness hole). Handle the `OverloadedType` case (iterate `OverloadedType.getOverloads(...)` or route through `assignType`, which understands overloads) rather than guarding it out. [unverified] runtime — type-guard semantics and the call path (`protocols.ts` protocol matching, `typeEvaluator.ts` property assignment) are structurally confirmed, but not executed; a sample with a Protocol plain setter + class overloaded setter would confirm.

[verified]

packages/pyright-internal/src/analyzer/properties.ts
```diff
@@ -343,10 +349,69 @@ function addGetMethodToPropertySymbolTable(evaluator: TypeEvaluator, propertyObj
     fields.set('__get__', getSymbol);
```
rchiodo, line 349: Copilot generated:
Minor edge cases (Skeptic/Architect/Advocate, all Low): (1) `setMethod = setOverloads.length > 1 ? ... : setOverloads[0]` assumes `getOverloads(fset)` is non-empty — safe for current callers but unguarded; (2) a lone `@overload` setter with no implementation stays a `FunctionType` (Overloaded flag set) and falls into the non-overloaded `__set__` branch, silently losing overload intent (invalid user program anyway); (3) in `combineSetterOverloads`, when `prevSetter` is already an `OverloadedType` and another `@overload` arrives, the prior implementation is dropped from `prevOverloads` (only matters for malformed multi-implementation source). All benign today; a one-line comment noting these assumptions would help future maintainers.

[verified]

packages/pyright-internal/src/analyzer/checker.ts
```diff
@@ -6987,7 +6987,7 @@ export class Checker extends ParseTreeWalker {
         overrideSymbol: Symbol,
```
rchiodo, line 6991: Copilot generated:
The override-compatibility paths (`_validatePropertyOverride` ~7010/7050 and `_validateMultipleInheritancePropertyOverride` ~6070/6112) `return`/skip when the accessor is not `isFunction(...)`. An **overloaded base-class setter** is an `OverloadedType`, so an incompatible subclass override of an overloaded setter is never flagged (missed `reportIncompatibleMethodOverride`). Either iterate the overloads or add a comment + negative test documenting that overloaded-setter override checking is intentionally not yet performed. Note: these two `propMethodInfo` arrays were left typed `FunctionType | undefined` even though `methodType` is now wider, so the callbacks return an `OverloadedType` typed as `FunctionType` — the annotation is now lying (the `isFunction` guard saves it at runtime). [unverified] runtime; structurally confirmed.

[verified]

packages/pyright-internal/src/tests/samples/property19.py
```diff
@@ -0,0 +1,39 @@
...
+
+    @t1.setter
+    def t1(self, value: str) -> None: ...
+
+    @t1.deleter
+    def t1(self) -> None: ...
+
+
+a = A()
+
```
rchiodo, line 26: Copilot generated:
Rules violation (`tests-no-partial-asserts`): `reveal_type(a.t1)  # int` — the `# int` is a comment, not an assertion, so the framework never verifies the resolved type. Sibling `property18.py:22/44` uses `reveal_type(expr, expected_text="int")`. Change to `reveal_type(a.t1, expected_text="int")`. This also hardens the test, which the Advocate noted currently only asserts an error count.

[verified]

packages/pyright-internal/src/analyzer/packageTypeVerifier.ts
```diff
@@ -876,7 +876,7 @@ export class PackageTypeVerifier {
             case TypeCategory.Class: {
                 // Properties require special handling.
                 if (TypeBase.isInstance(type) && ClassType.isPropertyClass(type)) {
-                    const propMethodInfo: [string, (c: ClassType) => FunctionType | undefined][] = [
+                    const propMethodInfo: [string, (c: ClassType) => FunctionType | OverloadedType | undefined][] = [
```
rchiodo, line 880: Copilot generated:
For an overloaded setter, the `isFunction(accessType)` branch (~895) that strips the `StaticMethod` flag to avoid a spurious "self not annotated" is skipped, yet the raw `OverloadedType` is still passed to `_getSymbolTypeKnownStatus`. This can produce spurious `verifytypes`/py.typed completeness noise for overloaded setters. Low priority (only affects `verifytypes`), but worth handling or documenting. [unverified] runtime; structurally confirmed.

[verified]

packages/pyright-internal/src/tests/samples/property19.py
rchiodo, line 26: 📍 packages/pyright-internal/src/tests/samples/property19.py:26
Review-rule violation (`tests-no-partial-asserts.md`): `reveal_type(a.t1)  # int` uses a prose comment that the test framework never validates, so this can pass even if the inferred type is wrong. Every other `reveal_type` in this suite (e.g. `property16.py:29`) uses the validated form. Change to `reveal_type(a.t1, expected_text="int")`. This is a confirmed rule violation and must be fixed; it cannot be downgraded.

[verified]

packages/pyright-internal/src/analyzer/properties.ts
```diff
@@ -479,7 +544,7 @@ export function assignProperty(
     const destObjectToBind = ClassType.cloneAsInstance(destClass);
```
rchiodo, line 563: 📍 packages/pyright-internal/src/analyzer/properties.ts (assignProperty)
The Skeptic and Advocate both flag a soundness hole: the `getFunction` annotation was widened to `FunctionType | OverloadedType`, but the `assignProperty` body still narrows with `isFunction(...)`. When the dest property has an overloaded setter the setter-compat check is skipped entirely (false negative); when the src has an overloaded setter but the dest's is plain, the `!isFunction(srcAccessType)` branch reports `missingSetter` even though a setter exists (false positive). Overloaded accessors should be specialized/bound and compared like the single-function case. [unverified] (structurally confirmed; behavioral effect not test-run).

[verified]

rchiodo:
GitHub cannot anchor PR review comments to unchanged lines in the diff. Falling back to a general PR comment for [packages/pyright-internal/src/analyzer/checker.ts:L6075](https://github.com/microsoft/pyright/blob/a42d4a5d9e99c07bffc4fafb09c68206b76733f5/packages/pyright-internal/src/analyzer/checker.ts#L6075).

📍 packages/pyright-internal/src/analyzer/checker.ts (_validatePropertyOverride / _validateMultipleInheritancePropertyOverride)
The added `isFunction(...)` guards make an `OverloadedType` setter fail the check, so property override validation is silently skipped for overloaded setters — a subclass that incompatibly overrides (or is overridden by) an overloaded-setter property won't get `reportIncompatibleMethodOverride`. Handle the overloaded case or fall back to the implementation signature, and add a test covering override of an overloaded-setter property (currently untested). [unverified] (structurally confirmed).

[verified]

packages/pyright-internal/src/analyzer/properties.ts
```diff
@@ -343,10 +349,69 @@ function addGetMethodToPropertySymbolTable(evaluator: TypeEvaluator, propertyObj
...
+        // The new setter is an overload. Append it to the accumulated overloads.
+        const overloads = [...prevOverloads, newSetter];
+        return overloads.length > 1 ? OverloadedType.create(overloads) : newSetter;
+    }
+
+    // The new setter is not an overload, so it's either a plain setter or the
+    // implementation for a set of setter overloads. If there are accumulated
+    // overloads, treat it as the implementation; otherwise it's a plain setter.
+    if (prevOverloads.length > 0) {
+        return OverloadedType.create(prevOverloads, newSetter);
```
rchiodo: 📍 packages/pyright-internal/src/analyzer/properties.ts (combineSetterOverloads)
The Architect notes this is a parallel reimplementation of the canonical `addOverloadsToFunctionType` (`decorators.ts:495`) and omits two things that path guarantees: (1) PEP 702 implementation-level `@deprecated` propagation — if the setter *implementation* is `@deprecated`, all overloads should inherit it, but this path only copies each overload's own `deprecatedMessage`; and (2) docstring propagation. `property20.py` only deprecates an overload, so the implementation-deprecation gap is untested. Recommend extracting and reusing the propagation block so overloaded setters behave identically to overloaded methods, plus a sample that deprecates the setter implementation. [unverified].

[verified]

rchiodo:
Solid core change for overloaded setters. The main gap is that widening `PropertyMethodInfo.methodType` to `FunctionType | OverloadedType` was not fully threaded through the consumers: `assignProperty` and the property-override loops still narrow with `isFunction(...)`, so overloaded setters get incorrect/skipped diagnostics. Worth completing those sites (and adding override coverage) plus tightening the new test's `reveal_type` to use `expected_text=`.

packages/pyright-internal/src/analyzer/properties.ts
rchiodo: Fixed in 9795fff. assignProperty now handles OverloadedType accessors: the guard accepts function or overloaded accessors, and bindFunctionToClassOrObject + assignType (both understand overloads) do the comparison. This closes the false-positive missingSetter when the source has an overloaded setter, and the silently-skipped setter-compat check when the dest setter is overloaded.

packages/pyright-internal/src/analyzer/properties.ts
rchiodo: Addressed in 9795fff with clarifying comments. (1) addSetMethodToPropertySymbolTable now documents that combineSetterOverloads only yields an OverloadedType when there is at least one overload, so setOverloads[0] is safe. (2)/(3) The lone-overload and multi-implementation cases are noted as benign/invalid-program behavior in combineSetterOverloads.

packages/pyright-internal/src/analyzer/checker.ts
rchiodo: Partially addressed in 9795fff. The propMethodInfo arrays are already typed FunctionType | OverloadedType | undefined, so there is no lying annotation. For the override-compatibility paths I added comments documenting that overloaded accessors are intentionally not yet validated there and the isFunction guard skips them (the documented-limitation option you offered). Unlike assignProperty (protocol matching), the override path needs declaration/related-info plumbing that does not generalize cleanly to OverloadedType, so I kept that out of scope to avoid risk.

packages/pyright-internal/src/tests/samples/property19.py
rchiodo: Fixed in 9795fff -- now reveal_type(a.t1, expected_text="int") so the inferred type is actually validated.

packages/pyright-internal/src/analyzer/packageTypeVerifier.ts
rchiodo: Fixed in 9795fff. packageTypeVerifier now strips the StaticMethod flag from each signature of an overloaded accessor (overloads + implementation), not just single-function accessors, avoiding spurious verifytypes self noise for overloaded setters.

packages/pyright-internal/src/tests/samples/property19.py
rchiodo: Fixed in 9795fff -- changed to reveal_type(a.t1, expected_text="int") so the framework validates the inferred type.

packages/pyright-internal/src/analyzer/properties.ts
rchiodo: Fixed in 9795fff. assignProperty now handles overloaded accessors (function or OverloadedType) and routes them through bindFunctionToClassOrObject + assignType, which understand overloads. This removes the false negative (dest overloaded setter skipped) and the false positive (src overloaded setter reported as missingSetter).

packages/pyright-internal/src/analyzer/properties.ts
rchiodo: Addressed in 9795fff. combineSetterOverloads now propagates an @deprecated setter *implementation* to all overloads per PEP 702 (mirroring addOverloadsToFunctionType), and property21.py covers the implementation-deprecation case. Docstring propagation is intentionally not mirrored: the synthesized __set__ descriptor does not surface the setter docstring, so it would have no observable effect. I kept this as a focused inline change rather than extracting the shared block to avoid threading the property-specific synthesis through the generic helper.

packages/pyright-internal/src/analyzer/properties.ts
```diff
@@ -343,10 +349,88 @@ function addGetMethodToPropertySymbolTable(evaluator: TypeEvaluator, propertyObj
...
+    } else {
+        setMethod = createSetMethodFromSetter(evaluator, fset, /* asOverload */ false);
+    }
+
+    const setSymbol = Symbol.createWithType(SymbolFlags.ClassMember, setMethod);
+    fields.set('__set__', setSymbol);
+}
+
+function createSetMethodFromSetter(evaluator: TypeEvaluator, fset: FunctionType, asOverload: boolean): FunctionType {
+    const setFunction = FunctionType.createSynthesizedInstance(
```
heejaechang, line 430: Edge case: a single `@overload` setter followed by a non-overload implementation produces `OverloadedType.create([oneOverload], impl)`, and `addSetMethodToPropertySymbolTable`'s `setOverloads.length > 1 ? ... : setOverloads[0]` then emits `__set__` as a lone `FunctionType` carrying `FunctionTypeFlags.Overloaded` but not wrapped in an `OverloadedType` and with no implementation. The overload-resolution path may not expect this shape. No sample exercises exactly one overload + impl. Consider adding a `property*.py` with one `@overload` setter plus an implementation and asserting the diagnostic count, or always wrapping in `OverloadedType` for consistency.

packages/pyright-internal/src/analyzer/checker.ts
```diff
@@ -7007,6 +7011,10 @@ export class Checker extends ParseTreeWalker {
                     this._evaluator.getTypeClassType()
                 );
 
+                // Overloaded accessors (e.g. overloaded property setters) are
+                // represented as an OverloadedType rather than a FunctionType.
```
heejaechang, line 7015: Previously `methodType` was always a `FunctionType`, so this branch always ran; now an overloaded base-class accessor causes the missing-method and incompatible-override checks to be bypassed (`if (!isFunction(baseClassMethodType)) return;`, and the parallel guard at ~6067). This is a deliberate, documented limitation (the inline comments are good) and net-new capability, so there's no regression — but it is a real reduction in `reportIncompatibleMethodOverride` coverage with no test. Acceptable to ship; consider a tracking note/test for follow-up override-compat support on overloaded accessors.

heejaechang:
Solid implementation with strong test design (the dual-run reportDeprecated pattern is nice). Two non-blocking notes below worth a look; otherwise this looks good to merge.

heejaechang (approved):
Approved via Review Center.

packages/pyright-internal/src/analyzer/checker.ts
rchiodo: Added a documented-limitation regression test in e5a30a5 (`property23.py`): a base class with an overloaded setter and a subclass that overrides it with an incompatible (`str`) setter, run with `reportIncompatibleMethodOverride = error`. It asserts 0 errors, locking in the current behavior that override-compatibility checking is skipped for overloaded accessors, and the sample/test comments call out that the expectation must be updated when override support for overloaded accessors is added. This gives us the tracking test you suggested so the limitation is visible and the gap is flagged the moment it's closed.

packages/pyright-internal/src/analyzer/properties.ts
rchiodo: Addressed in e5a30a5 by adding a regression test (`property22.py`) for the single-`@overload`-setter-plus-implementation case. I verified the current shape is sound: `addSetMethodToPropertySymbolTable` synthesizes `__set__` as a lone `FunctionType` with the `Overloaded` flag, and overload resolution handles it correctly — `a.t1 = 1` (matches the single overload) is accepted, `a.t1 = "str"` (matches only the implementation) is reported, and `reveal_type(a.t1)` is `int`. The test asserts exactly 1 error, locking in that behavior. I kept the existing shape rather than always wrapping in `OverloadedType`, since wrapping a single overload with no implementation would itself be an atypical `OverloadedType` shape and the test confirms the current path is correct.

PR #11500, opened 2026-06-15
Fix #11475: Cannot override a class's callable variable with a method

### LINKED ISSUE #11475: Cannot override a class's callable variable with a method

rchiodo:
## Description

Pylance/Pyright incorrectly flagged an error when a subclass overrode a base class's callable **variable** (e.g. `x: Callable[[], None]`) with a real **method**. Other type checkers allow this, because when you access the override method on an instance, its `self` parameter is bound, leaving a signature that matches the base callable attribute.

The base-class member here is a plain `Callable` attribute, not a method, so it has no implicit `self`. The override, being a method, *does* declare `self`. The override check compared the unbound method signature (which still has `self`) against the base callable, producing a spurious `reportIncompatibleMethodOverride` error. The fix is to bind the override method to the class before comparing, but only when the base member is actually a callable variable.

## Implementation

In `checker.ts`, within the override-validation path for function-typed members:

- Detect when the base member is a **callable variable** (all declarations are `DeclarationType.Variable`) and the override is a **method** (`DeclarationType.Function`).
- In that case, bind the override method's `self` via `bindFunctionToClassOrObject` and use the bound type for `validateOverrideMethod`.
- All other cases are unchanged, so normal method-vs-method override checking still applies.

## Testing

Added `methodOverride7.py` sample plus a `MethodOverride7` test in `typeEvaluator3.test.ts` (with `reportIncompatibleMethodOverride = 'error'`):

- `class B(A)` overrides callable attributes `hello` / `cb` with compatible methods — no error.
- `class C(A)` overrides `hello` with an incompatible signature (extra required parameter) — still produces exactly **1** error, confirming the bound comparison continues to catch genuine mismatches.

Ran the targeted Pyright type-evaluator tests to confirm the new case passes and the negative case still errors.

## Addresses

Fixes https://github.com/microsoft/pylance-release/issues/11475

---

### DIFF
--- a/packages/pyright-internal/src/analyzer/checker.ts
+++ b/packages/pyright-internal/src/analyzer/checker.ts
@@ -5862,7 +5862,11 @@ export class Checker extends ParseTreeWalker {
                 // If one base defines the member as a plain callable variable and a
                 // sibling base overrides it with a real method, bind the method's "self"
                 // so the two are compared as plain callables (mirrors the single-base
-                // override handling in `_validateBaseClassOverride`).
+                // override handling in `_validateBaseClassOverride`). This is
+                // load-bearing for parametered callables (e.g. `Callable[[int], None]`
+                // vs `def cb(self, value: int)`): without binding, the method's extra
+                // "self" causes a spurious arity mismatch. See the `DiamondParam*`
+                // samples in methodOverride7.py.
                 const childClassSelf = ClassType.cloneAsInstance(
                     selfSpecializeClass(childClassType, { useBoundTypeVars: true })
                 );
@@ -6631,24 +6635,48 @@ export class Checker extends ParseTreeWalker {
             childClassType
         );
 
-        if (boundOverrideType && isFunction(baseType) && isFunction(boundOverrideType)) {
+        if (!boundOverrideType) {
+            return { baseType, overrideType };
+        }
+
+        // Mark both the base callable and the bound override as static so the
+        // override comparison does not skip the first parameter as an unbound
+        // "self" (see the `i === 0` self/cls skip in typeEvaluator's
+        // `validateOverrideMethodInternal`). Without this, the bound override
+        // keeps its instance-method flag and its first real parameter would be
+        // silently dropped during the comparison.
+        const staticBaseType = isFunction(baseType) ? this._markFunctionStatic(baseType) : baseType;
+
+        if (isFunction(boundOverrideType)) {
             return {
-                baseType: FunctionType.cloneWithNewFlags(
-                    baseType,
-                    baseType.shared.flags | FunctionTypeFlags.StaticMethod
-                ),
-                overrideType: FunctionType.cloneWithNewFlags(
-                    boundOverrideType,
-                    boundOverrideType.shared.flags | FunctionTypeFlags.StaticMethod
-                ),
+                baseType: staticBaseType,
+                overrideType: this._markFunctionStatic(boundOverrideType),
             };
         }
 
-        if (boundOverrideType) {
-            return { baseType, overrideType: boundOverrideType };
+        if (isOverloaded(boundOverrideType)) {
+            // An overloaded method binds to an overloaded callable. Apply the
+            // same static normalization to every overload (and the
+            // implementation, if present) so an incompatible first parameter in
+            // an overloaded override is still caught rather than skipped as
+            // "self".
+            const staticOverloads = OverloadedType.getOverloads(boundOverrideType).map((overload) =>
+                this._markFunctionStatic(overload)
+            );
+            const impl = OverloadedType.getImplementation(boundOverrideType);
+            const staticImpl = impl && isFunction(impl) ? this._markFunctionStatic(impl) : impl;
+
+            return {
+                baseType: staticBaseType,
+                overrideType: OverloadedType.create(staticOverloads, staticImpl),
+            };
         }
 
-        return { baseType, overrideType };
+        return { baseType, overrideType: boundOverrideType };
+    }
+
+    private _markFunctionStatic(functionType: FunctionType): FunctionType {
+        return FunctionType.cloneWithNewFlags(functionType, functionType.shared.flags | FunctionTypeFlags.StaticMethod);
     }
 
     private _validateBaseClassOverride(
--- a/packages/pyright-internal/src/analyzer/checker.ts
+++ b/packages/pyright-internal/src/analyzer/checker.ts
@@ -6600,6 +6600,12 @@ export class Checker extends ParseTreeWalker {
     // not skip the first parameter as an unbound "self". All other cases are returned
     // unchanged. This matches the behavior of other type checkers, which allow overriding
     // a callable attribute with a method.
+    //
+    // Note: this validates only the read direction (the bound override is compatible with
+    // the base callable). A method is effectively read-only, so this intentionally does not
+    // attempt to preserve assignability of the mutable callable attribute; that asymmetry
+    // matches other type checkers and does not trigger `reportIncompatibleVariableOverride`
+    // because the override is a function, not a variable.
     private _getCallableVariableOverrideComparison(
         baseSymbol: Symbol,
         overrideSymbol: Symbol,
--- a/packages/pyright-internal/src/analyzer/checker.ts
+++ b/packages/pyright-internal/src/analyzer/checker.ts
@@ -6656,6 +6656,19 @@ export class Checker extends ParseTreeWalker {
         if (isFunctionOrOverloaded(baseType)) {
             const diagAddendum = new DiagnosticAddendum();
 
+            // Determine whether the base class member is a plain callable variable
+            // (e.g. `x: Callable[..., None]`) rather than an actual method. In that
+            // case, accessing the override method on an instance binds its "self"
+            // parameter, so we bind it here before comparing signatures. This matches
+            // the behavior of other type checkers, which allow overriding a callable
+            // attribute with a method.
+            const baseDecls = baseClassAndSymbol.symbol.getDeclarations();
+            const baseIsCallableVariable =
+                baseDecls.length > 0 && baseDecls.every((decl) => decl.type === DeclarationType.Variable);
+            const overrideIsMethod = overrideSymbol
+                .getDeclarations()
+                .some((decl) => decl.type === DeclarationType.Function);
+
             // Don't check certain magic functions or private symbols.
             // Also, skip this check if the class is a TypedDict. The methods for a TypedDict
             // are synthesized, and they can result in many overloads. We assume they
@@ -6674,10 +6687,22 @@ export class Checker extends ParseTreeWalker {
                 // false positives.
                 const enforceParamNameMatch = !SymbolNameUtils.isDunderName(memberName);
 
+                let overrideTypeForComparison: FunctionType | OverloadedType = overrideType;
+                if (baseIsCallableVariable && overrideIsMethod) {
+                    const boundOverrideType = this._evaluator.bindFunctionToClassOrObject(
+                        childClassSelf,
+                        overrideType,
+                        childClassType
+                    );
+                    if (boundOverrideType) {
+                        overrideTypeForComparison = boundOverrideType;
+                    }
+                }
+
                 if (
                     this._evaluator.validateOverrideMethod(
                         baseType,
-                        overrideType,
+                        overrideTypeForComparison,
                         childClassType,
                         diagAddendum,
                         enforceParamNameMatch
--- a/packages/pyright-internal/src/analyzer/checker.ts
+++ b/packages/pyright-internal/src/analyzer/checker.ts
@@ -140,6 +140,7 @@ import {
     EnumLiteral,
     FunctionParam,
     FunctionType,
+    FunctionTypeFlags,
     ModuleType,
     OverloadedType,
     Type,
@@ -5858,10 +5859,27 @@ export class Checker extends ParseTreeWalker {
             const diagAddendum = new DiagnosticAddendum();
 
             if (isFunctionOrOverloaded(overrideType)) {
-                if (
-                    !this._evaluator.validateOverrideMethod(
+                // If one base defines the member as a plain callable variable and a
+                // sibling base overrides it with a real method, bind the method's "self"
+                // so the two are compared as plain callables (mirrors the single-base
+                // override handling in `_validateBaseClassOverride`).
+                const childClassSelf = ClassType.cloneAsInstance(
+                    selfSpecializeClass(childClassType, { useBoundTypeVars: true })
+                );
+                const { baseType: overriddenTypeForComparison, overrideType: overrideTypeForComparison } =
+                    this._getCallableVariableOverrideComparison(
+                        overriddenClassAndSymbol.symbol,
+                        overrideClassAndSymbol.symbol,
                         overriddenType,
                         overrideType,
+                        childClassType,
+                        childClassSelf
+                    );
+
+                if (
+                    !this._evaluator.validateOverrideMethod(
+                        overriddenTypeForComparison,
+                        overrideTypeForComparison,
                         /* baseClass */ undefined,
                         diagAddendum,
                         /* enforceParamNameMatch */ true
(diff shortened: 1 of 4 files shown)


packages/pyright-internal/src/analyzer/checker.ts
```diff
@@ -6674,10 +6687,22 @@ export class Checker extends ParseTreeWalker {
                 // false positives.
```
rchiodo, line 6779: Copilot generated:
The Skeptic verified (before/after analyzer runs) a soundness hole: the bound override still reports `isInstanceMethod() === true`, so `validateOverrideMethodInternal` (typeEvaluator.ts:~27891) keeps skipping index 0 — which after binding is the *first real parameter*. Probe: `class A: cb: Callable[[int], None]` overridden by `def cb(self, value: str) -> None` now produces **0 errors** (was 1) because the `str`-vs-`int` first-param mismatch is silently dropped. Fix: after binding, clear the instance-method flag on the bound override (or use a comparison path that doesn't skip index 0), then add a negative test with an incompatible first-param *type*.

[verified]

rchiodo, line 6779: Copilot generated:
Single-source-of-truth gap (Architect): the bind-before-compare handling is inlined here but the sibling multiple-inheritance path `_validateMultipleInheritanceOverride` calls `validateOverrideMethod` at checker.ts:5862 with raw (unbound) types. A diamond like `class A: hello: Callable[[], None]` + `class B: def hello(self)` + `class C(A, B): ...` would still emit the spurious `reportIncompatibleMethodOverride`. Either extend the handling to that site, push the asymmetry handling down into `validateOverrideMethod` so all callers benefit, or consciously scope it out with a documenting test. [unverified] runtime repro — confirmable by adding the diamond sample.

[verified]

packages/pyright-internal/src/analyzer/checker.ts
```diff
@@ -6656,6 +6656,19 @@ export class Checker extends ParseTreeWalker {
         if (isFunctionOrOverloaded(baseType)) {
             const diagAddendum = new DiagnosticAddendum();
 
+            // Determine whether the base class member is a plain callable variable
+            // (e.g. `x: Callable[..., None]`) rather than an actual method. In that
```
rchiodo: Copilot generated:
Minor (Architect): `baseDecls`, `baseIsCallableVariable`, and `overrideIsMethod` are computed before the `_isMethodExemptFromLsp`/private-name/TypedDict early-returns and before the `isFunctionOrOverloaded(overrideType)` guard, so exempt/private/TypedDict members pay for `getDeclarations()` scans they never use. Move these inside the `if (isFunctionOrOverloaded(overrideType))` block where they're consumed.

[verified]

packages/pyright-internal/src/tests/samples/methodOverride7.py
```diff
@@ -0,0 +1,32 @@
...
+class B(A):
+    @override
+    def hello(self) -> None:
+        print("hi")
+
+    @override
+    def cb(self, value: int) -> None:
+        print(value)
+
+
```
rchiodo, line 32: Copilot generated:
The Skeptic and Rules reviewers converge: this test is weaker than it looks. `validateResults(analysisResults, 1)` checks only the error *count* — not which line or message — and the single negative case (`class C`) exercises only a *parameter-count* mismatch, the one comparison path the fix doesn't break. Add a negative case with an incompatible first-parameter *type* (e.g. `def cb(self, value: str)` overriding `Callable[[int], None]`), add an error-location marker, and use `validateResultsFromFile` so the test verifies location, not just count. This negative-type case would have caught the soundness hole flagged above.

[verified]

packages/pyright-internal/src/tests/typeEvaluator3.test.ts
```diff
@@ -1007,6 +1007,14 @@ test('MethodOverride6', () => {
     TestUtils.validateResults(analysisResults2, 3);
 });
 
+test('MethodOverride7', () => {
```
rchiodo, line 1010: Copilot generated:
A pyright-internal test is correct for an upstream microsoft/pyright PR. However, if this lands in pyrx's vendored copy, repo policy prefers a Pylance-harness regression test (`packages/pylance-internal/src/tests/`) for a pylance-release issue, with the pyright-internal test as additional coverage. Flagged by the Advocate, Architect, and Rules reviewers as low priority — confirm placement matches the target repo.

[verified]

packages/pyright-internal/src/analyzer/checker.ts
```diff
@@ -6656,6 +6656,19 @@ export class Checker extends ParseTreeWalker {
...
+            // parameter, so we bind it here before comparing signatures. This matches
+            // the behavior of other type checkers, which allow overriding a callable
+            // attribute with a method.
+            const baseDecls = baseClassAndSymbol.symbol.getDeclarations();
+            const baseIsCallableVariable =
+                baseDecls.length > 0 && baseDecls.every((decl) => decl.type === DeclarationType.Variable);
+            const overrideIsMethod = overrideSymbol
+                .getDeclarations()
+                .some((decl) => decl.type === DeclarationType.Function);
+
```
rchiodo: 📍 packages/pyright-internal/src/analyzer/checker.ts:6663
`baseDecls`, `baseIsCallableVariable`, and `overrideIsMethod` are computed unconditionally on every function-typed override — including the exempt/private/TypedDict members that `return` early and the common method-vs-method path. Move these scans inside (or just before) the `baseIsCallableVariable && overrideIsMethod` guard so the common path pays nothing extra.

packages/pyright-internal/src/tests/samples/methodOverride7.py
```diff
@@ -0,0 +1,32 @@
...
+class C(A):
+    # This should generate an error because the bound override signature
+    # is incompatible with the base callable variable.
+    @override
+    def hello(self, extra: int) -> None:
+        print("hi")
```
rchiodo, line 38: 📍 packages/pyright-internal/src/tests/samples/methodOverride7.py:1
The negative case (`class C`) only exercises an extra required parameter (arity). The bound comparison should also catch incompatible return types (`Callable[[], int]` overridden by `-> str`) and param types (`Callable[[int], None]` overridden by `(value: str)`). Adding at least one return-type and one param-type negative case would guard against a future weakening of `validateOverrideMethod` in the variance dimension going uncaught.

packages/pyright-internal/src/analyzer/checker.ts
```diff
@@ -6674,10 +6687,22 @@ export class Checker extends ParseTreeWalker {
...
+                let overrideTypeForComparison: FunctionType | OverloadedType = overrideType;
+                if (baseIsCallableVariable && overrideIsMethod) {
+                    const boundOverrideType = this._evaluator.bindFunctionToClassOrObject(
+                        childClassSelf,
+                        overrideType,
+                        childClassType
+                    );
+                    if (boundOverrideType) {
+                        overrideTypeForComparison = boundOverrideType;
+                    }
```
rchiodo: 📍 packages/pyright-internal/src/analyzer/checker.ts:6690
The base member is a mutable callable variable, so `a: A = B(); a.hello = lambda: None` is valid against `A` but not against `B` (a method is effectively read-only). This fix makes the read direction compatible but is silent on the write direction — which may be why Pyright flagged it. Allowing the override seems right (other checkers do too), but please confirm `reportIncompatibleVariableOverride` doesn't independently fire here, and consider a brief note that the read-direction match is the intended contract.

rchiodo:
Looks good overall — the bound-comparison approach is reasonable and matches other type checkers. A few non-blocking suggestions: hoist the new declaration scans inside the callable-variable guard to avoid extra work on the common override path, and broaden the negative test beyond arity (return-type and param-type mismatches).

packages/pyright-internal/src/analyzer/checker.ts
rchiodo: Fixed. Verified the hole empirically — a `def cb(self, value: str)` override of `cb: Callable[[int], None]` previously produced **0** errors. After binding, the override still reported `isInstanceMethod()` as true, so `validateOverrideMethodInternal` skipped index 0, which post-binding is the first *real* parameter. The fix marks both the base callable and the bound override as static before comparison, so index 0 is no longer skipped. Added a param-type negative case (`class D`) that now catches it.

rchiodo: Fixed. Confirmed the diamond repro: `class Mixin: def hello(self)` + `class A: hello: Callable[[], None]` + `class DiamondBA(Mixin, A)` did emit a spurious `reportIncompatibleMethodOverride`. Extended the same bind-before-compare handling to `_validateMultipleInheritanceOverride` via a shared helper (`_getCallableVariableOverrideComparison`). Added `DiamondAB`/`DiamondBA` regression coverage in both inheritance orders (no error).

packages/pyright-internal/src/analyzer/checker.ts
rchiodo: Fixed. The declaration scans now live in the shared helper, invoked only inside the `isFunctionOrOverloaded(overrideType)` block (after the exempt/private/TypedDict early-returns), so the common method-vs-method path and exempt members pay nothing extra.

packages/pyright-internal/src/tests/samples/methodOverride7.py
rchiodo: Fixed. Added negative cases for an incompatible first-parameter *type* (`class D`: `str` vs `Callable[[int], None]`) and an incompatible *return* type (`class E`: `-> str` vs `Callable[[], int]`), alongside the arity case (`class C`). Verified each error lands on the expected class line.

packages/pyright-internal/src/tests/typeEvaluator3.test.ts
rchiodo: This PR targets `microsoft/pyright` directly and the change lives in `packages/pyright-internal/src/analyzer/checker.ts`, so the pyright-internal test is the correct home here. No move needed.

packages/pyright-internal/src/analyzer/checker.ts
rchiodo: Fixed (same underlying point as the related thread). The declaration scans are now computed inside the shared `_getCallableVariableOverrideComparison` helper, invoked only inside the `isFunctionOrOverloaded(overrideType)` guard after the exempt/private/TypedDict early-returns.

packages/pyright-internal/src/tests/samples/methodOverride7.py
rchiodo: Fixed. Added return-type (`class E`) and param-type (`class D`) negative cases in addition to the arity case (`class C`), guarding against future weakening of `validateOverrideMethod` in the variance dimensions.

packages/pyright-internal/src/analyzer/checker.ts
rchiodo: Confirmed `reportIncompatibleVariableOverride` does not independently fire: with both `reportIncompatibleMethodOverride` and `reportIncompatibleVariableOverride` set to `error`, only the three intended method-override errors are reported and the valid override (`class B`) stays clean. The override is a method (effectively read-only), so the read-direction match is the intended contract — added a note documenting this on the helper.

packages/pyright-internal/src/analyzer/checker.ts
```diff
@@ -6574,6 +6592,65 @@ export class Checker extends ParseTreeWalker {
...
+                    boundOverrideType,
+                    boundOverrideType.shared.flags | FunctionTypeFlags.StaticMethod
+                ),
+            };
+        }
+
+        if (boundOverrideType) {
+            return { baseType, overrideType: boundOverrideType };
+        }
+
```
heejaechang, line 6674: 📍 packages/pyright-internal/src/analyzer/checker.ts:6644
[unverified] Overloaded-override asymmetry (raised independently by the Skeptic and Architect). When `bindFunctionToClassOrObject` returns an `OverloadedType` (override is an overloaded method over a callable variable), this fallback returns `{ baseType (unmarked), overrideType: boundOverrideType }` and marks *neither* side `StaticMethod`. The single-function branch deliberately marks both static to defeat the `i===0` self-skip in `validateOverrideMethodInternal` (`typeEvaluator.ts:~27923`); without it, each bound overload likely keeps its `InstanceMethod` flag and the real first parameter is skipped — a potential false negative for an incompatible overloaded override. Fix: apply the same static normalization in the overloaded branch, or explicitly bail/document overloaded overrides as unsupported. Add a sample overriding `cb: Callable[[int], None]` with an `@overload`ed method whose first param is incompatible and assert it still errors.

[verified]

packages/pyright-internal/src/analyzer/checker.ts
```diff
@@ -5858,10 +5859,27 @@ export class Checker extends ParseTreeWalker {
...
+                    selfSpecializeClass(childClassType, { useBoundTypeVars: true })
+                );
+                const { baseType: overriddenTypeForComparison, overrideType: overrideTypeForComparison } =
+                    this._getCallableVariableOverrideComparison(
+                        overriddenClassAndSymbol.symbol,
+                        overrideClassAndSymbol.symbol,
                         overriddenType,
                         overrideType,
+                        childClassType,
+                        childClassSelf
```
heejaechang, line 5880: 📍 packages/pyright-internal/src/analyzer/checker.ts:5862
The first call-site change (multiple-inheritance path) is not exercised by any test that fails without it. The Advocate verified both `DiamondAB`/`DiamondBA` produce 0 errors with and without this hunk, because every diamond only shares the 0-arg `hello`, which the pre-existing `isCallableAttrOverriddenByInstanceMethod` heuristic already handles. Either add a parametered diamond (one base with `cb: Callable[[int], None]`, another with `def cb(self, value: int)`) that is load-bearing here, or add a comment stating this is intentional parity hardening so a future maintainer doesn't delete it as dead code.

[verified]

packages/pyright-internal/src/analyzer/checker.ts
```diff
@@ -6574,6 +6592,65 @@ export class Checker extends ParseTreeWalker {
...
+    // and the override is a real method, accessing the override on an instance binds its
+    // "self" parameter, leaving a signature that should be compared against the base
+    // callable. This helper detects that asymmetry and returns adjusted types: the bound
+    // override and the base callable, both marked static so the override comparison does
+    // not skip the first parameter as an unbound "self". All other cases are returned
+    // unchanged. This matches the behavior of other type checkers, which allow overriding
+    // a callable attribute with a method.
+    //
+    // Note: this validates only the read direction (the bound override is compatible with
+    // the base callable). A method is effectively read-only, so this intentionally does not
```
heejaechang, line 6609: 📍 packages/pyright-internal/src/analyzer/checker.ts:6594
Architectural two-source-of-truth (structurally confirmed by the Architect). The evaluator already special-cases "callable attribute overridden by an instance method" at `typeEvaluator.ts:~27873` (hard-coded to base 0 params / override 1 param), which handles the no-arg cases (`hello`, `ret`) and which this checker-side transform now supersedes/disables for the parametered case (the heuristic requires `!isStaticMethod(base)`, which this PR violates by marking base static). They don't double-count, but the concept now lives in two places that can drift. Consider generalizing/removing the evaluator heuristic, or add a cross-reference comment at each site so the two mechanisms stay in sync.

[verified]

heejaechang:
Looks good overall — the fix correctly binds the override method's self only when the base member is a callable variable. A few non-blocking notes below about overloaded-override handling and test/code-path coverage.

heejaechang (approved):
Approved via Review Center.

packages/pyright-internal/src/analyzer/checker.ts
rchiodo: Fixed. Added a load-bearing parametered diamond: `CallableVarBase` (`cb: Callable[[int], None]`) + `MethodBase` (`def cb(self, value: int)`) inherited in both orders (`DiamondParamVM`/`DiamondParamMV`). Unlike the 0-arg `hello` diamonds, this case is genuinely exercised by the multiple-inheritance hunk — without binding the method's `self`, the comparison sees an extra positional parameter and emits a spurious `reportIncompatibleMethodOverride`; with binding it produces 0 errors (verified). Also added a comment at the call site documenting this so it isn't deleted as dead code.

packages/pyright-internal/src/analyzer/checker.ts
rchiodo: Fixed. The helper now applies the same static normalization to the overloaded branch: when `bindFunctionToClassOrObject` returns an `OverloadedType`, every bound overload (and the implementation, if present) is marked `StaticMethod`, and the base callable is marked static too. Without this, each bound overload kept its `InstanceMethod` flag and `validateOverrideMethodInternal` skipped index 0 (the first real parameter) — verified empirically as a false negative. Added `class F` to methodOverride7.py: an `@overload`ed `cb` whose first param type (`str`) is incompatible with the base `Callable[[int], None]`; it now correctly errors (expected count bumped 3→4).

PR #11501, opened 2026-06-15
Fix #11481: 1.1.410 detects type as module

### LINKED ISSUE #11481: 1.1.410 detects type as module
**Describe the bug**
1.1.410:
```
bug.py
  bug.py:5:7 - error: Module cannot be used as a type (reportGeneralTypeIssues)
  bug.py:5:19 - error: Module is not callable (reportCallIssue)
  bug.py:6:13 - information: Type of "ImageView" is "Module(".ImageView")"
2 errors, 0 warnings, 1 information
```

1.1.409:
```
bug.py
  bug.py:6:13 - information: Type of "ImageView" is "type[ImageView]"
0 errors, 0 warnings, 1 information
```

**Code or Screenshots**
```python

from pyqtgraph import ImageView

view: ImageView = ImageView()
reveal_type(ImageView)
```

```
pyqtgraph==0.14.0
```

running pyright 1.1.410 as a command-line tool via pyright-python.

rchiodo:
## Description

When a package re-exports (via a wildcard `from .sub import *`) a name that is both an implicitly-imported submodule **and** a class of the same name, Pylance 1.1.410 incorrectly resolved the name to the submodule instead of the class. This broke using the name as a type annotation or calling it.

Reproduced the issue with a package whose submodule (`ImageView.py`) defines a class (`ImageView`) re-exported through wildcard imports. The symbol ended up with multiple declarations: an implicit submodule alias and the class. The binder's multipart module-alias handling treated the wildcard re-export as a pure submodule re-export, shadowing the class even though the non-module declaration is the symbol's effective (last) declaration.

## Implementation

In `binder.ts`, the wildcard re-export is now only treated as a pure submodule re-export when the module alias is the symbol's **effective (last)** declaration. When a later non-module declaration (the class) wins, the binder falls through to create a normal alias declaration that resolves to the winning symbol.

```ts
const importedDecls = importedSymbol.getDeclarations();
if (importedDecls[importedDecls.length - 1] !== importedModuleAliasDecl) {
    return false;
}
```

## Testing

Added a fourslash regression test (`import.multipartModuleClassShadow.fourslash.ts`) that sets up a package re-exporting a class whose name matches a submodule. It verifies the name resolves to `type[ImageView]`, can be used as a type annotation and called, with no diagnostics.

Run with:

```
cd packages/pyright/packages/pyright-internal
node node_modules\jest\bin\jest fourSlashRunner.test.ts --runInBand -t import.multipartModuleClassShadow.fourslash
```

## Addresses

Fixes https://github.com/microsoft/pylance-release/issues/11481

---

### DIFF
--- a/packages/pyright-internal/src/analyzer/binder.ts
+++ b/packages/pyright-internal/src/analyzer/binder.ts
@@ -4362,6 +4362,20 @@ export class Binder extends ParseTreeWalker {
             return false;
         }
 
+        // The imported symbol may be both an implicitly-imported submodule and a
+        // class/function/variable of the same name (e.g. a package that re-exports
+        // a class whose name matches a submodule). In that case the non-module
+        // declaration appears later in the declaration list and "wins" when the
+        // symbol is resolved. Only treat this wildcard re-export as a pure submodule
+        // re-export when the module alias is the symbol's effective (last)
+        // declaration; otherwise fall through so a normal alias declaration is
+        // created that resolves to the winning symbol.
+        // See https://github.com/microsoft/pyright/issues/11481.
+        const importedDecls = importedSymbol.getDeclarations();
+        if (importedDecls[importedDecls.length - 1] !== importedModuleAliasDecl) {
+            return false;
+        }
+
         const existingModuleAliasDecl = this._getMultipartModuleAliasDeclaration(
             localSymbol,
             importedModuleAliasDecl.moduleName,
--- a/packages/pyright-internal/src/analyzer/binder.ts
+++ b/packages/pyright-internal/src/analyzer/binder.ts
@@ -4367,10 +4367,20 @@ export class Binder extends ParseTreeWalker {
         // a class whose name matches a submodule). In that case the non-module
         // declaration appears later in the declaration list and "wins" when the
         // symbol is resolved. Only treat this wildcard re-export as a pure submodule
-        // re-export when the module alias is the symbol's effective (last)
-        // declaration; otherwise fall through so a normal alias declaration is
-        // created that resolves to the winning symbol.
-        // See https://github.com/microsoft/pyright/issues/11481.
+        // re-export when the module alias is the symbol's last declaration;
+        // otherwise fall through so a normal alias declaration is created that
+        // resolves to the winning symbol.
+        //
+        // We compare against the raw last declaration (not getLastTypedDeclarationForSymbol)
+        // on purpose: a module alias is a DeclarationType.Alias, which hasTypeForDeclaration
+        // treats as untyped, so it never appears among a symbol's typed declarations.
+        // The evaluator resolves alias symbols (like this one, whose declarations are all
+        // imports) by declaration order, so the last declaration is the relevant "winner"
+        // here. When this guard falls through, the alias created by the caller resolves to
+        // the winning (e.g. class) declaration and intentionally has no submoduleFallback:
+        // the class shadows the submodule, so submodule member access through the
+        // re-exported name is no longer offered. The genuine-submodule case (where the
+        // module alias is the last declaration) keeps its module/submodule behavior.
         const importedDecls = importedSymbol.getDeclarations();
         if (importedDecls[importedDecls.length - 1] !== importedModuleAliasDecl) {
             return false;
--- a/packages/pyright-internal/src/tests/fourslash/import.multipartModuleClassShadow.fourslash.ts
+++ b/packages/pyright-internal/src/tests/fourslash/import.multipartModuleClassShadow.fourslash.ts
@@ -3,7 +3,6 @@
 // This verifies that when a package re-exports (via wildcard) a name that is
 // both an implicitly-imported submodule and a class/symbol of the same name,
 // the class/symbol "wins" rather than being shadowed by the submodule.
-// See https://github.com/microsoft/pyright/issues/11481.
 
 // @filename: pkg/__init__.py
 //// from .sub import *
@@ -22,5 +21,27 @@
 //// # used as a type annotation and called.
 //// view: ImageView = ImageView()
 //// reveal_type(ImageView, expected_text="type[ImageView]")
+
+// Companion case: a pure submodule re-export (no shadowing class) must still
+// resolve as a module so that submodule member access remains available. This
+// locks in that the multipart-alias behavior didn't regress.
+
+// @filename: pkg2/__init__.py
+//// from .sub2 import *
+
+// @filename: pkg2/sub2/__init__.py
+//// from . import mod
+
+// @filename: pkg2/sub2/mod.py
+//// def mod_func() -> None:
+////     pass
+
+// @filename: test2.py
+//// import pkg2
+////
+//// # mod has only a submodule alias declaration, so it stays a module and its
+//// # members remain accessible through the re-exported name.
+//// pkg2.mod.mod_func()
+//// reveal_type(pkg2.mod, expected_text='Module("..mod")')
 
 helper.verifyDiagnostics();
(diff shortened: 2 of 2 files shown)


packages/pyright-internal/src/analyzer/binder.ts
```diff
@@ -4362,6 +4362,20 @@ export class Binder extends ParseTreeWalker {
...
+        // class/function/variable of the same name (e.g. a package that re-exports
+        // a class whose name matches a submodule). In that case the non-module
+        // declaration appears later in the declaration list and "wins" when the
+        // symbol is resolved. Only treat this wildcard re-export as a pure submodule
+        // re-export when the module alias is the symbol's effective (last)
+        // declaration; otherwise fall through so a normal alias declaration is
+        // created that resolves to the winning symbol.
+        // See https://github.com/microsoft/pyright/issues/11481.
+        const importedDecls = importedSymbol.getDeclarations();
+        if (importedDecls[importedDecls.length - 1] !== importedModuleAliasDecl) {
```
rchiodo, line 4385: The guard keys off the raw last element of `getDeclarations()`, but the canonical "effective declaration" used by the evaluator is `getLastTypedDeclarationForSymbol` (over typed declarations only). These coincide for the repro but are two different notions of "winner," creating an implicit binder↔evaluator coupling. Consider reusing the canonical helper (or a clearly-named local predicate), or add a brief comment citing the evaluator's precedence so the two definitions provably agree.

packages/pyright-internal/src/tests/fourslash/import.multipartModuleClassShadow.fourslash.ts
```diff
@@ -0,0 +1,26 @@
...
+
+// @filename: test.py
+//// from pkg import ImageView
+////
+//// # ImageView should resolve to the class, not the submodule, so it can be
+//// # used as a type annotation and called.
+//// view: ImageView = ImageView()
+//// reveal_type(ImageView, expected_text="type[ImageView]")
+
+helper.verifyDiagnostics();
```
rchiodo, line 47: The test asserts the *class wins* direction but not that a *pure submodule re-export* (no shadowing class) still resolves as a module after this guard. A companion assertion in the same fixture would lock in that the original multipart-alias behavior didn't regress.

packages/pyright-internal/src/analyzer/binder.ts
rchiodo, line 4385: When this returns `false`, the caller's fallthrough creates a plain `symbolName` alias with no `submoduleFallback` (unlike the genuine-submodule branch), so member access through the re-exported name to the *submodule's* members is no longer available. This is likely the intended disambiguation (class wins), but it's an untested narrowing of capability — worth a deliberate note or a small test confirming the intent.

packages/pyright-internal/src/analyzer/binder.ts
```diff
@@ -4362,6 +4362,20 @@ export class Binder extends ParseTreeWalker {
...
         }
 
+        // The imported symbol may be both an implicitly-imported submodule and a
+        // class/function/variable of the same name (e.g. a package that re-exports
+        // a class whose name matches a submodule). In that case the non-module
+        // declaration appears later in the declaration list and "wins" when the
+        // symbol is resolved. Only treat this wildcard re-export as a pure submodule
+        // re-export when the module alias is the symbol's effective (last)
+        // declaration; otherwise fall through so a normal alias declaration is
+        // created that resolves to the winning symbol.
```
rchiodo: This repo prefers issue links in PR/commit metadata rather than inline (`docs/pylancewiki/review/issue-references.md`). Consider dropping the `// See https://github.com/.../issues/11481.` line and keeping the rest of the explanatory comment.

packages/pyright-internal/src/tests/fourslash/import.multipartModuleClassShadow.fourslash.ts
```diff
@@ -0,0 +1,26 @@
+/// <reference path="typings/fourslash.d.ts" />
+
+// This verifies that when a package re-exports (via wildcard) a name that is
+// both an implicitly-imported submodule and a class/symbol of the same name,
+// the class/symbol "wins" rather than being shadowed by the submodule.
+// See https://github.com/microsoft/pyright/issues/11481.
+
```
rchiodo, line 6: Same convention note as in binder.ts: prefer keeping the issue link in PR/commit metadata rather than inline. Consider dropping the `// See https://github.com/.../issues/11481.` line while keeping the descriptive comment above it.

rchiodo:
Sound, minimal fix with good regression coverage — approving. A couple of non-blocking follow-ups worth considering: (1) the new guard keys off the raw last `getDeclarations()` entry rather than the evaluator's canonical typed "effective declaration" notion, so a short comment (or reuse of the canonical helper) would lock in that the two agree; (2) consider a companion assertion that a pure submodule re-export (no shadowing class) still resolves as a module, to guard the original multipart-alias behavior. Also note the inline issue-URL comments — this repo prefers issue links in PR/commit metadata rather than in source/tests.

packages/pyright-internal/src/tests/fourslash/import.multipartModuleClassShadow.fourslash.ts
rchiodo: Added a companion fixture (pkg2) in the same test: a pure submodule re-export with no shadowing class. It asserts pkg2.mod still resolves as a module (reveal_type -> Module("..mod")) and that submodule member access (pkg2.mod.mod_func()) remains available, locking in that the multipart-alias behavior did not regress. Pushed in d4788db5f.

packages/pyright-internal/src/analyzer/binder.ts
rchiodo: Done -- dropped the inline // See ...issues/11481 line and kept the explanatory comment (issue link lives in PR/commit metadata). Pushed in d4788db5f.

packages/pyright-internal/src/analyzer/binder.ts
rchiodo: Not switching to getLastTypedDeclarationForSymbol -- that helper would actually break this. A module alias is a DeclarationType.Alias, which hasTypeForDeclaration treats as untyped, so the alias never appears among the symbol's typed declarations; the helper would return undefined (or the class decl) and never equal importedModuleAliasDecl, defeating the genuine-submodule branch. Took the alternative you offered: added a comment explaining we intentionally compare the raw last declaration because the evaluator resolves these all-import alias symbols by declaration order, so the two notions of winner provably agree here. Pushed in d4788db5f.

rchiodo: Good catch -- this is the intended disambiguation (class wins). Documented it explicitly in the binder comment: when the guard falls through, the resulting alias has no submoduleFallback, so submodule member access through the re-exported name is intentionally no longer offered. The new pkg2 companion test also pins the opposite direction (genuine submodule case keeps module/member behavior). Pushed in d4788db5f.

packages/pyright-internal/src/tests/fourslash/import.multipartModuleClassShadow.fourslash.ts
rchiodo: Done -- dropped the inline issue-link line in the test fixture too, kept the descriptive comment. Pushed in d4788db5f.

rgoya:
> ## Addresses
> Fixes https://github.com/microsoft/pylance-release/issues/11481
> 
> _Generated by fix_all_my_issues pipeline_

Looking forward to this fix! Putting link to the public Issue here: https://github.com/microsoft/pyright/issues/11481

bschnurr (approved):

PR #11502, opened 2026-06-15
Fix #11483: Incorrect type inference from context manager __exit__ return type with Generic

### LINKED ISSUE #11483: Incorrect type inference from context manager __exit__ return type with Generic

rchiodo:
## Description

When a generic context manager annotates its `__exit__` / `__aexit__` return type with a `TypeVar` (e.g. `class ContextManager(Generic[T])` where `__exit__` returns `T` bound to `bool | None`), Pyright determined exception-suppression behavior from the **unspecialized declared** return type instead of the solved type argument. This made code-flow narrowing after the `with` block wrong — `reveal_type(foo)` was `Literal[1]` instead of `Literal["str", 1]` for a `bool`-specialized manager.

Reproduced the issue with a generic context manager, then used the debugger to inspect `isExceptionContextManager` in `codeFlowEngine.ts`. At the `bool` case:

- `exitType.shared.declaredReturnType.shared.name` = `"T"` (the raw TypeVar — the bug)
- `exitType.priv.specializedTypes.returnType.shared.name` = `"bool"` (the solved type)

Because the declared type was the TypeVar `T`, the `ClassType.isBuiltIn(returnType, 'bool')` check never matched and `cmSwallowsExceptions` stayed `false`.

## Implementation

In `isExceptionContextManager` (`codeFlowEngine.ts`), use `FunctionType.getEffectiveReturnType(exitType, /* includeInferred */ false)` instead of reading `shared.declaredReturnType` directly. The effective return type prefers `priv.specializedTypes.returnType`, so generic managers are evaluated using the solved type argument. The existing `__aexit__` Coroutine-unwrapping (third type arg) and literal-`bool` handling are preserved.

## Testing

- Added sample `packages/pyright-internal/src/tests/samples/with7.py` covering generic sync and async context managers specialized to `None` and `bool`, asserting the narrowed type after the block via `assert_type`.
- Added `With7` test case to `checker.test.ts`.
- Red → green: `With7` failed before the fix (4 `assert_type` mismatches), passes after.
- `checker.test` (72) and `typeEvaluator1-8` (1163) suites pass.
- Prettier, ESLint, and `tsc --noEmit` clean.

## Addresses

Fixes https://github.com/microsoft/pyright/issues/11483

---


packages/pyright-internal/src/tests/samples/with7.py
```diff
@@ -0,0 +1,46 @@
...
+async def func1() -> None:
+    baz = "str"
+    async with AsyncContextManager[None]() as ctx:
+        baz = 1
+    assert_type(baz, Literal[1])
+
+    qux = "str"
+    async with AsyncContextManager[bool]() as ctx:
+        qux = 1
+    assert_type(qux, Literal["str", 1])
```
rchiodo: This fixes a `microsoft/pylance-release` issue, and the split policy (`.github/instructions/pylance-pyright-split.instructions.md`) expects the repro/regression test to start in the Pylance harness (`packages/pylance-internal/src/tests/`), with `pyright-internal` as additional coverage. Since the fix lands in the vendored subrepo, a pyright-internal test is natural — but without a Pylance-side guard, a future subrepo re-pull that drops this fix would go undetected by the product suite. Consider adding a Pylance-harness regression test (or note explicitly why it's skipped).

.github/plans/archived/FIX_11483_CONTEXT_MANAGER_GENERIC_EXIT_PLAN.md
```diff
@@ -0,0 +1,49 @@
+# Fix #11483 — Incorrect type inference from context manager `__exit__` return type with Generic
+## Status: ✅ Complete
+
+Status: ✅ Complete (pending pipeline push/PR)
+
+Issue: https://github.com/microsoft/pyright/issues/11483
+
```
rchiodo: Issue-reference inconsistency: the PR body says `Fixes https://github.com/microsoft/pylance-release/issues/11483` while this plan cites `microsoft/pyright/issues/11483`. Repo convention is `Addresses` for pylance-release issues and `Fixes` for pyrx/pyright issues, with the full URL. Reconcile the repo + verb + URL so the linkage is unambiguous.

rchiodo:
Fix looks correct and is well-tested. A couple of minor repo-convention/test-placement notes below; none are blocking.

packages/pyright-internal/src/tests/samples/with7.py
rchiodo: Thanks ΓÇö but the premise here is actually off: issue 11483 is a `microsoft/pyright` issue (https://github.com/microsoft/pyright/issues/11483), not a pylance-release one. `pylance-release#11483` returns 404; the mislabeled PR body (now corrected) was the source of that confusion. So this is a genuine pyright issue, fixed in pyright, with the regression test correctly living in `pyright-internal` (`with7.py`). This standalone `microsoft/pyright` repo has no `packages/pylance-internal` harness, so a Pylance-side guard can't be added here; when pyright is vendored into pyrx, `with7.py` travels with the subrepo and keeps guarding the fix. Noting explicitly why the Pylance-harness test is skipped, per your alternative.

.github/plans/archived/FIX_11483_CONTEXT_MANAGER_GENERIC_EXIT_PLAN.md
rchiodo: Good catch. Verified upstream: the issue is `microsoft/pyright#11483` (`pylance-release#11483` returns 404), so the plan's citation (pyright) was the correct one and the PR body was wrong. Fixed the PR body to `Fixes https://github.com/microsoft/pyright/issues/11483` ΓÇö correct repo + verb (`Fixes` for pyright) + full URL. Plan, PR body, and URL are now consistent.

heejaechang:
This change appears to be based on an out-of-date tree. The current working tree already resolves #11483 in codeFlowEngine.ts using applySolvedTypeVars, and both the With7 test and a richer with7.py (with SplitContextManager multi-TypeVar coverage) already exist. Please re-baseline onto the current subrepo state, confirm With7 is already green, and drop the redundant changes. Note also that packages/pyright is an upstream-owned subrepo per AGENTS.md.

heejaechang:
This PR appears stale relative to the current subrepo: the working tree already fixes #11483 in codeFlowEngine.ts (via applySolvedTypeVars), already has a With7 test in checker.test.ts, and already has a richer with7.py (including a multi-TypeVar SplitContextManager case). As-is, this would revert the better fix, create a duplicate test, and delete regression coverage. Please re-baseline onto current main and confirm With7 is already green before re-submitting.

heejaechang:
This change appears to be based on a stale baseline. The current subrepo already fixes #11483 in `codeFlowEngine.ts` (via `applySolvedTypeVars` on the declared return type), and `with7.py` / `checker.test.ts` already contain `With7` coverage on disk. As written this diff would revert the existing fix, drop a return-type guard, delete the `SplitContextManager` multi-TypeVar coverage, and add a duplicate `With7` test. Please re-baseline onto the current branch, confirm `With7` is already green, and drop the redundant edits before re-submitting.

rchiodo:
@heejaechang you're right — thanks for catching this. I verified your claim against current `upstream/main` and it holds:

- `codeFlowEngine.ts` already fixes #11483 via `applySolvedTypeVars`/`buildSolution` on the declared `__exit__`/`__aexit__` return type (#11527). My branch's `getEffectiveReturnType` approach was branched off an older base and would have reverted it.
- `with7.py` and the `With7` test in `checker.test.ts` already exist upstream, including the `SplitContextManager` multi-TypeVar case.

I re-baselined `fix11483` onto current `main` and dropped all the redundant/conflicting edits:

- `codeFlowEngine.ts` and `checker.test.ts` now match `main` exactly (no revert, no duplicate `With7` test, `SplitContextManager` coverage preserved).
- Removed the `.github/plans/archived/*.md` artifact — it's a pyrx pipeline file that doesn't belong in upstream-owned pyright.

The only remaining diff vs `main` is a small **additive** extension to `with7.py`: a generic `AsyncContextManager[T]` regression case covering `__aexit__` specialized to both `bool` (exception possibly suppressed) and `None` (never suppressed). Upstream's `with7.py` only covers the sync path, so this just strengthens the existing fix's coverage without touching any product code.

`git diff main HEAD` is now a single file, +22 lines in `with7.py`. If you'd prefer this PR be closed instead (since the fix already landed via #11527), happy to do that too.

bschnurr:
The core fix for #11483 has already merged in #11488. The remaining diff here adds async context-manager regression cases to `with7.py`; if those cases are worth retaining, they could be extracted into a test-only follow-up. Otherwise, this PR can be closed as superseded.

bschnurr:

packages/pyright-internal/src/tests/samples/with7.py
```diff
@@ -36,3 +36,25 @@ def test_multi_typevar_context_manager() -> None:
...
     reveal_type(bar, expected_text="Literal['str', 1]")
+
+
+class AsyncContextManager(Generic[T]):
+    async def __aenter__(self) -> Self: ...
+
+    async def __aexit__(self, exc_type, exc_val, exc_tb) -> T: ...
+
+
+async def test_generic_async_context_manager() -> None:
```
bschnurr, line 47: > **Warning** · Non-blocking recommendation

📍 packages/pyright-internal/src/tests/samples/with7.py:47
The complete PR diff only adds async tests, while the title and description claim a production fix in `codeFlowEngine.ts`. Include the missing implementation change or revise the PR metadata to accurately describe this as test-only coverage.

[verified]

bschnurr:

**Verification:** Isolated verification observed failures whose relationship to this PR could not be determined\: Offline dependency bootstrap; this review is not fully verified.

**Summary:** No meaningful test could run. The required offline \`npm ci\` failed because \`yoctocolors-cjs\` was absent from the sandbox cache, leaving Jest unavailable. Consequently, the targeted context-manager tests were not run and confidence is low. The new tests cover direct generic sync, multi-TypeVar, and async managers, but not inherited generic specialization.

**Test runs:** 1 passed, 1 failed, 1 not run

- :warning: **Not run** | <code>Checker With1-With7 context-manager tests</code> | <code>node ./node_modules/jest/bin/jest.js src/tests/checker.test.ts --runInBand -t &#x27;^With[1-7]$&#x27;</code>
- :x: **Failed** | relationship unknown | <code>Offline dependency bootstrap</code> | <code>npm ci --offline</code>
- :white_check_mark: **Passed** | <code>Sandbox dependency and test discovery</code> | <code>printf &#x27;%s\n&#x27; &#x27;--- sandbox profile ---&#x27;; printf &#x27;%s\n&#x27; &quot;${AUTOMATION_SANDBOX_PROFILE:-unset}&quot;; printf &#x27;%s\n&#x27; &#x27;--- changed files ---&#x27;; git diff --name-status HEAD^...HEAD; printf &#x27;%s\n&#x27; &#x27;--- relevant diff ---&#x27;; git diff --unified=20 HEAD^...HEAD -- packages/pyright-internal/src/analyzer/codeFlowEngine.ts packages/pyright-internal/src/tests/checker.test.ts packages/pyright-internal/src/tests/samples/with7.py; printf &#x27;%s\n&#x27; &#x27;--- dependency state ---&#x27;; test -x packages/pyright-internal/node_modules/.bin/jest &amp;&amp; echo package-jest-present || echo package-jest-missing; test -x node_modules/.bin/prettier &amp;&amp; echo root-prettier-present || echo root-prettier-missing; test -f package-lock.json &amp;&amp; echo root-lock-present || echo root-lock-missing</code>

<details><summary>:warning: <code>Checker With1-With7 context-manager tests</code> diagnostic output</summary>

<pre>Jest dependencies were unavailable after offline bootstrap failed with ENOTCACHED for yoctocolors-cjs.</pre>
</details>

<details><summary>:x: <code>Offline dependency bootstrap</code> diagnostic output</summary>

<pre>npm error code ENOTCACHED
npm error request to https://packagefeedproxy.microsoft.io/npm/yoctocolors-cjs failed: cache mode is &#x27;only-if-cached&#x27; but no cached response is available.</pre>
</details>
</details>

bschnurr (approved):
Approved via Review Center.

heejaechang (approved):
Approved via Review Center.

PR #11503, opened 2026-06-15
Fix #11453: no error on accessing __qualname__ of instance

### LINKED ISSUE #11453: no error on accessing __qualname__ of instance
## Code or Screenshots

```python
class Foo:
    pass

foo = Foo()
foo.__qualname__

```

**Expected behavior**
type error: Cannot access attribute "\_\_qualname\_\_" for class "Foo"

**Actual**
no type error

Image

https://pyright-play.net/?code=MYGwhgzhAEBiD28BcAoa7oAdIRSgZotALxyIAUAlAYgHQD69AjgK5ggB2YAtgKaMogA

## VS Code extension or command-line
Pylance v2026.2.1

rchiodo:
## Description

Pyright/Pylance failed to report the expected "cannot access attribute" diagnostic when `__qualname__` was accessed on an **instance** (e.g. `MyClass().__qualname__`). At runtime, `__qualname__` is exposed via the metaclass `type`, not as an instance attribute, so instance access raises `AttributeError`. Class-object access (`MyClass.__qualname__`) is and should remain valid.

The existing sample `classes3.py` already documented this exact bug at the `instance.__qualname__` line. Reproduced it (no diagnostic emitted), then traced the root cause: the binder unconditionally injects an implicit `__qualname__` symbol into each class scope's symbol table (`binder.ts` ~L510). In a Class scope, `_addSymbolToCurrentScope` stamps it with `SymbolFlags.ClassMember`. Instance member-access resolution (`getClassMemberIterator`) treats any `isClassMember()` symbol as accessible, so the error was suppressed. Confirmed via debugger that the symbol had `isClassMember()=true` / `isInstanceMember()=false`, and via CPython that `A().__qualname__` raises `AttributeError`.

## Implementation

Threaded an optional `isClassMember` parameter (default `true`) through `_addImplicitSymbolToCurrentScope` ΓåÆ `_addSymbolToCurrentScope`. The `ClassMember` flag is now only set when the scope is a Class **and** `isClassMember` is true. The `__qualname__` injection passes `/* isClassMember */ false`, so it stays name-resolvable inside the class body (`print(__qualname__)`) but is no longer exposed as a class/instance member. `Class.__qualname__` still resolves via the metaclass `type.__qualname__: str`. `__doc__`/`__module__` are unchanged (they are legitimately instance-accessible).

## Testing

- Updated `classes3.py`: `instance.__qualname__` now expected error; added `reveal_type(..., expected_text="str")` guards for class-object access; added subclass, nested/inner class, and `type[_T]` generic cases. Bumped `Classes3` expected error count 3 ΓåÆ 6.
- Added `completions.qualname.fourslash.ts`: instance completions exclude `__qualname__` (but include `__doc__`/`__module__`); class completions include `__qualname__`.
- `typeEvaluator1-8` + `checker` (9 suites / 1234 tests) pass; fourslash completions/hover (333 tests) pass.
- Prettier + ESLint clean on changed files.

## Addresses

Fixes https://github.com/microsoft/pyright/issues/11453

---

### DIFF
--- a/.github/plans/archived/FIX_11453_QUALNAME_INSTANCE_ACCESS_PLAN.md
+++ b/.github/plans/archived/FIX_11453_QUALNAME_INSTANCE_ACCESS_PLAN.md
@@ -1,0 +1,126 @@
+# FIX 11453 — No error on accessing `__qualname__` of an instance
+## Status: ✅ Complete
+
+Status: 🟢 Active — fix implemented, verified & reviewed; ready to commit (PR/archival pipeline-deferred)
+Owner: PylanceFix/PylanceCode (fix-loop)
+Branch: `fix11453`
+Worktree: `C:\Users\rchiodo\.fix_loop\repos\microsoft__pyright\.worktrees_8765\fix11453`
+
+## Context
+
+- Issue: https://github.com/microsoft/pyright/issues/11453
+- Title: "no error on accessing `__qualname__` of instance"
+- Kind: bugfix
+- User impact: Pyright/Pylance fails to report the expected "cannot access attribute" diagnostic when `__qualname__` is accessed on an **instance** (e.g. `SomeClass().__qualname__`). `__qualname__` is a class-level attribute (defined on `type`), not an instance attribute, so instance access should be flagged.
+- Triage (PylanceTriage handoff): Verified defect. The binder injects an implicit `__qualname__` class member (including on `object`), so instance access resolves it and skips the expected "cannot access attribute" diagnostic.
+
+### Non-goals
+
+- Do not change behavior for **class-level** access (`SomeClass.__qualname__` must remain valid).
+- Do not change `__name__`/`__doc__`/other dunder handling beyond what is strictly required for parity with the fix (verify before touching).
+- No broad refactors of the binder or type evaluator.
+
+## Constraints & policies
+
+- This is a `microsoft/pyright` issue (worktree is the pyright repo). PR convention: "Fixes <full URL>".
+- Keep diffs surgical; pyright core is upstream-owned.
+- Test placement: add a pyright-internal regression test under `packages/pyright-internal/src/tests/` (sample `.py` + `validateResults` count, or fourslash if interaction-based). Follow existing `typeEvaluator*`/`checker` sample-test conventions.
+- Do not reduce type precision elsewhere; the change must be narrowly scoped to instance access of `__qualname__`.
+- Test policy: tests are the spec; new test must encode the expected diagnostic.
+
+## Strategy
+
+Phases (high level — verify file/function names before editing; these are hypotheses, not directives):
+
+1. **Reproduce**: Write a minimal sample showing `instance.__qualname__` should error and `Class.__qualname__` should not. Confirm current behavior produces no diagnostic on the instance access.
+2. **Locate root cause**: Per triage, an implicit `__qualname__` member is injected as a class member (hypothesis: in the binder, alongside other synthesized dunders, possibly applied to `object`/all classes). Verify where `__qualname__` becomes accessible on instances. Confirm whether it is the binder injection or member-access resolution that fails to treat it as class-only.
+3. **Minimal fix**: Make `__qualname__` resolve only for class-object access, not instance access — mirroring how Python exposes it via `type` (class) rather than instances. Prefer the approach that keeps `Class.__qualname__` valid and only suppresses instance access.
+4. **Validate**: Run the new regression test plus the relevant existing suites (`typeEvaluator*`, `checker`) to ensure no precision regressions.
+
+Verification approach: sample-based test with `validateResults` expected error count; confirm class access stays clean.
+
+## Acceptance Contract
+
+1. **Observable change**: Accessing `__qualname__` on an instance is flagged with the expected "cannot access attribute" (reportAttributeAccessIssue) diagnostic, matching runtime behavior where instances do not expose `__qualname__`.
+2. **Success criteria**:
+   - [ ] `instance.__qualname__` (e.g. `MyClass().__qualname__`) → reports an attribute-access error.
+   - [ ] `MyClass.__qualname__` (class-object access) → no error, type `str`.
+3. **Regression guard**:
+   - [ ] Existing `__name__`/class dunder behavior unchanged.
+   - [ ] No new false positives in `typeEvaluator*` / `checker` suites.
+4. **Scope boundary**: Only `__qualname__` instance-vs-class access semantics. No changes to unrelated dunders or to the binder's broader synthesized-member logic unless proven required.
+
+## Work items
+
+| # | Work item | Status | Owner | Exit criteria |
+|---|-----------|--------|-------|---------------|
+| 1 | Reproduce no-error-on-instance behavior | ✅ Done | impl | Confirmed current build emits no diagnostic for `instance.__qualname__` |
+| 2 | Verify root cause (binder implicit `__qualname__` injection) | ✅ Done | impl | Located exact injection/resolution site; confirmed root cause |
+| 3 | Implement minimal fix (class-only access) | ✅ Done | impl | `instance.__qualname__` errors; `Class.__qualname__` valid |
+| 4 | Add regression test (pyright-internal) | ✅ Done | impl | New sample + `validateResults` asserts expected error/no-error counts |
+| 5 | Run targeted + relevant existing suites | ✅ Done | impl | New test passes; no regressions in `typeEvaluator*`/`checker` |
+| 6 | Format & lint | ✅ Done | impl | `npm run check` clean for changed files |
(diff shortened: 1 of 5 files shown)


rchiodo:
It looks to me like this has introduced a whole number of failures along the lines of:

dd-trace-py (https://github.com/DataDog/dd-trace-py)
+   .../projects/dd-trace-py/ddtrace/contrib/internal/asyncio/patch.py:73:32 - error: "__qualname__" is not a known attribute of "None" (reportOptionalMemberAccess)
+   .../projects/dd-trace-py/ddtrace/vendor/debtcollector/_utils.py:137:52 - error: Cannot access attribute "__qualname__" for class "MethodType"
+     Attribute "__qualname__" is unknown (reportAttributeAccessIssue)
- 14755 errors, 750 warnings, 0 informations
+ 14757 errors, 750 warnings, 0 informations

pydantic (https://github.com/pydantic/pydantic)
+ .../projects/pydantic/pydantic/_internal/_validate_call.py
+   .../projects/pydantic/pydantic/_internal/_validate_call.py:26:98 - error: Cannot access attribute "__qualname__" for class "MethodType"
+     Attribute "__qualname__" is unknown (reportAttributeAccessIssue)
- 10 errors, 0 warnings, 0 informations
+ 11 errors, 0 warnings, 0 informations

rchiodo:
Good catch — fixed in 73f54616c.

Root cause: the implicit __qualname__ symbol was injected into the class scope *before* the class suite was walked. For typeshed classes that legitimately declare __qualname__ (`function`, `MethodType`, `type`, `partial`, etc.), that pre-created **non-class-member** symbol got the explicit declaration merged into it, but `_bindNameValueToScope` only sets the `ClassMember` flag when it *creates* a symbol — so the flag was never restored. The legitimate member was then hidden, producing the new `reportAttributeAccessIssue` errors on valid `instance.__qualname__` accesses across the mypy_primer projects.

Fix: inject the implicit `__qualname__` **after** walking the suite. Now an explicit declaration creates the symbol as a proper class member first, and the implicit injection only adds a non-class-member symbol for classes that don't declare one. Instance access on a plain class still errors (the original #11453 behavior), while `function.__qualname__` / `MethodType.__qualname__` / explicit `__qualname__` members resolve again.

Added regression coverage to `classes3.py` for a function object's `__qualname__` and a class that explicitly declares `__qualname__`. `Classes3` and `completions.qualname` both pass locally.

rchiodo:
Follow-up on the mypy_primer `__qualname__` diff (the run posted after 73f54616c still lists `__qualname__` errors, but they're now a different, expected set):

- **Fixed regressions (false positives that hid legitimate members):** the `MethodType` / `function` / `partial`-via-`_lru_cache_wrapper` etc. cases from the original report are gone. These typeshed classes *explicitly declare* `__qualname__: str`, so `instance.__qualname__` resolves again. Verified: `Classes3` and `completions.qualname` pass, and dd-trace-py/pydantic no longer show the `MethodType.__qualname__` error.

- **Remaining `__qualname__` diffs are intended true positives of #11453:** `object`, `str`, `None`, and `functools.partial` do **not** declare `__qualname__` as a class/instance attribute (it lives on the metaclass `type`). At runtime `object().__qualname__`, `"x".__qualname__`, and `partial(f).__qualname__` all raise `AttributeError`, so flagging them (`reportAttributeAccessIssue` / `reportOptionalMemberAccess`) is the correct new behavior — these are pre-existing latent bugs in jax/streamlit/spack/dd-trace-py that the fix now surfaces.

Mechanism of the fix (73f54616c): the implicit `__qualname__` symbol is now injected *after* walking the class suite. A class that explicitly declares `__qualname__` therefore creates the class-member symbol first; the implicit injection (with `isClassMember=false`) finds the existing symbol and leaves its `ClassMember` flag intact. Plain classes that don't declare it get a non-class-member symbol, so instance access still errors as intended.

packages/pyright-internal/src/tests/samples/classes3.py
```diff
@@ -61,4 +63,57 @@ def method1(self) -> str:
...
+
+class WithQualname:
+    # A class that explicitly declares `__qualname__` exposes it on instances.
+    __qualname__ = "WithQualname"
+
+
+# This should not generate an error because the attribute is explicitly declared
+# as a class member.
+WithQualname().__qualname__
+reveal_type(WithQualname().__qualname__, expected_text="str")
```
heejaechang, line 125: The `WithQualname` block asserts `WithQualname().__qualname__` is valid `str`, but at runtime `type.__new__` strips a class-body `__qualname__` assignment from `__dict__`, so `WithQualname().__qualname__` raises `AttributeError`. This is consistent with pyright's general modeling of class-body assignments as instance-accessible class variables, so changing the diagnostic may be out of scope — but the inline comment "exposes it on instances" is factually inaccurate and slightly contradicts this PR's stated principle. Please reword the comment to note the runtime would `AttributeError` (and that this reflects pyright's class-variable modeling), so the test doesn't read as enshrining runtime behavior it doesn't match.

packages/pyright-internal/src/analyzer/binder.ts
```diff
@@ -507,14 +507,28 @@ export class Binder extends ParseTreeWalker {
...
+                // incorrectly resolve instead of reporting an attribute-access error.
+                //
+                // This is injected after walking the suite so that a class that explicitly
+                // declares `__qualname__` (e.g. `type`, `function`, and other typeshed
+                // classes) keeps the class-member symbol created for that declaration. If it
+                // were injected first, the non-class-member symbol would be created up front
+                // and the explicit declaration would merge into it without restoring the
+                // class-member flag, incorrectly hiding the legitimate member.
+                this._addImplicitSymbolToCurrentScope('__qualname__', node, 'str', /* isClassMember */ false);
+
```
heejaechang, line 540: Moving the `__qualname__` injection to after `this.walk(node.d.suite)` means that for symbols with a real declaration (typeshed `function`/`type`, or a user `__qualname__ = "..."`), the synthetic empty-range Intrinsic declaration is now appended last via `addDeclaration`. Consumers that select the "last" declaration for hover / go-to-definition could now resolve to the empty range instead of the real declaration. `reveal_type` still yields `str`, so existing tests won't catch this. Worth verifying "Go to Definition" on `func1.__qualname__` lands on the real declaration in a post-patch build.

.github/plans/archived/FIX_11453_QUALNAME_INSTANCE_ACCESS_PLAN.md
```diff
@@ -0,0 +1,126 @@
+# FIX 11453 — No error on accessing `__qualname__` of an instance
+## Status: ✅ Complete
+
+Status: 🟢 Active — fix implemented, verified & reviewed; ready to commit (PR/archival pipeline-deferred)
+Owner: PylanceFix/PylanceCode (fix-loop)
+Branch: `fix11453`
+Worktree: `C:\Users\rchiodo\.fix_loop\repos\microsoft__pyright\.worktrees_8765\fix11453`
+
+## Context
+
```
heejaechang, line 10: Repo/verb mismatch: the plan and pyright-internal test placement treat this as `microsoft/pyright` issue #11453 (matching the "Fixes" verb), but the PR footer points at `microsoft/pylance-release`. Per repo convention, `pylance-release` issues use "Addresses" and only `pyrx` issues use "Fixes". Please reconcile the repo and verb before merge.

heejaechang:
Solid, surgical fix for instance `__qualname__` access. A couple of test/metadata cleanups and one navigation behavior worth a quick verification, but nothing blocking.

heejaechang (approved):
Approved via Review Center.

packages/pyright-internal/src/tests/samples/classes3.py
rchiodo: @C:\Users\rchiodo\AppData\Local\Temp\r1.txt

packages/pyright-internal/src/analyzer/binder.ts
rchiodo: @C:\Users\rchiodo\AppData\Local\Temp\r2.txt

.github/plans/archived/FIX_11453_QUALNAME_INSTANCE_ACCESS_PLAN.md
rchiodo: @C:\Users\rchiodo\AppData\Local\Temp\r3.txt

heejaechang:
Looks good — clean root-cause fix in the binder (injecting `__qualname__` after the suite walk as a non-class member, while preserving real declarations). Test coverage is solid. The `WithQualname`/annotation-only no-error cases diverge from CPython runtime, but that divergence is an intentional, clearly-documented consequence of pyright's general class-variable modeling rather than something this PR introduces. Approving.

heejaechang (approved):
Approved via Review Center.

PR #11516, opened 2026-06-20
Fix symlinked partial stub packages

tamird:
Partial stub packages are overlaid onto their runtime packages. When a
file in the stub package is a symbolic link, its directory entry reports
that it is a link rather than a file. `ReadOnlyAugmentedFileSystem`
currently converts that entry to a virtual directory, so import
resolution ignores the `.pyi` and falls back to the runtime `.py`.

Resolve symbolic links before constructing virtual directory entries.
Dangling links and special files are omitted rather than exposed as
directories. This preserves the partial stub overlay for environments
that install wheel files as symbolic links.

### DIFF
--- a/packages/pyright-internal/src/readonlyAugmentedFileSystem.ts
+++ b/packages/pyright-internal/src/readonlyAugmentedFileSystem.ts
@@ -13,6 +13,7 @@ import { FileSystem, MkDirOptions, Stats, VirtualDirent } from './common/fileSys
 import { FileWatcher, FileWatcherEventHandler } from './common/fileWatcher';
 import { Uri } from './common/uri/uri';
 import { UriMap } from './common/uri/uriMap';
+import { tryStat } from './common/uri/uriUtils';
 import { Disposable } from 'vscode-jsonrpc';
 
 interface MappedEntry {
@@ -70,12 +71,20 @@ export class ReadOnlyAugmentedFileSystem implements FileSystem {
         const mappedEntry = this._getOriginalEntry(uri);
         if (mappedEntry) {
             const originalUri = this._getInternalOriginalUri(uri);
-            const filteredEntries = this.realFS
-                .readdirEntriesSync(originalUri)
-                .filter((e) => mappedEntry.filter(originalUri.combinePaths(e.name), this.realFS))
-                .map((e) => new VirtualDirent(e.name, e.isFile(), uri.getFilePath()));
-            for (const entry of filteredEntries) {
-                entries.set(entry.name, entry);
+            for (const entry of this.realFS.readdirEntriesSync(originalUri)) {
+                const originalEntryUri = originalUri.combinePaths(entry.name);
+                if (!mappedEntry.filter(originalEntryUri, this.realFS)) {
+                    continue;
+                }
+
+                // Mapped entries are virtual, so resolve types that readdir cannot classify, including symlinks and
+                // DT_UNKNOWN entries.
+                const target = entry.isFile() || entry.isDirectory() ? entry : tryStat(this.realFS, originalEntryUri);
+                if (!target || (!target.isFile() && !target.isDirectory())) {
+                    continue;
+                }
+
+                entries.set(entry.name, new VirtualDirent(entry.name, target.isFile(), uri.getFilePath()));
             }
         }
 
--- a/packages/pyright-internal/src/tests/importResolver.test.ts
+++ b/packages/pyright-internal/src/tests/importResolver.test.ts
@@ -68,6 +68,52 @@ describe('Import tests with fake venv', () => {
                 );
             });
 
+            test('symlinked partial stub file exists', () => {
+                const stubSource = combinePaths(normalizeSlashes('/'), 'wheel', 'partialStub.pyi');
+                const files = [
+                    {
+                        path: stubSource,
+                        content: 'def test(): ...',
+                    },
+                    {
+                        path: combinePaths(libraryRoot, 'myLib-stubs', 'py.typed'),
+                        content: 'partial\n',
+                    },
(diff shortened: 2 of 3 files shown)


packages/pyright-internal/src/tests/importResolver.test.ts
```diff
@@ -68,6 +68,52 @@ describe('Import tests with fake venv', () => {
...
+                assert(importResult.isImportFound);
+                assert(importResult.isStubFile);
+                assert(
+                    importResult.resolvedUris.some(
+                        (resolvedUri) =>
+                            resolvedUri.getFilePath() === combinePaths(libraryRoot, 'myLib', 'partialStub.pyi')
+                    )
+                );
+            });
+
```
rchiodo, line 116: 📍 packages/pyright-internal/src/tests/importResolver.test.ts:110
Review-rule violation (`tests-no-partial-asserts`): this test uses `importResult.resolvedUris.some(...)` while every peer test in this file ("partial stub file exists", "partial stub __init__ exists") uses `assert.strictEqual(1, importResult.resolvedUris.filter(...).length)`. The `.some()` form is a weaker partial assertion that passes even on duplicate/coincidental matches. Align with the established count-based pattern. This is a documented review rule and must be fixed before merge (cannot be downgraded).

[verified]

packages/pyright-internal/src/tests/pyrightFileSystem.test.ts
```diff
@@ -59,6 +59,33 @@ test('virtual file exists', () => {
...
+
+    const fs = new PyrightFileSystem(testFs);
+    const ps = new PartialStubService(fs);
+    ps.processPartialStubPackages([libraryRootUri], [libraryRootUri]);
+
+    const entries = fs.readdirEntriesSync(libraryRootUri.combinePaths('myLib'));
+    const stubFile = entries.find((entry) => entry.name === 'partialStub.pyi');
+    assert(stubFile?.isFile());
+    assert(!entries.some((entry) => entry.name === 'missing.pyi'));
+});
```
rchiodo, line 89: 📍 packages/pyright-internal/src/tests/pyrightFileSystem.test.ts:84
Review-rule violation (`tests-no-partial-asserts`): every other `readdirEntriesSync` test in this file leads with `assert.strictEqual(N, entries.length)` before inspecting entries, but this test omits the count check and uses `entries.find(...)`/`entries.some(...)`. Additionally `assert(stubFile?.isFile())` silently evaluates `assert(undefined)` (a cryptic failure) if the entry is missing. Add a leading entry-count assertion and a non-optional `assert.ok(stubFile, ...)` guard before `stubFile.isFile()`. Documented review rule — must be fixed before merge.

[verified]

packages/pyright-internal/src/readonlyAugmentedFileSystem.ts
```diff
@@ -70,12 +71,19 @@ export class ReadOnlyAugmentedFileSystem implements FileSystem {
...
+                    continue;
+                }
+
+                // Mapped entries are virtual, so expose a symlink target's type rather than the backing symlink.
+                const target = entry.isSymbolicLink() ? tryStat(this.realFS, originalEntryUri) : entry;
+                if (!target || (!target.isFile() && !target.isDirectory())) {
+                    continue;
+                }
+
+                entries.set(entry.name, new VirtualDirent(entry.name, target.isFile(), uri.getFilePath()));
```
rchiodo, line 87: 📍 packages/pyright-internal/src/readonlyAugmentedFileSystem.ts:79
[unverified] `DT_UNKNOWN` regression (Skeptic): the backing `RealFileSystem.readdirEntriesSync` returns raw Node `fs.Dirent`s with no `lstat` backfill. On filesystems that report `DT_UNKNOWN` (some NFS/CIFS/FUSE, older XFS without ftype=1), an entry returns `false` for `isFile()`, `isDirectory()`, AND `isSymbolicLink()`, so it takes the non-symlink path (`target = entry`) and is then dropped by the `!isFile() && !isDirectory()` guard — turning a real `.pyi` stub from "present but mistyped" into "vanished". Consider resolving unknown types too: `const target = entry.isFile() || entry.isDirectory() ? entry : tryStat(this.realFS, originalEntryUri);` which covers both symlinks and `DT_UNKNOWN` without ever dropping a real file. (Backing-FS behavior structurally confirmed; the `DT_UNKNOWN` trigger could not be reproduced on Windows.)

[verified]

rchiodo:
The fix itself looks correct and is well-targeted. Main ask: align the two new tests with the repo's documented exact-assertion pattern used by their peers (count-based asserts + a non-optional entry guard). There's also one optional robustness note on the readdir change for filesystems that report unknown dirent types.

packages/pyright-internal/src/tests/importResolver.test.ts
tamird: Done.

packages/pyright-internal/src/tests/pyrightFileSystem.test.ts
tamird: Done.

packages/pyright-internal/src/readonlyAugmentedFileSystem.ts
tamird: Done. Mapped entries now stat any dirent that isn't already known to be a file or directory, so symlinks and DT_UNKNOWN share the fallback while dangling links and special files remain omitted.

rchiodo (approved):
Approved via Review Center.

tamird:
Thanks for the review! It would be great to see this in a release soon.

rchiodo:
Thanks for the PR

PR #11520, opened 2026-06-23
Evaluate truthiness of literals in static boolean expressions

### LINKED ISSUE #11505: `Code is structurally unreachable` different for `while 1` and `while True`.
**Describe the bug**
`Code is structurally unreachable` different for `while 1` and `while True`.

Both should be `Code is structurally unreachable` 
The below `while 1` one should have `Code is structurally unreachable`, like the `while True` one

**Code or Screenshots**

Image

Image

VS Code, Pylance, and Pyright Playground, all of them exist this issue.

velikodniy:
## Summary

Currently, `bool`-like expressions evaluator supports only `None` as a falsy literal.
So that, this code does not report `print(x)` as unreachable:

```python
while 1:
    x = 1
print(x)
```

This PR adds support of other literal types listed below

| Expression form | Examples | Static result |
| --- | --- | --- |
| Bool keywords | `True`, `False` | `true` / `false` |
| `None` | `None` | `false` |
| Ellipsis | `...` | `true` |
| Numbers | `0`, `0.0`, `0j` / `1`, `3.14`, `1j` | `false` / `true` |
| Strings & bytes | `""`, `b""` / `"x"`, `b"x"` | `false` / `true` |
| List / set / tuple / dict | `[]`, `()`, `{}` / `[1]`, `(1,)`, `{1}`, `{1: 2}` | `false` / `true` |
| `not` of a foldable operand | `not 1`, `not []`, `not not 1` | negated |
| `and` / `or` of foldable operands | `1 and 0`, `0 or "x"` | combined truthiness |
| Container with a concrete element | `[1, *x]`, `{**d, "k": 1}` | `true` (definitely non-empty) |
| f-strings | `f"{x}"` | `undefined` (runtime-dependent) |
| Spread/comprehension-only containers | `[*x]`, `[i for i in y]`, `{**d}` | `undefined` (length unknown) |
| Negative or non-literal operands | `-1`, `x`, `f()`, `Decimal(0)`, `set()` | `undefined` (a negative number isn't a single node)|

## Tests

Added `staticExpressionLiteral1.py` with positive and negative cases.
One of the tests is updated because `if 1` is determined now.

Fixes #11505

### DIFF
--- a/packages/pyright-internal/src/analyzer/staticExpressions.ts
+++ b/packages/pyright-internal/src/analyzer/staticExpressions.ts
@@ -84,12 +84,13 @@ function _evaluateStaticBoolOrBoolLikeExpression(
 
     if (node.nodeType === ParseNodeType.UnaryOperation) {
         if (node.d.operator === OperatorType.Not) {
-            const value = evaluateStaticBoolLikeExpression(
+            const value = _evaluateStaticBoolOrBoolLikeExpression(
                 node.d.expr,
                 execEnv,
                 definedConstants,
                 typingImportAliases,
-                sysImportAliases
+                sysImportAliases,
+                evaluateLeafAsBool
             );
             if (value !== undefined) {
                 return !value;
--- a/packages/pyright-internal/src/analyzer/staticExpressions.ts
+++ b/packages/pyright-internal/src/analyzer/staticExpressions.ts
@@ -90,7 +90,9 @@ function _evaluateStaticBoolOrBoolLikeExpression(
                 definedConstants,
                 typingImportAliases,
                 sysImportAliases,
-                evaluateLeafAsBool
+                // `not x` always forces a truthiness context on its operand and yields real `bool`.
+                // So the operand is folded with the bool-like leaf even for strict callers.
+                _evaluateBoolLikeLiteral
             );
             if (value !== undefined) {
                 return !value;
--- a/packages/pyright-internal/src/analyzer/staticExpressions.ts
+++ b/packages/pyright-internal/src/analyzer/staticExpressions.ts
@@ -10,7 +10,16 @@
 
 import { ExecutionEnvironment, PythonPlatform } from '../common/configOptions';
 import { PythonReleaseLevel, PythonVersion } from '../common/pythonVersion';
-import { ArgCategory, ExpressionNode, NameNode, NumberNode, ParseNodeType, TupleNode } from '../parser/parseNodes';
+import {
+    ArgCategory,
+    DictionaryEntryNode,
+    ExpressionNode,
+    NameNode,
+    NumberNode,
+    ParseNodeType,
+    StringListNode,
+    TupleNode,
+} from '../parser/parseNodes';
 import { KeywordType, OperatorType } from '../parser/tokenizerTypes';
 
 // Returns undefined if the expression cannot be evaluated
@@ -22,13 +31,54 @@ export function evaluateStaticBoolExpression(
     typingImportAliases?: string[],
     sysImportAliases?: string[]
 ): boolean | undefined {
+    return _evaluateStaticBoolOrBoolLikeExpression(
+        node,
+        execEnv,
+        definedConstants,
+        typingImportAliases,
+        sysImportAliases,
+        _evaluateBoolConstant
+    );
+}
+
+// Similar to evaluateStaticBoolExpression except that it handles other non-bool
+// values that are statically falsy or truthy (like "None", "...", and
+// numeric/string/container literals).
+export function evaluateStaticBoolLikeExpression(
+    node: ExpressionNode,
+    execEnv: ExecutionEnvironment,
+    definedConstants: Map<string, boolean | string>,
+    typingImportAliases?: string[],
+    sysImportAliases?: string[]
+): boolean | undefined {
+    return _evaluateStaticBoolOrBoolLikeExpression(
+        node,
+        execEnv,
+        definedConstants,
+        typingImportAliases,
+        sysImportAliases,
+        _evaluateBoolLikeLiteral
+    );
+}
+
+// Shared implementation of the two functions above.
+// The "evaluatePrimitive" callback evaluates leaf expressions.
+function _evaluateStaticBoolOrBoolLikeExpression(
+    node: ExpressionNode,
+    execEnv: ExecutionEnvironment,
+    definedConstants: Map<string, boolean | string>,
+    typingImportAliases: string[] | undefined,
+    sysImportAliases: string[] | undefined,
+    evaluateLeafAsBool: (node: ExpressionNode) => boolean | undefined
+): boolean | undefined {
     if (node.nodeType === ParseNodeType.AssignmentExpression) {
-        return evaluateStaticBoolExpression(
+        return _evaluateStaticBoolOrBoolLikeExpression(
             node.d.rightExpr,
             execEnv,
             definedConstants,
             typingImportAliases,
-            sysImportAliases
+            sysImportAliases,
+            evaluateLeafAsBool
         );
     }
 
@@ -48,19 +98,21 @@ export function evaluateStaticBoolExpression(
     } else if (node.nodeType === ParseNodeType.BinaryOperation) {
         // Is it an OR or AND expression?
         if (node.d.operator === OperatorType.Or || node.d.operator === OperatorType.And) {
-            const leftValue = evaluateStaticBoolExpression(
+            const leftValue = _evaluateStaticBoolOrBoolLikeExpression(
                 node.d.leftExpr,
                 execEnv,
                 definedConstants,
                 typingImportAliases,
-                sysImportAliases
+                sysImportAliases,
+                evaluateLeafAsBool
             );
-            const rightValue = evaluateStaticBoolExpression(
+            const rightValue = _evaluateStaticBoolOrBoolLikeExpression(
                 node.d.rightExpr,
                 execEnv,
                 definedConstants,
                 typingImportAliases,
-                sysImportAliases
+                sysImportAliases,
+                evaluateLeafAsBool
             );
 
             if (leftValue === undefined || rightValue === undefined) {
@@ -139,12 +191,6 @@ export function evaluateStaticBoolExpression(
                 }
             }
         }
-    } else if (node.nodeType === ParseNodeType.Constant) {
-        if (node.d.constType === KeywordType.True) {
-            return true;
-        } else if (node.d.constType === KeywordType.False) {
-            return false;
-        }
     } else if (node.nodeType === ParseNodeType.Name) {
         if (node.d.value === 'TYPE_CHECKING') {
             return true;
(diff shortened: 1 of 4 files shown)


velikodniy:

packages/pyright-internal/src/analyzer/staticExpressions.ts
```diff
@@ -21,14 +30,55 @@ export function evaluateStaticBoolExpression(
...
+        typingImportAliases,
+        sysImportAliases,
+        _evaluateBoolConstant
+    );
+}
+
+// Similar to evaluateStaticBoolExpression except that it handles other non-bool
+// values that are statically falsy or truthy (like "None", "...", and
+// numeric/string/container literals).
+export function evaluateStaticBoolLikeExpression(
```
rchiodo, line 47: The shared walker threads `evaluateLeafAsBool` through the assignment and `and`/`or` branches, but the `not` (UnaryOperation) branch still calls the public `evaluateStaticBoolLikeExpression` directly rather than using the callback. As a result the strict `evaluateStaticBoolExpression("not 1")` now folds to `false` while `evaluateStaticBoolExpression("1")` stays `undefined`, so the leaf-evaluator parameter doesn't actually govern all leaves. The runtime result is always a real bool so this is semantically harmless, but the inconsistency is undocumented and leaks to strict-mode callers. Either thread `evaluateLeafAsBool` into the `not` branch or add a comment explaining why `not` is intentionally always bool-like.

packages/pyright-internal/src/analyzer/staticExpressions.ts
```diff
@@ -21,14 +30,55 @@ export function evaluateStaticBoolExpression(
...
-        return evaluateStaticBoolExpression(
+        return _evaluateStaticBoolOrBoolLikeExpression(
             node.d.rightExpr,
             execEnv,
             definedConstants,
             typingImportAliases,
-            sysImportAliases
+            sysImportAliases,
+            evaluateLeafAsBool
         );
```
rchiodo, line 82: The shared-helper comment refers to an `"evaluatePrimitive"` callback, but the actual parameter is named `evaluateLeafAsBool`. Update the comment to match the parameter name.

rchiodo:
Looks good — approving. Two minor maintainability notes inline (one behavioral consistency question on the `not` branch, one stale doc-comment).

rchiodo (approved):
Approved via Review Center.

packages/pyright-internal/src/analyzer/staticExpressions.ts
velikodniy: Fixed. I should've been more attentive :)

packages/pyright-internal/src/analyzer/staticExpressions.ts
velikodniy: Updated to use `_evaluateStaticBoolOrBoolLikeExpression` in the shared walker

rchiodo:
I'm wondering about the mypy primer differences. 

Like this one:

pytest (https://github.com/pytest-dev/pytest)
-   .../projects/pytest/src/_pytest/pytester.py:1270:10 - error: Function with declared return type "Item" must return value on all code paths
-     "None" is not assignable to "Item" (reportReturnType)
- 290 errors, 11 warnings, 0 informations
+ 289 errors, 11 warnings, 0 informations

That seems good to me, but we should double check. Is that expected now because we realized the function never returns?

velikodniy:
I'm looking into this to make sure it's not a regression

rchiodo:
GitHub cannot anchor PR review comments to unchanged lines in the diff. Falling back to a general PR comment for [packages/pyright-internal/src/analyzer/staticExpressions.ts:L97](https://github.com/microsoft/pyright/blob/64ca1ebfc4cc2f06b760390c7a7a4b21cf2c23f0/packages/pyright-internal/src/analyzer/staticExpressions.ts#L97).

The `not` branch now threads the *strict* leaf evaluator (`_evaluateBoolConstant`) for strict callers. Pre-PR, the `not` branch always delegated to the lenient bool-like evaluator, so strict `evaluateStaticBoolExpression("not None")` folded to `true` (None → false ⟹ not None → true). Post-PR it now returns `undefined`, contradicting the PR's own semantics table. Impact is low (precision loss only; both branches still analyzed, no false reachability), but it's a silent behavior change for strict callers. Since Python's `not` always forces a truthiness context on its operand, consider evaluating the `not` operand with the bool-like leaf (`_evaluateBoolLikeLiteral`) unconditionally to restore `not None → true` and make strict `not 0/""/[]` precise. If the conservative behavior is intentional, add a one-line comment so the next reader doesn't "re-fix" it.

rchiodo:
Looks solid overall — the shared-walker refactor and literal truthiness handling are clean and well-tested. One subtle strict-mode behavior change in the `not` branch is worth a note/comment before merge, but it's non-blocking (precision-only, no incorrect reachability).

rchiodo (approved):
Approved via Review Center.

velikodniy:
> That seems good to me, but we should double check. Is that expected now because we realized the function never returns?

Regarding the pytest case specifically.

The code is:

```python
    def getitem(
        self, source: str | os.PathLike[str], funcname: str = "test_func"
    ) -> Item:
        items = self.getitems(source)
        for item in items:
            if item.name == funcname:
                return item
        assert 0, f"{funcname!r} item not found in module:\n{source}\nitems: {items}"
```

Before there were two errors:

- Function with declared return type "str" must return value on all code paths
- "None" is not assignable to "str" (reportReturnType)

It's because the type checker couldn't evaluate `0` as falsy. It assumed that assert might not trigger, so the function could return `None` implicitly. This led to the errors above. Now it's clear that `assert` always raises an exception, so the errors has gone.

I'll check other cases and post a brief follow-up.

velikodniy:
> The not branch now threads the strict leaf evaluator (_evaluateBoolConstant) for strict callers.

Ah, I see what you mean! I've updated the code to process the `not` operand as `bool`-like.

velikodniy:
Regarding mypy_primer. Almost all of the changes are expected behaviour:

- **apprise**: the result of `_fetch` is never `None`, it happens only when `target` is truthy but it's `None`
- **pytest**: explained above
- **core**: `while 1` in `find_moov()`
- **pip**: `while 1` in `ask()`
- **werkzeug**: `if 0` caused `Never`
- **sympy**: `while 1`

Though, there's a regression in `zope.interface`. I'm still investigating what caused it

packages/pyright-internal/src/tests/typeEvaluator8.test.ts
```diff
@@ -963,6 +963,12 @@ test('StaticExpression2', () => {
     TestUtils.validateResults(analysisResults, 0);
 });
 
+test('StaticExpressionLiteral1', () => {
+    const analysisResults = TestUtils.typeAnalyzeSampleFiles(['staticExpressionLiteral1.py']);
+
+    TestUtils.validateResults(analysisResults, 4);
```
rchiodo, line 969: 📍 packages/pyright-internal/src/tests/typeEvaluator8.test.ts:969
This asserts 4 errors but the sample produces 7. I verified by running the type checker on `strict_not_folding()` in isolation: it emits exactly 3 `assert_type` mismatches (`not 0`, `not 1`, `not []` → `bool`); combined with `negative()`'s 4, the total is 7, so this test fails CI. Update the expected count to 7 (after fixing the sample assertions below), or — preferably — replace the indirect `assert_type` cases with a direct unit test of `evaluateStaticBoolExpression`/`evaluateStaticBoolLikeExpression`, which are the functions this PR actually changes.

[verified]

packages/pyright-internal/src/tests/samples/staticExpressionLiteral1.py
```diff
@@ -0,0 +1,150 @@
...
+def strict_not_folding() -> None:
+    a = not None
+    assert_type(a, Literal[True])
+
+    b = not 0
+    assert_type(b, Literal[True])
+
+    c = not 1
+    assert_type(c, Literal[False])
+
```
rchiodo: 📍 packages/pyright-internal/src/tests/samples/staticExpressionLiteral1.py:138
These three `assert_type` calls are wrong and produce errors: `not 0`, `not 1`, and `not []` infer `bool`, not `Literal[True]`/`Literal[False]`. I confirmed this against the real type evaluator. pyright only folds `not` to a bool *literal* when the operand's truthiness is known from its type (which is why `not None` and `not not None` below DO pass — `None` is a singleton). Either change these three to `assert_type(..., bool)`, or drop them. Note this function tests the type evaluator's `not`-narrowing, not the `staticExpressions.ts` code this PR changes — consider a direct unit test of the changed functions instead.

[verified]

packages/pyright-internal/src/analyzer/staticExpressions.ts
```diff
@@ -48,19 +101,21 @@ export function evaluateStaticBoolExpression(
...
-            const rightValue = evaluateStaticBoolExpression(
+            const rightValue = _evaluateStaticBoolOrBoolLikeExpression(
                 node.d.rightExpr,
                 execEnv,
                 definedConstants,
                 typingImportAliases,
-                sysImportAliases
+                sysImportAliases,
+                evaluateLeafAsBool
             );
```
rchiodo, line 119: 📍 packages/pyright-internal/src/analyzer/staticExpressions.ts:101
The `not` branch documents why it ignores the caller's `evaluateLeafAsBool` (always real bool), but the `and`/`or` branch silently does the opposite — it threads the caller's callback so strict `1 and 1` stays `undefined`. This divergence is correct but load-bearing and non-obvious; a one-line comment on the `and`/`or` branch (or a sentence contrasting the two in the function header) would prevent a future refactor from "harmonizing" them and regressing strict reachability.

[verified]

rchiodo:
Core change to `staticExpressions.ts` looks solid. The blocker is the new test: `strict_not_folding()` adds 3 unexpected `assert_type` errors (`not 0/1/[]` infer `bool`, not literal), so the file emits 7 errors while the test expects 4. Fix the sample assertions and the expected count together — ideally with a direct unit test of the changed functions.

velikodniy:
A quick update.

> This asserts 4 errors but the sample produces 7

I believe it's false-positive. It's 4 errors when I run locally and in the CI test jobs.

> These three assert_type calls are wrong and produce errors: not 0, not 1, and not [] infer bool

All of them produce `True`/`False` deterministically, so `bool` wouldn't work.

> replace the indirect assert_type cases with a direct unit test of evaluateStaticBoolExpression/evaluateStaticBoolLikeExpression, which are the functions this PR actually changes

My intention was to add high-level property-based tests. Though, I can add more tests if needed.

Regarding the mypy_primer cases. All of them show improvement in type inference, so it's a win. The only exception is the zope case. I spent some time yesterday trying to reproduce it. It looks like type inference is fragile.

The PR makes pyright statically treat the empty-dict literal in `attrs = attrs or {}` as falsy. This perturbs the control-flow graph of `InterfaceClass.__new__` somehow and tips its recursive type inference from the permissive `Any | Unknown` fallback onto a concrete instance type for `Interface`. So that every class `IFoo(Interface)` now wrongly errors with `"Argument to class must be a base class"`.

@rchiodo, seemingly the comments about the tests are false-positive and the zope case is an issue not related to this PR (even though it was triggered by it). Shall we merge? What do you think?

rchiodo:
Core change to staticExpressions.ts looks sound, but the test won't pass: strict_not_folding() relies on pyright folding `not <literal>` to a bool literal, which it does not do (only actual bool-literal operands fold). That function needs to be removed or rewritten before merge.

rchiodo:
Core feature change to `staticExpressions.ts` looks reasonable, but the new test is broken: `strict_not_folding()` asserts literal `not`-folding that pyright's type evaluator does not perform (it folds `not` to a bool literal only when the operand's truthiness is known from its *type*, which holds for `None` but not for `0`, `[]`, `{}`, etc.). The expected error count `4` matches `negative()` alone; the extra `assert_type` failures make the total 11. Easiest fix: drop `strict_not_folding()` (or replace it with a direct unit test of `evaluateStaticBoolExpression`/`evaluateStaticBoolLikeExpression`, the functions this PR actually changes), keeping the expected count at 4.

velikodniy:
> but the test won't pass

> The expected error count 4 matches negative() alone; the extra assert_type failures make the total 11.

The tests are green in the CI workflows

rchiodo:
@velikodniy, sorry but most of the code review comments on this are generated by AI. It can sometimes get stuck in a logical loop.

I'm running the tests now. If they're green, I'll submit.

rchiodo:
Thanks for the PR :)

PR #11522, opened 2026-06-24
Pull Pylance with Pyright 1.1.410

bschnurr:


build/azuredevops/azure-pipelines.yml
```diff
@@ -41,9 +41,9 @@ extends:
                 fetchTags: true
                 persistCredentials: True
               - task: NodeTool@0
-                displayName: Use Node 22.x
+                displayName: Use Node 24.15.0
                 inputs:
-                  versionSpec: 22.x
+                  versionSpec: 24.15.0
```
rchiodo, line 46: 📍 build/azuredevops/azure-pipelines.yml:44
Node version single-source-of-truth is now broken (raised independently by Advocate, Skeptic, and Architect). The pipelines hard-code `versionSpec: 24.15.0` (here and in `azure-pipelines-release.yml`), but the repo root `.nvmrc` still reads `22` (verified) and `@types/node` is bumped to `^25`. This contradicts the documented rule: "Always use the Node.js version specified in `.nvmrc`. Do not hard-code a Node version." Result: CI builds on Node 24, local `nvm use` selects Node 22, and the type surface describes Node 25 — a green CI no longer matches the declared dev runtime. Fix in this PR: bump `.nvmrc` to the intended major (24) and align `@types/node`; ideally have the pipeline read from `.nvmrc` rather than hard-coding the version in four NodeTool tasks.

[verified]

rchiodo (approved):
Approved via Review Center.

rchiodo (approved):
Approved via Review Center.

heejaechang (approved):

rchiodo:
GitHub cannot anchor PR review comments to unchanged lines in the diff. Falling back to a general PR comment for [packages/pyright-internal/src/languageService/completionProvider.ts:L460](https://github.com/microsoft/pyright/blob/bf586c17dd2c74691b16350a7938a5b297271d0d/packages/pyright-internal/src/languageService/completionProvider.ts#L460).

Moving MRU recording off passive `resolveCompletionItem` into `recordCompletionAccepted` correctly fixes per-keystroke oscillation, but `recordCompletionAccepted`'s only in-repo caller is the unit test. Its production caller (the accept command) lives in Pylance's `asynchronousFeatures.ts`, which isn't part of standalone Pyright. As a result, the pyright standalone server's `_mostRecentCompletions` never grows and MRU-based ranking becomes inert for standalone Pyright consumers (previously fed via the `completionItem/resolve` flow). Either wire an accept/resolve path in the pyright server or document MRU as Pylance-only so the dead `static` isn't mistaken for live behavior.

rchiodo (approved):
Approved via Review Center.

PR #11531, opened 2026-06-30
[WIP] Harden source file lifetime boundaries

bschnurr:
## Summary

This PR tightens source-file lifetime boundaries so parse trees, tokenizer output, source text, import results, and evaluator-owned parse-node references are released at the correct invalidation and cleanup points while preserving Pyright's established incremental-edit semantics.

Key changes:
- Preserve the historical distinction between direct open state updates and language-server edit notifications:
  - direct `Program.setFileOpened` updates open contents and does not itself recreate the evaluator;
  - `BackgroundAnalysisProgram.updateOpenFileContents` still follows the previous behavior of calling `markFilesDirty([uri], true)` after `setFileOpened`, so normal editor updates dirty dependents and recreate the evaluator through the existing dirty path.
  - The edited file drops stale source-owned syntax immediately and reparses with globally unique parse-node IDs.
  - New files and repeated pre-analysis direct opens avoid evaluator churn.
- Release full source-owned syntax-tier state at explicit release boundaries: parser output, tokenizer output, parsed source text, module symbol table, source-owned import results, and ignore-comment maps.
- Clear evaluator-owned retainers more completely when the evaluator is explicitly invalidated, with cache stats/probes to verify old parse generations are unloaded on those boundaries.
- Treat file removal and program disposal as explicit lifetime boundaries:
  - removed files are collected with `_isFileNeeded`-compatible reachability before compaction and detached in one graph pass;
  - surviving chained/notebook dependents are dirtied when their chain edge is removed;
  - program disposal clears evaluator/source retainers and edit-mode pre-edit snapshots.
- Clear import resolver caches under memory pressure via `Program.emptyCache()`, including `_cachedPythonSearchPaths`, import-result caches, module-name caches, parent-directory caches, stdlib module caches, and filesystem caches.
- Harden edit-mode/background-analysis teardown so edit-created files, parked background analysis workers, and abandoned edit-mode snapshots do not retain source state.

Behavioral note:
- `markAllFilesDirty(true)` now recreates the evaluator when it actually dirties files. Previously the `evenIfContentsAreSame` path could leave the old evaluator alive because the recursive dirty set stayed empty.
- Ordinary close of still-needed tracked/imported files preserves syntax/evaluator state; syntax release happens through explicit cache-pressure, removal, content replacement, and disposal boundaries.

## Validation

- `cd packages\pyright-internal && npm run build --silent`
- `cd packages\pyright-internal && npx jest service.test sourceFileLifetime.probe.test --forceExit --runInBand`
- `cd packages\pyright-internal && npx jest service.test chainedSourceFiles.test --forceExit`
- `cd packages\pyright-internal && npx jest workspaceEditUtils.test --forceExit`
- `cd packages\pyright-internal && npx jest importResolver.test importResolverSupport.test --forceExit`
- Targeted Prettier and ESLint checks for touched files
- Opt-in source lifetime heap probe with `PYRIGHT_RUN_HEAP_PROBES=1` and Node `--expose-gc`
- Synthetic 1000-file probe: `markAllFilesDirty(true)` ~0.8 ms; closed bulk-untrack ~4.7 ms with zero remaining program files
- Compared local CLI against `pyright@1.1.411` on `Q:\dev\numpy`; diagnostics matched exactly


packages/pyright-internal/src/analyzer/program.ts
```diff
@@ -444,19 +486,43 @@ export class Program {
...
+        const contentsChanged = sourceFileInfo.sourceFile.setClientVersion(version, contents);
+        if (wasKnownFile && (!wasOpenByClient || contentsChanged)) {
+            // Reset the diagnostic version so we force an update to the diagnostics,
+            // which can change based on whether the file is open or its contents changed.
+            // We do not set the version to undefined here because that implies there are
+            // no diagnostics currently reported for this file.
+            sourceFileInfo.diagnosticsVersion = 0;
+        }
+
+        if (wasKnownFile && (hadObservedContents || hadAnalysisState || hadDependencyGraph) && contentsChanged) {
```
bschnurr: This is the main ownership shift for open-file edits: `setFileOpened` now owns dirtying dependents and evaluator invalidation, so callers like `updateOpenFileContents` do not need to separately call `markFilesDirty`. The guards keep brand-new or never-observed files from over-invalidating the evaluator.

packages/pyright-internal/src/analyzer/program.ts
```diff
@@ -1209,46 +1313,16 @@ export class Program {
...
 
-                // Remove any shadowed files corresponding to this file.
-                fileInfo.shadowedBy.forEach((shadowedFile) => {
-                    shadowedFile.mutate((s) => (s.shadows = s.shadows.filter((f) => f !== fileInfo)));
-                });
-                fileInfo.mutate((s) => (s.shadowedBy = []));
+                // A removed file can still be reachable through live dependency graph
+                // edges. Detach it before dropping the list/map entry so parse trees
+                // and source-owned import results do not remain rooted by importers.
+                this._detachSourceFileInfoForRemoval(fileInfo);
```
bschnurr: This detachment is important for lifetime cleanup. Removing a file from `_sourceFileList`/`_sourceFileMap` is not enough if live importers still reference its `SourceFileInfo` through dependency, shadow, chained, or builtins edges.

packages/pyright-internal/src/analyzer/sourceFile.ts
```diff
@@ -475,13 +493,37 @@ export class SourceFile {
...
+
+    releaseClosedFileSyntax(): boolean {
+        if (this._writableData.isBindingInProgress || this._writableData.isCheckingInProgress) {
+            return false;
+        }
+
+        // These fields all belong to the full-syntax tier. Diagnostics and dependency
+        // graph summaries have already been materialized elsewhere, so closed files can
+        // release these objects without losing the compact state Program still needs.
+        const hadSyntax =
```
bschnurr, line 509: This method defines the "full syntax tier" boundary. After diagnostics and compact dependency summaries have been materialized, closed files can release parser output, tokenizer output, parsed source text, source-owned imports, and ignore-comment maps without losing the state `Program` still needs.

packages/pyright-internal/src/analyzer/importResolver.ts
```diff
@@ -139,6 +149,10 @@ export class ImportResolver {
     }
 
     invalidateCache() {
+        // This is used both for configuration/filesystem invalidation and for
+        // Program.emptyCache() memory pressure cleanup, so every resolver-owned
+        // cache that can retain import results or search-path metadata belongs here.
+        this._cachedPythonSearchPaths = undefined;
```
bschnurr, line 155: `_cachedPythonSearchPaths` is included here because `invalidateCache` is now also used by `Program.emptyCache()` for memory-pressure cleanup. Search-path metadata can be expensive and should follow the same cache lifetime as resolver import/module caches.

bschnurr (commented):
Added reviewer notes for the less obvious lifetime and cache-invalidation changes.

packages/pyright-internal/src/analyzer/backgroundAnalysisProgram.ts
```diff
@@ -255,8 +254,26 @@ export class BackgroundAnalysisProgram {
 
         this._disposed = true;
         this._program.dispose();
-        this._backgroundAnalysis?.shutdown();
-        this._backgroundAnalysis?.dispose();
+
+        // Edit mode parks the active background analysis in _preEditAnalysis.
```
bschnurr, line 259: Edit mode can park the active background analysis in `_preEditAnalysis`. Disposal must collect both active and parked analyses so the background worker is not retained after the foreground program has disposed its source state.

packages/pyright-internal/src/analyzer/program.ts
```diff
@@ -1265,9 +1339,62 @@ export class Program {
             }
         }
```
heejaechang, line 1316: 📍 packages/pyright-internal/src/analyzer/program.ts:1296
The rewrite combines `_detachSourceFileInfoForRemoval` — which does a **full `this._sourceFileList.forEach` scan with `.includes()`/`.filter()` per removed file** — with an unconditional `i = 0;` loop restart after every removal. For bulk removals (`setTrackedFiles([])`, large-workspace refresh) this is O(n²)–O(n³), regressing the teardown/branch-switch paths this PR targets. The baseline was a single in-place forward scan. Prefer a single-pass edge unlink over the bounded edge sets (`imports`/`importedBy`/`shadows`/`shadowedBy`/`chainedSourceFile`) without resetting `i` to 0. Please also benchmark bulk-untrack latency, not just diagnostic equality.

packages/pyright-internal/src/analyzer/sourceFile.ts
```diff
@@ -490,8 +532,7 @@ export class SourceFile {
         this._writableData.noCircularDependencyConfirmed = false;
         this._writableData.isCheckingNeeded = true;
```
heejaechang, line 534: 📍 packages/pyright-internal/src/analyzer/sourceFile.ts:527
`markDirty` previously cleared only `moduleSymbolTable`+`lineCount`, **retaining** `parserOutput`/`tokenizerOutput`/`parsedFileContents`. It now calls `_releaseSyntaxCaches(false)`, dropping the parse tree. Because `markAllFilesDirty(true)` and `markFilesDirty(...)` call `markDirty()` directly on every listed file, config/library/builtins refreshes now force a **full reparse** instead of rebind-without-reparse. Consider isolating syntax-drop into a distinct method (e.g. `markDirtyAndDropSyntax`) used only on close/cache-pressure, and add a test asserting parse-tree retention across `markAllFilesDirty` for unchanged files.

heejaechang:
GitHub cannot anchor PR review comments to unchanged lines in the diff. Falling back to a general PR comment for [packages/pyright-internal/src/analyzer/program.ts:L530](https://github.com/microsoft/pyright/blob/5c408e2d87f729f9e02ddbae27d9c61ec4736c80/packages/pyright-internal/src/analyzer/program.ts#L530).

📍 packages/pyright-internal/src/analyzer/program.ts:502
Potential reopen gap: `setFileClosed` → `releaseClosedFileSyntax()` sets `isBindingNeeded = true` but not `isCheckingNeeded`. On reopen with identical contents, `setFileOpened` sees `contentsChanged === false` and skips both evaluator recreation and dirtying. The added tests cover close-without-reopen and reopen-with-changes, but not close → reopen with **same** contents → diagnostics still present. Add that test (open a file with a type error, close, reopen identical text, drain `analyze()`, assert the diagnostic reappears) to lock down that reopen reliably restores full diagnostics.

packages/pyright-internal/src/tests/service.test.ts
```diff
@@ -682,6 +761,770 @@ test('file changes cause semantic update', () => {
...
+    assert.strictEqual(sourceFileInfo.sourceFile.isCheckingRequired(), false);
+    assert.strictEqual(sourceFileInfo.sourceFile.didContentsChangeOnDisk(), false);
+});
+
+test('setFileClosed drops tracked syntax without invalidating evaluator caches', () => {
+    const code = `
+// @filename: test.py
+//// # module lead
+//// # lifetime comment retained on token
+//// def f():
```
heejaechang, line 812: 📍 packages/pyright-internal/src/tests/service.test.ts:793
In `setFileClosed drops tracked syntax without invalidating evaluator caches`, `assert(writableData.parsedFileContents?.includes('lifetime comment retained on token'))` is a partial `.includes()` check on fully-deterministic content defined by the `////` fourslash lines in the same test. Replace with `assert.strictEqual(writableData.parsedFileContents, <exact expected string>)` — capture the exact value first.

packages/pyright-internal/src/tests/service.test.ts
```diff
@@ -682,6 +761,770 @@ test('file changes cause semantic update', () => {
...
+        if (name !== 'evaluatorGeneration') {
+            assert.strictEqual(value, 0, `${name} should be cleared on evaluator disposal`);
+        }
+    });
+    assert((program.evaluator as any).getEvaluatorCacheStats().evaluatorGeneration > oldEvaluatorGeneration);
+    assert.strictEqual(sourceFileInfo.sourceFile.getParserOutput(), undefined);
+});
+
+test('setFileOpened defers evaluator disposal during edit mode', () => {
+    const code = `
```
heejaechang, line 1171: 📍 packages/pyright-internal/src/tests/service.test.ts:1108
In `emptyCache drops retained parse tree and parsed contents for closed tracked file`, `assert(writableData.parsedFileContents?.includes('lifetime source text'))` is again a partial `.includes()` check on a fully-known fourslash-defined string. Replace with an exact `assert.strictEqual` comparison.

.github/plan/source-file-lifetime-redesign.md
```diff
@@ -0,0 +1,269 @@
+# Source file lifetime and memory redesign
```
heejaechang: 📍 .github/plan/source-file-lifetime-redesign.md:1
Two concerns: (1) These docs land in `.github/plan/` (singular), but the repo plan convention is `.github/plans/` (see `.github/plans/README.md`, `archived/`, `long_running/`). As-is they won't be discovered by anyone following the documented plan workflow. (2) The redesign doc embeds changelog-style stats (diff sizes `+426/-42`, todo counts, "Lifetime-model doc size: 83 lines" while the model doc is actually 103 lines — already internally inconsistent). Per the repo documentation-writing policy, drop the stale metrics and move the docs to `.github/plans/`.

rchiodo:
GitHub cannot anchor PR review comments to unchanged lines in the diff. Falling back to a general PR comment for [packages/pyright-internal/src/tests/service.test.ts:L1](https://github.com/microsoft/pyright/blob/5c408e2d87f729f9e02ddbae27d9c61ec4736c80/packages/pyright-internal/src/tests/service.test.ts#L1).

Two `.includes()` partial assertions on `parsedFileContents` violate the repo's `tests-no-partial-asserts` rule (docs/pylancewiki/review/tests-no-partial-asserts.md), which explicitly prohibits `.includes(...)`: `assert(writableData.parsedFileContents?.includes('lifetime comment retained on token'))` (in `setFileClosed drops tracked syntax...`) and `...includes('lifetime source text')` (in the `emptyCache` test). Since these are presence/retention preconditions, the simplest compliant fix is a non-partial existence check (`assert(writableData.parsedFileContents !== undefined)`), or an exact `assert.strictEqual(writableData.parsedFileContents, <full expected contents>)` if the specific text actually matters to the assertion.

packages/pyright-internal/src/analyzer/program.ts
```diff
@@ -1265,9 +1339,62 @@ export class Program {
...
+        // shadow files, chained files, and builtins. Clear both sides so removal
+        // is a graph operation, not just a deletion from Program's file table.
+        fileInfo.imports.forEach((importedFile) => {
+            importedFile.mutate((s) => (s.importedBy = s.importedBy.filter((fi) => fi !== fileInfo)));
+        });
+
+        this._sourceFileList.forEach((sourceFileInfo) => {
+            if (sourceFileInfo === fileInfo) {
+                return;
+            }
```
rchiodo: Potential performance regression in `_removeUnneededFiles`: it now restarts the outer scan with `i = 0` after every removal, and `_detachSourceFileInfoForRemoval` walks the entire `_sourceFileList` (with multiple `.includes()` per element) once per removed file. Removing k of n files becomes roughly O(k·n·E) versus the previous single linear pass with targeted edge unlink. This path runs at the top of `getDiagnostics`, so bulk-removal events (`setTrackedFiles([])`, branch switch, venv change) can stall on large workspaces. Prefer collecting victims first and doing one bidirectional edge-rebuild/compaction pass, unlinking only via the existing `imports`/`importedBy`/`shadowedBy`/`chainedSourceFile` back-references instead of a full-list scan. At minimum, quantify the cost on a large workspace before relying on the restart approach.

packages/pyright-internal/src/analyzer/program.ts
```diff
@@ -1265,9 +1339,62 @@ export class Program {
...
+                sourceFileInfo.imports.includes(fileInfo) ||
+                sourceFileInfo.importedBy.includes(fileInfo) ||
+                sourceFileInfo.shadows.includes(fileInfo) ||
+                sourceFileInfo.shadowedBy.includes(fileInfo) ||
+                sourceFileInfo.builtinsImport === fileInfo ||
+                sourceFileInfo.chainedSourceFile === fileInfo
+            ) {
+                sourceFileInfo.mutate((s) => {
+                    s.imports = s.imports.filter((fi) => fi !== fileInfo);
+                    s.importedBy = s.importedBy.filter((fi) => fi !== fileInfo);
```
rchiodo: `_detachSourceFileInfoForRemoval` now clears `chainedSourceFile` on surviving files (e.g. file B chained to the removed file) but never marks B dirty. The old code left the pointer dangling and didn't act on it, so this isn't a strict regression — but now that the edge is actively broken, B can keep analysis computed under the old chain (notebook-cell name resolution into a now-removed earlier cell). `updateChainedUri` dirties dependents on chain change; this removal path should do the same. Narrow to chained/notebook scenarios.

.github/plan/source-file-lifetime-redesign.md
```diff
@@ -0,0 +1,269 @@
...
+| Heap proof | Rooted opt-in WeakRef/GC probe proves old syntax/source/import/declaration objects and old evaluator `TypeResult`s collect across whole-evaluator invalidation paths. | The change has object-collection evidence, not just cache-size assertions. |
+
+Tracked stats for this slice:
+
+| Metric | Value | Notes |
+| --- | ---: | --- |
+| Completed investigation todos | 28 | Evaluator/source-lifetime/duplicate-data/over-invalidation/memory-scan/set-open/open-update/lifecycle-hardening/cache-pressure/removal-dispose/import-diagnostic-resource/diagnostic-cache/cache-usage-owner investigation items tracked in session state. |
+| Evaluator cache/stat fields exposed | 23 | `EvaluatorCacheStats` covers generation, maps, entry counts, stacks, speculative state, and prefetched types. |
+| Import resolver cache/stat fields exposed | 11 | `ImportResolverCacheStats` and `ParentDirectoryCacheStats` cover import results, module-name results, python search paths, stdlib modules, and parent-directory cache entries. |
```
rchiodo: These two planning docs read as private working notes rather than durable design docs: they contain machine-local absolute paths (`Q:\dev\numpy`, `Q:\dev\benchmark-lsp\heap-proof\...`), session bookkeeping ("Completed investigation todos: 28"), and point-in-time stat tables that go stale immediately (the meta-doc claims the model doc is 83 lines while the diff adds 103). The folder name `.github/plan/` is also nonstandard — the repo convention is `.github/plans/`. Consider trimming these to a durable, path-free design note (and moving under `.github/plans/`), or keeping them out of the PR.

packages/pyright-internal/src/analyzer/program.ts
```diff
@@ -502,47 +579,67 @@ export class Program {
...
+        if (
+            fileUris.some((fileUri) => {
+                const fileName = fileUri.fileName;
+                return (
+                    (fileName === 'builtins.pyi' || fileName === '__builtins__.pyi') &&
+                    this.getSourceFileInfo(fileUri) !== undefined
+                );
+            })
+        ) {
+            this.markAllFilesDirty(evenIfContentsAreSame);
```
rchiodo, line 608: The builtins predicate `fileName === 'builtins.pyi' || fileName === '__builtins__.pyi'` is now duplicated — once in `markFilesDirty` and twice in `setFileOpened` (the base had it once). Consider extracting an `isBuiltinsModuleFile(fileName)` helper to keep a single source of truth.

rchiodo:
Solid, well-tested lifetime-hardening work with strong heap-probe evidence. Before merge, please address the partial-assertion rule violation in the new tests, and take a look at the O(k·n) removal rescan and the chained-file dirtying gap. The remaining notes are non-blocking.

packages/pyright-internal/src/tests/service.test.ts
heejaechang, line 1171: `assert(writableData.parsedFileContents?.includes('lifetime source text'))` is again a partial `.includes()` check on deterministic fourslash-defined content, violating `tests-no-partial-asserts.md`. Replace with an exact `assert.strictEqual` comparison.

packages/pyright-internal/src/analyzer/sourceFile.ts
```diff
@@ -490,8 +532,7 @@ export class SourceFile {
...
         this._writableData.isBindingNeeded = true;
-        this._writableData.moduleSymbolTable = undefined;
-        this._writableData.lineCount = undefined;
+        this._releaseSyntaxCaches(/* preserveLineCount */ false);
 
         this._fireFileDirtyEvent();
```
heejaechang, line 539: `markDirty()` now calls `_releaseSyntaxCaches(/* preserveLineCount */ false)`, dropping `parserOutput`/`tokenizerOutput`/`parsedFileContents`/`imports` where the baseline only cleared `moduleSymbolTable`+`lineCount`. Because `markAllFilesDirty(true)`/`markFilesDirty(...)` call `markDirty()` directly, config/library/builtins refreshes now force a full reparse of directly-dirtied files. This appears intentional, but consider isolating the syntax-drop into a close/cache-pressure-only path and adding a parse-tree-retention test across `markAllFilesDirty` to bound the blast radius.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -687,10 +690,12 @@ export function createTypeEvaluator(
...
             return result.finally(() => {
                 cancellationToken = oldToken;
+                completePendingDisposalIfInactive();
             });
         } finally {
             if (!isThenable(result)) {
                 cancellationToken = oldToken;
+                completePendingDisposalIfInactive();
             }
         }
```
heejaechang, line 704: `completePendingDisposalIfInactive()` is invoked only from `runWithCancellationToken`'s unwind. If `disposeEvaluator()` ever fires with a non-empty stack on a path that does not unwind through `runWithCancellationToken`, `isDisposalPending` stays `true` forever and `deferredClassCompletions`/`returnTypeInferenceTypeCache`/`prefetched`/speculative state/stacks are never cleared — silently reintroducing the leak this PR targets. The invariant "every non-empty stack implies we are inside `runWithCancellationToken`" is load-bearing and undocumented; assert it or drain pending disposal from a second site.

packages/pyright-internal/src/analyzer/program.ts
```diff
@@ -502,47 +579,67 @@ export class Program {
...
     }
 
     markFilesDirty(fileUris: Uri[], evenIfContentsAreSame: boolean) {
+        if (
+            fileUris.some((fileUri) => {
+                const fileName = fileUri.fileName;
+                return (
+                    (fileName === 'builtins.pyi' || fileName === '__builtins__.pyi') &&
+                    this.getSourceFileInfo(fileUri) !== undefined
+                );
```
heejaechang: The old code recreated the evaluator only `if (markDirtySet.size > 0)`, which stayed empty in the `evenIfContentsAreSame` branch — so `markAllFilesDirty(true)` previously did not recreate the evaluator. The new `foundDirtyFile` gating makes it always recreate. This is probably more correct, but it adds an evaluator-recreate to several hot refresh paths; please call it out explicitly in the PR description.

heejaechang:
Solid lifetime-hardening work with strong evidence (the WeakRef/GC probe is excellent). Before merge: (1) replace the two partial `.includes()` assertions on deterministic fourslash content with exact comparisons; (2) reconsider the O(n^2) detach+`i=0` restart on bulk-removal paths this PR targets; (3) move the plan docs from `.github/plan/` to the conventional `.github/plans/` and drop the stale metrics table.

packages/pyright-internal/src/analyzer/importResolver.ts
```diff
@@ -80,6 +80,16 @@ export function createImportedModuleDescriptor(moduleName: string): ImportedModu
...
 type CachedImportResults = Map<string, ImportResult>;
 
+export interface ImportResolverCacheStats {
+    cachedImportResultRoots: number;
+    cachedImportResults: number;
+    cachedModuleNameRoots: number;
+    cachedModuleNameResults: number;
+    cachedPythonSearchPaths: number;
+    stdlibModules: number;
+    parentDirectoryCache: ParentDirectoryCacheStats;
```
rchiodo, line 90: 📍 packages/pyright-internal/src/analyzer/importResolver.ts:83
Unresolved (`exports-only-when-used`): the Rules reviewer confirmed against base that `ImportResolverCacheStats` is new in this PR and no file outside this module imports it by name — `service.test.ts`/`importResolver.test.ts` consume `getCacheStats()` structurally without a named type import, and no public-API documentation exception applies. Either drop `export` (keep it module-local) or add a type-annotated external consumer. (`ParentDirectoryCacheStats` and `EvaluatorCacheStats` are fine — both have named-import consumers.)

[verified]

.github/plan/source-file-lifetime-redesign.md
rchiodo: 📍 .github/plan/source-file-lifetime-redesign.md:1
Unresolved from prior review (rchiodo + heejaechang). These two docs are committed under `.github/plan/` (singular) while the repo convention is `.github/plans/`, so they won't be discovered by the documented plan workflow. They also embed machine-local paths (`Q:\dev\numpy`, `Q:\dev\benchmark-lsp\heap-proof\…`), session bookkeeping ("Completed investigation todos: 28"), and self-contradicting stats ("Lifetime-model doc size: 83 lines" while the model doc is 103 lines in this same diff). Trim to a durable, path-free design note and move under `.github/plans/`, or keep them out of the PR.

[verified]

packages/pyright-internal/src/analyzer/sourceFile.ts
```diff
@@ -490,8 +532,7 @@ export class SourceFile {
         this._writableData.noCircularDependencyConfirmed = false;
         this._writableData.isCheckingNeeded = true;
         this._writableData.isBindingNeeded = true;
-        this._writableData.moduleSymbolTable = undefined;
-        this._writableData.lineCount = undefined;
+        this._releaseSyntaxCaches(/* preserveLineCount */ false);
```
rchiodo, line 538: 📍 packages/pyright-internal/src/analyzer/sourceFile.ts:533
Unresolved and broader than first reported (heejaechang): `markDirty()` now calls `_releaseSyntaxCaches(/* preserveLineCount */ false)`, dropping `parserOutput`/`tokenizerOutput`/`parsedFileContents` where the base cleared only `moduleSymbolTable`+`lineCount`. The Skeptic/Architect verified the impact: because `markAllFilesDirty(true)` calls `markDirty()` on every file (and the new `foundDirtyFile` gate then recreates the evaluator), a single routine config/library/builtins/venv event now forces a **whole-workspace reparse AND discards the type cache**, versus base's rebind-without-reparse with evaluator retained. This is the highest-impact regression in the PR. Split full-syntax release into a distinct `markDirtyAndDropSyntax()` used only by close/removal/cache-pressure, keep `markDirty` parse-retaining, and add a regression test asserting parse-tree retention across `markAllFilesDirty(true)` for unchanged files.

[verified]

packages/pyright-internal/src/analyzer/program.ts
rchiodo, line 1316: 📍 packages/pyright-internal/src/analyzer/program.ts:1297
Unresolved from prior review (rchiodo + heejaechang, all three non-Rules reviewers re-confirmed): `_removeUnneededFiles` still resets the scan with `i = 0` after every removal, and `_detachSourceFileInfoForRemoval` still does a full `this._sourceFileList.forEach` with per-element `.includes()`/`.filter()` per removed file — ≈O(k·n·E) at the top of `getDiagnostics`, so bulk events (`setTrackedFiles([])`, branch switch, venv change) can stall large workspaces. **Important fixer nuance from the Skeptic:** the `i = 0` restart is load-bearing (it replaced the deleted inline orphan-import cascade removal), so do NOT just delete it — that would leak orphaned imports at already-passed indices. Collect victims first, then do one bidirectional edge-unlink/compaction pass over the bounded back-references (`imports`/`importedBy`/`shadows`/`shadowedBy`/`chainedSourceFile`) without resetting `i`. Quantify bulk-untrack latency, not just diagnostic equality.

[verified]

packages/pyright-internal/src/analyzer/program.ts
```diff
@@ -1265,9 +1339,62 @@ export class Program {
...
+
+        fileInfo.mutate((s) => {
+            s.imports = [];
+            s.importedBy = [];
+            s.shadows = [];
+            s.shadowedBy = [];
+            s.builtinsImport = undefined;
+            s.chainedSourceFile = undefined;
+            s.effectiveFutureImports = undefined;
+        });
```
rchiodo: 📍 packages/pyright-internal/src/analyzer/program.ts:1372
Unresolved (rchiodo): `_detachSourceFileInfoForRemoval` clears `s.chainedSourceFile = undefined` on surviving files but never marks them dirty. While `_removeUnneededFiles` recreates the evaluator (clearing type-cache staleness), the survivor isn't `markReanalysisRequired`'d, so a notebook cell B chained to a now-removed earlier cell can retain bind/check state computed under the broken chain until an unrelated edit. `updateChainedUri` already dirties dependents on chain change; mirror that here. Narrow to chained/notebook (CellDocs) scenarios and not a strict regression, so Low priority.

[verified]

packages/pyright-internal/src/analyzer/program.ts
rchiodo, line 608: 📍 packages/pyright-internal/src/analyzer/program.ts:635
Unresolved (rchiodo): the builtins predicate `fileName === 'builtins.pyi' || fileName === '__builtins__.pyi'` now appears 3× (once in `markFilesDirty`, twice in `setFileOpened`), up from one occurrence in base. Extract an `isBuiltinsModuleFile(fileName)` helper for a single source of truth.

[verified]

rchiodo:
Overall the lifetime-boundary work is well-tested and the heap probes are convincing. The main things to resolve before merge: (1) the committed `.github/plan/*` docs embed machine-local paths and self-contradicting stats and use a non-standard `.github/plan/` path — trim/relocate or drop them; (2) `markDirty` now releases full syntax (parse tree + type cache) on every dirty, which via `markAllFilesDirty(true)` can force whole-workspace reparse on routine config/library/builtins events — please validate this isn't a latency regression and consider splitting close/cache-pressure release from `markDirty`; (3) `_removeUnneededFiles` bulk-removal complexity; and (4) a few test-hygiene / unused-export items below.

bschnurr:
Addressed the review feedback in `e10a8ab03`:

- Reworked `_removeUnneededFiles` to collect victims first, detach graph edges in one pass, then compact from the end. This removes the `i = 0` restart and per-removed-file full-list scan. A synthetic 1000-file closed bulk-untrack probe completes in ~2.6 ms with no remaining program files.
- Split ordinary `markDirty` from `markDirtyAndDropSyntax`; `markAllFilesDirty(true)` now retains the private parser output while still forcing reanalysis/evaluator recreation. Added regression coverage.
- Added exact `parsedFileContents` assertions in the lifecycle tests.
- Removed the private `.github/plan` docs from the PR rather than moving stale session notes into `.github/plans`.
- Added `isBuiltinsModuleFile` helper for the duplicated builtins predicate.
- Added an assertion documenting that active evaluator disposal must unwind through `runWithCancellationToken`.
- Made `ImportResolverCacheStats` module-local.
- Dirtied surviving chained dependents when removal clears their `chainedSourceFile` edge.

packages/pyright-internal/src/analyzer/program.ts
```diff
@@ -1260,14 +1325,121 @@ export class Program {
...
 
+    private _collectUnneededFiles() {
+        const neededFiles = new Set<SourceFileInfo>();
+
+        const markNeededRecursive = (fileInfo: SourceFileInfo | undefined) => {
+            if (!fileInfo || neededFiles.has(fileInfo) || fileInfo.sourceFile.isFileDeleted()) {
+                return;
+            }
+
+            neededFiles.add(fileInfo);
```
rchiodo: `_collectUnneededFiles` roots reachability only on `isTracked || isOpenByClient` and follows `imports`/`builtinsImport`/`chainedSourceFile`/`shadowedBy`, but never roots on (or follows) `shadows`. The retained `_isFileNeeded` treated `shadows.length > 0` as a standalone "needed" condition, so an orphan shadow/stub file that loses its importer can now be removed where base kept it. This may be intentional (more aggressive lifetime GC), but it's uncovered by tests — `emptyCache preserves live partial stub files` exercises `emptyCache`, not the `_removeUnneededFiles` removal path. Add a regression test that establishes a shadow relationship, removes the importer, and asserts the intended survive/remove outcome (or re-add `shadows.length > 0` as a reachability root). A short comment explaining why removal intentionally bypasses `_isFileNeeded` would also keep the two "needed" definitions from drifting.

rchiodo:
Most prior-review flags appear already resolved in this revision (markDirty no longer drops syntax — that's split into markDirtyAndDropSyntax; _removeUnneededFiles now collects victims first and avoids the i=0 restart; _detachSourceFileInfosForRemoval already marks survivors dirty on chain change; isBuiltinsModuleFile helper extracted; the cited test asserts use exact assert.strictEqual, not partial .includes(); and ImportResolverCacheStats is module-local, not exported). The one remaining concern is the shadow-file reachability narrowing in file removal noted inline.

rchiodo (approved):
Approved via Review Center.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -718,6 +772,57 @@ export function createTypeEvaluator(
         effectiveTypeCache = new Map<number, Map<string, EffectiveTypeResult>>();
```
heejaechang, line 770: `disposeEvaluator()` still wipes the durable maps — including `functionRecursionMap` (the live return-inference recursion guard) and `codeFlowAnalyzerCache` — *before* the new `isEvaluatorActive()` gate. This PR newly makes reentrant disposal a first-class, tested path (`emptyCache()`→`_createNewEvaluator()`→`disposeEvaluator()` while on-stack), so the deferral logic now deliberately preserves `returnTypeInferenceContextStack` et al. via `isDisposalPending` while discarding the map that pairs with them — an internally inconsistent application of the PR's own deferral principle. If a memory-pressure clear lands mid recursive return-type inference, the recursion cutoff resets to 0 for the rest of the in-flight unwind, risking runaway recursion. Fix by either deferring `functionRecursionMap`/`codeFlowAnalyzerCache` alongside the stacks under `isDisposalPending`, or adding their `.size` to `isEvaluatorActive()`. Separately, the deferred-disposal drain fires only from `runWithCancellationToken`'s unwind; consider draining from a second site or documenting the load-bearing invariant. [unverified runtime claim]

packages/pyright-internal/src/analyzer/program.ts
```diff
@@ -502,47 +583,63 @@ export class Program {
...
             }
+
+            sourceFileInfo.sourceFile.releaseClosedFileSyntax();
         }
 
         return this._removeUnneededFiles();
     }
 
-    markAllFilesDirty(evenIfContentsAreSame: boolean) {
+    markAllFilesDirty(evenIfContentsAreSame: boolean, recreateEvaluator = true) {
```
heejaechang, line 574: On an unchanged-contents close, the new unconditional `releaseClosedFileSyntax()` drops the syntax tier but the close path keeps the evaluator and its type cache (the `setFileClosed drops tracked syntax…` test itself asserts `program.evaluator === oldEvaluator` and unchanged stats while nulling `parserOutput`/`moduleSymbolTable`). Type/effective-type entries keyed by the closed file's `node.id`s reference declarations, which reference parse nodes — so those entries can transitively keep the "dropped" parse subtrees alive, meaning the reclaim this PR advertises may not actually happen for the most common lifetime event. The opt-in heap probe covers open/dirty/emptyCache paths but not the plain unchanged-close path. Add a heap probe proving the parse tree is collectable on unchanged close; if it isn't, either clear the closed file's type-cache entries or document the tradeoff explicitly.

packages/pyright-internal/src/analyzer/program.ts
```diff
@@ -1260,14 +1325,121 @@ export class Program {
...
+
+            fileInfo.imports.forEach(markNeededRecursive);
+            markNeededRecursive(fileInfo.builtinsImport);
+            markNeededRecursive(fileInfo.chainedSourceFile);
+            fileInfo.shadowedBy.forEach(markNeededRecursive);
+        };
+
+        this._sourceFileList.forEach((fileInfo) => {
+            if (fileInfo.isTracked || fileInfo.isOpenByClient) {
+                markNeededRecursive(fileInfo);
```
heejaechang: `_collectUnneededFiles` introduces a second, diverging definition of "needed" while `_isFileNeeded`/`_isImportNeededRecursive` remain authoritative for parse/bind/check elsewhere. Two concrete divergences: (1) base `_isFileNeeded` retains any file with `shadows.length > 0` regardless of reachability, but `_collectUnneededFiles` only roots at tracked/open and reaches shadow targets via `shadowedBy` — so an implementation shadowed by a now-unreachable stub gets removed here but kept by the rest of the engine; (2) `markNeededRecursive` early-returns on `isFileDeleted()`, dropping files imported only through a deleted-but-tracked importer, whereas base intentionally keeps them. Both changes may be cleaner, but they are uncalled-out and untested. Either compute the removal set from `_isFileNeeded`, or add shadow + deleted-importer tests proving parity.

packages/pyright-internal/src/analyzer/sourceFileInfo.ts
```diff
@@ -213,6 +225,21 @@ export class SourceFileInfo implements extensibility.SourceFileInfo {
...
+        data.imports = [];
+        data.importedBy.length = 0;
+        data.importedBy = [];
+        data.shadows.length = 0;
+        data.shadows = [];
+        data.shadowedBy.length = 0;
+        data.shadowedBy = [];
+        data.builtinsImport = undefined;
+        data.chainedSourceFile = undefined;
+        data.effectiveFutureImports = undefined;
```
heejaechang, line 272: The full edge inventory (`imports`/`importedBy`/`shadows`/`shadowedBy`/`builtinsImport`/`chainedSourceFile`/`effectiveFutureImports`) is now enumerated independently in three places (`_clearWriteableDataReferences` here and `Program._detachSourceFileInfosForRemoval`'s two branches). Adding a future edge means updating all three or silently leaking a retainer — exactly the failure class this PR exists to prevent. Consider a single `clearAllEdges(writeableData)` helper as the one source of truth.

heejaechang:
No confirmed blockers. Worth addressing before merge: (1) `disposeEvaluator()` wipes durable maps (e.g. `functionRecursionMap`) before the new `isEvaluatorActive()` gate, which is inconsistent with the PR's own reentrant-disposal deferral; (2) the unchanged-close path keeps the evaluator/type cache, which can transitively retain the parse subtrees this PR aims to reclaim — add a heap probe on that path; (3) `_collectUnneededFiles` introduces a second definition of "needed" that may diverge from `_isFileNeeded` for shadowed-impl and deleted-importer cases.

heejaechang (approved):
Approved via Review Center.

rchiodo:
🔒 Automated review in progress — @rchiodo is auto-reviewing this PR via the Review Center.

Other automated reviewers defer to this claim to avoid duplicate, conflicting feedback. The claim is refreshed while the review is active; if it is not refreshed within the ownership timeout it is treated as vacant and another reviewer may take over. This is an automated marker.

rchiodo:
🔒 Automated review in progress — @rchiodo is auto-reviewing this PR via the Review Center.

Other automated reviewers defer to this claim to avoid duplicate, conflicting feedback. The claim is refreshed while the review is active; if it is not refreshed within the ownership timeout it is treated as vacant and another reviewer may take over. This is an automated marker.

packages/pyright-internal/src/analyzer/program.ts
```diff
@@ -502,47 +583,63 @@ export class Program {
...
+                    /* forceRebinding */ false,
+                    /* recreateEvaluator */ false
+                );
+                this._createNewEvaluator();
             }
+
+            sourceFileInfo.sourceFile.releaseClosedFileSyntax();
         }
 
         return this._removeUnneededFiles();
```
rchiodo, line 571: `setFileClosed` now unconditionally calls `releaseClosedFileSyntax()`, which drops `parserOutput`, `tokenizerOutput`, `moduleSymbolTable`, `imports`, and ignore-maps and sets `isBindingNeeded = true` — even when `_removeUnneededFiles` will keep the file because another open file still imports it. The previous code preserved the parse tree for still-needed closed files. Result: closing a heavily-imported module (e.g. tab churn in workspace-diagnostics mode) now forces a re-parse + re-bind on the next `analyze()`. Consider gating `releaseClosedFileSyntax()` on the file actually becoming unneeded (no live importer) rather than every close.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -718,6 +772,57 @@ export function createTypeEvaluator(
...
+        printExpressionSpaceCount = 0;
+        incompleteGenCount = 0;
+        speculativeTypeTracker.clear();
+        symbolResolutionStack.length = 0;
+        suppressedNodeStack.length = 0;
+        assignClassToSelfStack.length = 0;
+        returnTypeInferenceContextStack.length = 0;
+        signatureTrackerStack.length = 0;
+        isDisposalPending = false;
     }
```
rchiodo, line 828: "The evaluator is active" is now encoded in three independent places — `isEvaluatorActive()` (reads `cancellationToken` + 5 stacks + speculative context), `clearInactiveRetainers()` (clears that same set), and the scattered stack-push sites. There is no single source of truth, so adding a new context stack later and forgetting to update `isEvaluatorActive` would either corrupt an in-flight evaluation or silently leak. Additionally, `assert(runWithCancellationTokenDepth > 0, ...)` couples two independently-maintained notions of "active"; if an active-evaluator disposal ever occurs outside a `runWithCancellationToken` scope the assert throws (and `isDisposalPending` would never complete). No confirmed in-repo path triggers this today, but consider consolidating the activity stacks into one registry that both the predicate and the cleaner iterate, and driving deferred completion from the stack-unwind/speculative-pop points rather than only the cancellation `finally`.

packages/pyright-internal/src/analyzer/program.ts
```diff
@@ -502,47 +583,63 @@ export class Program {
...
+        if (recreateEvaluator && (foundDirtyFile || markDirtySet.size > 0)) {
             this._createNewEvaluator();
         }
     }
 
     markFilesDirty(fileUris: Uri[], evenIfContentsAreSame: boolean) {
+        if (
+            fileUris.some((fileUri) => {
+                return isBuiltinsModuleFile(fileUri.fileName) && this.getSourceFileInfo(fileUri) !== undefined;
+            })
```
rchiodo, line 606: `markAllFilesDirty(true)` now recreates the evaluator whenever it dirties any file (via the new `foundDirtyFile` gate). This is a correct latent-bug fix — the old `evenIfContentsAreSame` path left `markDirtySet` empty and never recreated the evaluator, leaving stale analysis alive. But it now fires evaluator recreation on the most common reanalysis triggers (config/library invalidation in `service.ts`, `backgroundAnalysisBase.ts`, `invalidateAndForceReanalysis`). Please add a perf sanity check on a real config/library-change reanalysis flow, not just the numpy diff-parity run, before merge.

packages/pyright-internal/src/analyzer/program.ts
```diff
@@ -1260,14 +1325,121 @@ export class Program {
...
+        if (filesToRemove.size > 0) {
+            const markDirtySet = new Set<string>();
+            if (this._detachSourceFileInfosForRemoval(filesToRemove, markDirtySet)) {
+                shouldRecreateEvaluator = true;
+            }
+
+            for (let i = this._sourceFileList.length - 1; i >= 0; i--) {
+                const fileInfo = this._sourceFileList[i];
+                if (filesToRemove.has(fileInfo)) {
+                    this._removeSourceFileFromListAndMap(fileInfo.uri, i);
```
rchiodo, line 1313: Subtle GC-semantics divergence: the old `_isFileNeeded` kept any file with `shadows.length > 0` alive unconditionally, whereas the new `_collectUnneededFiles` keeps a shadowing file only if it's reachable from a tracked/open root. In practice a shadowing stub is itself imported (so it stays reachable — confirmed by the `emptyCache preserves live partial stub files` test), but equivalence was not exhaustively proven for a shadower whose only inbound edge is the `shadows` relation. Suggest adding a test that constructs `F.shadows = [B]` with no root importing `F` and `B` otherwise unreachable, asserting old and new keep/drop `F` identically.

packages/pyright-internal/src/analyzer/typeEvaluatorTypes.ts
```diff
@@ -415,6 +415,32 @@ export interface ExpectedTypeResult {
...
+    speculativeTypeCache: number;
+    speculativeTypeCacheEntries: number;
+    activeDependentTypes: number;
+    symbolResolutionStack: number;
+    suppressedNodeStack: number;
+    assignClassToSelfStack: number;
+    returnTypeInferenceContextStack: number;
+    signatureTrackerStack: number;
+}
+
```
rchiodo, line 443: The `getEvaluatorCacheStats`/`*CacheStats` surface is baked into shipped production interfaces (`TypeEvaluator`, `ImportResolver`, `ParentDirectoryCache`, `SpeculativeTypeTracker`) but has zero production callers — it exists to let leak tests assert "every field except `evaluatorGeneration` is 0 after disposal." That makes the stats struct an implicit "every retainer" contract: good for leak detection, but it requires authors to add new caches in 3+ places and ships a test-instrumentation method on the public interface. Consider moving it behind a dev/test-only surface.

packages/pyright-internal/src/analyzer/program.ts
```diff
@@ -1260,14 +1325,121 @@ export class Program {
...
+            if (
+                imports.length !== sourceFileInfo.imports.length ||
+                importedBy.length !== sourceFileInfo.importedBy.length ||
+                shadows.length !== sourceFileInfo.shadows.length ||
+                shadowedBy.length !== sourceFileInfo.shadowedBy.length ||
+                builtinsImport !== sourceFileInfo.builtinsImport ||
+                chainedSourceFileChanged
+            ) {
+                sourceFileInfo.mutate((s) => {
+                    s.imports = imports;
```
rchiodo: `_detachSourceFileInfosForRemoval` builds four filtered arrays (`imports`/`importedBy`/`shadows`/`shadowedBy`) for every surviving file on every removal pass and only commits if a length changed. On a large program where few files are removed this is O(total edges) allocation per pass. The measured 2.6 ms/1000-removals suggests it's tolerable, but a cheap pre-check (does this survivor reference any removed file?) before allocating would avoid the churn. Not a correctness issue.

packages/pyright-internal/src/analyzer/sourceFile.ts
```diff
@@ -475,13 +493,37 @@ export class SourceFile {
...
+            this._writableData.typeIgnoreLines.size > 0 ||
+            this._writableData.typeIgnoreAll !== undefined ||
+            this._writableData.pyrightIgnoreLines.size > 0;
+
+        if (!hadSyntax) {
+            return false;
+        }
+
+        this._releaseSyntaxCaches(/* preserveLineCount */ true);
         this._writableData.isBindingNeeded = true;
```
rchiodo, line 526: There are now four syntax-drop entry points with subtly different semantics — `dropParseAndBindingInfo`, `releaseClosedFileSyntax(): boolean`, `markDirtyAndDropSyntax`, and `clearPreEditStateForDispose` — all over the shared `_releaseSyntaxCaches`. `releaseClosedFileSyntax`'s boolean return (whether anything was released) isn't self-documenting. Add a one-line "call this when…" doc to each, and note the boolean's meaning at the declaration.

rchiodo:
Solid, well-tested lifetime hardening. The main thing to resolve before merge is the unconditional `releaseClosedFileSyntax()` in `setFileClosed` (re-parse/re-bind regression for still-needed closed files). The rest are maintainability/perf observations on the new evaluator-disposal mechanism that are worth a deliberate decision but aren't blocking.

packages/pyright-internal/src/analyzer/program.ts
bschnurr: Addressed in 5369656b. _collectUnneededFiles now uses _isFileNeeded-compatible iterative reachability rather than a separate root/follow definition, and I added shadow coverage for both unrooted removal and rooted preservation.

packages/pyright-internal/src/analyzer/program.ts
bschnurr: Fixed in 5369656b. The collector now runs an iterative _isFileNeeded-style pass with the pending removal set, so shadow/deleted/importer semantics stay aligned with the existing predicate instead of drifting into a second definition.

packages/pyright-internal/src/analyzer/program.ts
bschnurr: Added regression coverage in service.test: setTrackedFiles removes unrooted shadow files and setTrackedFiles preserves rooted shadow files. The unrooted shadower/shadowed pair is removed together; rooted shadow files remain.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
bschnurr: Fixed in 5369656b. unctionRecursionMap and codeFlowAnalyzerCache are no longer cleared before the active-evaluator gate; they are cleared by inactive cleanup after the evaluator unwinds, alongside the stacks they pair with.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
bschnurr: Partially addressed in 5369656b: recursion/code-flow maps now follow the same active-disposal deferral as the stacks, and the cancellation-scope assertion documents the current invariant. I did not add a full activity-stack registry in this PR because that is a broader evaluator refactor.

packages/pyright-internal/src/analyzer/program.ts
bschnurr: Addressed by narrowing close behavior in 5369656b. Ordinary unchanged close no longer calls 
eleaseClosedFileSyntax; it preserves syntax/evaluator state for still-known tracked files. Syntax release is now left to explicit cache-pressure, removal, content replacement, and disposal boundaries.

packages/pyright-internal/src/analyzer/program.ts
bschnurr: Fixed in 5369656b. setFileClosed no longer unconditionally drops syntax, so tab close of an imported/tracked module will not force a reparse/rebind on the next analysis pass.

packages/pyright-internal/src/analyzer/sourceFileInfo.ts
bschnurr: Addressed in 5369656b. Edge filtering now lives on SourceFileInfo.detachFromRemovedFiles, and removed files use SourceFileInfo.clearForDispose, so the edge inventory is centralized with SourceFileInfo rather than repeated in Program branches.

packages/pyright-internal/src/analyzer/program.ts
bschnurr: Added a fresh sanity probe after the latest changes: synthetic 1000-file markAllFilesDirty(true) completes in ~0.8 ms. I also called this evaluator-recreate behavior out explicitly in the PR description.

packages/pyright-internal/src/analyzer/program.ts
bschnurr: Addressed in 5369656b. _detachSourceFileInfosForRemoval now performs a cheap reference pre-check before calling SourceFileInfo.detachFromRemovedFiles, avoiding filtered-array allocation for survivors with no removed references.

packages/pyright-internal/src/analyzer/sourceFile.ts
bschnurr: Added one-line docs in sourceFile.ts for dropParseAndBindInfo, 
eleaseClosedFileSyntax (including the boolean meaning), and markDirtyAndDropSyntax. clearPreEditStateForDispose already has the dispose-specific comment.

packages/pyright-internal/src/analyzer/typeEvaluatorTypes.ts
bschnurr: I left the cache stats on internal interfaces for this PR because they are used by the leak/lifetime regression tests and this package does not expose these analyzer interfaces as a public API surface. Agree that adding future evaluator caches should update the stats contract; that is intentional for these tests.

packages/pyright-internal/src/analyzer/sourceFileInfo.ts
```diff
@@ -171,6 +171,50 @@ export class SourceFileInfo implements extensibility.SourceFileInfo {
...
+        const chainedSourceFileChanged = chainedSourceFile !== data.chainedSourceFile;
+        const changed =
+            imports.length !== data.imports.length ||
+            importedBy.length !== data.importedBy.length ||
+            shadows.length !== data.shadows.length ||
+            shadowedBy.length !== data.shadowedBy.length ||
+            builtinsImport !== data.builtinsImport ||
+            chainedSourceFileChanged;
+
+        if (changed) {
```
rchiodo, line 205: The full edge inventory (`imports`/`importedBy`/`shadows`/`shadowedBy`/`builtinsImport`/`chainedSourceFile`) is enumerated independently in three places: `_clearAllEdges` (dispose), `detachFromRemovedFiles` (filter), and `Program._detachSourceFileInfosForRemoval`'s `.some(...)` membership guard. A future edge kind must be added to all three or a retainer silently leaks. Since the detach path filters rather than clears, a single clear-helper can't cover it — but a shared edge-target enumeration (e.g. `forEachEdgeTarget`) could unify the list across the clear/filter/membership sites.

rchiodo:
Overall this is sound. The main recurring theme worth a deliberate decision before finalizing: "active evaluator state", "evaluator retainers", and "source-file edge kinds" are each enumerated in multiple independent sites, so a future addition that misses one site reintroduces exactly the silent-leak failure class this PR targets. Consider a single source of truth for each. Also add a perf sanity check for the new markAllFilesDirty(true) evaluator-recreation behavior on a real config/library reanalysis flow.

rchiodo (approved):
Approved via Review Center.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -705,19 +714,117 @@ export function createTypeEvaluator(
         return typeCache.size;
```
rchiodo, line 714: The evaluator activity-set is now enumerated in three methods — `isEvaluatorActive()`, `clearInactiveRetainers()`, and `getEvaluatorCacheStats()` — all listing the same seven-stack set (`cancellationToken`, speculative context, `symbolResolutionStack`, `suppressedNodeStack`, `assignClassToSelfStack`, `returnTypeInferenceContextStack`, `signatureTrackerStack`). Adding a future retainer stack requires editing all three in lockstep, or the deferral gate silently drifts and reintroduces the leak class this PR exists to prevent. Consider a single shared activity/depth list consumed by all three. Non-blocking maintainability debt.

packages/pyright-internal/src/analyzer/sourceFileInfo.ts
```diff
@@ -171,6 +171,50 @@ export class SourceFileInfo implements extensibility.SourceFileInfo {
...
+            this._writableData.imports = imports;
+            this._writableData.importedBy = importedBy;
+            this._writableData.shadows = shadows;
+            this._writableData.shadowedBy = shadowedBy;
+            this._writableData.builtinsImport = builtinsImport;
+            this._writableData.chainedSourceFile = chainedSourceFile;
+        }
+
+        return { changed, chainedSourceFileChanged };
+    }
```
rchiodo, line 216: The full edge inventory is enumerated independently across multiple sites: `_clearAllEdges` (clear), `detachFromRemovedFiles` (filter), and the `.some(...)` guard in `program.ts:_detachSourceFileInfosForRemoval`. A future edge kind must be added to all of them or a retainer silently leaks. Note the asymmetry is currently implicit — `effectiveFutureImports` is a `ReadonlySet<string>` (not a removable `SourceFileInfo` edge) so it correctly appears only in `_clearAllEdges`. Consider driving all sites from one shared edge descriptor and making the `effectiveFutureImports` exclusion explicit. Non-blocking.

packages/pyright-internal/src/analyzer/program.ts
```diff
@@ -1260,15 +1328,56 @@ export class Program {
...
+
+            this._sourceFileList.forEach((fileInfo) => {
+                if (
+                    !filesToRemove.has(fileInfo) &&
+                    !this._isFileNeeded(fileInfo, /* skipFileNeededCheck */ false, filesToRemove)
+                ) {
+                    filesToRemove.add(fileInfo);
+                    removedFile = true;
+                }
+            });
```
rchiodo, line 1340: `_collectUnneededFiles`'s `while (removedFile)` fixpoint re-scans all files per pass, so a deep abandoned import/chain cascade `A→B→C→…` that becomes collectable one hop per pass can cost up to O(n²) scans. This is a net improvement over the prior restart loop and only runs when files are actually removed, but the ~4.7 ms/1000-file benchmark reflects shallow graphs, not deep single-file-per-pass cascades. Non-blocking — worth a short comment noting the non-linear worst case, and if it ever shows in a profile, switch to reverse-reachability (mark-from-roots, sweep the complement) for single-pass O(n·E).

packages/pyright-internal/src/analyzer/program.ts
```diff
@@ -502,47 +583,66 @@ export class Program {
...
+                    /* forceRebinding */ false,
+                    /* recreateEvaluator */ false
+                );
+                this._createNewEvaluator();
             }
+
+            // Syntax release happens through explicit cache-pressure and removal
+            // boundaries. A closed file can still be needed by tracked files or
+            // live importers, and dropping syntax here would force a reparse/rebind
+            // on the next analysis pass while the evaluator still owns the old cache.
```
rchiodo, line 568: Redundant double `_createNewEvaluator()`: on `setFileClosed` where the on-disk contents changed and the file then becomes unneeded, the `didContentsChangeOnDisk()` branch recreates the evaluator, then `_removeUnneededFiles()` recreates it again. Correctness is unaffected (one extra evaluator allocation per such close); consider letting `_removeUnneededFiles` own the single recreate, or guarding the close-path recreate. Non-blocking.

rchiodo:
Overall this looks solid and safe to merge. The main recurring theme worth addressing is drift risk: the evaluator activity-set and the SourceFileInfo edge inventory are each enumerated across multiple sites, so adding a future stack/edge kind requires editing all of them in lockstep. Consolidating each behind a single shared enumeration would remove that hazard.

rchiodo (approved):
Approved via Review Center.

bschnurr:
Follow-up correction pushed in \\4b6a381e4\\: normal open-document edits now preserve the existing evaluator/type cache, matching the original incremental-analysis design. The PR still drops source-owned stale parse output for the edited file, but evaluator disposal/heap probes are now limited to explicit invalidation paths such as chained URI changes, mark-files-dirty, mark-all-files-dirty, empty cache, removal, and dispose.

packages/pyright-internal/src/analyzer/backgroundAnalysisProgram.ts
```diff
@@ -128,7 +128,6 @@ export class BackgroundAnalysisProgram {
     updateOpenFileContents(uri: Uri, version: number | null, contents: string, options: OpenFileOptions) {
         this._backgroundAnalysis?.setFileOpened(uri, version, contents, options);
         this._program.setFileOpened(uri, version, contents, options);
-        this.markFilesDirty([uri], /* evenIfContentsAreSame */ true);
     }
```
rchiodo, line 133: Removing `this.markFilesDirty([uri], /* evenIfContentsAreSame */ true)` from `updateOpenFileContents` drops the only importer-invalidation path on a `didChange`. Verified against base: this call was the sole route (`languageServerBase` → `service.updateOpenFileContents` → here → `_markFileDirtyRecursive`, which walks `importedBy`), and `program.ts:setFileOpened` never walks `importedBy` on a content change. Combined with the deliberate evaluator preservation across edits, an importer `b.py` of an edited `a.py` is never re-checked and keeps serving cached types keyed to `a.py`'s old declarations — stale cross-file diagnostics in both `openFilesOnly` and `workspace` modes. Restore importer dirtying on a real content change (invalidating the edited file's dependent type-cache entries if you want to preserve the evaluator), or prove the invalidation happens elsewhere. Add a non-edit-mode, multi-file regression test (edit `a.py`, assert `b.py`'s diagnostics / `reveal_type` update) — the existing dependent test is edit-mode-scoped, where the checker doesn't run, so it cannot catch this.

packages/pyright-internal/src/analyzer/program.ts
```diff
@@ -444,19 +482,21 @@ export class Program {
...
+
+        const contentsChanged = sourceFileInfo.sourceFile.setClientVersion(version, contents);
+        if (wasKnownFile && (!wasOpenByClient || contentsChanged)) {
+            // Reset the diagnostic version so we force an update to the diagnostics,
+            // which can change based on whether the file is open or its contents changed.
+            // We do not set the version to undefined here because that implies there are
+            // no diagnostics currently reported for this file.
+            sourceFileInfo.diagnosticsVersion = 0;
+        }
     }
```
rchiodo, line 500: `setFileOpened` now preserves the evaluator across open-file edits, but the edited file reparses with globally-unique node IDs (`markDirtyAndDropSyntax`), so the preserved evaluator's `typeCache`/`effectiveTypeCache` entries keyed by prior generations' node IDs are never matched again and are never reclaimed until an unrelated evaluator recreate. On long single-file editing sessions this grows the steady-state working set — the opposite of this PR's goal. Since you already track `evaluatorGeneration`, prune the edited file's stale-generation entries on reparse, or otherwise bound the preservation.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -705,19 +714,117 @@ export function createTypeEvaluator(
...
+    }
+
+    function isEvaluatorActive() {
+        const speculativeCacheStats = speculativeTypeTracker.getCacheStats();
+        return (
+            cancellationToken !== undefined ||
+            speculativeCacheStats.speculativeContextStack > 0 ||
+            symbolResolutionStack.length > 0 ||
+            suppressedNodeStack.length > 0 ||
+            assignClassToSelfStack.length > 0 ||
```
rchiodo, line 800: The evaluator activity-set (`cancellationToken`, speculative context, `symbolResolutionStack`, `suppressedNodeStack`, `assignClassToSelfStack`, `returnTypeInferenceContextStack`, `signatureTrackerStack`) is enumerated independently in `isEvaluatorActive()`, `clearInactiveRetainers()`, and `getEvaluatorCacheStats()`. This is exactly the "add an 8th stack and one site silently drifts" failure class this PR aims to eliminate, now reproduced inside the evaluator's disposal logic — a drift here means incorrect cleanup or a missed retainer. Factor it into a single named predicate/descriptor consumed by all three sites. Separately, `isEvaluatorActive()` calls `speculativeTypeTracker.getCacheStats()` (which scans the whole `_speculativeTypeCache`) just to read `speculativeContextStack > 0`; a cheap context-depth accessor would avoid the incidental scan.

packages/pyright-internal/src/analyzer/sourceFileInfo.ts
```diff
@@ -213,6 +257,21 @@ export class SourceFileInfo implements extensibility.SourceFileInfo {
...
+        data.importedBy = [];
+        data.shadows.length = 0;
+        data.shadows = [];
+        data.shadowedBy.length = 0;
+        data.shadowedBy = [];
+        data.builtinsImport = undefined;
+        data.chainedSourceFile = undefined;
+        data.effectiveFutureImports = undefined;
+        data.diagnosticsVersion = undefined;
+    }
```
rchiodo, line 274: The reference-edge inventory (`imports`/`importedBy`/`shadows`/`shadowedBy`/`builtinsImport`/`chainedSourceFile`) is spelled out independently in three sites: `_clearAllEdges`, `detachFromRemovedFiles`, and the `.some(…)` guard in `Program._detachSourceFileInfosForRemoval`. Adding a future edge means one site silently leaks a retainer — the same drift risk this PR targets. Drive all three from one shared edge descriptor / `forEachEdge` helper. Also consider renaming/re-scoping `_clearAllEdges`, since it additionally clears the non-edge fields `effectiveFutureImports` and `diagnosticsVersion` (the `effectiveFutureImports` omission from the removal path is correct — it's a `ReadonlySet<string>`, not a `SourceFileInfo` edge — so make that inclusion/exclusion explicit rather than implicit).

rchiodo:
The lifetime-hardening direction is sound, but the removal of importer dirtying from `updateOpenFileContents` introduces a cross-file staleness regression that must be addressed (restore dependent invalidation on real content changes, or prove it happens elsewhere, and add a non-edit-mode multi-file regression test). A couple of secondary concerns about unbounded stale type-cache growth and duplicated invariant lists are worth resolving while you're here.

bschnurr:
Deep-research follow-up pushed in \\c8cb0400\\: I restored the historical BackgroundAnalysisProgram.updateOpenFileContents -> markFilesDirty([uri], true) call. That means normal language-server edit notifications again dirty dependents and invalidate the evaluator through the existing dirty path; only direct Program.setFileOpened preserves evaluator identity. Tests and the heap probe were updated to reflect that distinction.

rchiodo:
Verified against the current tree: the claimed importer-invalidation regression does not exist — `BackgroundAnalysisProgram.updateOpenFileContents` still calls `markFilesDirty([uri], true)`, which dirties dependents and recreates the evaluator. Main remaining feedback is DRY consolidation of hand-maintained field/edge inventories so this hardening work doesn't itself become the next 'add a field, silently leak' hazard.

rchiodo (approved):
Approved via Review Center.

PR #11532, opened 2026-07-01
Adapt stubPath special-casing to latest changes

rchl:
In https://github.com/microsoft/pyright/pull/11480 I've removed redundant `workspace/configuration` requests but I've missed this special-case logic in vscode extension that removes `stubPath` property if it's at default value.

Without this change, the server would receive the default `typings` value and complain that the path is not correct (in logs only, I believe).

### DIFF
--- a/packages/vscode-pyright/src/extension.ts
+++ b/packages/vscode-pyright/src/extension.ts
@@ -126,7 +126,7 @@ export async function activate(context: ExtensionContext) {
                     }
 
                     for (const [i, item] of params.items.entries()) {
-                        if (item.section === 'python.analysis') {
+                        if (item.section === 'python') {
                             const analysisConfig = workspace.getConfiguration(
                                 item.section,
                                 item.scopeUri ? Uri.parse(item.scopeUri) : undefined
--- a/packages/vscode-pyright/src/extension.ts
+++ b/packages/vscode-pyright/src/extension.ts
@@ -127,7 +127,7 @@ export async function activate(context: ExtensionContext) {
 
                     for (const [i, item] of params.items.entries()) {
                         if (item.section === 'python') {
-                            const analysisConfig = workspace.getConfiguration(
+                            const pythonConfig = workspace.getConfiguration(
                                 item.section,
                                 item.scopeUri ? Uri.parse(item.scopeUri) : undefined
                             );
@@ -135,8 +135,8 @@ export async function activate(context: ExtensionContext) {
                             // If stubPath is not set, remove it rather than sending default value.
                             // This lets the server know that it's unset rather than explicitly
                             // set to the default value (typings) so it can behave differently.
-                            if (!isConfigSettingSetByUser(analysisConfig, 'stubPath')) {
-                                delete (result[i] as any).stubPath;
+                            if (!isConfigSettingSetByUser(pythonConfig, 'analysis.stubPath')) {
+                                delete (result[i] as any).analysis.stubPath;
                             }
                         }
                     }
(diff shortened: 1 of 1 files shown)


rchiodo:
The section rename is correct but incomplete — the guard and delete inside the loop still assume the old 'python.analysis' shape. Both need to be updated to operate on the nested `analysis.stubPath` so the default `typings` value is actually stripped (and a user's custom stubPath is preserved). A middleware test would prevent this section→shape coupling from regressing.

rchiodo:
The migration to section `python` is only half-done. Since the server now reads stubPath nested under `analysis`, both the user-set guard and the `delete` need to target `analysis.stubPath` (guarding against a missing `analysis`). As written the delete no-ops and the default `typings` is still forwarded, so the reported log complaint remains.

rchiodo:
The section switch to `python` needs matching updates to the stubPath handling below it: both the user-set guard and the `delete` still assume the old top-level `python.analysis` shape, so they now target the wrong key/depth (`result[i].stubPath` and `python.stubPath`) instead of `result[i].analysis.stubPath` / `python.analysis.stubPath`. As written, the default `typings` value is still forwarded, so the fix is incomplete.

packages/vscode-pyright/src/extension.ts
```diff
@@ -126,17 +126,17 @@ export async function activate(context: ExtensionContext) {
...
                                 item.scopeUri ? Uri.parse(item.scopeUri) : undefined
                             );
 
                             // If stubPath is not set, remove it rather than sending default value.
                             // This lets the server know that it's unset rather than explicitly
                             // set to the default value (typings) so it can behave differently.
-                            if (!isConfigSettingSetByUser(analysisConfig, 'stubPath')) {
-                                delete (result[i] as any).stubPath;
+                            if (!isConfigSettingSetByUser(pythonConfig, 'analysis.stubPath')) {
+                                delete (result[i] as any).analysis.stubPath;
```
rchiodo: 📍 packages/vscode-pyright/src/extension.ts:139
The delete lacks optional chaining: `delete (result[i] as any).analysis.stubPath`. If `result[i].analysis` is ever `undefined`/`null`, this throws `TypeError` inside the config middleware and rejects the entire batch, degrading all config resolution rather than just stubPath. In practice `python.analysis.stubPath` is registered with a `typings` default so `analysis` is normally present, but this should match the server's own defensive access (`server.ts:128` uses `pythonSection?.analysis`). Suggest: `delete (result[i] as any).analysis?.stubPath;`.

packages/vscode-pyright/src/extension.ts
```diff
@@ -126,17 +126,17 @@ export async function activate(context: ExtensionContext) {
...
                             }
```
rchiodo, line 140: 📍 packages/vscode-pyright/src/extension.ts:129
No regression test covers this middleware's strip-vs-preserve contract — the exact logic that silently regressed to a no-op in a prior revision. Consider adding a middleware test asserting (a) an unset stubPath → the default `typings` is stripped from `result[i].analysis`, and (b) a user-set value survives. This pins the implicit client/server config-shape coupling so a future refactor can't silently regress it.

rchiodo:
The migration correctly forwards section `python` and targets the nested `analysis.stubPath`, matching the server's `pythonSection?.analysis.stubPath` read. Two small, non-blocking suggestions inline: guard the delete with optional chaining and add a middleware regression test for the strip-vs-preserve contract.

rchiodo (approved):
Approved via Review Center.

rchiodo (approved):
Approved via Review Center.

rchiodo:

heejaechang (approved):
Approved via Review Center.

rchiodo (approved):
Approved via Review Center.

StellaHuang95 (approved):
Approved via Review Center.

StellaHuang95 (approved):
Approved via Review Center.

PR #11537, opened 2026-07-09
Add standalone Type Server Protocol (TSP) server to Pyright

rchiodo:
## Summary

Adds a standalone **Type Server Protocol (TSP)** implementation to Pyright, so Pyright can serve TSP on its own. The implementation is ported from Pylance's `packages/typeServer`, with the Pylance-specific pieces (notebook support, virtual-file overlays, stub generation) copied over so Pyright has equivalent functionality.

## What's included

- **`packages/pyright-internal/src/typeServer/`** — the TSP module: protocol definitions (`protocol/`), the server (`server.ts`), the program wrapper (`programWrapper.ts`), type/declaration conversion (`typeServerConversion*`), cancellation, notebook support (`notebook*`), a virtual-file overlay filesystem, and stub generation.
- **`packages/pyright-typeserver`** — a new package that bundles a distributable `pyright-typeserver` binary (run as `pyright-typeserver --stdio`).
- **Tests** — in-proc TSP tests, notebook TSP tests, and virtual-file redirect tests under `src/tests/typeServer/`.
- **Docs** — `docs/type-server.md` plus sidebar/internals links.
- **CI** — the validation workflow builds the new package.

The change is almost entirely additive; the only pre-existing files touched are `.github/workflows/validation.yml` and the docs index.

## Design notes

- **Fully synchronous.** Unlike Pylance's async wrapper, Pyright's TSP server runs synchronously. Cancellation is token/file-based, matching normal Pyright, so requests can be cancelled mid-flight.
- **No `IProgramSnapshot` concept.** Pyright doesn't model program snapshots the way Pylance does. The port drops that abstraction, folds it into the regular `IProgram`, and uses a monotonic snapshot counter purely for staleness detection.
- **Dead code removed.** Client-side-only conversion helpers and unused provider/factory plumbing from the Pylance original were dropped during the port.

## Validation

- `packages/pyright-internal` builds clean (`tsc`), eslint reports 0 errors, prettier applied.
- The Jest `typeServer` test suites pass.
- Validated end-to-end against **Pylance's own TSP client test suite** (microsoft/pyrx `typeServer.external.test.ts`): **33/33 passing** when Pylance drives Pyright's built type-server bundle over TSP. A companion PR fixes three pre-existing bugs in that pyrx suite — microsoft/pyrx#8838.

## Commits

- Add standalone TSP (Type Server Protocol) server to Pyright
- Add notebook support to Pyright TSP server
- Wire up pyright-typeserver package build, CI, and docs
- Make TSP server synchronous and merge `IProgramSnapshot` into `IProgram`
- Remove unused client-side protocol-to-Pyright type deserialization
- Remove dead code from typeServer


heejaechang:

packages/pyright-internal/src/typeServer/enums.ts
```diff
@@ -0,0 +1,759 @@
...
+    isAny,
+    isClass,
+    isClassInstance,
+    isFunction,
+    isFunctionOrOverloaded,
+    isInstantiableClass,
+    maxTypeRecursionCount,
+    Type,
+    TypeBase,
+    UnknownType,
```
heejaechang, line 37: **Duplicated evaluator semantics.** ~760 lines of intricate, spec-sensitive enum logic (functional Enum forms, `enum.member`/`nonmember`, 3.13 class-member behavior, `_value_`/`name` magic) are forked here from `analyzer/enums.ts`, which exports the identical symbol set. The only real delta is routing file-info through `evaluator.getSymbolLookup().getFileInfo(...)`. Since `ITypeServerEvaluator extends TypeEvaluator`, the analyzer's canonical functions already accept a compatible evaluator — prefer reusing them directly (with a thin file-info adapter) rather than copying the module. As written, when Pyright's enum semantics evolve this copy drifts, which is exactly the failure the 'identical to Pyright's other front ends' claim is meant to prevent. `typeGuards.ts` has the same issue. If reuse isn't feasible now, please file tracked debt and add pointer comments linking the copies.

packages/pyright-internal/src/typeServer/server.ts
```diff
@@ -0,0 +1,782 @@
...
+                workspace.service.updateOpenFileContents(
+                    realUri,
+                    doc?.version ?? null,
+                    virtualContent,
+                    IPythonMode.None
+                );
+            }
+            workspace.service.invalidateAndScheduleReanalysis(InvalidatedReason.Reanalyzed);
+        }
+    }
```
heejaechang, line 736: **Virtual-file redirect forces `IPythonMode.None` on possibly-notebook files.** `_onSetVirtualFileRedirect`/`_onRemoveVirtualFileRedirect` call `updateOpenFileContents(realUri, …, IPythonMode.None)`. If `realUri` is an open notebook cell (opened as `IPythonMode.CellDocs`), this resets its mode and detaches it from the cell chain. `ProgramWrapper.updateFileContents` already shows the correct approach (read and preserve `fileInfo.ipythonMode`). Django stubs aren't cells today so this is latent, but the mismatch with the wrapper's own method is a trap worth closing.

packages/pyright-internal/src/typeServer/protocol/typeServerProtocol.ts
```diff
@@ -0,0 +1,1388 @@
+/*
```
heejaechang, line 1: **Duplicated wire protocol — single-source-of-truth gap.** This `.ts` (and the generated `tsp.json`/`tsp.schema.json`) is a hand-maintained copy of pyrx's canonical shared TSP. Version negotiation catches gross mismatches but not silent field-level drift within a version, which would break Pylance↔Pyright interop over the wire. This copy is unavoidable for a standalone upstream Pyright, but the PR should state explicitly what the single source of truth is (is `tsp.json` the neutral contract both repos regenerate from, or is one `.ts` authoritative?) and ideally add a pointer comment or a parity check so the two cannot silently diverge.

packages/pyright-internal/src/typeServer/server.ts
```diff
@@ -0,0 +1,782 @@
...
+    private async _getProgram(uri: Uri): Promise<ProgramWrapper | undefined> {
+        const workspace = await this.getWorkspaceForFile(uri);
+        if (workspace) {
+            return makeProgram(workspace.service.backgroundAnalysisProgram.program) as ProgramWrapper;
+        }
+        return undefined;
+    }
+
+    private async _onGetSnapshot() {
+        return this._globalTypeCache.snapshot;
```
heejaechang, line 601: **Snapshot-cache identity is timing-dependent.** `_onGetSnapshot()` returns `_globalTypeCache.snapshot`, but request handlers validate `program.getSnapshot(token) !== params.snapshot` against the wrapper's own `ITypeCache`. They agree only because `onWorkspaceCreated` runs `makeProgram(program, this._globalTypeCache)` before `_getProgram` calls `makeProgram(program)` with no cache arg, relying on the `WeakMap` to return the global-cache wrapper. This does not fire today (the program instance is stable) — but if a program is ever replaced or a wrapper is built cache-less, the client livelocks on permanent `ServerCancelled`. Consider enforcing the invariant: pass `this._globalTypeCache` from `_getProgram`, or make `makeProgram` reject re-wrapping a program with a different cache.

heejaechang:
Looks solid and safe to merge — the change is almost entirely additive and the type-server suites pass. The main theme worth the author's attention is duplication of spec-sensitive logic ported from Pylance/analyzer (enums/typeGuards, and the hand-copied wire protocol), plus one latent notebook-mode trap in the virtual-file redirect path. None block merge, but addressing the duplication reduces long-term drift risk against the 'identical to Pyright's other front ends' guarantee.

heejaechang (approved):
Approved via Review Center.

packages/pyright-internal/src/tests/typeServer/typeServer.inProc.test.ts
```diff
@@ -0,0 +1,254 @@
...
+
+            await context.refreshSnapshot();
+            const type = await context.sendRequestWithSnapshot(TypeServerProtocol.GetDeclaredTypeRequest.type, {
+                arg,
+            });
+
+            assert(type !== undefined);
+            assert(typeof type.id === 'number');
+            assert(typeof type.kind === 'number');
+        });
```
heejaechang, line 245: Weak assertions in the new TSP type tests. `getComputedType`, `getExpectedType`, and `getDeclaredType` only assert `typeof type.id === 'number'` / `typeof type.kind === 'number'`. The inputs are fully deterministic (`__arg__ = 1` → `int`; `x: C = C()` → `C`; `__declared__: str = 'x'` → `str`), so the tests would still pass even if type conversion returned the wrong kind/class. Assert the concrete `TypeKind` and class name (the sibling `typeServer.virtualFileRedirect.test.ts` already shows the pattern via `getClassTypeName()` + `assert.strictEqual(..., 'int')`); `type.id` can stay a `typeof` check since it's opaque.

heejaechang:
Solid additive port — thanks for the thorough validation against the real client suite. I've dropped the style-rule nits (readonly modifiers, helper placement, dynamic require, duplicated helper) as non-blocking noise, and collapsed the many duplicated/latent findings. Three non-blocking notes remain below for robustness and test strength; none block merge.

heejaechang (approved):
Approved via Review Center.

packages/pyright-internal/src/tests/typeServer/typeServer.inProc.test.ts
```diff
@@ -0,0 +1,254 @@
...
+// @filename: main.py
+//// def takes_int(x: int) -> None:
+////     pass
+////
+//// [|/*arg*/__arg__|] = 1
+//// takes_int(__arg__)
+`;
+
+        await withInProcTypeServer(code, async (context) => {
+            await context.openFileForMarker('arg');
```
heejaechang, line 186: The `getComputedType`/`getExpectedType`/`getDeclaredType` tests only assert `typeof type.kind === 'number'` on fully deterministic inputs (`__arg__ = 1`→int, `x: C = C()`→C, `__declared__: str`→str). These would pass even if conversion returned the wrong kind. Assert the concrete `TypeKind` (and class name via `getClassTypeName()`, as the sibling `virtualFileRedirect.test.ts` already does). `type.id` is opaque and can stay a `typeof` check.

packages/pyright-internal/src/typeServer/typeEvalUtils.ts
```diff
@@ -0,0 +1,114 @@
...
+    const scope = getScopeForNode(nodeForScope);
+    if (!scope) {
+        return undefined;
+    }
+
+    return scope.lookUpSymbol(name);
+}
+
+export function getEffectiveTypeOfDeclaration(
+    evaluator: TypeEvaluator | undefined,
```
heejaechang, line 84: `getSymbolFromScope` here is byte-identical to the copy at the bottom of `programWrapper.ts` (which already imports from this module). Export this one and import it there instead of maintaining two copies.

packages/pyright-internal/src/typeServer/enums.ts
```diff
@@ -0,0 +1,759 @@
+/*
+ * enums.ts
+ * Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT license.
+ *
+ * Provides special-case logic for the Enum class.
+ */
+
```
heejaechang, line 15: This module (and `typeGuards.ts`) is a large fork of `analyzer/enums.ts`, differing only by routing file-info through `evaluator.getSymbolLookup().getFileInfo(...)`. Add a short "forked from analyzer/enums.ts; keep in sync" signpost plus a tracked issue so upstream enum-semantics changes don't drift silently, even if full reuse is deferred.

heejaechang:
GitHub cannot anchor PR review comments to unchanged lines in the diff. Falling back to a general PR comment for [packages/pyright-internal/src/typeServer/server.ts:L9931](https://github.com/microsoft/pyright/blob/b70ba06a083d78960062dc6ce179a582ef7064b6/packages/pyright-internal/src/typeServer/server.ts#L9931).

`_onSetVirtualFileRedirect` / `_onRemoveVirtualFileRedirect` pass literal `IPythonMode.None` to `updateOpenFileContents`, whereas `ProgramWrapper.updateFileContents` reads and preserves `fileInfo.ipythonMode`. This is latent today (redirect `realUri` targets are module files, never notebook cells), but if a `realUri` is ever an open cell it would reset the mode and detach it from the cell chain. Mirror the wrapper's read-and-preserve pattern to close the trap.

heejaechang:
Solid additive port of the TSP server into Pyright — well isolated and validated end-to-end. A few non-blocking suggestions: tighten a couple of deterministic test assertions, de-duplicate one shared helper, and add a fork-drift signpost on the copied enum logic.

heejaechang (approved):
Approved via Review Center.

rchiodo:
Please take a look at all of Heejae's comments.

packages/pyright-internal/src/typeServer/server.ts
rchiodo: Fixed in 1e4cd05 — both `_onSetVirtualFileRedirect` and `_onRemoveVirtualFileRedirect` now preserve the file's existing IPython mode via a new `_getExistingIPythonMode(workspace, uri)` helper (reads `program.getSourceFileInfo(uri)?.ipythonMode`) instead of hard-coding `IPythonMode.None`, matching `ProgramWrapper.updateFileContents`. This closes the notebook-cell reset/detach trap even though Django stubs aren't cells today.

packages/pyright-internal/src/typeServer/server.ts
rchiodo: Fixed in 1e4cd05 — enforced the invariant both ways the review suggested. `_getProgram` now passes `this._globalTypeCache` explicitly to `makeProgram`, and `makeProgram` now rejects re-wrapping a program with a different cache (`fail(...)` when an already-wrapped program is asked to use a mismatched `ITypeCache`, via a new `ProgramWrapper.typeCache` getter). This guarantees `_onGetSnapshot()` and the request handlers read the same cache/snapshot, so the permanent-`ServerCancelled` livelock cannot occur even if a program is ever re-wrapped cache-less.

packages/pyright-internal/src/typeServer/protocol/typeServerProtocol.ts
rchiodo: Addressed in 1e4cd05 by documenting the single source of truth in the `typeServerProtocol.ts` header: within Pyright this `.ts` file is authoritative and `tsp.json`/`tsp.schema.json` are GENERATED from it by `generate_json.py` (do not hand-edit the JSON). Across repos this file is a synchronized copy of the canonical shared TSP also consumed by Pylance; version negotiation guards gross mismatches but not silent field-level drift, so any wire change must be mirrored and coordinated cross-repo. Left the wire types untouched per the protocol-change policy.

packages/pyright-internal/src/typeServer/typeEvalUtils.ts
rchiodo: Fixed in 1e4cd05 — `getSymbolFromScope` is now exported from `typeEvalUtils.ts` and imported in `programWrapper.ts`; removed the byte-identical copy (and the now-unused `ParseNodeType` import there).

packages/pyright-internal/src/tests/typeServer/typeServer.inProc.test.ts
rchiodo: Fixed in 1e4cd05 — all three tests now assert the concrete `TypeKind` and class name: `getComputedType`→`TypeKind.Class`/'int', `getExpectedType`→`TypeKind.Class`/'C', `getDeclaredType`→`TypeKind.Class`/'str', using a local `getClassTypeName()` helper mirroring the sibling `virtualFileRedirect.test.ts`. `type.id` stays a `typeof` check since it's opaque.

packages/pyright-internal/src/tests/typeServer/typeServer.inProc.test.ts
rchiodo: Fixed in 1e4cd05 — the `getComputedType`/`getExpectedType`/`getDeclaredType` tests now assert the concrete `TypeKind.Class` and class name ('int'/'C'/'str') via a local `getClassTypeName()` helper (same pattern as `virtualFileRedirect.test.ts`), so a wrong-kind conversion would now fail. `type.id` remains a `typeof` check.

packages/pyright-internal/src/typeServer/enums.ts
rchiodo: Addressed in 1e4cd05 — added a 'forked from analyzer/enums.ts; keep in sync' signpost to the header of `typeServer/enums.ts` (and a matching one to `typeServer/typeGuards.ts`), explicitly calling out that the only substantive delta is routing file-info through the `ITypeServerEvaluator` and that the copies must be kept in sync when the canonical semantics change. Full reuse of the analyzer functions is deferred; the signposts reference the type-server deduplication follow-up so upstream enum-semantics changes don't drift silently.

packages/pyright-internal/src/typeServer/enums.ts
rchiodo: Addressed in 1e4cd05 with tracked-debt + pointer comments rather than immediate reuse. Both `typeServer/enums.ts` and `typeServer/typeGuards.ts` now carry a header signpost stating they are forked from their `analyzer/*` counterparts, that the only substantive delta is routing file-info through the type-server evaluator, and that the copies must be kept in sync (referencing the type-server deduplication follow-up). Direct reuse of the analyzer functions with a thin file-info adapter is a good future step but is deferred here to keep this PR scoped; the signposts prevent silent drift in the meantime.

heejaechang:
Looks solid — approving. Clean additive port, tests pass, and external validation is green. A batch of `readonly-by-default` / test-strictness convention notes were raised by the first pass; they carry no functional impact and several sibling comments in that batch misread the diff, so I'm not blocking on them. If you want to satisfy those conventions, adding `readonly` to the never-reassigned private fields in programWrapper.ts / typeCache.ts / notebookUriMapper.ts is a trivial optional follow-up.

heejaechang (approved):
Approved via Review Center.

PR #11541, opened 2026-07-11
Fix __class__ binding in class-body lambdas

### LINKED ISSUE #11511: `__class__` inside a lambda in class definition is reported as not defined
**Describe the bug**
The code below is correct in Python (`A.f()` returns the class), but Pyright reports an error.

As an additional reference, there was a similar issue in Ruff: https://github.com/astral-sh/ruff/issues/20562

**Code or Screenshots**
```python
class A:
    f = lambda: __class__
```

`"__class__" is not defined  (reportUndefinedVariable)`

VS Code, same in  [playground](https://pyright-play.net/?code=MYGwhgzhAECCBcBYAUNN0Bm0C81wFsAjAEzHmgH0LRIIqg)

kingrubic:
## Summary
- bind the implicit PEP 3135 `__class__` symbol in lambda scopes enclosed by a class
- allow lambda nodes to back intrinsic declarations
- add a regression sample for `class A: get_class = lambda: __class__`

## Testing
- targeted regression test
- complete `typeEvaluator1.test.ts` (159 passed)
- complete `pyright-internal` suite (62 suites, 2541 tests passed)
- build, ESLint, Prettier, typecheck

Fixes #11511

### DIFF
--- a/packages/pyright-internal/src/tests/samples/lambda16.py
+++ b/packages/pyright-internal/src/tests/samples/lambda16.py
@@ -1,0 +1,15 @@
+# This sample tests the implicit __class__ closure in a lambda defined in a
+# class body.
+
+# pyright: strict, reportUnknownLambdaType=false
+
+from typing import Callable
+
+
+class A:
+    get_class = lambda: __class__
+
+
+# This should not create an implicit __class__ binding because the lambda is
+# defined outside of a class body.
+get_class: Callable[[], type] = lambda: __class__
--- a/packages/pyright-internal/src/analyzer/binder.ts
+++ b/packages/pyright-internal/src/analyzer/binder.ts
@@ -703,6 +703,12 @@ export class Binder extends ParseTreeWalker {
             () => {
                 AnalyzerNodeInfo.setScope(node, this._currentScope);
 
+                const enclosingClass = ParseTreeUtils.getEnclosingClass(node);
+                if (enclosingClass) {
+                    // Lambdas create the same implicit __class__ closure as named functions.
+                    this._addImplicitSymbolToCurrentScope('__class__', node, '__class__');
+                }
+
                 this._deferBinding(() => {
                     // Create a start node for the lambda.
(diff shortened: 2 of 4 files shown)


kingrubic:

rchiodo:

packages/pyright-internal/src/tests/samples/lambda16.py
```diff
@@ -0,0 +1,8 @@
+# This sample tests the implicit __class__ closure in a lambda defined in a
+# class body.
+
+# pyright: strict
+
+
+class A:
+    get_class = lambda: __class__
```
rchiodo, line 10: 📍 packages/pyright-internal/src/tests/samples/lambda16.py:7
The sample only asserts the happy path. Since the only real risk of this change is *over-binding* `__class__`, add a negative case — a module-level or free-function `g = lambda: __class__` wired as a second sample expecting `validateResults(..., 1)` — so a future regression that over-broadens the binding is actually caught. The Skeptic verified this case is correct today (CPython raises `NameError`, and `getEnclosingClass` returns `undefined`), but it isn't locked by a test.

[verified]

rchiodo:
Looks good — correct, minimal fix for the PEP 3135 `__class__` binding in class-body lambdas. Approving. One optional suggestion: add a negative sample to guard against future over-binding.

rchiodo (approved):
Approved via Review Center.

kingrubic:
Thanks for the suggestion. I rebased onto the current `main` and added a negative regression case for a module-level lambda referencing `__class__`, asserting that it is still reported as undefined. I also updated the expected diagnostic count and ran the focused test:

- `npm run test:norebuild -- --runInBand typeEvaluator1.test.ts -t Lambda16` (passed)
- `git diff --check` (passed)

The updated commit has been pushed to the PR branch.

rchiodo (approved):
Approved via Review Center.

PR #11547, opened 2026-07-16
Fix false positive reportUnnecessaryComparison for bytes-like types

### LINKED ISSUE #11433: False positive: Condition will always evaluate to False since the types "bytearray" and "bytes" have no overlap
[Playground link](https://pyright-play.net/?strict=true&code=CYQwLiCMAEC80CMCeYCmIBOGRIBQBYBKAKFAgCY5EVUBnAk4gSwDNoyo54PyAuY6IOgAHDEwB2YXACIAYgHsArhmgAvVBnm1omVNBZKV6zbQCE0xhyrI0mbHiLEAbiAA2ivfATSAOgAZ-QL9pYmY2K1h4F3dUfiERMUkZBWU1DS0dDD0DVOMtc0IgA)

```python
data1 = bytearray(4)
data2 = bytes(4)

if data1 == data2:
    print("Four zeros are four zeros!")
```
Results in the warning `Condition will always evaluate to False since the types "bytearray" and "bytes" have no overlap`. 

But this is incorrect, comparing `bytearray` and `bytes` with `==` returns `True` if they have the same content.

A related issue occurs with literals:

```python
data = bytearray(4)
value = b"\0\0\0\0"

if data == value:
    print("Four zeros are four zeros!")
```

Results in the warning `Condition will always evaluate to False since the types "bytearray" and "Literal[b"\x00\x00\x00\x00"]" have no overlap`.

pctablet505:
Fixes #11433

`reportUnnecessaryComparison` treats two disjoint built-in class instances as never comparable with `==`/`!=`. That's a reasonable default, but it doesn't hold for `bytes`, `bytearray` and `memoryview`: they're unrelated for assignability purposes, yet their `__eq__` implementations do support cross-type content comparisons (`bytearray(4) == bytes(4)` is `True` at runtime).

This adds a narrow carve-out in `isTypeComparable` for this specific group of types, reusing the existing `typePromotions` table that already models the relationship elsewhere (it's the same list consulted for the `bytes`/`bytearray`/`memoryview` assignability promotion). Comparability isn't gated behind `disableBytesTypePromotions`, since that setting is about assignment compatibility, not about whether `__eq__` can return `True`.

Added test cases to `comparison1.py` covering `bytearray`/`bytes`, a `bytearray`-vs-literal-bytes comparison, and all pairings of `bytes`/`bytearray`/`memoryview`, alongside a couple of genuinely-disjoint builtin cases (`int`/`str`, `list`/`dict`) to confirm the general diagnostic still fires where it should.

### DIFF
--- a/packages/pyright-internal/src/analyzer/typeEvaluator.ts
+++ b/packages/pyright-internal/src/analyzer/typeEvaluator.ts
@@ -26059,6 +26059,32 @@ export function createTypeEvaluator(
         );
     }
 
+    // The "bytes" entry of "typePromotions" (see below) groups together the
+    // "bytes-like" types that support cross-type content comparisons via
+    // "==" and "!=", even though they're otherwise unrelated (non-overlapping)
+    // types. Unlike the numeric promotions ("int" -> "float" -> "complex"),
+    // where two disjoint literal subtypes (e.g. "Literal[1]" and "Literal[2]")
+    // must still be treated as non-comparable, every pairing within this
+    // group is mutually comparable, so it's safe to treat fullName membership
+    // in this specific set as sufficient without consulting literal values.
+    const bytesLikeTypeNames = new Set(['builtins.bytes', ...(typePromotions.get('builtins.bytes') ?? [])]);
+
+    function isBytesLikeType(type: ClassType) {
+        return type.shared.mro.some((mroClass) => isClass(mroClass) && bytesLikeTypeNames.has(mroClass.shared.fullName));
+    }
+
+    // Determines whether the two types are both members of the "bytes-like"
+    // group of types ("bytes", "bytearray" and "memoryview"). Pyright models
+    // this relationship as a "type promotion" (used elsewhere for
+    // assignability), but that grouping is also useful for determining
+    // whether two otherwise-disjoint built-in types can be compared with
+    // "==" or "!=", since these types share compatible "__eq__" semantics
+    // regardless of whether the "disableBytesTypePromotions" setting allows
+    // the assignment.
+    function isTypePromotionRelated(leftType: ClassType, rightType: ClassType) {
+        return isBytesLikeType(leftType) && isBytesLikeType(rightType);
+    }
+
     // Determines whether the two types are potentially comparable -- i.e.
     // their types overlap in such a way that it makes sense for them to
     // be compared with an == or != operator. The functional also supports
@@ -26165,6 +26191,17 @@ export function createTypeEvaluator(
                         }
 
                         return boolVal === (intVal === 1);
+                    }
+
+                    // Types like "bytes", "bytearray" and "memoryview" are otherwise
+                    // disjoint but are still comparable with "==" and "!=" because
+                    // their "__eq__" methods support cross-type content comparisons.
+                    // Pyright models this relationship as a "type promotion" (used
+                    // elsewhere for assignability), so reuse that same list here
+                    // regardless of the "disableBytesTypePromotions" setting, since
+                    // comparability isn't affected by that assignability toggle.
+                    if (isTypePromotionRelated(leftType, rightType)) {
+                        return true;
                     }
 
                     return false;
--- a/packages/pyright-internal/src/analyzer/typeEvaluator.ts
+++ b/packages/pyright-internal/src/analyzer/typeEvaluator.ts
@@ -27497,6 +27497,32 @@ export function createTypeEvaluator(
         );
     }
 
+    // The "bytes" entry of "typePromotions" (see below) groups together the
+    // "bytes-like" types that support cross-type content comparisons via
+    // "==" and "!=", even though they're otherwise unrelated (non-overlapping)
+    // types. Unlike the numeric promotions ("int" -> "float" -> "complex"),
+    // where two disjoint literal subtypes (e.g. "Literal[1]" and "Literal[2]")
+    // must still be treated as non-comparable, every pairing within this
+    // group is mutually comparable, so it's safe to treat fullName membership
+    // in this specific set as sufficient without consulting literal values.
+    const bytesLikeTypeNames = new Set(['builtins.bytes', ...(typePromotions.get('builtins.bytes') ?? [])]);
+
+    function isBytesLikeType(type: ClassType) {
+        return type.shared.mro.some((mroClass) => isClass(mroClass) && bytesLikeTypeNames.has(mroClass.shared.fullName));
+    }
+
+    // Determines whether the two types are both members of the "bytes-like"
+    // group of types ("bytes", "bytearray" and "memoryview"). Pyright models
+    // this relationship as a "type promotion" (used elsewhere for
+    // assignability), but that grouping is also useful for determining
+    // whether two otherwise-disjoint built-in types can be compared with
+    // "==" or "!=", since these types share compatible "__eq__" semantics
+    // regardless of whether the "disableBytesTypePromotions" setting allows
+    // the assignment.
+    function isTypePromotionRelated(leftType: ClassType, rightType: ClassType) {
+        return isBytesLikeType(leftType) && isBytesLikeType(rightType);
+    }
+
     // Determines whether the two types are potentially comparable -- i.e.
     // their types overlap in such a way that it makes sense for them to
     // be compared with an == or != operator. The functional also supports
@@ -27603,6 +27629,17 @@ export function createTypeEvaluator(
                         }
 
                         return boolVal === (intVal === 1);
+                    }
+
+                    // Types like "bytes", "bytearray" and "memoryview" are otherwise
+                    // disjoint but are still comparable with "==" and "!=" because
+                    // their "__eq__" methods support cross-type content comparisons.
+                    // Pyright models this relationship as a "type promotion" (used
+                    // elsewhere for assignability), so reuse that same list here
+                    // regardless of the "disableBytesTypePromotions" setting, since
+                    // comparability isn't affected by that assignability toggle.
+                    if (isTypePromotionRelated(leftType, rightType)) {
+                        return true;
                     }
 
                     return false;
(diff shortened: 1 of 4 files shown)


pctablet505:

rchiodo:

rchiodo (approved):
Approved via Review Center.

StellaHuang95 (approved):
Approved via Review Center.

StellaHuang95 (approved):
Approved via Review Center.

StellaHuang95:

rchiodo (approved):
Approved via Review Center.

rchiodo (approved):
Approved via Review Center.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -26167,6 +26193,17 @@ export function createTypeEvaluator(
...
+                    // Types like "bytes", "bytearray" and "memoryview" are otherwise
+                    // disjoint but are still comparable with "==" and "!=" because
+                    // their "__eq__" methods support cross-type content comparisons.
+                    // Pyright models this relationship as a "type promotion" (used
+                    // elsewhere for assignability), so reuse that same list here
+                    // regardless of the "disableBytesTypePromotions" setting, since
+                    // comparability isn't affected by that assignability toggle.
+                    if (isTypePromotionRelated(leftType, rightType)) {
+                        return true;
+                    }
```
StellaHuang95, line 27643: > **Issue** · Please address or respond

📍 packages/pyright-internal/src/analyzer/typeEvaluator.ts:26203

This verified carve-out also runs for `is` and `is not`, suppressing valid always-false diagnostics for distinct bytes-like classes. Gate it with `!assumeIsOperator` and add identity-comparison regression cases.

[verified]

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -26059,6 +26059,32 @@ export function createTypeEvaluator(
...
+    // "bytes-like" types that support cross-type content comparisons via
+    // "==" and "!=", even though they're otherwise unrelated (non-overlapping)
+    // types. Unlike the numeric promotions ("int" -> "float" -> "complex"),
+    // where two disjoint literal subtypes (e.g. "Literal[1]" and "Literal[2]")
+    // must still be treated as non-comparable, every pairing within this
+    // group is mutually comparable, so it's safe to treat fullName membership
+    // in this specific set as sufficient without consulting literal values.
+    const bytesLikeTypeNames = new Set(['builtins.bytes', ...(typePromotions.get('builtins.bytes') ?? [])]);
+
+    function isBytesLikeType(type: ClassType) {
```
StellaHuang95, line 27510: > **Info** · Optional note

📍 packages/pyright-internal/src/analyzer/typeEvaluator.ts:26059

This duplicates the MRO lookup already provided by `derivesFromStdlibClass`, violating the documented core-utility reuse rule. Use that existing helper for `bytes`, `bytearray`, and `memoryview` instead of maintaining a separate full-name set and MRO walk.

[verified]

packages/pyright-internal/src/analyzer/typeEvaluator.ts
StellaHuang95, line 27643: > **Warning** · Non-blocking recommendation

📍 packages/pyright-internal/src/analyzer/typeEvaluator.ts:26203

The corresponding Pylance async evaluator has its own `isTypeComparable` implementation, so downstream integration must mirror this decision or demonstrate parity through async-mode tests if TSP owns the behavior. Otherwise Pyright and Pylance can produce different user-visible results.

[verified]

packages/pyright-internal/src/tests/typeEvaluator6.test.ts
```diff
@@ -652,7 +652,7 @@ test('Comparison1', () => {
 
     configOptions.diagnosticRuleSet.reportUnnecessaryComparison = 'error';
     const analysisResults2 = TestUtils.typeAnalyzeSampleFiles(['comparison1.py'], configOptions);
-    TestUtils.validateResults(analysisResults2, 7);
+    TestUtils.validateResults(analysisResults2, 9);
```
StellaHuang95, line 685: > **Warning** · Non-blocking recommendation

📍 packages/pyright-internal/src/tests/typeEvaluator6.test.ts:655

`isTypeComparable` also feeds container-element narrowing, but the added coverage checks only diagnostics. Add a targeted bytes-like narrowing case to ensure this broader consumer remains correct.

[verified]

heejaechang (approved):
Approved via Review Center.

heejaechang:
### 🔴 Pyright CLI QA — `b0f855ed` — verdict: red

12 focused check(s) completed.

**`main.py`**

```python
from typing import Literal, assert_type, reveal_type


def membership(flag: bool) -> None:
    value = b"a" if flag else b"b"
    if value in (b"a",):
        reveal_type(value)
        assert_type(value, Literal[b"a"])


def comparison(flag: bool) -> None:
    value = b"a" if flag else None
    if value == b"b":
        print("impossible")
```

**`pyrightconfig.json`**

```json
{
  "typeCheckingMode": "basic",
  "pythonVersion": "3.13",
  "reportUnnecessaryComparison": "error",
  "include": [
    "main.py"
  ]
}
```

```text
```

**Base `85837efea752fbb1cd04c261a51cded566e21e57`**

```json
{
  "exit_code": 1,
  "diagnostics": [
    {
      "file": "main.py",
      "severity": "error",
      "rule": "reportUnnecessaryComparison",
      "message": "Condition will always evaluate to False since the types \"Literal[b\"a\"] | None\" and \"Literal[b\"b\"]\" have no overlap",
      "range": {
        "start": [
          12,
          7
        ],
        "end": [
          12,
          20
        ]
      }
    },
    {
      "file": "main.py",
      "severity": "information",
      "rule": null,
      "message": "Type of \"value\" is \"Literal[b\"a\"]\"",
      "range": {
        "start": [
          6,
          20
        ],
        "end": [
          6,
          25
        ]
      }
    }
  ]
}
```

**PR `b0f855ed2c6034d97a9f6e27ac2ad7322fd917b2`**

```json
{
  "exit_code": 1,
  "diagnostics": [
    {
      "file": "main.py",
      "severity": "information",
      "rule": null,
      "message": "Type of \"value\" is \"Literal[b\"a\", b\"b\"]\"",
      "range": {
        "start": [
          6,
          20
        ],
        "end": [
          6,
          25
        ]
      }
    },
    {
      "file": "main.py",
      "severity": "error",
      "message": "\"assert_type\" mismatch: expected \"Literal[b\"a\"]\" but received \"Literal[b\"a\", b\"b\"]\"",
      "range": {
        "start": [
          7,
          20
        ],
        "end": [
          7,
          25
        ]
      }
    }
  ]
}
```

</details>

rchiodo:
I confirmed the bytes-literal regression reported by the CLI QA check. The new exception treats distinct `bytes` literals as comparable, not just different bytes-like classes. This affects both consumers of `isTypeComparable`:

```python
from typing import Literal

def f(x: Literal[b"a", b"b"]) -> Literal[b"a"]:
    if x in (b"a",):
        return x
    return b"a"
```

At the merge base, the positive branch narrows `x` to `Literal[b"a"]`. At the current PR head it retains `Literal[b"a", b"b"]`, producing an incorrect return-type error. Comparisons of `x: Literal[b"a"] | int` against `b"b"` also lose valid `reportUnnecessaryComparison` diagnostics. Both regressions reproduce with bytes type promotions enabled and disabled.

I am preparing a fix that limits the exception to different bytes-like classes, plus regression coverage for literal narrowing and disjoint comparisons while retaining the intended cross-type equality behavior and identity diagnostics. I will also check the existing review threads before pushing the follow-up.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
rchiodo: Addressed in b0f855ed2 and preserved in d5aa8b04e: the bytes-like exception is gated by `!assumeIsOperator`, and `comparison1.py::func9` covers identity comparisons between the three distinct bytes-like classes. The follow-up now also exercises both bytes-promotion settings, retaining the pre-existing assignability behavior when promotions are enabled.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
rchiodo: Addressed in b0f855ed2: `isBytesLikeType` uses the existing `derivesFromStdlibClass` helper for bytes, bytearray, and memoryview. The follow-up in d5aa8b04e retains that reuse and adds the same-generic-class exclusion needed to preserve disjoint bytes literals.

packages/pyright-internal/src/tests/typeEvaluator6.test.ts
rchiodo: Covered by `typeNarrowingIn1.py::func24` and expanded in d5aa8b04e with precise bytes-literal narrowing for tuple/list membership, `not in` branches, a literal return type, and literal bytes matched against bytearray/memoryview containers. Both bytes-promotion settings are exercised. The new literal-narrowing checks fail on the previous PR head and pass with the fix; all 1,284 evaluator/checker tests pass.

rchiodo:
Pushed d5aa8b04e to fix the bytes-literal regression. The exception now applies only to different bytes-like classes, preserving cross-type equality while keeping distinct `bytes` literals disjoint.

Regression coverage now includes literal tuple/list membership, `not in` branches, a literal return type, impossible comparisons involving nonliteral unions, and `==`/`!=` in both operand orders for all three bytes-like pairings. Comparison and narrowing cases run with bytes promotions enabled and disabled. The new regression cases fail before the fix and pass afterward.

Validation: all 1,284 tests across the eight evaluator suites and checker suite pass; core TypeScript type-check, changed-file ESLint/Prettier, and diff whitespace checks also pass.

Resolved the review threads about identity gating, shared MRO helper reuse, and container-element narrowing coverage. The Pylance async-parity thread remains open: this repository contains only the shared Pyright implementation, so I cannot claim downstream async parity from these checks. The prior macOS CI failure was a language-server `didopendocument` timeout; fresh CI should still be green before merging.

PR #11560, opened 2026-07-23
Fix false positive when assigning to an attribute of a union with divergent member types

### LINKED ISSUE #11559: TypeNarrowing on assignment leads to false positive
**Describe the bug**
Hello, I stumbled upon a rather confusing false positive.
It seems assigning to fields shared between multiple types through a union, with a value valid for both (like empty list) the right hand side type is narrowed to one of the field types.

**Code or Screenshots**

```python
class A:
    a: list[int]
class B:
    a: list[str]
def foo(a: A | B):
    a.a = [] # <- empty list is narrowed to list[str]
#   Cannot assign to attribute "a" for class "A"
#   Expression of type "list[str]" cannot be assigned to attribute "a" of class "A"
#     "list[str]" is not assignable to "list[int]"
#       Type parameter "_T@list" is invariant, but "str" is not the same as "int"
#       Consider switching from "list" to "Sequence" which is covariant
#   Pylance: reportAttributeAccessIssue
```

Found error in pyright 1.1.411 as well as in VSCode pylance

Best regards,
Robin

rchiodo:
## Summary

Fixes #11559.

Assigning to an attribute whose base expression is a union produced a false positive when the union''s subtypes declared that attribute with different invariant types:

```python
class A:
    a: list[int]

class B:
    a: list[str]

def foo(a: A | B):
    a.a = []  # Previously errored: "list[str]" is not assignable to "list[int]"
```

## Root cause

When the member-access assignment target has a union base, `getDeclaredTypeForExpression` looped over the (sorted) subtypes and committed to a single arbitrary subtype''s declared member type as the expected type for bidirectional inference of the RHS. The set-path (`getTypeOfMemberAccessWithBaseType`) then validated that single committed type against *every* subtype, so the type chosen for one subtype (`list[str]`) failed against the other (`list[int]`).

## Fix

For union bases, the declared member type is now computed for each subtype and compared with `isTypeSame`. If the subtypes diverge, `getDeclaredTypeForExpression` returns `undefined`, so the assigned value is evaluated without an expected type:

- An empty container (`[]`) infers as `list[Unknown]`, which is assignable to every subtype, so the false positive is gone.
- Concrete values (`[1]`, `["hi"]`) retain their element type and still correctly error against the incompatible subtype.
- When the subtypes agree on the declared type (e.g. both `list[int]`), that type is still returned, preserving bidirectional inference and narrowing.

The descriptor-setter / specialization / function-binding logic that finalizes the declared type was extracted into a `resolveDeclaredMemberType` helper so it can be reused both inside the subtype loop and for the non-union path.

## Tests

Added `memberAccess29.py` and the `MemberAccess29` test, covering both the divergent union case (`A | B`) and the same-type union case (`C | D`, which must still infer `list[int]`).

### DIFF
--- a/packages/pyright-internal/src/analyzer/typeEvaluator.ts
+++ b/packages/pyright-internal/src/analyzer/typeEvaluator.ts
@@ -2832,6 +2832,56 @@ export function createTypeEvaluator(
         return false;
     }
 
+    // Given a member symbol and the context in which it was accessed, computes
+    // the declared type of the member (applying descriptor setter types, partial
+    // specialization, and function binding as appropriate). Returns undefined if
+    // the symbol has no declared type.
+    function resolveDeclaredMemberType(
+        symbol: Symbol,
+        classOrObjectBase: ClassType | undefined,
+        memberAccessClass: Type | undefined,
+        useDescriptorSetterType: boolean,
+        bindFunction: boolean,
+        selfType: ClassType | TypeVarType | undefined
+    ): Type | undefined {
+        let declaredType = getDeclaredTypeOfSymbol(symbol)?.type;
+        if (!declaredType) {
+            return undefined;
+        }
+
+        // If it's a descriptor, we need to get the setter type.
+        if (useDescriptorSetterType && isClassInstance(declaredType)) {
+            const setter = getBoundMagicMethod(declaredType, '__set__');
+            if (setter && isFunction(setter) && setter.shared.parameters.length >= 2) {
+                declaredType = FunctionType.getParamType(setter, 1);
+
+                if (isAnyOrUnknown(declaredType)) {
+                    return undefined;
+                }
+            }
+        }
+
+        if (classOrObjectBase) {
+            if (memberAccessClass && isInstantiableClass(memberAccessClass)) {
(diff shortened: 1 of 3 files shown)


rchiodo:
**mypy_primer note:** The remaining sympy churn is expected and net-positive. mypy_primer''s base is the PR head''s parent (`$GITHUB_SHA^1`), so this diff isolates just this commit with no `main` drift.

Net effect on sympy: **-98 diagnostics** (removes 153, adds 55).
- **Removed (153):** mostly `reportOptionalSubscript` / `reportOptionalMemberAccess` cascade false positives.
- **Added (55):** mostly `reportOperatorIssue` on sympy''s dynamically-typed `Basic`/`Expr` classes, plus a few `reportAttributeAccessIssue`.

Root cause: baseline pyright committed to an **arbitrary, sort-order-dependent** union subtype''s member type as the assignment''s expected type (the very bug behind #11559). The fix stops doing that for divergent unions, which shifts RHS inference across sympy''s dynamic hierarchy. The django churn from the earlier iteration is gone after the `divergesOnlyByTypeArgs` refinement.

heejaechang:

bschnurr (approved):

heejaechang (approved):
Approved via Review Center.

Issue #11026, opened 2025-10-13
False positive of `reportUnnecessaryComparison` on match case with `int(<float>)`


q-wertz:
Note: if you are reporting a wrong signature of a function or a class in the standard library, then the typeshed tracker is better suited for this report: https://github.com/python/typeshed/issues.

If you have a question about typing or a behavior that you’re seeing in Pyright (as opposed to a bug report or enhancement request), consider posting to the [Pyright discussion forum](https://github.com/microsoft/pyright/discussions).

**Describe the bug**

In the following code
```python
sr = int(2.0e6)

match sr:
    case int(2.0e6):
        a = "asdf"
    case int(2.4e6):
        a = "cvbn"
    case _:
        raise ValueError

print(a)
```
a `reportUnnecessaryComparison` error is shown in strict mode ([Playground](https://pyright-play.net/?strict=true&code=M4JwBAvGCWB2AuAKATAOgAwFMBsBKAUPgLYCG8AxgBZigBc%2BYjY5JwmMCKGOu9T-YEpDAAiVgBMAZiIZMWbDkjQAWHnwGMhUEeQBuAI1gz%2B89gH11GkCWgKAaiQA2AV0wBRECAD2IQgAcQOCQSXCA)).

As soon as it is replaced by integers directly the error is gone.
```python
sr = 2000000

match sr:
    case 2000000:
        a = "asdf"
    case 2400000:
        a = "cvbn"
    case _:
        raise ValueError

print(a)
```

Not sure if some float/int thing I don't get right but I think it also should not fail in the first example.

**Code or Screenshots**
If possible, provide a minimal, self-contained code sample (surrounded by triple back ticks) to demonstrate the issue. The code should define or import all referenced symbols.

```python
sr = int(2.0e6)

match sr:
    case int(2.0e6):
        a = "asdf"
    case int(2.4e6):
        a = "cvbn"
    case _:
        raise ValueError

print(a)
```

If your code relies on symbols that are imported from a third-party library, include the associated import statements and specify which versions of those libraries you have installed.

**VS Code extension or command-line**
Are you running pyright as a VS Code extension, a language server in another editor, integrated into Pylance, or the command-line tool? Which version?

erictraut:
[...]

PR #11570, opened 2026-07-28
Fix bool narrowing for numeric literal patterns

hsusul:
## Summary

- Preserve `Literal[False]` and `Literal[True]` when matching the numeric literal patterns `0` and `1`.
- Remove those bool literals from the fallthrough branch because numeric literal patterns compare with equality.
- Keep singleton bool patterns identity-based, so `case False` does not match `Literal[0]`.

## Problem

Numeric literal patterns use equality at runtime. This means `case 0` matches both `0` and `False`, and `case 1` matches both `1` and `True`.

```python
from typing import Literal

def f(value: Literal[0, False]) -> None:
    match value:
        case 0:
            reveal_type(value)  # Literal[0, False]
        case _:
            reveal_type(value)  # Never
```

Pyright previously narrowed the matched branch to `Literal[0]` and the fallthrough branch to `Literal[False]`, which is unsound relative to runtime behavior.

## Root cause

Literal-pattern narrowing relied on assignability between the pattern literal and subject literal. That does not capture Python's equality relationship between `0`/`False` and `1`/`True`.

## Solution

Handle the asymmetric numeric-pattern equality case explicitly:

- an int literal pattern `0` or `1` retains the equal bool literal in the positive branch;
- the same bool literal is removed from the negative branch;
- bool singleton patterns remain unchanged because they use identity rather than equality.

## Tests

Added focused coverage for:

- `case 0` over `Literal[0, False]`;
- `case 1` over `Literal[1, True]`;
- `case False` retaining the identity-based distinction from `Literal[0]`.

## Validation

- `npm run typecheck`
- `npm run build:cli:dev`
- `npm run check`
- `npx jest -t 'MatchLiteral1' --forceExit --runInBand`
- `npx jest src/tests/typeEvaluator6.test.ts --runInBand`
- `npm test -- --runInBand` (62 suites, 2549 tests)
- CLI repro on Python 3.10 and 3.13

## Compatibility and risk

The change is limited to literal-pattern narrowing for int patterns against bool literal subjects. Tests cover both positive and negative branches and the inverse singleton-bool boundary.

## Documentation

No documentation changes are needed; this aligns analyzer behavior with Python's existing structural pattern-matching semantics.

## Related issues and prior work

#11026 and #11028 addressed broader literal-pattern narrowing involving non-literal and supertype subjects. This change covers the distinct bool/int literal equality edge case.

### DIFF
--- a/packages/pyright-internal/src/analyzer/patternMatching.ts
+++ b/packages/pyright-internal/src/analyzer/patternMatching.ts
@@ -682,7 +682,8 @@ function narrowTypeBasedOnLiteralPattern(
                 isLiteralType(literalType) &&
                 isClassInstance(expandedSubtype) &&
                 isLiteralType(expandedSubtype) &&
-                evaluator.assignType(literalType, expandedSubtype)
+                (evaluator.assignType(literalType, expandedSubtype) ||
+                    isIntLiteralPatternEqualToBool(literalType, expandedSubtype))
             ) {
                 return undefined;
             }
@@ -708,6 +709,16 @@ function narrowTypeBasedOnLiteralPattern(
     }
 
     return evaluator.mapSubtypesExpandTypeVars(type, /* options */ undefined, (expandedSubtype, unexpandedSubtype) => {
+        if (
+            isClassInstance(literalType) &&
+            isLiteralType(literalType) &&
+            isClassInstance(expandedSubtype) &&
+            isLiteralType(expandedSubtype) &&
+            isIntLiteralPatternEqualToBool(literalType, expandedSubtype)
+        ) {
+            return expandedSubtype;
+        }
+
         if (evaluator.assignType(expandedSubtype, literalType)) {
             // We have to be careful here because the runtime uses an equality
             // check, but the expandedSubtype could be a superclass that is not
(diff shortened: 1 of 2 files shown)


rchiodo:

rchiodo (approved):
Approved via Review Center.

rchiodo:
### The sympy mypy_primer diffs are not caused by this PR — they're pre-existing pyright non-determinism

I built pyright at this PR's base (`729aabfa3`, == current `main`) and at the PR head (`37d3fc744`) and ran both on the full sympy source in an identical environment.

**The two builds produce identical diagnostics on sympy:**
- Base vs PR: **0 added, 0 removed** (75,789 diagnostics each), comparing full multi-line messages.
- Determinism check (same binary run twice): also **0 diff**.

So on a deterministic machine the change has **zero effect** on sympy — a real, PR-caused diff would have reproduced.

**Why that's expected from the code:** the new `isIntLiteralPatternEqualToBool` path only activates when a numeric literal pattern `0`/`1` is matched against a **bool-literal** subject (`Literal[True]`/`Literal[False]`) in a `match`/`case`. In all of sympy:
- The only `case 0:`/`case 1:` is `trigonometric.py`'s `match n % 4:`, whose subject is `int` (not a bool literal) → the new path never triggers.
- None of the flagged files (`subscheck.py`, `systems.py`, `crv.py`, `crv_types.py`, `drv.py`, `drv_types.py`, `frv.py`, `joint_rv_types.py`) contain any `match`/`case`.

**What the primer diff actually is:** analysis-order / recursion-resolution non-determinism over sympy's recursive `Basic`/`Expr`/`Self` hierarchy. The signatures are the usual tells — none of which a literal-pattern-narrowing change can produce:
- `Basic*` ↔ `Basic` (the `*` marks a TypeVar/`Self`-solved type)
- `Self@Basic` dropping in/out of unions
- union member reordering (e.g. `Unknown | Self@Basic | bool | Expr | Basic` → `Unknown | Expr | Basic | bool`)
- `Add` / `Infinity | NegativeInfinity | Float` appearing/disappearing in large unions

**Bottom line:** expected primer noise, not a new bug.

rchiodo:
Thanks for the PR

PR #11574, opened 2026-07-29
Add cross-type-checker benchmark regression tracking

bschnurr:
## Summary
- add a pinned 9-package corpus for repeatable Pyright performance measurements
- record checked-in GitHub-hosted Linux x64/Python 3.14.6 baseline results and compare PR candidates against them
- add PR and weekly workflows, manual triggers, trusted PR comments, retained JSON results, and a multi-checker HTML report

All nine packages have successful baseline measurements. NumPy is measured over its production-source scope; SciPy was removed from the gated corpus after exceeding the 600-second hosted-runner timeout.

## Validation
- `python3 -m unittest test_compare_benchmarks.py test_typecheck_benchmark.py` (23 tests)
- baseline self-comparison
- Prettier
- Python compilation
- `git diff --check`


rchiodo:

.github/workflows/typecheck_benchmark_pr.yml
```diff
@@ -0,0 +1,84 @@
...
+      - '.github/workflows/typecheck_benchmark_pr.yml'
+      - '.github/workflows/typecheck_benchmark_comment.yml'
+      - 'build/benchmark/**'
+      - 'packages/pyright/**'
+      - 'packages/pyright-internal/src/**'
+      - 'packages/pyright-internal/typeshed-fallback/**'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
```
rchiodo, line 15: > **Issue** · Please address or respond

This gate compares an `ubuntu-latest` candidate against a baseline captured on a 32-core WSL2 machine. The comparator checks only `platform` and `architecture`, so the 10% wall-time threshold is applied across materially different hardware despite the README warning against cross-machine comparisons. Generate the baseline on the same runner class, or make this comparison informational until that baseline exists.

.github/workflows/typecheck_benchmark_pr.yml
```diff
@@ -0,0 +1,84 @@
+name: Type checker benchmark
+
+env:
+  NODE_VERSION: '22.x'
+  PYTHON_VERSION: '3.14'
```
rchiodo: > **Issue** · Please address or respond

`3.14` resolves to the newest available patch, while the baseline records `python_version: "3.14.6"` and the comparator requires exact equality. A routine Python patch update will fail the benchmark regardless of performance. Pin the workflow to the baseline's exact patch, or compare Python versions at major/minor granularity.

build/benchmark/install_envs.json
```diff
@@ -0,0 +1,149 @@
...
+        {
+            "github_url": "https://github.com/numpy/numpy",
+            "commit": "db6ccacf630dd90f7c498e0d924bd47b66c83746",
+            "check_paths": [
+                "numpy"
+            ],
+            "install": false,
+            "deps": [
+                "pytest",
+                "hypothesis"
```
rchiodo, line 63: > **Warning** · Non-blocking recommendation

Packages configured with `install: true` can fail or time out during their editable install or dependency setup on hosted runners. That returns empty metrics, which the comparator treats as a blocking benchmark failure rather than an environment-preparation failure. Avoid making dependency-install instability a performance gate, or remove such packages from the gated CI corpus.

.github/workflows/typecheck_benchmark_pr.yml
```diff
@@ -0,0 +1,84 @@
...
+
+      - name: Install build prerequisites
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y build-essential
+          npm install
+
+      - name: Build Pyright CLI
+        working-directory: packages/pyright
+        run: npm run build
```
rchiodo: > **Warning** · Non-blocking recommendation

The gate applies a 10% threshold to one measured run on a shared runner. Although the harness records statistics, this configuration cannot distinguish normal timing variance from a regression. Use multiple runs and a robust statistic, or add a variance guard before failing the comparison.

build/benchmark/typecheck_benchmark.py
```diff
@@ -0,0 +1,1335 @@
...
+            match = re.match(r"(\d+)\s+maximum resident set size", stripped)
+            if match:
+                peak_bytes = int(match.group(1))
+                continue
+            if re.match(r"\d+\s+\w", stripped):
+                continue
+            in_time_output = False
+        filtered_lines.append(line)
+    peak_mb = round(peak_bytes / (1024 * 1024), 1) if peak_bytes else 0.0
+    return peak_mb, "".join(filtered_lines)
```
rchiodo, line 278: > **Warning** · Non-blocking recommendation

The build and `--skip-pyright-build` paths validate only `dist/pyright.js`, but `_pyright_command()` also requires `packages/pyright/index.js`. If the entry point is absent or renamed, the benchmark silently marks Pyright as unavailable and the later comparison fails unclearly. Validate `PYRIGHT_ENTRY_POINT` alongside the bundle in `prepare_local_pyright`.

.github/workflows/typecheck_benchmark_comment.yml
```diff
@@ -0,0 +1,82 @@
...
+            fs.writeFileSync('benchmark-report.zip', Buffer.from(download.data))
+
+      - name: Extract benchmark report
+        run: unzip -q benchmark-report.zip -d benchmark-report
+
+      - name: Post benchmark comment
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const fs = require('fs')
```
rchiodo, line 123: > **Issue** · Please address or respond

This write-authorized `workflow_run` consumes `pr_number.txt` and `report.md` from an artifact produced by PR code. A fork can overwrite both files, causing this trusted job to post attacker-controlled Markdown to an arbitrary PR or update an existing marker comment. Derive the PR number from the workflow-run payload/API rather than the artifact, and render or validate the report in this trusted workflow.

.github/workflows/typecheck_benchmark_pr.yml
```diff
@@ -0,0 +1,84 @@
...
+        run: |
+          mkdir -p build/benchmark/results
+          printf '%s\n' '${{ github.event.pull_request.number }}' > build/benchmark/results/pr_number.txt
+
+      - name: Run benchmark
+        env:
+          PYTHONNOUSERSITE: '1'
+        run: |
+          python build/benchmark/typecheck_benchmark.py \
+            -c pyright -r 1 -w 0 -t 600 --memory-limit-mb 8192 \
```
rchiodo: > **Issue** · Please address or respond

This gates an `ubuntu-latest` run against a baseline captured on a 32-core WSL2 machine, but the comparator checks only `platform` and `architecture`. The absolute wall-time threshold will therefore flag runner-performance differences as Pyright regressions. Generate the baseline on the same runner class, or keep this comparison informational until such a baseline exists.

.github/workflows/typecheck_benchmark_pr.yml
```diff
@@ -0,0 +1,84 @@
...
+          sudo apt-get install -y build-essential
+          npm install
+
+      - name: Build Pyright CLI
+        working-directory: packages/pyright
+        run: npm run build
+
+      - name: Record pull request
+        run: |
+          mkdir -p build/benchmark/results
```
rchiodo: > **Warning** · Non-blocking recommendation

The gate applies a 10% threshold to one measured run on a shared runner. Normal host variance can therefore create or hide regressions, despite the harness already supporting repeated runs and summary statistics. Use multiple runs with a robust statistic or add a variance guard before failing the check.

build/benchmark/install_envs.json
```diff
@@ -0,0 +1,149 @@
...
+                "attrs",
+                "pydantic",
+                "SQLAlchemy",
+                "types-setuptools",
+                "types-atomicwrites",
+                "types-certifi",
+                "types-croniter",
+                "types-PyYAML",
+                "types-requests",
+                "types-python-slugify",
```
rchiodo, line 46: > **Warning** · Non-blocking recommendation

These editable-install packages make the performance gate depend on dependency preparation succeeding. When installation fails, the benchmark emits empty metrics and the comparator reports a candidate failure, turning a transient install/build problem into a performance regression. Either use a corpus that does not require unstable editable installs in CI or classify preparation failures separately from regressions.

build/benchmark/compare_benchmarks.py
```diff
@@ -0,0 +1,270 @@
...
+            continue
+        if old_package.get("commit") != new_package.get("commit"):
+            failures.append(
+                f"{package}/{checker}: package commit changed from "
+                f"{old_package.get('commit')} to {new_package.get('commit')}"
+            )
+            rows.append(
+                {
+                    "package": package,
+                    "checker": checker,
```
rchiodo, line 233: > **Info** · Optional note

A failed candidate clone produces a package result without `commit`, so this branch reports that the commit changed to `None` rather than reporting the clone/environment failure. Check the package error or missing commit before comparing commits so the report identifies the actual failure.

.github/workflows/typecheck_benchmark_comment.yml
```diff
@@ -0,0 +1,82 @@
...
+            )
+            if (!report) {
+              core.setFailed('The benchmark report artifact was not found')
+              return
+            }
+            const download = await github.rest.actions.downloadArtifact({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              artifact_id: report.id,
+              archive_format: 'zip',
```
rchiodo, line 66: > **Info** · Optional note

If benchmark comparison cannot load either JSON file, it exits before writing `report.md`; the artifact is still uploaded, but this workflow then fails at `readFileSync`. Handle a missing report explicitly so failed benchmark preparation produces a useful comment or is skipped cleanly.

.github/workflows/typecheck_benchmark_comment.yml
```diff
@@ -0,0 +1,82 @@
...
+              fs.readFileSync('benchmark-report/pr_number.txt', 'utf8').trim()
+            )
+            const body = `${marker}\n${report}`
+            const comments = await github.paginate(github.rest.issues.listComments, {
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: issueNumber,
+            })
+            const previous = comments.find(
+              (comment) => comment.user?.type === 'Bot' && comment.body?.includes(marker)
```
rchiodo, line 134: > **Issue** · Please address or respond

This privileged `workflow_run` job trusts both the PR number and Markdown report from an artifact produced by untrusted PR code. A fork can replace `pr_number.txt` and `report.md` to post attacker-controlled content to another PR or overwrite an existing marked bot comment. Derive the PR from `workflow_run.pull_requests` (and validate the head SHA), and render or validate the report in this trusted workflow instead.

.github/workflows/typecheck_benchmark_pr.yml
```diff
@@ -0,0 +1,84 @@
...
+        run: |
+          python build/benchmark/typecheck_benchmark.py \
+            -c pyright -r 1 -w 0 -t 600 --memory-limit-mb 8192 \
+            --skip-pyright-build --os-name linux-x64 \
+            --output build/benchmark/results
+
+      - name: Compare with baseline
+        run: |
+          set +e
+          python build/benchmark/compare_benchmarks.py \
```
rchiodo, line 96: > **Issue** · Please address or respond

This gates `ubuntu-latest` measurements against a baseline captured on a 32-CPU WSL2 host, but the comparator checks neither `cpu_count` nor `platform_details`. The resulting 10% threshold measures runner-class differences rather than Pyright regressions. Generate the baseline on the same hosted runner class, or keep this comparison informational until such a baseline exists.

.github/workflows/typecheck_benchmark_pr.yml
```diff
@@ -0,0 +1,84 @@
+name: Type checker benchmark
+
+env:
+  NODE_VERSION: '22.x'
+  PYTHON_VERSION: '3.14'
+
```
rchiodo, line 7: > **Issue** · Please address or respond

`setup-python` resolves `3.14` to the latest patch release, while the comparator requires an exact match with the baseline's `3.14.6`. The next 3.14 patch release will fail every comparison before metrics are evaluated. Pin the workflow to the baseline patch or compare Python versions at major/minor granularity.

.github/workflows/typecheck_benchmark_pr.yml
rchiodo: > **Warning** · Non-blocking recommendation

The gate applies its 10% timing and RSS threshold to one measured run on a shared runner. The harness has statistics support, but this invocation provides no variance signal, so normal host noise can fail the gate. Use multiple runs with a robust statistic or make the comparison resilient to measurement variance.

build/benchmark/typecheck_benchmark.py
```diff
@@ -0,0 +1,1375 @@
...
+    except subprocess.TimeoutExpired as exc:
+        raise BenchmarkError(
+            f"Pyright build timed out after {BUILD_TIMEOUT} seconds"
+        ) from exc
+    except OSError as exc:
+        raise BenchmarkError(f"Unable to run npm: {exc}") from exc
+
+    if result.returncode != 0:
+        raise BenchmarkError(f"Pyright build failed with exit code {result.returncode}")
+    if not PYRIGHT_BUNDLE.is_file():
```
rchiodo: > **Warning** · Non-blocking recommendation

The skip-build path validates only `dist/pyright.js`, whereas `_pyright_command` also requires `index.js`. If the entry point is missing or renamed, `--skip-pyright-build` succeeds and every package is later reported as Pyright not being installed. Validate the entry point alongside the bundle so this fails clearly before benchmarking.

.github/workflows/typecheck_benchmark_comment.yml
```diff
@@ -0,0 +1,82 @@
...
+              artifact_id: report.id,
+              archive_format: 'zip',
+            })
+            fs.writeFileSync('benchmark-report.zip', Buffer.from(download.data))
+
+      - name: Extract benchmark report
+        run: unzip -q benchmark-report.zip -d benchmark-report
+
+      - name: Post benchmark comment
+        uses: actions/github-script@v7
```
rchiodo: > **Info** · Optional note

If the benchmark or comparison step fails before writing `report.md`, the always-uploaded artifact can still contain `pr_number.txt` but no report. This job then throws on `readFileSync` instead of handling the failed benchmark result. Check for the report before reading it and skip or post a clear soft-failure message.

.github/workflows/typecheck_benchmark_pr.yml
bschnurr: Addressed in 326f31a0e. The checked baseline now comes from `ubuntu-latest`, results record `runner_class: github-ubuntu-latest`, and the comparator also requires matching `cpu_count`.

.github/workflows/typecheck_benchmark_pr.yml
bschnurr: Addressed in 326f31a0e. The workflow now pins Python `3.14.6`, matching the baseline and exact-version comparison.

build/benchmark/install_envs.json
bschnurr: Addressed in 326f31a0e. Clone/install preparation failures are reported as yellow unmeasured rows and no longer count as performance regressions.

.github/workflows/typecheck_benchmark_pr.yml
bschnurr: Addressed with a variance guard. CI uses a 20% threshold based on observed repeat `ubuntu-latest` variation; the README documents the distinction from the comparator’s local default.

build/benchmark/typecheck_benchmark.py
bschnurr: Addressed in 326f31a0e. Both normal-build and `--skip-pyright-build` paths now validate `index.js` and `dist/pyright.js` before benchmarking, with focused test coverage.

.github/workflows/typecheck_benchmark_comment.yml
bschnurr: Addressed in 326f31a0e. The trusted job derives the PR from `workflow_run.pull_requests`, validates the PR head SHA, extracts only bounded JSON, checks out default-branch code, and renders escaped Markdown itself.

.github/workflows/typecheck_benchmark_pr.yml
bschnurr: Addressed. The baseline was replaced with measurements from the same GitHub-hosted runner class; 326f31a0e additionally records and compares runner class and CPU count.

.github/workflows/typecheck_benchmark_pr.yml
bschnurr: Addressed with the 20% hosted-runner variance guard, selected from observed repeat-run variation and documented in the README.

build/benchmark/install_envs.json
bschnurr: Addressed in 326f31a0e. Package-level preparation errors remain visible in the report but are not classified as performance regressions.

build/benchmark/compare_benchmarks.py
bschnurr: Addressed in 326f31a0e. Candidate package errors are checked before commit compatibility, so clone/install failures are reported as preparation failures rather than `commit changed to None`.

.github/workflows/typecheck_benchmark_comment.yml
bschnurr: Addressed in 326f31a0e. Trusted rendering creates a fixed safe failure report when comparison cannot produce `report.md`.

.github/workflows/typecheck_benchmark_comment.yml
bschnurr: Addressed in 326f31a0e. No PR number or Markdown is trusted from the artifact; PR identity and head SHA come from GitHub, and trusted default-branch code renders escaped output from bounded JSON.

.github/workflows/typecheck_benchmark_pr.yml
bschnurr: Addressed. Baselines now contain GitHub-hosted measurements, and 326f31a0e enforces matching `runner_class` plus `cpu_count`.

.github/workflows/typecheck_benchmark_pr.yml
bschnurr: Addressed in 326f31a0e by pinning `actions/setup-python` to `3.14.6`.

.github/workflows/typecheck_benchmark_pr.yml
bschnurr: Addressed with a documented 20% CI variance guard based on observed shared-runner variation.

build/benchmark/typecheck_benchmark.py
bschnurr: Addressed in 326f31a0e. `prepare_local_pyright` validates both the entry point and bundle in both build modes, with a regression test.

.github/workflows/typecheck_benchmark_comment.yml
bschnurr: Addressed in 326f31a0e. The privileged workflow no longer reads artifact Markdown and emits a safe fallback report when trusted comparison cannot render one.

build/benchmark/test_compare_benchmarks.py
```diff
@@ -0,0 +1,147 @@
...
+        )
+
+    def test_rejects_package_scope_mismatch(self) -> None:
+        candidate = _result(10.0, 100.0)
+        candidate["results"][0]["exclude_directories"] = ["tests"]
+        with redirect_stdout(io.StringIO()):
+            failures = compare_benchmarks.compare(
+                _result(10.0, 100.0), candidate, 10.0
+            )
+
```
rchiodo, line 313: > **Issue** · Please address or respond

`test_renders_markdown_summary_and_failure` only checks fragments of deterministic rendered output. Assert the complete expected report so formatting, status rows, and failure rendering are covered together.

build/benchmark/test_typecheck_benchmark.py
```diff
@@ -0,0 +1,360 @@
...
+
+        self.assertEqual(calls, [True, False, False])
+        self.assertIn("Check...", stdout.getvalue())
+        self.assertNotIn("Warmup", stdout.getvalue())
+
+    def test_only_first_warmup_captures_output(self) -> None:
+        captures: list[bool] = []
+
+        def run_checker(
+            checker: str,
```
rchiodo, line 527: > **Issue** · Please address or respond

This test only asserts fragments of deterministic stdout. Assert the full expected output (or the exact emitted labels) so changes to the warmup/check sequencing cannot slip through.

.github/workflows/typecheck_benchmark_pr.yml
```diff
@@ -0,0 +1,89 @@
...
+          npm install
+
+      - name: Build Pyright CLI
+        working-directory: packages/pyright
+        run: npm run build
+
+      - name: Run benchmark
+        env:
+          PYTHONNOUSERSITE: '1'
+        run: |
```
rchiodo, line 85: > **Warning** · Non-blocking recommendation

The gate still evaluates one measured run with no warmup on a shared runner, so timing and high-memory-package deltas can be dominated by host variance. Use multiple runs with a robust statistic, or add a variance guard before failing the comparison.

build/benchmark/baselines/latest-linux-x64.json
```diff
@@ -0,0 +1,405 @@
...
+  },
+  "memory_measurement": "/proc/<pid>/status",
+  "memory_limit_mb": 8192,
+  "type_checkers": [
+    "pyright"
+  ],
+  "type_checker_versions": {
+    "pyright": "1.1.411"
+  },
+  "package_count": 9,
```
rchiodo, line 26: > **Warning** · Non-blocking recommendation

The PR description says NumPy timed out and is recorded as a failed datum, but this nine-package baseline includes a successful NumPy metric. Please reconcile the description with the checked-in baseline so the gated corpus is unambiguous.

.github/workflows/typecheck_benchmark_comment.yml
```diff
@@ -0,0 +1,140 @@
...
+    runs-on: ubuntu-latest
+    if: ${{ github.event.workflow_run.event == 'pull_request' }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.repository.default_branch }}
+
+      - name: Download benchmark results
+        id: download
+        uses: actions/github-script@v7
```
rchiodo: > **Issue** · Please address or respond

📍 .github/workflows/typecheck_benchmark_comment.yml:24
The Skeptic verified this is a regression introduced by the (correct) security rewrite: `workflow_run.pull_requests` is empty for **fork** PRs, so `length !== 1 → core.setFailed` turns every external-contributor PR touching these paths into a red, failed comment workflow that never posts a comment. On an open-source repo like pyright that receives most contributions from forks, this is noisy and confusing. Treat `length === 0` as a neutral skip (`core.info(...)` + `return`) instead of `setFailed`.

[verified]

build/benchmark/test_compare_benchmarks.py
```diff
@@ -0,0 +1,147 @@
...
+        self.assertEqual(failures, [])
+        report = compare_benchmarks.render_markdown(
+            _result(10.0, 100.0), candidate, 10.0
+        )
+        self.assertIn("🟡 **1 package(s) could not be prepared", report)
+        self.assertIn("🟡 Preparation failed", report)
```
rchiodo: > **Issue** · Please address or respond

📍 build/benchmark/test_compare_benchmarks.py:77
Rules confirmed this partial-assertion pattern persists and is pervasive (also L126-131, L141-143, and `test_skip_build_requires_entry_point_and_bundle` / `test_zero_warmups_uses_one_captured_check` in the sibling file). Per `tests-no-partial-asserts.md`, assert the full rendered report/string with `assertEqual` (capturing the actual value first if unknown), or add the required "genuinely non-deterministic output" justification comment. This is a documented review-rule violation and cannot be downgraded or waived for this PR.

[verified]

build/benchmark/baselines/latest-linux-x64.json
```diff
@@ -0,0 +1,405 @@
...
+    "source_file_url": "https://github.com/lolpack/type_coverage_py/blob/85667d6f090ce9648d88cd7a9777b492f3b95f1c/typecheck_benchmark/daily_runner.py"
+  },
+  "memory_measurement": "/proc/<pid>/status",
+  "memory_limit_mb": 8192,
+  "type_checkers": [
+    "pyright"
+  ],
+  "type_checker_versions": {
+    "pyright": "1.1.411"
+  },
```
rchiodo, line 25: > **Warning** · Non-blocking recommendation

📍 build/benchmark/baselines/latest-linux-x64.json:20
All four reviewers flagged that the PR description contradicts the shipped gating baseline: the description claims a "10-package corpus" and that NumPy "exceeded the 600-second timeout … recorded as a failed datum," but this file records `package_count: 9`, `packages_failed: 0`, and `numpy` `ok: true, 35.422s` (and `install_envs.json` lists 9). Reviewers can't tell which packages are load-bearing or whether this is a healthy run. Correct the description to match the data, or regenerate a baseline matching the narrative.

[verified]

bschnurr:
## Type checker benchmark

🟢 **No performance regressions detected.**

Regression threshold: `20.0%`
Variance guard: `>1.0s` time and `>100.0 MB` memory

| Package | Checker | Time | Time delta | Peak memory | Memory delta | Status |
| --- | --- | ---: | ---: | ---: | ---: | --- |
| ansible | pyright | 19.481s | -1.1% | 930.6 MB | -0.4% | 🟢 Pass |
| click | pyright | 3.237s | +0.1% | 255.9 MB | -1.6% | 🟢 Pass |
| homeassistant | pyright | 169.768s | -6.5% | 4064.2 MB | +0.5% | 🟢 Pass |
| numpy | pyright | 34.410s | -2.9% | 1446.7 MB | +0.7% | 🟢 Pass |
| pandas | pyright | 36.705s | -7.6% | 1516.0 MB | +0.1% | 🟢 Pass |
| pytest | pyright | 12.676s | -9.0% | 752.7 MB | -0.8% | 🟢 Pass |
| requests | pyright | 2.587s | -7.0% | 239.7 MB | +1.1% | 🟢 Pass |
| torch | pyright | 182.568s | -16.2% | 4056.9 MB | +0.5% | 🟢 Pass |
| transformers | pyright | 190.657s | -12.1% | 4035.5 MB | +1.4% | 🟢 Pass |

build/benchmark/test_compare_benchmarks.py
bschnurr: Addressed in 09e5bba37. The deterministic Markdown test now asserts the complete report, including summary, threshold, variance guard, table row, and escaped failure text.

build/benchmark/test_typecheck_benchmark.py
bschnurr: Addressed in 09e5bba37. The zero-warmup test now asserts the complete stdout sequence, including the validation heading, discarded check, measured runs, and mean.

.github/workflows/typecheck_benchmark_pr.yml
bschnurr: Addressed in 09e5bba37. The CI gate now requires deltas to exceed both 20% and an explicit absolute variance guard (>1.0s for time or >100 MB for RSS); both report paths disclose and apply the same guard.

build/benchmark/baselines/latest-linux-x64.json
bschnurr: Addressed. The PR description now matches the checked data: nine successful packages, NumPy measured over production sources, and SciPy removed after its hosted-runner timeout.

.github/workflows/typecheck_benchmark_comment.yml
bschnurr: Addressed in 09e5bba37. A workflow run with no associated PR now logs a neutral skip, and all extraction/render/comment steps are gated on discovery of a PR number.

build/benchmark/test_compare_benchmarks.py
bschnurr: Addressed in 09e5bba37. Partial deterministic report, failure-list, exception, and runner-stdout assertions were replaced with complete equality assertions. The focused suite now passes 23 tests.

build/benchmark/baselines/latest-linux-x64.json
bschnurr: Addressed. The PR description now states the actual nine-package corpus and successful NumPy datum, and explains that SciPy was removed after exceeding the timeout.

build/benchmark/typecheck_benchmark.py
```diff
@@ -0,0 +1,1388 @@
...
+    config_path.write_text("\n".join(lines) + "\n", encoding="utf-8")
+    return config_path
+
+
+def _failure_message(result: ProcessResult) -> str:
+    combined_output = f"{result['stderr']}\n{result['stdout']}".strip()
+    for line in combined_output.splitlines():
+        if line.strip():
+            return line.strip()[:300]
+    return f"Fatal error (exit code {result['returncode']})"
```
rchiodo, line 666: > **Warning** · Non-blocking recommendation

The fixed mypy `exclude` pattern drops `testing` from the pytest corpus even though Pyright analyzes it. This makes cross-type-checker results measure different scopes; derive mypy exclusions from each package's `exclude_directories` or remove the fixed regex.

rchiodo (approved):
Approved via Review Center.

.github/workflows/typecheck_benchmark_pr.yml
```diff
@@ -0,0 +1,92 @@
...
+          python build/benchmark/typecheck_benchmark.py \
+            -c pyright -r 1 -w 0 -t 1800 --memory-limit-mb 8192 \
```
rchiodo, line 87: > **Issue** · Please address or respond

The candidate uses `-t 1800`, which writes `timeout_s: 1800`, but the checked-in baseline records `timeout_s: 600`. Since the comparator treats `timeout_s` as part of the environment contract, every PR run will report an environment mismatch and fail regardless of measured performance. Align the workflow and baseline timeout values (or remove timeout from the contract if that is intentional).

.github/workflows/typecheck_benchmark_comment.yml
```diff
@@ -0,0 +1,149 @@
...
+            build/benchmark/baselines/latest-linux-x64.json \
+            candidate.json \
+            --threshold-percent 20 \
+            --time-noise-floor-seconds 1 \
+            --memory-noise-floor-mb 100 \
+            --markdown-output report.md
+          if [[ ! -s report.md ]]; then
+            printf '%s\n' \
+              '## Type checker benchmark' \
+              '' \
```
rchiodo, line 113: > **Warning** · Non-blocking recommendation

This report renderer duplicates the gate thresholds from the PR workflow. If either workflow changes independently, the PR comment can report no regression while the benchmark gate fails. Define the threshold and noise-floor values in one shared source, such as workflow-level environment variables or comparator configuration.

.github/workflows/typecheck_benchmark_comment.yml
```diff
@@ -0,0 +1,149 @@
...
+                  name
+                  for name in archive.namelist()
+                  if PurePosixPath(name).name == 'latest-linux-x64.json'
+                  and '..' not in PurePosixPath(name).parts
+              ]
+              if len(matches) != 1:
+                  raise RuntimeError(f'Expected one candidate result, found {len(matches)}')
+              contents = archive.read(matches[0])
+          if len(contents) > 5 * 1024 * 1024:
+              raise RuntimeError('Candidate result exceeds 5 MB')
```
rchiodo: > **Info** · Optional note

`archive.read()` decompresses the full artifact member before enforcing the 5 MB limit. Because this privileged `workflow_run` workflow consumes an artifact produced by PR code, check the member's uncompressed `ZipInfo.file_size` before reading it (or stream with a bounded read) to avoid a compressed artifact exhausting runner memory.

build/benchmark/typecheck_benchmark.py
bschnurr: Addressed in 23f2679aa. The fixed mypy test-directory exclusion was removed, so mypy now receives the same resolved paths as every other checker; an exact config test covers the pytest `testing` path.

build/benchmark/compare_benchmarks.py
```diff
@@ -0,0 +1,388 @@
...
+    time_noise_floor_s: float = 0.0,
+    memory_noise_floor_mb: float = 0.0,
+) -> tuple[list[str], list[ComparisonRow]]:
+    failures: list[str] = []
+    rows: list[ComparisonRow] = []
+    for field in (
+        "platform",
+        "architecture",
+        "runner_class",
+        "cpu_count",
```
StellaHuang95, line 136: > **Issue** · Please address or respond

`timeout_s` is part of the environment contract, but the checked-in baseline uses `600` while the PR workflow generates candidates with `-t 1800`. Every benchmark run will therefore fail comparison before evaluating performance. Align the workflow and baseline timeouts, or remove `timeout_s` from the comparison contract.

.github/workflows/typecheck_benchmark_pr.yml
```diff
@@ -0,0 +1,92 @@
...
+            --threshold-percent 20 \
+            --time-noise-floor-seconds 1 \
+            --memory-noise-floor-mb 100 \
+            --markdown-output build/benchmark/results/report.md
+          comparison_status=$?
+          cat build/benchmark/results/report.md >> "$GITHUB_STEP_SUMMARY"
+          exit "$comparison_status"
+
+      - name: Upload candidate results
+        if: always()
```
StellaHuang95, line 106: > **Warning** · Non-blocking recommendation

This comparison reads the baseline from the PR checkout, so a PR can update `baselines/latest-linux-x64.json` with inflated measurements and make its own regression pass the required gate. Compare against the base/default-branch baseline, or protect baseline changes with an appropriate review policy.

build/benchmark/compare_benchmarks.py
```diff
@@ -0,0 +1,388 @@
...
+
+
+def _packages_by_name(data: dict[str, Any]) -> dict[str, dict[str, Any]]:
+    return {
+        package["package_name"]: package
+        for package in data.get("results", [])
+        if package.get("package_name")
+    }
+
+
```
StellaHuang95, line 104: > **Warning** · Non-blocking recommendation

The comment workflow consumes a PR-authored artifact, but result entries and metric fields are not validated. For example, `results: [123]` raises `AttributeError`, and an `ok: true` metric without `execution_time_s` raises `KeyError`. Validate list/dict shapes and treat malformed or incomplete metrics as failed rows so the report can render a useful result.

build/benchmark/compare_benchmarks.py
```diff
@@ -0,0 +1,388 @@
...
+        if new_package.get("error"):
+            rows.append(
+                {
+                    "package": package,
+                    "checker": checker,
+                    "execution_time_s": None,
+                    "time_delta": None,
+                    "peak_memory_mb": None,
+                    "memory_delta": None,
+                    "status": "Preparation failed",
```
StellaHuang95, line 221: > **Warning** · Non-blocking recommendation

Only baseline package/checker pairs are examined. A newly added corpus package or checker with no corresponding baseline is silently omitted from comparison, so it is not gated at all. At minimum document that corpus additions require a regenerated baseline; preferably emit a comparison failure or warning for candidate-only entries.

StellaHuang95:

.github/workflows/typecheck_benchmark_pr.yml
StellaHuang95, line 87: > **Issue** · Please address or respond

The checked-in baseline cannot satisfy this workflow: it records `timeout_s=600` while this command uses `1800`, and its pandas scope excludes `tests` and `_testing` while `install_envs.json` does not. Regenerate or align the baseline, workflow, and corpus configuration, and add a fast contract test to prevent this drift.

build/benchmark/compare_benchmarks.py
```diff
@@ -0,0 +1,388 @@
...
+                }
+            )
+            continue
+        if old_package.get("commit") != new_package.get("commit"):
+            failures.append(
```
StellaHuang95, line 226: > **Issue** · Please address or respond

A preparation failure for a previously successful baseline package adds no failure, allowing the regression gate to pass even when every candidate package is unmeasured. Preserve the distinct report status, but append a comparison failure whenever a baseline-successful package cannot be prepared.

build/benchmark/typecheck_benchmark.py
```diff
@@ -0,0 +1,1378 @@
...
+        return {
+            "ok": False,
+            "execution_time_s": result["execution_time_s"],
+            "peak_memory_mb": result["peak_memory_mb"],
+            "oom_killed": False,
+            "error_message": f"Timed out after {timeout} seconds",
+        }
+
+    combined_output = f"{result['stderr']}\n{result['stdout']}"
+    if capture_output and "errors prevented further checking" in combined_output:
```
StellaHuang95, line 824: > **Warning** · Non-blocking recommendation

Measured runs discard output, and fatal text paired with exit code 1 is rejected only when output is captured. Capture enough output to validate every run, or implement checker-specific fatal detection that remains effective during measured runs.

build/benchmark/compare_benchmarks.py
```diff
@@ -0,0 +1,388 @@
...
+        "architecture",
+        "runner_class",
+        "cpu_count",
+        "python_version",
+        "memory_limit_mb",
+        "runs_per_package",
+        "warmup_runs",
+        "timeout_s",
+    ):
+        if baseline.get(field) != candidate.get(field):
```
StellaHuang95, line 145: > **Warning** · Non-blocking recommendation

The compatibility contract accepts results from different `ubuntu-latest` images because it ignores `platform_details` and records only a manually assigned runner class. Record and compare a stable hosted-image identity, such as the runner image version, so infrastructure changes are not reported as Pyright regressions.

build/benchmark/typecheck_benchmark.py
```diff
@@ -0,0 +1,1378 @@
...
+            )
+    return resolved_paths, missing
+
+
+def _new_config_path(package_path: Path, checker: str, suffix: str) -> Path:
+    file_descriptor, path = tempfile.mkstemp(
+        prefix=f".typecheck-benchmark-{checker}-",
+        suffix=suffix,
+        dir=package_path,
+    )
```
StellaHuang95, line 612: > **Warning** · Non-blocking recommendation

All nine packages install into one mutable Python environment, so dependency resolution for an earlier package may affect later measurements and make results order-dependent. Isolate package environments or reset dependencies between packages, and add an experiment demonstrating order independence.

.github/plans/archived/type-checker-benchmark.md
```diff
@@ -0,0 +1,68 @@
+# Pyright Type Checker Benchmark Plan
```
StellaHuang95, line 1: > **Issue** · Please address or respond

This completed implementation plan remains in the active plans directory. Move it to `.github/plans/archived/` and mark it complete so active plans remain actionable.

.github/workflows/typecheck_benchmark_pr.yml
```diff
@@ -0,0 +1,92 @@
...
+      - name: Run benchmark
+        env:
+          PYTHONNOUSERSITE: '1'
+        run: |
+          python build/benchmark/typecheck_benchmark.py \
+            -c pyright -r 1 -w 0 -t 1800 --memory-limit-mb 8192 \
+            --skip-pyright-build --os-name linux-x64 \
+            --output build/benchmark/results
+
+      - name: Compare with baseline
```
rchiodo, line 91: > **Issue** · Please address or respond

The candidate run uses `-t 1800`, but the checked-in baseline records `timeout_s: 600`; `compare_benchmarks.py` treats that as a failure. The baseline also records pandas exclusions that are absent from `install_envs.json`, causing a scope mismatch. Align the workflow profile and baseline (or regenerate the baseline) so PR runs can produce comparable results.

build/benchmark/compare_benchmarks.py
```diff
@@ -0,0 +1,388 @@
...
+                    "time_delta": None,
+                    "peak_memory_mb": None,
+                    "memory_delta": None,
+                    "status": "Missing",
+                }
+            )
+            continue
+        if new_package.get("error"):
+            rows.append(
+                {
```
rchiodo, line 214: > **Warning** · Non-blocking recommendation

A package preparation error only produces an informational row and does not add a comparison failure. Consequently, a candidate where every previously measured package fails dependency installation passes the PR gate without collecting any performance data. Strict PR comparisons should fail when a successful baseline package cannot be prepared, while weekly/local reporting can remain informational.

.github/workflows/typecheck_benchmark_comment.yml
```diff
@@ -0,0 +1,149 @@
...
+          import json
+          import zipfile
+          from pathlib import PurePosixPath
+
+          with zipfile.ZipFile('benchmark-report.zip') as archive:
+              matches = [
+                  name
+                  for name in archive.namelist()
+                  if PurePosixPath(name).name == 'latest-linux-x64.json'
+                  and '..' not in PurePosixPath(name).parts
```
rchiodo: > **Warning** · Non-blocking recommendation

The trusted workflow consumes an artifact produced by untrusted PR code, but `json.loads` accepts non-finite constants such as `NaN` and this only checks that the top-level value is an object. A `NaN` measurement makes percentage comparisons non-regressions. Reject non-standard JSON constants and validate measurement fields are finite numbers before rendering the trusted comment.

build/benchmark/typecheck_benchmark.py
```diff
@@ -0,0 +1,1378 @@
...
+    config_path = _new_config_path(package_path, checker, ".ini")
+    lines = ["[mypy]"]
+    if check_paths:
+        lines.append(f"files = {', '.join(check_paths)}")
+    lines.extend(["check_untyped_defs = True", ""])
+    config_path.write_text("\n".join(lines), encoding="utf-8")
+    return config_path
+
+
+def _write_ty_config(package_path: Path, check_paths: list[str] | None) -> Path:
```
rchiodo, line 641: > **Warning** · Non-blocking recommendation

Dependencies are unpinned and are installed into one shared environment for the whole corpus. Dependency releases or packages installed for earlier corpus entries can change later measurements without appearing in the comparison contract. Pin and record resolved dependencies, or isolate each package's environment, to make baseline comparisons reproducible.

build/benchmark/render_benchmark_html.py
```diff
@@ -0,0 +1,175 @@
...
+            "<tr>"
+            f"<th scope=\"row\"><span class=\"checker\">{html.escape(checker)}</span>"
+            f"<small>{html.escape(str(version))}</small></th>"
+            f"<td>{tested}</td><td class=\"{'failed' if failed else 'passed'}\">{failed}</td>"
+            f"<td><div class=\"bar\" style=\"--size:{share:.1f}%\"></div>"
+            f"<strong>{total:.1f}s</strong></td>"
+            f"<td>{float(aggregate.get('p50_execution_time_s', 0.0)):.1f}s</td>"
+            f"<td>{float(aggregate.get('p50_peak_memory_mb', 0.0)):.0f} MB</td>"
+            "</tr>"
+        )
```
rchiodo, line 97: > **Warning** · Non-blocking recommendation

The report selects a "fastest" checker from measurements collected by separate matrix jobs on different hosted runners, so runner variation is included in the ranking. Run checkers on the same host or present these as independent trends without a fastest designation.

build/benchmark/test_typecheck_benchmark.py
```diff
@@ -0,0 +1,389 @@
...
+            "oom_killed": False,
+        }
+        with (
+            patch.object(
+                benchmark, "_checker_command", return_value=["node", "pyright.js"]
+            ),
+            patch.object(
+                benchmark,
+                "run_process_with_timeout",
+                return_value=process_result,
```
rchiodo, line 382: > **Issue** · Please address or respond

Replace the substring assertion with an exact assertion for the deterministic mocked error message (`Fatal error (exit code 2)`). The current partial assertion can pass while the expected failure contract changes.

.github/workflows/typecheck_benchmark_pr.yml
bschnurr: Aligned the PR profile and bootstrap baseline: both now use the same 600-second timeout and the corpus configuration includes the pandas tests/_testing exclusions. The workflow also checks out the trusted base baseline first and falls back to this PR baseline only for initial bootstrap.

build/benchmark/compare_benchmarks.py
bschnurr: Addressed. PR comparisons now pass --fail-on-preparation-error, and the comparator fails when a package that was measurable in the baseline cannot be prepared in the candidate. Weekly/local reports remain informational by default.

.github/workflows/typecheck_benchmark_comment.yml
bschnurr: Addressed. The trusted workflow rejects non-standard JSON constants during parsing, and the default-branch comparator validates the full result shape plus finite timing and memory fields before rendering the report.

build/benchmark/render_benchmark_html.py
bschnurr: Addressed. The HTML no longer selects or labels a fastest checker; it explicitly presents separate hosted-runner measurements as independent trends.

build/benchmark/test_typecheck_benchmark.py
bschnurr: Addressed. The test now asserts the complete deterministic message exactly: Fatal error (exit code 2).

.github/workflows/typecheck_benchmark_pr.yml
bschnurr: Addressed: the PR workflow and baseline both use timeout_s=600, and the comparator retains timeout as part of the environment contract.

.github/workflows/typecheck_benchmark_comment.yml
bschnurr: Addressed: both the gate and trusted renderer invoke the comparator defaults, which are the single source for the 20% threshold and 1s/100MB noise floors.

.github/workflows/typecheck_benchmark_comment.yml
bschnurr: Addressed: extraction checks ZipInfo.file_size against the 5 MB limit before archive.read().

build/benchmark/compare_benchmarks.py
bschnurr: Addressed: workflow and baseline timeout values are aligned at 600 seconds.

.github/workflows/typecheck_benchmark_pr.yml
bschnurr: Addressed: the PR workflow checks out the base SHA into benchmark-baseline and prefers that trusted baseline. The PR snapshot is used only for the one-time bootstrap while main has no baseline.

build/benchmark/compare_benchmarks.py
bschnurr: Addressed: the comparator validates result/list/dict shape, required metric fields, booleans, and finite numeric measurements before analysis or trusted rendering.

build/benchmark/compare_benchmarks.py
bschnurr: Addressed: candidate-only package/checker pairs now emit explicit No baseline rows, and the README documents that baseline regeneration is required before they are regression-gated.

.github/workflows/typecheck_benchmark_pr.yml
bschnurr: Addressed: timeout and pandas scope are aligned across workflow/config/baseline, with a fast workflow/baseline profile contract test.

build/benchmark/compare_benchmarks.py
bschnurr: Addressed: preparation failures retain their report status and fail strict PR comparisons via --fail-on-preparation-error.

build/benchmark/typecheck_benchmark.py
bschnurr: Addressed: every validation, warmup, and measured checker invocation captures output and applies fatal-message/exit-code checks.

build/benchmark/compare_benchmarks.py
bschnurr: Addressed: results record ImageOS as runner_image and the comparator requires an exact runner_image match.

.github/plans/archived/type-checker-benchmark.md
bschnurr: Addressed: the completed plan is now under .github/plans/archived/ and marked complete.

build/benchmark/typecheck_benchmark.py
bschnurr: Addressed in 007778677 and 2823516fd. Each package installs with pip --target into its own directory outside the checkout; checker processes receive only that target first on PYTHONPATH with user site disabled. The new order-independence test runs two package setups and asserts distinct install and checker environments.

build/benchmark/typecheck_benchmark.py
bschnurr: Addressed in 007778677 and 2823516fd. Corpus dependencies now use a package-specific pip --target directory, checker processes receive the matching isolated PYTHONPATH, and dependency_isolation is recorded in the comparison contract. A focused two-package test demonstrates that install and checker environments remain distinct.

PR #11578, opened 2026-08-01
Fix TypedDict narrowing for literal key unions

### LINKED ISSUE #11575: False positive `reportTypedDictNotRequiredAccess` when indexing `TypedDict` after `if key in typed_dict` guard inside a `Literal` loop
#### **Environment**
- **Pyright Version**: 1.1.411
- **Python Version**: 3.13.14

---

#### **Bug Description**

When iterating over a tuple of string literal keys (e.g. `for key in ("anyOf", "oneOf"):`) and guarding a key lookup on a `total=False` (or `NotRequired`) `TypedDict` using `if key in typed_dict:`, Pyright fails to perform type narrowing on `typed_dict[key]`. 

Instead, Pyright raises a false positive `reportTypedDictNotRequiredAccess` warning:
```text
Could not access item in TypedDict
  "anyOf" is not a required key in "JsonSchema", so access may result in runtime exception
  "oneOf" is not a required key in "JsonSchema", so access may result in runtime exception
```

Even though `if key in typed_dict:` guarantees at runtime that `key` exists before indexing `typed_dict[key]`, Pyright fails to apply narrowing when `key` is a union of literal string types.

---

#### **Minimal Code Example (Repro)**

[pyright playground](https://pyright-play.net/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMoAqiApgCYAiSAxjAFD1UA2AhgM6tQBSrYKAylQAWxCMwAURBGUo0ANLDAxmjALwAxZa2IBKAFy0oUZijgB5YLqiMkrGAG0A5N14DhohwF0DUXsXOXrW0dnfiERZk96WlJiYCgEcCpidgB9VjDRMXS3ZksQ13DtKABaAD4oADlffUNgXCgAa2I4AHdcUkwUKDEAImMzYB75Ht9zHr1vQyQ4ptb2zqhs8JrDVZBiADdiZRT4KTFZtpBSIqgAYkISHziew-aezA4egBkkGGIQZUd%2B8wd5B1GwE8PUmq1YAFcAEZpDJsKAqRawux3Y4eQwXdY4PCSaTUGBVGAAJWIAEdwUh1qQAIJUJLsUGGBKoGBiRjEFBZKEwnKsbTaWhAA)

```python
from typing import TypedDict


class JsonSchema(TypedDict, total=False):
  anyOf: list['JsonSchema']
  oneOf: list['JsonSchema']


def process_schema(schema: JsonSchema) -> None:
  for keyword in ("anyOf", "oneOf"):
    if keyword in schema:
      reveal_type(keyword)  # Type of "keyword" is "Literal['anyOf', 'oneOf']"
      sub_schemas = schema[keyword]  # reportTypedDictNotRequiredAccess
      print(len(sub_schemas))

```

---

#### **Expected Behavior**

Pyright should recognize `if keyword in schema:` as a valid narrowing guard for `TypedDict` keys when `keyword` is a `Literal["anyOf", "oneOf"]` (or a `str` union of valid `TypedDict` keys), permitting safe indexing `schema[keyword]`.

subotac:
Fixes #11575.

  Preserve the correlation between literal key alternatives and narrowed TypedDict variants after a membership check. This allows guarded indexed access while retaining diagnostics for unrelated optional
  keys.

  Tests:
  - TypeNarrowingTypedDict1–3
  - complete typeEvaluator1 test suite
  - TypeScript compilation
  - scoped ESLint and Prettier checks
  - git diff --check

### DIFF
--- a/packages/pyright-internal/src/analyzer/typeGuards.ts
+++ b/packages/pyright-internal/src/analyzer/typeGuards.ts
@@ -677,9 +677,12 @@ export function getTypeNarrowingCallback(
                                     const constraintIndex = keyNarrowingInfo.conditionIndexMap.get(
                                         literalKeyType.priv.literalValue as string
                                     )!;
-                                    return addConditionToType(narrowedTypeForKey, [
-                                        { typeVar: keyNarrowingInfo.conditionTypeVar, constraintIndex },
-                                    ]);
+                                    const condition = [{ typeVar: keyNarrowingInfo.conditionTypeVar, constraintIndex }];
+                                    return mapSubtypes(narrowedTypeForKey, (subtype) =>
+                                        TypeCondition.isCompatible(getTypeCondition(subtype), condition)
+                                            ? addConditionToType(subtype, condition)
+                                            : undefined
+                                    );
                                 })
                             );
                         } else {
--- a/packages/pyright-internal/src/analyzer/typeGuards.ts
+++ b/packages/pyright-internal/src/analyzer/typeGuards.ts
@@ -592,8 +592,45 @@ export function getTypeNarrowingCallback(
                     testExpression.d.operator === OperatorType.In ? isPositiveTest : !isPositiveTest;
 
                 return (type: Type) => {
+                    let narrowedType = narrowTypeForContainerType(evaluator, type, rightType, adjIsPositiveTest);
+
+                    if (
+                        adjIsPositiveTest &&
+                        someSubtypes(
+                            rightType,
+                            (subtype) => isClassInstance(subtype) && ClassType.isTypedDictClass(subtype)
+                        )
+                    ) {
+                        const literalKeyTypes = getStringLiteralTypes(narrowedType);
+                        if (literalKeyTypes && literalKeyTypes.length > 1) {
+                            const conditionTypeVar = createTypedDictKeyConditionTypeVar(
+                                testExpression,
+                                literalKeyTypes
+                            );
+                            const conditionIndexMap = new Map(
+                                literalKeyTypes.map((keyType, index) => [keyType.priv.literalValue as string, index])
+                            );
+
+                            narrowedType = mapSubtypes(narrowedType, (subtype) => {
+                                if (
+                                    isClassInstance(subtype) &&
+                                    ClassType.isBuiltIn(subtype, 'str') &&
+                                    typeof subtype.priv.literalValue === 'string'
+                                ) {
+                                    const constraintIndex = conditionIndexMap.get(subtype.priv.literalValue);
+                                    if (constraintIndex !== undefined) {
+                                        return addConditionToType(subtype, [
+                                            { typeVar: conditionTypeVar, constraintIndex },
+                                        ]);
+                                    }
+                                }
+                                return subtype;
+                            });
+                        }
+                    }
+
                     return {
-                        type: narrowTypeForContainerType(evaluator, type, rightType, adjIsPositiveTest),
+                        type: narrowedType,
                         isIncomplete: !!rightTypeResult.isIncomplete,
                     };
                 };
@@ -609,17 +646,41 @@ export function getTypeNarrowingCallback(
                 const leftTypeResult = evaluator.getTypeOfExpression(testExpression.d.leftExpr);
                 const leftType = leftTypeResult.type;
 
-                if (isClassInstance(leftType) && ClassType.isBuiltIn(leftType, 'str') && isLiteralType(leftType)) {
-                    const adjIsPositiveTest =
-                        testExpression.d.operator === OperatorType.In ? isPositiveTest : !isPositiveTest;
+                const literalKeyTypes = getStringLiteralTypes(leftType);
+                const adjIsPositiveTest =
+                    testExpression.d.operator === OperatorType.In ? isPositiveTest : !isPositiveTest;
+                if (literalKeyTypes && (literalKeyTypes.length === 1 || adjIsPositiveTest)) {
                     return (type: Type) => {
-                        return {
-                            type: narrowTypeForTypedDictKey(
+                        let narrowedType: Type;
+                        if (literalKeyTypes.length === 1) {
+                            narrowedType = narrowTypeForTypedDictKey(
                                 evaluator,
                                 type,
-                                ClassType.cloneAsInstantiable(leftType),
+                                ClassType.cloneAsInstantiable(literalKeyTypes[0]),
                                 adjIsPositiveTest
-                            ),
+                            );
+                        } else {
+                            const conditionTypeVar = createTypedDictKeyConditionTypeVar(
+                                testExpression,
+                                literalKeyTypes
+                            );
+                            narrowedType = combineTypes(
+                                literalKeyTypes.map((literalKeyType, constraintIndex) =>
+                                    addConditionToType(
+                                        narrowTypeForTypedDictKey(
+                                            evaluator,
+                                            type,
+                                            ClassType.cloneAsInstantiable(literalKeyType),
+                                            /* isPositiveTest */ true
+                                        ),
+                                        [{ typeVar: conditionTypeVar, constraintIndex }]
+                                    )
+                                )
+                            );
+                        }
+
+                        return {
+                            type: narrowedType,
                             isIncomplete: !!leftTypeResult.isIncomplete,
                         };
                     };
@@ -2237,6 +2298,41 @@ export function narrowTypeForContainerElementType(evaluator: TypeEvaluator, refe
             return referenceSubtype;
         });
     });
+}
+
+function getStringLiteralTypes(type: Type): ClassType[] | undefined {
+    const literalTypes: ClassType[] = [];
+    let isValid = true;
+
+    doForEachSubtype(type, (subtype) => {
+        if (
+            isClassInstance(subtype) &&
+            ClassType.isBuiltIn(subtype, 'str') &&
+            typeof subtype.priv.literalValue === 'string'
+        ) {
+            literalTypes.push(subtype);
+        } else {
+            isValid = false;
+        }
+    });
+
+    if (!isValid || literalTypes.length === 0) {
+        return undefined;
+    }
+
+    return literalTypes.sort((a, b) => (a.priv.literalValue as string).localeCompare(b.priv.literalValue as string));
+}
+
+function createTypedDictKeyConditionTypeVar(node: ExpressionNode, literalKeyTypes: ClassType[]): TypeVarType {
+    const typeVar = TypeVarType.cloneForScopeId(
+        TypeVarType.createInstance('__typedDictKeyNarrowing'),
+        `typedDictKeyNarrowing.${node.id}`,
+        /* scopeName */ undefined,
+        /* scopeType */ undefined
+    );
+    typeVar.shared.isSynthesized = true;
+    literalKeyTypes.forEach((literalKeyType) => TypeVarType.addConstraint(typeVar, literalKeyType));
+    return typeVar;
 }
 
 // Attempts to narrow a type based on whether it is a TypedDict with
--- a/packages/pyright-internal/src/analyzer/typeGuards.ts
+++ b/packages/pyright-internal/src/analyzer/typeGuards.ts
@@ -592,41 +592,35 @@ export function getTypeNarrowingCallback(
                     testExpression.d.operator === OperatorType.In ? isPositiveTest : !isPositiveTest;
 
                 return (type: Type) => {
+                    const keyNarrowingInfo = createTypedDictKeyNarrowingInfo(testExpression, type);
                     let narrowedType = narrowTypeForContainerType(evaluator, type, rightType, adjIsPositiveTest);
 
                     if (
                         adjIsPositiveTest &&
+                        keyNarrowingInfo &&
+                        keyNarrowingInfo.literalKeyTypes.length > 1 &&
                         someSubtypes(
                             rightType,
                             (subtype) => isClassInstance(subtype) && ClassType.isTypedDictClass(subtype)
                         )
                     ) {
-                        const literalKeyTypes = getStringLiteralTypes(narrowedType);
-                        if (literalKeyTypes && literalKeyTypes.length > 1) {
-                            const conditionTypeVar = createTypedDictKeyConditionTypeVar(
-                                testExpression,
-                                literalKeyTypes
-                            );
-                            const conditionIndexMap = new Map(
-                                literalKeyTypes.map((keyType, index) => [keyType.priv.literalValue as string, index])
-                            );
-
-                            narrowedType = mapSubtypes(narrowedType, (subtype) => {
-                                if (
-                                    isClassInstance(subtype) &&
-                                    ClassType.isBuiltIn(subtype, 'str') &&
-                                    typeof subtype.priv.literalValue === 'string'
-                                ) {
-                                    const constraintIndex = conditionIndexMap.get(subtype.priv.literalValue);
-                                    if (constraintIndex !== undefined) {
-                                        return addConditionToType(subtype, [
-                                            { typeVar: conditionTypeVar, constraintIndex },
-                                        ]);
-                                    }
+                        narrowedType = mapSubtypes(narrowedType, (subtype) => {
+                            if (
+                                isClassInstance(subtype) &&
+                                ClassType.isBuiltIn(subtype, 'str') &&
+                                typeof subtype.priv.literalValue === 'string'
+                            ) {
+                                const constraintIndex = keyNarrowingInfo.conditionIndexMap.get(
+                                    subtype.priv.literalValue
+                                );
+                                if (constraintIndex !== undefined) {
+                                    return addConditionToType(subtype, [
+                                        { typeVar: keyNarrowingInfo.conditionTypeVar, constraintIndex },
+                                    ]);
                                 }
-                                return subtype;
-                            });
-                        }
+                            }
+                            return subtype;
+                        });
                     }
 
                     return {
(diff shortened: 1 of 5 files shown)


subotac:

rchiodo:

packages/pyright-internal/src/analyzer/typeGuards.ts
```diff
@@ -609,17 +646,41 @@ export function getTypeNarrowingCallback(
...
+                                        [{ typeVar: conditionTypeVar, constraintIndex }]
+                                    )
+                                )
+                            );
+                        }
+
+                        return {
+                            type: narrowedType,
                             isIncomplete: !!leftTypeResult.isIncomplete,
                         };
```
rchiodo, line 695: > **Warning** · Non-blocking recommendation

When a literal-key union includes an undeclared TypedDict key, `narrowTypeForTypedDictKey` drops that alternative and the condition filter then suppresses the usual invalid-key diagnostic for `value[key]`. Please preserve the diagnostic for undeclared alternatives (and add coverage for it), rather than treating `key in value` as proof that every literal is a declared key.

[verified]

packages/pyright-internal/src/analyzer/typeGuards.ts
```diff
@@ -592,8 +592,45 @@ export function getTypeNarrowingCallback(
...
+                                    if (constraintIndex !== undefined) {
+                                        return addConditionToType(subtype, [
+                                            { typeVar: conditionTypeVar, constraintIndex },
+                                        ]);
+                                    }
+                                }
+                                return subtype;
+                            });
+                        }
+                    }
```
rchiodo, line 624: > **Warning** · Non-blocking recommendation

The two callbacks independently derive the literal-key list and assign condition indexes from different inputs (`narrowedType` here versus `leftType` below). Those indexes are the correlation contract, so a future change that makes the lists differ could associate a key with the wrong TypedDict variant. Please centralize the canonical literal-key/index computation, or derive both from the same source.

[verified]

packages/pyright-internal/src/tests/samples/typeNarrowingTypedDict3.py
```diff
@@ -19,6 +19,17 @@ class MyDict1(TypedDict, total=False):
...
 
+def read_present_keys(value: MyDict1):
+    for key in ("key1", "key2"):
+        if key in value:
+            reveal_type(key, expected_text="Literal['key1', 'key2']")
+            value[key]
+
+            # This should generate an error because checking the variable key
+            # doesn't prove that this specific key is present.
+            value["key1"]
```
rchiodo, line 30: > **Warning** · Non-blocking recommendation

Please add focused coverage for the new correlation behavior beyond the exact-key happy path: an undeclared literal in the key union, assignment/deletion through a narrowed variable key, and a union of TypedDicts on the right side of `in`. These exercise the new condition filter in additional indexed-access paths.

[verified]

rchiodo (approved):
Approved via Review Center.

packages/pyright-internal/src/analyzer/typeGuards.ts
subotac: Good catch. Addressed in babaa5277: undeclared literal-key alternatives are now retained so indexed access still reports the invalid key, and the sample includes regression coverage.

packages/pyright-internal/src/analyzer/typeGuards.ts
subotac: Agreed. Addressed in babaa5277: the sorted literal list, synthesized condition TypeVar, and key-to-constraint map now come from one helper.

packages/pyright-internal/src/tests/samples/typeNarrowingTypedDict3.py
subotac: Added in babaa5277: the sample now covers an undeclared key, guarded assignment/deletion, and a union of TypedDicts. The full `typeEvaluator1.test.ts` suite passes (159/159).

packages/pyright-internal/src/analyzer/typeGuards.ts
```diff
@@ -609,17 +640,47 @@ export function getTypeNarrowingCallback(
...
+                                    ]);
+                                })
+                            );
+                        }
+
+                        return {
+                            type: narrowedType,
                             isIncomplete: !!leftTypeResult.isIncomplete,
                         };
                     };
```
rchiodo, line 696: > **Warning** · Non-blocking recommendation

The multi-literal path applies to any container, although `narrowTypeForTypedDictKey` leaves non-TypedDict subtypes unchanged. This can create redundant conditioned copies for literal unions checked against normal lists, dicts, or sets. Please gate this path on a TypedDict subtype, mirroring the other branch, and add coverage for a literal union checked against a non-TypedDict container.

[verified]

packages/pyright-internal/src/analyzer/typedDicts.ts
```diff
@@ -1521,7 +1523,12 @@ export function getTypeOfIndexedTypedDict(
     let diag = new DiagnosticAddendum();
     let allDiagsInvolveNotRequiredKeys = true;
 
+    const conditionFilter = getTypeCondition(baseType);
     const resultingType = mapSubtypes(indexType, (subtype) => {
+        if (!TypeCondition.isCompatible(getTypeCondition(subtype), conditionFilter)) {
+            return undefined;
+        }
+
```
rchiodo, line 1537: > **Warning** · Non-blocking recommendation

The condition filter is applied only in the indexed-read path. Assignment and deletion use separate TypedDict paths, and the current mutation sample uses same-typed keys, so it cannot detect a lost key/value correlation. Please add coverage with differently typed keys, or explicitly document that write-path correlation is out of scope.

[verified]

rchiodo (approved):
Approved via Review Center.

packages/pyright-internal/src/analyzer/typeGuards.ts
subotac: You're right, this also conditions ordinary containers. I'll gate the multi-literal path on a TypedDict subtype and add a regular-dict case.

packages/pyright-internal/src/analyzer/typedDicts.ts
subotac: The helper covers get, set, and delete, but you're right that the same-typed sample doesn't prove write correlation. With different value types, the assigned value needs the same condition filter; I'll add that coverage along with required-key deletion.

packages/pyright-internal/src/tests/samples/typeNarrowingTypedDict3.py
```diff
@@ -19,6 +19,58 @@ class MyDict1(TypedDict, total=False):
...
+    second: str
+
+
+def mutate_present_keys(value: MutableDict):
+    for key in ("first", "second"):
+        if key in value:
+            value[key] = value[key]
+
+            # This should generate an error for the required key alternative.
+            del value[key]
```
rchiodo, line 65: > **Warning** · Non-blocking recommendation

The write-path test is self-referential (`value[key] = value[key]`), so both sides carry the same condition and it cannot detect broken key/value correlation. Add a mismatched assignment case, such as assigning an `int` through the `second: str` alternative, and assert that it produces an error.

rchiodo (approved):
Approved via Review Center.

StellaHuang95 (approved):
Approved via Review Center.

StellaHuang95 (approved):
Approved via Review Center.

packages/pyright-internal/src/tests/samples/typeNarrowingTypedDict3.py
subotac: Added mismatched assignments in both directions, so the test now rejects `int` for `second: str` and `str` for `first: int` instead of only copying a narrowed read back to itself. The analyzer now preserves the correlated variants through the write.

subotac:
The current head is approved, but the Validation and mypy_primer workflows are still waiting for maintainer approval. Could someone start them when convenient?

StellaHuang95:

packages/pyright-internal/src/analyzer/typeGuards.ts
```diff
@@ -2239,6 +2310,59 @@ export function narrowTypeForContainerElementType(evaluator: TypeEvaluator, refe
...
+        /* scopeName */ undefined,
+        /* scopeType */ undefined
+    );
+    conditionTypeVar.shared.isSynthesized = true;
+    literalKeyTypes.forEach((literalKeyType) => TypeVarType.addConstraint(conditionTypeVar, literalKeyType));
+
+    return {
+        literalKeyTypes,
+        conditionTypeVar,
+        conditionIndexMap: new Map(
```
StellaHuang95, line 2360: > **Warning** · Non-blocking recommendation

`TypeCondition.combine` deduplicates using the TypeVar shared name and constraint index, but every synthesized TypeVar here shares `__typedDictKeyNarrowing`. Conditions from separate membership tests with the same constraint index can therefore be collapsed, corrupting nested-guard correlation. Make the shared name unique per test node, or update the comparison to use the scoped identity, and add nested-guard coverage. [verified]

StellaHuang95, line 2360: > **Warning** · Non-blocking recommendation

This correlation relies on both narrowing branches deriving the same sorted literal-key set from the same test node, so their constraint indices align. Please document that invariant near `createTypedDictKeyNarrowingInfo`; a future caller that supplies a narrowed subset could otherwise silently produce incorrect indexed types. [verified]

StellaHuang95 (approved):
Approved via Review Center.

rchiodo (approved):
Approved via Review Center.

rchiodo (approved):
Approved via Review Center.

rchiodo (approved):
Approved via Review Center.

heejaechang (approved):
Approved via Review Center.

bschnurr (approved):
Approved via Review Center.

subotac:
The PR is still clean and approved, but Validation and mypy_primer have no jobs or check runs for this SHA; both workflow records ended with no executable job and no logs, while only license/cla is attached. Could a maintainer rerun or approve those workflows?

rchiodo:
Not sure why the tests aren't running. Going to close and reopen.

rchiodo:
I found a type-safety regression in the current head (747e0bd48110d8d524066cef6c7e49975055cd85): a value carried across loop iterations is incorrectly treated as correlated with the current key, allowing an incompatible TypedDict assignment.

```python
from typing import TypedDict


class D(TypedDict):
    a: int
    b: str


def overwrite(d: D):
    previous = 0
    for key in ("a", "b"):
        if key in d:
            d[key] = previous
            previous = d[key]
```

**Before:** the merge base (`de94b525d69e6b9554f39947d6c1eaefab5c41ab`) reports an incompatible assignment at `d[key] = previous`.

**After:** this PR reports no errors. I also reproduced the missing diagnostic with the standard sample-test harness: expecting one error fails with `Expected 1 errors, got 0`.

At runtime, calling `overwrite` with `{"a": 1, "b": "ok"}` produces `{"a": 0, "b": 0}`, violating `b: str`.

The new [RHS condition filter in typedDicts.ts](https://github.com/microsoft/pyright/blob/747e0bd48110d8d524066cef6c7e49975055cd85/packages/pyright-internal/src/analyzer/typedDicts.ts#L1526-L1532) relies on conditions identified by the membership-test node. That node is reused across loop iterations, so the condition retained by `previous` from an earlier iteration is treated as if it describes the current `key`. Incompatible RHS alternatives are then filtered out before assignment compatibility is checked.

Please handle these correlations conservatively across loop backedges and add a regression case for a loop-carried value, rather than only the same-iteration `value[key] = value[key]` case. I think this should be fixed before merging.

rchiodo (changes requested):
Please fix the regression

heejaechang:
### 🟡 Pyright CLI QA — `747e0bd4` — verdict: yellow

42 focused check(s) completed.

#### Needs review: `reportArgumentType` no longer reported

The base reports `reportArgumentType` for this example; the PR does not. Is this change intended?

**`main.py`**

```python
from typing import Literal, TypedDict, TypeVar

T = TypeVar("T", int, str)

class Data(TypedDict):
    a: int
    b: str

def add(left: T, right: T) -> T:
    return left + right

def read(data: Data, key: Literal["a", "b"]):
    if key in data:
        return add(data[key], 1)  # Invalid when key selects the str entry.

read({"a": 1, "b": "text"}, "b")
```

**`pyrightconfig.json`**

```json
{
  "include": [
    "main.py"
  ],
  "pythonVersion": "3.13",
  "typeCheckingMode": "basic",
  "reportUnnecessaryComparison": "warning"
}
```

```text
```

**Base `a07bbdeae085f59396acee8739c7d862dd3a1ff1`**

```json
{
  "exit_code": 1,
  "diagnostics": [
    {
      "file": "main.py",
      "severity": "error",
      "rule": "reportArgumentType",
      "message": "Argument of type \"int | str\" cannot be assigned to parameter \"left\" of type \"T@add\" in function \"add\"\n  Type \"int | str\" is not assignable to type \"int\"\n    \"str\" is not assignable to \"int\"",
      "range": {
        "start": [
          13,
          19
        ],
        "end": [
          13,
          28
        ]
      }
    }
  ]
}
```

**PR `747e0bd48110d8d524066cef6c7e49975055cd85`**

```json
{
  "exit_code": 0,
  "diagnostics": []
}
```

</details>

#### Needs review: `reportAssignmentType` no longer reported

The base reports `reportAssignmentType` for this example; the PR does not. Is this change intended?

**`main.py`**

```python
from typing_extensions import TypedDict

class Data(TypedDict, extra_items=int):
    x: int

def read(d: Data):
    if "missing" not in d:
        reveal_type(d)
        invalid: str = d["x"]  # Must report int assigned to str.
        return invalid

read({"x": 1})  # The branch is reachable for this valid input.
```

**`pyrightconfig.json`**

```json
{
  "include": [
    "main.py"
  ],
  "pythonVersion": "3.13",
  "typeCheckingMode": "basic",
  "reportUnnecessaryComparison": "warning"
}
```

```text
```

**Base `a07bbdeae085f59396acee8739c7d862dd3a1ff1`**

```json
{
  "exit_code": 1,
  "diagnostics": [
    {
      "file": "main.py",
      "severity": "error",
      "rule": "reportAssignmentType",
      "message": "Type \"int\" is not assignable to declared type \"str\"\n  \"int\" is not assignable to \"str\"",
      "range": {
        "start": [
          8,
          23
        ],
        "end": [
          8,
          29
        ]
      }
    },
    {
      "file": "main.py",
      "severity": "information",
      "rule": null,
      "message": "Type of \"d\" is \"Data\"",
      "range": {
        "start": [
          7,
          20
        ],
        "end": [
          7,
          21
        ]
      }
    }
  ]
}
```

**PR `747e0bd48110d8d524066cef6c7e49975055cd85`**

```json
{
  "exit_code": 0,
  "diagnostics": []
}
```

</details>

**`main.py`**

```python
from typing import Literal, TypedDict, assert_type

class Data(TypedDict):
    a: int
    b: int

def read(d: Data, key: Literal["a", "b"]):
    if key in d:
        assert_type(d[key], int)
        reveal_type(d[key])
```

**`pyrightconfig.json`**

```json
{
  "include": [
    "main.py"
  ],
  "pythonVersion": "3.13",
  "typeCheckingMode": "basic",
  "reportUnnecessaryComparison": "warning"
}
```

```text
```

**Base `a07bbdeae085f59396acee8739c7d862dd3a1ff1`**

```json
{
  "exit_code": 0,
  "diagnostics": [
    {
      "file": "main.py",
      "severity": "information",
      "rule": null,
      "message": "Type of \"d[key]\" is \"int\"",
      "range": {
        "start": [
          9,
          20
        ],
        "end": [
          9,
          26
        ]
      }
    }
  ]
}
```

**PR `747e0bd48110d8d524066cef6c7e49975055cd85`**

```json
{
  "exit_code": 1,
  "diagnostics": [
    {
      "file": "main.py",
      "severity": "information",
      "rule": null,
      "message": "Type of \"d[key]\" is \"int*\"",
      "range": {
        "start": [
          9,
          20
        ],
        "end": [
          9,
          26
        ]
      }
    },
    {
      "file": "main.py",
      "severity": "error",
      "message": "\"assert_type\" mismatch: expected \"int\" but received \"int*\"",
      "range": {
        "start": [
          8,
          20
        ],
        "end": [
          8,
          26
        ]
      }
    }
  ]
}
```

</details>

**`main.py`**

```python
from typing_extensions import TypedDict

class Left(TypedDict, closed=True):
    left: int

class Right(TypedDict, closed=True):
    right: str

def read(value: Left | Right) -> int | None:
    if "left" in value:
        assert_type(value, Left)
        return value["left"]
    return None
```

**`pyrightconfig.json`**

```json
{
  "include": [
    "main.py"
  ],
  "pythonVersion": "3.13",
  "typeCheckingMode": "basic",
  "reportUnnecessaryComparison": "warning"
}
```

```text
```

**Base `a07bbdeae085f59396acee8739c7d862dd3a1ff1`**

```json
{
  "exit_code": 0,
  "diagnostics": []
}
```

**PR `747e0bd48110d8d524066cef6c7e49975055cd85`**

```json
{
  "exit_code": 1,
  "diagnostics": [
    {
      "file": "main.py",
      "severity": "error",
      "rule": "reportGeneralTypeIssues",
      "message": "Could not access item in TypedDict\n  \"left\" is not a defined key in \"Right\"",
      "range": {
        "start": [
          12,
          15
        ],
        "end": [
          12,
          28
        ]
      }
    },
    {
      "file": "main.py",
      "severity": "error",
      "message": "\"assert_type\" mismatch: expected \"Left\" but received \"Left | Right\"",
      "range": {
        "start": [
          11,
          20
        ],
        "end": [
          11,
          25
        ]
      }
    }
  ]
}
```

</details>

**`main.py`**

```python
from typing import Literal, TypeGuard, overload, assert_type

@overload
def is_value(value: object, kind: Literal["int"]) -> TypeGuard[int]: ...
@overload
def is_value(value: object, kind: Literal["str"]) -> TypeGuard[str]: ...
def is_value(value: object, kind: Literal["int", "str"]) -> bool:
    return isinstance(value, int if kind == "int" else str)

def check(value: object, kind: Literal["int", "str"]):
    if is_value(value, kind):
        assert_type(value, int | str)
    else:
        assert_type(value, object)
```

**`pyrightconfig.json`**

```json
{
  "include": [
    "main.py"
  ],
  "pythonVersion": "3.13",
  "typeCheckingMode": "basic",
  "reportUnnecessaryComparison": "warning"
}
```

```text
```

**Base `a07bbdeae085f59396acee8739c7d862dd3a1ff1`**

```json
{
  "exit_code": 0,
  "diagnostics": []
}
```

**PR `747e0bd48110d8d524066cef6c7e49975055cd85`**

```json
{
  "exit_code": 1,
  "diagnostics": [
    {
      "file": "main.py",
      "severity": "error",
      "message": "\"assert_type\" mismatch: expected \"int | str\" but received \"object\"",
      "range": {
        "start": [
          11,
          20
        ],
        "end": [
          11,
          25
        ]
      }
    }
  ]
}
```

</details>

**`main.py`**

```python
from enum import IntEnum, StrEnum
from typing import Literal, assert_type, assert_never

class Status(IntEnum):
    OK = 200
    FAILED = 500

def check(status: Status):
    if status == 200:
        assert_type(status, Literal[Status.OK])
    else:
        assert_type(status, Literal[Status.FAILED])

class Verb(StrEnum):
    GET = "GET"
    POST = "POST"

def dispatch(verb: Verb):
    if verb == "GET":
        assert_type(verb, Literal[Verb.GET])
    elif verb == "POST":
        assert_type(verb, Literal[Verb.POST])
    else:
        assert_never(verb)
```

**`pyrightconfig.json`**

```json
{
  "include": [
    "main.py"
  ],
  "pythonVersion": "3.13",
  "typeCheckingMode": "basic",
  "reportUnnecessaryComparison": "warning"
}
```

```text
```

**Base `a07bbdeae085f59396acee8739c7d862dd3a1ff1`**

```json
{
  "exit_code": 0,
  "diagnostics": []
}
```

**PR `747e0bd48110d8d524066cef6c7e49975055cd85`**

```json
{
  "exit_code": 1,
  "diagnostics": [
    {
      "file": "main.py",
      "severity": "error",
      "rule": "reportArgumentType",
      "message": "Argument of type \"Verb\" cannot be assigned to parameter \"arg\" of type \"Never\" in function \"assert_never\"\n  Type \"Verb\" is not assignable to type \"Never\"",
      "range": {
        "start": [
          23,
          21
        ],
        "end": [
          23,
          25
        ]
      }
    },
    {
      "file": "main.py",
      "severity": "error",
      "message": "\"assert_type\" mismatch: expected \"Literal[Status.FAILED]\" but received \"Status\"",
      "range": {
        "start": [
          11,
          20
        ],
        "end": [
          11,
          26
        ]
      }
    },
    {
      "file": "main.py",
      "severity": "error",
      "message": "\"assert_type\" mismatch: expected \"Literal[Status.OK]\" but received \"Status\"",
      "range": {
        "start": [
          9,
          20
        ],
        "end": [
          9,
          26
        ]
      }
    },
    {
      "file": "main.py",
      "severity": "error",
      "message": "\"assert_type\" mismatch: expected \"Literal[Verb.GET]\" but received \"Verb\"",
      "range": {
        "start": [
          19,
          20
        ],
        "end": [
          19,
          24
        ]
      }
    },
    {
      "file": "main.py",
      "severity": "error",
      "message": "\"assert_type\" mismatch: expected \"Literal[Verb.POST]\" but received \"Verb\"",
      "range": {
        "start": [
          21,
          20
        ],
        "end": [
          21,
          24
        ]
      }
    }
  ]
}
```

</details>

PR #11579, opened 2026-08-02
Fix false positive reportOverlappingOverload for recursive TypeVar lower bounds

crypt0lith:
## Summary

Fixes a regression introduced by #11415.

This caused a false positive `reportOverlappingOverload` on valid overloads whose return types are generic in a recursive way. The occurs check rejected an assignment when recording the lower bound would create a cyclic constraint. It did not account for the case where `constraints` is undefined: this is used when the caller is only testing assignability, not solving for the TypeVar, so nothing is propagated and the cyclic lower bound is satisfiable. Overload return-consistency checking uses such an assignability test, so rejecting in that scenario reported valid overloads as overlapping with an incompatible return type.

## Root cause

Overload return-consistency checking calls `assignType(returnType, prevReturnType, /* constraints */ undefined, ...)` (`checker.ts`) purely to test whether the earlier overload's return is assignable to the later one. When the two returns are generic such that this test must solve `V :> C[V]` (a TypeVar required to be a supertype of a type that mentions it), the assignment reaches the occurs check in `assignUnconstrainedTypeVar`.

The occurs check treats any such cyclic lower bound as unsatisfiable and returns `false`. This is correct when solving for the TypeVar, since propagating the self-reference expands without terminating. On the assignability test nothing is propagated, and the constraint is satisfiable. Returning `false` there reports a valid assignment as invalid. This surfaces as `overloadReturnTypeMismatch` under the `reportOverlappingOverload` rule.

## Fix

Gate the rejection on `constraints` being defined. When `constraints` is undefined the caller is only testing assignability, so instead of rejecting, widen the lower bound to `Unknown` (a valid solution) to escape the infinite recursion and let the assignability test proceed as normal. The constraint-solving path is unchanged and continues to reject as before, preserving the original hang fix.

## Test

Adds `func37` to `overloadOverlap1`, a narrow-before-wide overload pair whose returns induce the `V :> C[V]` constraint. It reports a false positive without the fix (the sample's expected error count fails) and is clean with it.


crypt0lith:

rchiodo:

rchiodo (approved):
Approved via Review Center.

StellaHuang95 (approved):
Approved via Review Center.

StellaHuang95 (approved):
Approved via Review Center.

StellaHuang95:

rchiodo (approved):
Approved via Review Center.

rchiodo (approved):
Approved via Review Center.

heejaechang (approved):
Approved via Review Center.

heejaechang (approved):
Approved via Review Center.

crypt0lith:
Rebased onto `main` to resolve merge conflict in `packages/pyright-internal/src/tests/samples/overloadOverlap1.py` (renamed `func35` to `func37`)

heejaechang:
### 🔴 Pyright CLI QA — `979506cf` — verdict: red

23 focused check(s) completed.

#### Needs review: `reportOverlappingOverload` no longer reported

The base reports `reportOverlappingOverload` for this example; the PR does not. Is this change intended?

**`main.py`**

```python
from typing import Callable, assert_type, overload, reveal_type

@overload
def factory[T](tag: str, value: T) -> Callable[[T], list[T]]: ...
@overload
def factory[T](tag: object, value: T) -> Callable[[T], T]: ...
def factory[T](tag: object, value: T) -> Callable[[T], T | list[T]]:
    def callback(x: T) -> T | list[T]:
        return [x] if isinstance(tag, str) else x
    return callback

def use(tag: object) -> int:
    callback = factory(tag, 0)
    assert_type(callback, Callable[[int], int])
    reveal_type(callback)
    return callback(1)

assert_type(factory("wrap", 0)(1), list[int])
assert_type(use("wrap"), int)
```

**`pyrightconfig.json`**

```json
{
  "pythonVersion": "3.13",
  "reportOverlappingOverload": "error",
  "reportUnnecessaryTypeIgnoreComment": "error"
}
```

```text
```

**Base `37c03866d7929e4745e802c37e7e9bcabeda080b`**

```json
{
  "exit_code": 1,
  "diagnostics": [
    {
      "file": "main.py",
      "severity": "error",
      "rule": "reportOverlappingOverload",
      "message": "Overload 1 for \"factory\" overlaps overload 2 and returns an incompatible type",
      "range": {
        "start": [
          3,
          4
        ],
        "end": [
          3,
          11
        ]
      }
    },
    {
      "file": "main.py",
      "severity": "information",
      "rule": null,
      "message": "Type of \"callback\" is \"(int) -> int\"",
      "range": {
        "start": [
          14,
          16
        ],
        "end": [
          14,
          24
        ]
      }
    }
  ]
}
```

**PR `979506cfee2748b4be8d9d60848c305c7a8de904`**

```json
{
  "exit_code": 0,
  "diagnostics": [
    {
      "file": "main.py",
      "severity": "information",
      "rule": null,
      "message": "Type of \"callback\" is \"(int) -> int\"",
      "range": {
        "start": [
          14,
          16
        ],
        "end": [
          14,
          24
        ]
      }
    }
  ]
}
```

</details>

crypt0lith:
> The base reports `reportOverlappingOverload` for this example; the PR does not. Is this change intended?

That false positive `reportOverlappingOverload` is precisely what this PR exists to address, so yes the change is intended.

heejaechang:
I understand the false-positive argument when a recursive TypeVar constraint shows up during assignability testing rather than actual type solving.

What I'm struggling with is the tradeoff. For concrete incompatible types, we still report an error. With this change, it seems like recursive TypeVar cases go from "can't satisfy this constraint" to "don't report an error".

In a case like T :> list[T], there doesn't seem to be any obvious finite solution. Wouldn't this change suppress what could be a legitimate reportOverlappingOverload diagnostic?

Put another way, why is it better to assume this is a false positive rather than a real overlap issue? Have you seen many examples where the current behavior reports an error incorrectly, versus catching a genuinely problematic overload?

crypt0lith:
In your repro you used this:

```python
@overload
def factory[T](tag: str, value: T) -> Callable[[T], list[T]]: ...
@overload
def factory[T](tag: object, value: T) -> Callable[[T], T]: ...
def factory[T](tag: object, value: T) -> Callable[[T], T | list[T]]:
    def callback(x: T) -> T | list[T]:
        return [x] if isinstance(tag, str) else x
    return callback
```

This is legitimate Python and correctly typed. On the base, this produced a false-positive diagnostic; on the PR, this does not produce that same diagnostic. Pre-regression, that would also not have produced the same diagnostic.

During overlapping overload validation, we are not looking for a finite solution, we are looking for a finite _least_ solution. When one does not exist, as in the recursive generic case, that lower bound can be safely widened to `Unknown` to allow the assignability test to continue as before. It is not changing assignability test semantics. It is simply not failing at the occurs check, to preserve pre-regression behavior as well as the hang fix which the occurs check was implemented to address.

heejaechang:
This used to pass, but was that intentional behavior, or did the old check miss an unsafe overlap that the hang fix happened to expose?

In the repro, `use("wrap")` is typed as returning `int`, but actually returns `[1]`. Isn't that what this diagnostic is meant to catch?

My question is whether reporting the diagnostic is correct for this example, rather than whether this restores the previous behavior.

crypt0lith:
#11415 added the occurs check to fix a hang in the constraint solver. It wasn't meant to change how overload overlaps are evaluated, but it did: overload pairs that were never reported before now get `reportOverlappingOverload`. This PR undoes that side effect and keeps the hang fix. Rejection still happens whenever constraints are being solved.

I compared the behavior across a range of overload shapes. This PR gives the same results as 1.1.409 (before #11415) on all of them, so it restores the previous behavior without changing anything else. mypy also leaves the example from the QA comment unreported.

Whether Pyright should report overlaps like this one is a fair question, but it's a policy change and should be proposed and reviewed on its own, not inherited as a side effect of a hang fix. That's outside this PR's scope.

heejaechang:
Regardless of how the behavior changed, my question is whether the diagnostic is correct. Restoring the previous behavior doesn't by itself mean we're fixing a false positive.


You said this is a false positive, so can you give us an example where the diagnostic is incorrect? Then we can see whether we can remove it for that class of false positives while keeping it for cases where it correctly catches an unsafe overlap.


I don't think debating the process helps here. Let's focus on the examples and what the product should report.

crypt0lith:
The occurs check from #11415 treats `R :> R | Awaitable[R]` as having "no finite solution" (see its code comment), but `object` satisfies it. In assignability-only checks any solution suffices, so rejecting it produces false positives: e.g. sympy rings.py:907–908, where `PolyElement[Er]` is reported as violating its `RingElement` bound, which the same file uses without error at lines 340/499/543/686, and which 1.1.409 didn't report.

I'd appreciate a maintainer's call explicitly on whether this approach is acceptable or should be narrowed.

heejaechang:
The SymPy example you gave doesn't show `reportOverlappingOverload`, even before this PR. It shows `reportInvalidTypeArguments`. That diagnostic serves a different purpose, although both can be affected by the same code path.

I think we should handle the cases like this:

| Case | Assessment | Desired outcome |
|---|---|---|
| SymPy's two reported bound errors | False positives | Fix those `reportInvalidTypeArguments` errors |
| `factory` and `func37` unsafe overloads | Correct diagnostics | Keep `reportOverlappingOverload` |
| Original recursive-inference hang | Real bug | Keep the hang fix |

We should fix the SymPy false positives without losing the diagnostics that correctly catch unsafe overloads.

crypt0lith:
The intention of this PR is to restore the overload evaluation policy after it was unintentionally affected by #11415. I can't agree that #11415 was some "happy accident" that fixed the way Pyright evaluates overloads. The requested changes would involve reverting the single thing the PR exists to address, plus fundamental changes to overload validation which would invite even more scrutiny. If those are blocking changes, consider closing this PR.

PR #11584, opened 2026-08-04
Support enum assignment to complete literal unions

### LINKED ISSUE #11573: Treat literal of all enum members equal to type itself
**Describe the bug**
Treat Literal of all members of an enum the same way as the annotation of just that enum.

**Code or Screenshots**
https://discuss.python.org/t/treat-literal-of-all-elements-of-enum-equal-to-the-enum-in-type-annotations/108317
Here is the discuss.python.org link where i asked and showcased this issue. It seems to be a feature in most other type checkers already, with pyright being one of the exceptions.

Code re-pasted from there:
```py
import enum
import typing as t

class Importance(enum.IntEnum):
    LOW = 1
    MEDIUM = 2
    HIGH = 3

type AllT = t.Literal[Importance.LOW, Importance.MEDIUM, Importance.HIGH]

class Task[T: Importance]:
    ...

def manage_task(task: Task[AllT]) -> t.Any:
    ...

task: Task[Importance] = ...

manage_task(task)  # error
```

vetrovk:
Fixes #11573.

The typing specification treats a non-Flag enum as equivalent to the union of all of its literal members.

Pyright already handled assignment from an enum literal union to the enum type, but the reverse direction failed because a broad enum source was tested independently against each destination literal.

Expand non-literal enums to their complete literal-member union before union assignability checks. Apply the same normalization symmetrically for `assert_type`.

The change preserves existing behavior for:
- incomplete literal unions
- `enum.Flag`
- enums without known members
- underlying `int` and `str` literals

Tests cover ordinary Enum, IntEnum, StrEnum, the generic case from the issue, symmetric `assert_type` checks, incomplete unions, Flag, and underlying literals.

Validation:
- `typeEvaluator3.test.ts`: 172/172 passed
- `npm run build:cli:dev`
- Prettier
- ESLint
- `git diff --check`
- conformance and generic reproducers

### DIFF
--- a/packages/pyright-internal/src/analyzer/typeEvaluator.ts
+++ b/packages/pyright-internal/src/analyzer/typeEvaluator.ts
@@ -9016,9 +9016,17 @@ export function createTypeEvaluator(
         // The spec is unclear on whether this is the correct behavior, but it seems to be
         // what mypy does -- and what various library authors expect.
         const arg0Type = stripTypeGuard(arg0TypeResult.type);
+        let arg0TypeForComparison = arg0Type;
+
+        if (isClassInstance(arg0Type) && arg0Type.priv.literalValue === undefined && ClassType.isEnumClass(arg0Type)) {
+            const expandedLiteralTypes = enumerateLiteralsForType(evaluatorInterface, arg0Type);
+            if (expandedLiteralTypes && expandedLiteralTypes.length > 0) {
+                arg0TypeForComparison = combineTypes(expandedLiteralTypes);
+            }
+        }
 
         if (
-            !isTypeSame(assertedType, arg0Type, {
+            !isTypeSame(assertedType, arg0TypeForComparison, {
                 treatAnySameAsUnknown: true,
                 ignorePseudoGeneric: true,
                 ignoreConditions: true,
@@ -25185,6 +25193,20 @@ export function createTypeEvaluator(
         }
 
         if (isUnion(destType)) {
+            if (isClassInstance(srcType) && srcType.priv.literalValue === undefined && ClassType.isEnumClass(srcType)) {
+                const expandedLiteralTypes = enumerateLiteralsForType(evaluatorInterface, srcType);
+                if (expandedLiteralTypes && expandedLiteralTypes.length > 0) {
+                    return assignType(
+                        destType,
+                        combineTypes(expandedLiteralTypes),
+                        diag,
+                        constraints,
+                        flags,
+                        recursionCount
+                    );
+                }
+            }
+
             // If both the source and dest are unions, use assignFromUnionType which has
             // special-case logic to handle this case.
             if (isUnion(srcType)) {
--- a/packages/pyright-internal/src/analyzer/typeEvaluator.ts
+++ b/packages/pyright-internal/src/analyzer/typeEvaluator.ts
@@ -27698,6 +27698,19 @@ export function createTypeEvaluator(
                 // one we choose) or one or both include gradual types (Any, etc.),
                 // in which case we'll want to stick with the declared subtype.
                 if (assignType(assignedSubtype, declaredSubtype)) {
+                    // Preserve an assigned enum literal rather than widening it back to the
+                    // declared enum type when the enum has only one member.
+                    if (
+                        isClassInstance(assignedSubtype) &&
+                        assignedSubtype.priv.literalValue !== undefined &&
+                        ClassType.isEnumClass(assignedSubtype) &&
+                        isClassInstance(declaredSubtype) &&
+                        declaredSubtype.priv.literalValue === undefined &&
+                        ClassType.isSameGenericClass(assignedSubtype, declaredSubtype)
+                    ) {
+                        return assignedSubtype;
+                    }
+
                     // We need to be careful with TypedDict types that have
                     // narrowed fields. In this case, we want to return the
                     // assigned type.
--- a/packages/pyright-internal/src/analyzer/typeEvaluator.ts
+++ b/packages/pyright-internal/src/analyzer/typeEvaluator.ts
@@ -9017,6 +9017,7 @@ export function createTypeEvaluator(
         // what mypy does -- and what various library authors expect.
         const arg0Type = stripTypeGuard(arg0TypeResult.type);
         let arg0TypeForComparison = arg0Type;
+        let assertedTypeForComparison = assertedType;
 
         if (isClassInstance(arg0Type) && arg0Type.priv.literalValue === undefined && ClassType.isEnumClass(arg0Type)) {
             const expandedLiteralTypes = enumerateLiteralsForType(evaluatorInterface, arg0Type);
@@ -9026,7 +9027,18 @@ export function createTypeEvaluator(
         }
 
         if (
-            !isTypeSame(assertedType, arg0TypeForComparison, {
+            isClassInstance(assertedType) &&
+            assertedType.priv.literalValue === undefined &&
+            ClassType.isEnumClass(assertedType)
+        ) {
+            const expandedLiteralTypes = enumerateLiteralsForType(evaluatorInterface, assertedType);
+            if (expandedLiteralTypes && expandedLiteralTypes.length > 0) {
+                assertedTypeForComparison = combineTypes(expandedLiteralTypes);
+            }
+        }
+
+        if (
+            !isTypeSame(assertedTypeForComparison, arg0TypeForComparison, {
                 treatAnySameAsUnknown: true,
                 ignorePseudoGeneric: true,
                 ignoreConditions: true,
--- a/packages/pyright-internal/src/analyzer/typeEvaluator.ts
+++ b/packages/pyright-internal/src/analyzer/typeEvaluator.ts
@@ -2340,6 +2340,19 @@ export function createTypeEvaluator(
         });
     }
 
+    function expandEnumTypeToLiteralUnion(type: Type): Type {
+        return mapSubtypes(type, (subtype) => {
+            if (isClassInstance(subtype) && subtype.priv.literalValue === undefined && ClassType.isEnumClass(subtype)) {
+                const expandedLiteralTypes = enumerateLiteralsForType(evaluatorInterface, subtype);
+                if (expandedLiteralTypes && expandedLiteralTypes.length > 0) {
+                    return combineTypes(expandedLiteralTypes);
+                }
+            }
+
+            return subtype;
+        });
+    }
+
     function solveAndApplyConstraints(
         type: Type,
         constraints: ConstraintTracker,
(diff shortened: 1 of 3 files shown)


rchiodo:

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -9016,9 +9016,29 @@ export function createTypeEvaluator(
...
+            assertedType.priv.literalValue === undefined &&
+            ClassType.isEnumClass(assertedType)
+        ) {
+            const expandedLiteralTypes = enumerateLiteralsForType(evaluatorInterface, assertedType);
+            if (expandedLiteralTypes && expandedLiteralTypes.length > 0) {
+                assertedTypeForComparison = combineTypes(expandedLiteralTypes);
+            }
+        }
+
+        if (
```
rchiodo, line 9052: > **Warning** · Non-blocking recommendation

This normalizes only when an entire asserted or actual type is a bare enum instance. `assert_type(value, Literal[E.A, E.B] | int)` still fails for `value: E | int`, even though the corresponding union assignment expands its enum subtype. Please normalize enum members within unions here as well, and add coverage for this shape.

rchiodo, line 9052: > **Warning** · Non-blocking recommendation

The guarded enum-to-literal-union expansion is now duplicated for both `assert_type` operands and the `assignType` source. Please consider centralizing this normalization helper so future enum edge-case adjustments cannot drift across the three copies.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -25185,6 +25205,20 @@ export function createTypeEvaluator(
...
+                const expandedLiteralTypes = enumerateLiteralsForType(evaluatorInterface, srcType);
+                if (expandedLiteralTypes && expandedLiteralTypes.length > 0) {
+                    return assignType(
+                        destType,
+                        combineTypes(expandedLiteralTypes),
+                        diag,
+                        constraints,
+                        flags,
+                        recursionCount
+                    );
```
rchiodo: > **Info** · Optional note

A single-member enum's complete literal-member union is represented as a lone `Literal[Single.ONLY]`, not a `UnionType`, so this union-only branch never runs. `single_value: Literal[Single.ONLY] = broad_single_enum` therefore remains rejected. Please decide whether the intended rule includes this degenerate complete union and add a regression test.

packages/pyright-internal/src/tests/samples/enum1.py
```diff
@@ -15,6 +15,76 @@ class TestEnum3(Enum):
...
+
+
+def test_flag_literal_union(value: TestFlagLiteralUnion) -> None:
+    # This should generate an error because Flag values can be combined.
+    full_union: Literal[TestFlagLiteralUnion.A, TestFlagLiteralUnion.B] = value
+
+
 a = TestEnum1["A"]
 aa = TestEnum1.A
 reveal_type(aa.name, expected_text="Literal['A']")
```
rchiodo, line 130: > **Warning** · Non-blocking recommendation

Please add boundary coverage for TypeVar/Optional inference (for example, reveal the result of `def f[T](x: T | None) -> T` called with a broad enum) and for single-member enum assignment and `assert_type`. These cases exercise the constraint-solving and non-union destination boundaries of the new normalization.

rchiodo:
When this change is integrated into pyrx, mirror the enum normalization in Pylance's separate async evaluator to preserve sync/async parity.

rchiodo (approved):
Approved via Review Center.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -25184,6 +25199,11 @@ export function createTypeEvaluator(
             return true;
         }
 
+        const expandedEnumSrcType = expandEnumTypeToLiteralUnion(srcType);
+        if (expandedEnumSrcType !== srcType) {
+            return assignType(destType, expandedEnumSrcType, diag, constraints, flags, recursionCount);
```
rchiodo, line 25221: > **Issue** · Please address or respond

Expanding the source before checking the assignment context breaks invariant type-argument comparisons: `list[Color]` to `list[Color]`, `dict[Color, int]` to itself, and `list[Color | int]` to itself now fail because the source becomes a literal union while the destination remains `Color`. Please preserve identity for invariant comparisons while enabling the intended enum-to-complete-literal-union widening, and add regressions for these container self-assignments.

packages/pyright-internal/src/tests/samples/enum1.py
```diff
@@ -15,6 +15,105 @@ class TestEnum3(Enum):
...
+
+def test_single_member_enum(
+    value: TestSingleMemberEnum, literal_value: Literal[TestSingleMemberEnum.ONLY]
+) -> None:
+    full_union: Literal[TestSingleMemberEnum.ONLY] = value
+    assert_type(value, Literal[TestSingleMemberEnum.ONLY])
+    assert_type(literal_value, TestSingleMemberEnum)
+
+
+def test_enum_without_known_members(value: Enum) -> None:
```
rchiodo, line 74: > **Warning** · Non-blocking recommendation

The added `test_generic_enum_literal_union` covers class-generic container assignment, not constraint solving through `T | None`. Please add a regression such as `def f[T](x: T | None) -> T` called with a multi-member enum, asserting the revealed type remains `TestEnum3` rather than a literal union. This pins the intended boundary that enum expansion does not leak into TypeVar inference.

[verified]

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -25184,6 +25216,11 @@ export function createTypeEvaluator(
             return true;
         }
 
+        const expandedEnumSrcType = expandEnumTypeToLiteralUnionForDestination(srcType, destType);
+        if (expandedEnumSrcType !== srcType) {
```
rchiodo, line 25220: > **Info** · Optional note

This executes before checking the destination shape, so every assignment from a bare enum source enumerates its members even when the destination cannot accept the resulting literal union. Since `assignType` is a hot path, add a cheap precondition that permits union destinations and single literal destinations, while preserving the single-member-enum case.

[verified]

rchiodo (approved):
Approved via Review Center.

rchiodo:
The mypy-primer found a bunch of differences. Can you double check those are expected with the new change? Sympy ones can usually be ignored since that seems to be flakey  but the other ones look like real differences.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -25184,6 +25216,11 @@ export function createTypeEvaluator(
...
+            return assignType(destType, expandedEnumSrcType, diag, constraints, flags, recursionCount);
+        }
```
rchiodo, line 25222: > **Warning** · Non-blocking recommendation

This hot-path expansion enumerates every member of a bare enum even when the destination cannot be its literal union (for example, enum-to-same-enum assignments). Add a cheap destination precondition before calling `expandEnumTypeToLiteralUnionForDestination`, while retaining support for a single literal destination so the single-member-enum case continues to work.

[verified]

packages/pyright-internal/src/tests/samples/enum1.py
```diff
@@ -15,6 +15,116 @@ class TestEnum3(Enum):
...
+
+def test_str_enum_literal_union(value: TestStrEnumLiteralUnion) -> None:
+    full_union: Literal[TestStrEnumLiteralUnion.A, TestStrEnumLiteralUnion.B] = value
+    # This should generate an error because enum literals are distinct from
+    # literals of their underlying values.
+    underlying_values: Literal["a", "b"] = value
+
+
+class TestFlagLiteralUnion(Flag):
+    A = auto()
```
rchiodo, line 119: > **Warning** · Non-blocking recommendation

Please add a `T | None` constraint-solving regression case that passes a multi-member enum and verifies `T` remains the bare enum type. The current generic-container test exercises invariant type arguments, but it does not pin this separate inference boundary against future expansion leakage.

[verified]

rchiodo:
The single-member-enum narrowing is an intentional behavior change beyond the main assignment fix; document that effect explicitly in the PR description.

rchiodo (approved):
Approved via Review Center.

StellaHuang95 (approved):
Approved via Review Center.

StellaHuang95 (approved):
Approved via Review Center.

rchiodo:
`IntFlag` subclasses bypass the new Flag exclusion, allowing unsound assignment to a complete-looking literal union.

rchiodo (approved):
Approved via Review Center.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -2340,6 +2340,36 @@ export function createTypeEvaluator(
         });
     }
 
+    function expandEnumTypeToLiteralUnion(type: Type): Type {
+        return mapSubtypes(type, (subtype) => {
+            if (isClassInstance(subtype) && subtype.priv.literalValue === undefined && ClassType.isEnumClass(subtype)) {
+                const expandedLiteralTypes = enumerateLiteralsForType(evaluatorInterface, subtype);
```
rchiodo, line 2346: > **Issue** · Please address or respond

`enumerateLiteralsForType` excludes only direct `Flag` bases, so `IntFlag` and indirect `Flag` subclasses can be expanded and accepted as exhaustive literal unions even though combined values remain valid. Exclude the complete `Flag` inheritance family and add `IntFlag` plus indirect-subclass regressions.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -25184,6 +25216,11 @@ export function createTypeEvaluator(
             return true;
         }
 
+        const expandedEnumSrcType = expandEnumTypeToLiteralUnionForDestination(srcType, destType);
```
rchiodo, line 25219: > **Warning** · Non-blocking recommendation

This runs in the central `assignType` path and enumerates broad enum sources before determining whether the destination can contain the complete literal union. Please add a cheap destination-shape gate or demonstrate that this enumeration is sufficiently cached for enum-heavy workloads.

rchiodo, line 25219: > **Warning** · Non-blocking recommendation

Pylance has a structurally separate async evaluator counterpart for this logic. Before this change is vendored into Pylance, ensure that implementation receives equivalent normalization or add TSP-mode parity coverage.

vetrovk:
Thanks for the detailed review. I checked the latest main before applying the follow-up changes and found that #11597 has now been merged and covers the same enum literal expansion behavior along with the Flag/IntFlag, destination gating, TypeVar, invariant container, and narrowing cases discussed here.
Rebasing this PR would therefore either duplicate the merged implementation or leave no meaningful diff. I'll close this as superseded by #11597. Thanks again for the review and feedback.

PR #11593, opened 2026-08-06
Implement PEP 800 disjoint-base semantics

rchiodo:
## Summary
- recognize explicit `@disjoint_base` metadata and implicit disjoint bases from nonempty `__slots__`, including synthesized dataclass slots
- propagate effective disjoint bases through inheritance and reject class definitions with unrelated candidate bases
- reject `@disjoint_base` on functions, TypedDict classes, and Protocol classes with targeted diagnostics
- add upstream-derived PEP 800 conformance coverage for nominal classes, NamedTuple, inherited bases, slots, and invalid decorator targets

## Tests
- `npx jest typeEvaluator6.test --runInBand -t "DisjointBase1" --forceExit`
- `npm run typecheck`
- targeted class, Protocol, TypedDict, NamedTuple, slots, and dataclass-slots regression tests (isinstance narrowing for disjoint bases is intentionally deferred to a follow-up)


StellaHuang95:

packages/pyright-internal/src/analyzer/checker.ts
```diff
@@ -5659,6 +5661,43 @@ export class Checker extends ParseTreeWalker {
...
+                return;
+            }
+
+            const candidate = ClassType.getDisjointBase(baseClass);
+            if (!candidate) {
+                // The base class is invalid or its disjoint base is unknown.
+                return;
+            }
+
+            if (!candidates.some((existingCandidate) => ClassType.isSameGenericClass(existingCandidate, candidate))) {
```
StellaHuang95, line 5715: > **Warning** · Non-blocking recommendation

Returning on an unresolved or unknown base suppresses a definite conflict among the remaining known bases. For example, `class C(Left, Right, Unresolved)` will not report the incompatible `Left`/`Right` disjoint bases regardless of base order. Please preserve diagnostics for known-incompatible candidates where the unknown base cannot make them related.

packages/pyright-internal/src/analyzer/types.ts
```diff
@@ -1255,6 +1258,36 @@ export namespace ClassType {
...
+
+        const candidates = classType.shared.mro.filter(
+            (mroClass): mroClass is ClassType => isInstantiableClass(mroClass) && isDisjointBase(mroClass)
+        );
+
+        return candidates.find((candidate) =>
+            candidates.every((otherCandidate) => ClassType.isDerivedFrom(candidate, otherCandidate))
+        );
+    }
+
```
StellaHuang95, line 1324: > **Warning** · Non-blocking recommendation

The unique-most-derived reduction is duplicated here and in `_validateDisjointBaseClass`. These implementations define the same PEP 800 rule and can drift if its tie-break semantics change; please extract the shared reduction into a single helper.

packages/pyright-internal/src/tests/typeEvaluator6.test.ts
```diff
@@ -1055,3 +1055,9 @@ test('Decorator7', () => {
 
     TestUtils.validateResults(analysisResults, 0);
 });
+
+test('DisjointBase1', () => {
+    const analysisResults = TestUtils.typeAnalyzeSampleFiles(['disjointBase1.py']);
+
+    TestUtils.validateResults(analysisResults, 8);
+});
```
StellaHuang95, line 1062: > **Info** · Optional note

The PR description says this includes `isinstance`-narrowing regression coverage, but this diff adds only class-definition validation and no narrowing consumer or test. Please either include that coverage or clarify that narrowing support is intentionally deferred.

StellaHuang95 (approved):
Approved via Review Center.

packages/pyright-internal/src/analyzer/types.ts
rchiodo: Fixed — extracted the unique-most-derived reduction into `ClassType.getMostDerivedDisjointBase(candidates)`. Both `getDisjointBase` and `_validateDisjointBaseClass` now call it, so the PEP 800 tie-break rule lives in one place.

packages/pyright-internal/src/analyzer/types.ts
```diff
@@ -1255,6 +1258,43 @@ export namespace ClassType {
...
 
+    export function isDisjointBase(classType: ClassType) {
+        // Dataclass decorators can synthesize a __slots__ definition lazily.
+        classType.shared.synthesizeMethodsDeferred?.();
+
+        return (
+            !!(classType.shared.flags & ClassTypeFlags.DisjointBase) ||
+            ClassType.isBuiltIn(classType, 'object') ||
+            !!classType.shared.localSlotsNames?.length
+        );
```
StellaHuang95, line 1293: > **Issue** · Please address or respond

`isDisjointBase` is exported from `ClassType`, but its only current use is the unqualified same-module call in `getDisjointBase`. Please make it module-local, or add its external consumer in this change, to avoid unnecessarily expanding the public API.

[verified]

rchiodo: Fixed — made isDisjointBase a module-local (non-exported) function within the ClassType namespace. Its only callers are the unqualified same-module calls in getDisjointBase (direct check + MRO ilter), so dropping xport keeps them working without expanding the public API. Confirmed there are no ClassType.isDisjointBase references anywhere in the codebase; typecheck passes.

packages/pyright-internal/src/analyzer/types.ts
```diff
@@ -1255,6 +1258,45 @@ export namespace ClassType {
...
+    }
+
+    export function getDisjointBase(classType: ClassType): ClassType | undefined {
+        if (isDisjointBase(classType)) {
+            return classType;
+        }
+
```
StellaHuang95, line 1300: > **Issue** · Please address or respond

Returning `undefined` for any MRO containing `Any`/`Unknown` also discards known disjoint bases inherited by this class. For example, `class Weird(Left, unresolved_base): ...` followed by `class C(Weird, Right): ...` will not report the incompatible `Left`/`Right` bases. Preserve a most-derived known candidate when one exists, and add a regression case for this transitive shape.

rchiodo: Fixed — `getDisjointBase` no longer bails out to `undefined` when the MRO contains `Any`/`Unknown`. It now filters the known disjoint-base candidates from the MRO and returns the most-derived one via `getMostDerivedDisjointBase`. An unknown class can introduce an *additional* unknown disjoint base but cannot make two already-known disjoint bases compatible, so preserving the most-derived known candidate is sound and no longer discards inherited candidates.

Added a regression case in `disjointBase1.py` for the transitive shape:

```python
class LeftWithUnknown(Left, _unknown_base()):  # ok
    pass

class LeftWithUnknownAndRight(LeftWithUnknown, Right):  # now reported
    pass
```

`Left` is now correctly recovered from `LeftWithUnknown`'s MRO (which contains an unresolved base), so the `Left`/`Right` conflict is reported. Bumped the `DisjointBase1` expected error count 10 → 11.

StellaHuang95 (approved):
Approved via Review Center.

packages/pyright-internal/src/analyzer/checker.ts
```diff
@@ -5659,6 +5661,46 @@ export class Checker extends ParseTreeWalker {
...
+                continue;
+            }
+
+            if (!candidates.some((existingCandidate) => ClassType.isSameGenericClass(existingCandidate, candidate))) {
+                candidates.push(candidate);
+            }
+        }
+
+        if (candidates.length > 1 && !ClassType.getMostDerivedDisjointBase(candidates)) {
+            this._evaluator.addDiagnostic(
```
StellaHuang95, line 5721: > **Info** · Optional note

The diagnostic can include `"object"` alongside genuinely incompatible disjoint bases (for example, `class C(Left, Plain, Right)`). Since `object` is compatible with every disjoint base, including it makes the error misleading; filter it from the reported base names.

StellaHuang95 (approved):
Approved via Review Center.

packages/pyright-internal/src/analyzer/checker.ts
rchiodo: Fixed — `object` is now filtered from the reported incompatible-base names. Since `object` is a disjoint base but compatible with every other disjoint base, including it (e.g. for `class C(Left, Plain, Right)`, where `Plain` resolves to `object`) made the message misleading. The conflict detection itself is unchanged; only the displayed `bases` list now excludes `object` via `ClassType.isBuiltIn(candidate, 'object')`. Added a regression sample `LeftPlainRight(Left, Plain, Right)` and bumped the `DisjointBase1` expected error count 11 -> 12.

packages/pyright-internal/src/analyzer/types.ts
```diff
@@ -1255,6 +1258,43 @@ export namespace ClassType {
...
+        return getMostDerivedDisjointBase(candidates);
+    }
+
+    // Applies the PEP 800 reduction rule to a set of disjoint base candidates:
+    // returns the unique candidate that is a subclass of every other candidate,
+    // or undefined if no such candidate exists.
+    export function getMostDerivedDisjointBase(candidates: ClassType[]): ClassType | undefined {
+        return candidates.find((candidate) =>
+            candidates.every((otherCandidate) => ClassType.isDerivedFrom(candidate, otherCandidate))
```
StellaHuang95: > **Info** · Optional note

`isDerivedFrom` treats an `Any`/`Unknown` base as derived from anything. A directly decorated class with an unknown base can therefore satisfy the reduction rule and suppress a conflict with an unrelated disjoint base, contrary to the stated unknown-base invariant. Please document this interaction or add regression coverage for it.

[verified]

packages/pyright-internal/src/analyzer/types.ts
```diff
@@ -1255,6 +1258,43 @@ export namespace ClassType {
...
+    function isDisjointBase(classType: ClassType) {
+        // The disjoint-base property applies only to nominal classes.
+        if (ClassType.isProtocolClass(classType) || ClassType.isTypedDictClass(classType)) {
+            return false;
+        }
+
+        return (
+            !!(classType.shared.flags & ClassTypeFlags.DisjointBase) ||
+            ClassType.isBuiltIn(classType, 'object') ||
+            !!classType.shared.localSlotsNames?.length
```
StellaHuang95: > **Info** · Optional note

Detection of synthesized `@dataclass(slots=True)` slots relies on `localSlotsNames` having already been populated when the checker runs. Please make that ordering dependency explicit or ensure slot synthesis is triggered here, so a future synthesis-timing change cannot silently stop recognizing slotted dataclasses as disjoint bases.

[verified]

StellaHuang95 (approved):
Approved via Review Center.

StellaHuang95 (approved):
Approved via Review Center.

StellaHuang95:
The supplied diff includes changes already present on main; please rebase so the PR reflects only its actual disjoint-base delta.

StellaHuang95 (approved):
Approved via Review Center.

StellaHuang95:
GitHub cannot anchor PR review comments to unchanged lines in the diff. Falling back to a general PR comment for [packages/pyright-internal/src/analyzer/types.ts:L713](https://github.com/microsoft/pyright/blob/acffba7a0a62fc5bd48d5e7c40a835ecfda62aba/packages/pyright-internal/src/analyzer/types.ts#L713).

> **Info** · Optional note

`ClassDetailsShared.hasNonEmptySlots` is optional while `Scope.hasNonEmptySlots` defaults to `false`; `typeEvaluator.ts` can therefore assign `undefined` and relies on coercion at the current read site. Please make the shared representation consistently non-optional with a `false` default to avoid future direct reads treating the state inconsistently.

[verified]

StellaHuang95:
GitHub cannot anchor PR review comments to unchanged lines in the diff. Falling back to a general PR comment for [packages/pyright-internal/src/analyzer/scope.ts:L272](https://github.com/microsoft/pyright/blob/acffba7a0a62fc5bd48d5e7c40a835ecfda62aba/packages/pyright-internal/src/analyzer/scope.ts#L272).

> **Info** · Optional note

`hasNonEmptySlots` is exposed as a public mutable field but also has a setter, while the sibling `slotsNames` uses direct field access. Please choose one convention (for example, add a getter and encapsulate the field, or remove the setter) so the new state has a single, clear access pattern.

[verified]

StellaHuang95 (approved):
Approved via Review Center.

packages/pyright-internal/src/analyzer/dataClasses.ts
```diff
@@ -89,6 +89,100 @@ import {
...
+        if (synthesizeMethodsDeferred && !classType.shared.synthesizeMethodsDeferred) {
+            classType.shared.synthesizeMethodsDeferred = synthesizeMethodsDeferred;
+        }
+    }
+
+    classType.shared.localSlotsNames = localSlotsNames;
+    classType.shared.hasNonEmptySlots = localSlotsNames.length > 0;
+}
+
 // Validates fields for compatibility with a dataclass and synthesizes
```
StellaHuang95, line 186: > **Warning** · Non-blocking recommendation

`localSlotsNames` and `hasNonEmptySlots` are now computed independently here and in `synthesizeDataClassMethods`, with different exclusion rules. Because both are guarded by `=== undefined`, whichever path runs first determines the result, making slot behavior potentially evaluation-order dependent. Please consolidate the filtering logic or add coverage proving both paths produce identical results for KW_ONLY, ClassVar, InitVar, and inherited-field cases.

[verified]

StellaHuang95 (approved):
Approved via Review Center.

PR #11595, opened 2026-08-06
Fix heterogeneous TypeVarTuple constraint solving

rchiodo:
## Summary

- widen repeated `TypeVarTuple` constraints per position in non-invariant inference
- preserve precise common supertypes when one element type subsumes another, otherwise infer an element-wise union
- preserve invariant generic matching, tuple-length checks, and tuple shape metadata
- cover the upstream `generics_typevartuple_basic.py` line 90 and `generics_typevartuple_args.py` line 76 conformance cases

## Tests

- `cd packages/pyright-internal && npm run build`
- `cd packages/pyright-internal && npx jest typeEvaluator6.test typeEvaluator2.test -t "TypeVarTuple|Solver" --forceExit --runInBand`
- `cd packages/pyright-internal && npx jest typeEvaluator6.test typeEvaluator2.test -t "TypeVarTuple(4|11|32)$|Solver35$" --forceExit --runInBand`

### DIFF
--- a/packages/pyright-internal/src/analyzer/constraintSolver.ts
+++ b/packages/pyright-internal/src/analyzer/constraintSolver.ts
@@ -1413,7 +1413,9 @@ function widenTypeForTypeVarTuple(
 
         if (evaluator.assignType(typeArg1.type, typeArg2.type, undefined, undefined, undefined, recursionCount)) {
             widenedType = typeArg1.type;
-        } else if (evaluator.assignType(typeArg2.type, typeArg1.type, undefined, undefined, undefined, recursionCount)) {
+        } else if (
+            evaluator.assignType(typeArg2.type, typeArg1.type, undefined, undefined, undefined, recursionCount)
+        ) {
             widenedType = typeArg2.type;
         } else {
             widenedType = combineTypes([typeArg1.type, typeArg2.type], {
--- a/packages/pyright-internal/src/analyzer/constraintSolver.ts
+++ b/packages/pyright-internal/src/analyzer/constraintSolver.ts
@@ -849,7 +849,7 @@ function assignUnconstrainedTypeVar(
                         newLowerBound = adjSrcType;
                     }
                 } else if (isTypeVarTuple(destType)) {
-                    const widenedType = widenTypeForTypeVarTuple(evaluator, curLowerBound, adjSrcType);
+                    const widenedType = widenTypeForTypeVarTuple(evaluator, curLowerBound, adjSrcType, isInvariant);
                     if (!widenedType) {
                         diag?.addMessage(
                             LocAddendum.typeAssignmentMismatch().format(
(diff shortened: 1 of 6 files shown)


StellaHuang95:

packages/pyright-internal/src/analyzer/constraintSolver.ts
```diff
@@ -1367,11 +1367,55 @@ function widenTypeForTypeVarTuple(evaluator: TypeEvaluator, type1: Type, type2:
...
+        } else if (evaluator.assignType(typeArg2.type, typeArg1.type)) {
+            widenedType = typeArg2.type;
+        } else {
+            widenedType = combineTypes([typeArg1.type, typeArg2.type], {
+                maxSubtypeCount: maxSubtypeCountForTypeVarLowerBound,
+            });
+        }
+
+        return {
+            type: widenedType,
```
StellaHuang95, line 1427: > **Warning** · Non-blocking recommendation

Could we thread the caller's `recursionCount` through `widenTypeForTypeVarTuple` and into these `assignType` calls? Restarting at the default depth can weaken recursion guarding for recursively structured tuple element types.

packages/pyright-internal/src/analyzer/constraintSolver.ts
```diff
@@ -1344,12 +1344,12 @@ function typeVarOccursIn(typeVar: TypeVarType, type: Type): boolean {
...
-    // the same type" if a TypeVarTuple is used in multiple locations.
-    // This is problematic for a number of reasons, but in the interest
-    // of sticking to the spec, we'll enforce that here.
-
+function widenTypeForTypeVarTuple(
+    evaluator: TypeEvaluator,
+    type1: Type,
+    type2: Type,
+    isInvariant: boolean
+): Type | undefined {
```
StellaHuang95, line 1359: > **Info** · Optional note

Please retain a brief rationale for allowing element-wise widening in non-invariant contexts and for using `isInvariant` as the gate. The removed comment documented the prior spec-driven behavior, so without a replacement this intentional divergence is not apparent to future maintainers.

StellaHuang95:
The equivalent Pylance constraint-solver path still needs the same non-invariant TypeVarTuple widening and regression coverage; otherwise this upstream fix will not reach Pylance users.

StellaHuang95 (approved):
Approved via Review Center.

packages/pyright-internal/src/analyzer/constraintSolver.ts
rchiodo: Fixed — added a rationale comment above the `isInvariant` gate explaining that invariant contexts still honor the spec's "exactly the same type" rule (bail out on divergence), while non-invariant contexts widen element-wise, mirroring how normal TypeVars widen incompatible lower bounds into a union.

rchiodo:
Re: the Pylance constraint-solver path — this PR targets the upstream Pyright repo. Pylance vendors Pyright as a subrepo, so this fix reaches Pylance once it's pulled downstream into the pyrx monorepo. The equivalent Pylance-side widening + regression coverage will be handled there as part of the subrepo update, since it can't be modified from this repo.

StellaHuang95:
The equivalent async Pylance constraint-solver path still needs a tracked follow-up to avoid divergent TypeVarTuple inference between Pyright and Pylance.

StellaHuang95 (approved):
Approved via Review Center.

rchiodo:
Tracked — the downstream async Pylance parity work is now captured in https://github.com/microsoft/pyrx/issues/9082, including equivalent non-invariant TypeVarTuple widening, recursion-depth guarding, regression coverage, and sync/async inference parity.

StellaHuang95 (approved):
Approved via Review Center.

PR #11596, opened 2026-08-06
Fix defaulted subclass type specialization

rchiodo:
## Summary

Generic class values remain unspecialized so they can still be subscripted, while their `typeForm` records the default-specialized type. During `type[...]` assignability, Pyright was building the inheritance chain from that unspecialized class value, so absent source type arguments were treated permissively and a defaulted subclass could be accepted through the wrong specialization.

When every retained class type parameter has an explicit, non-gradual default, specialize the temporary source used for class assignability with `specializeWithDefaultTypeArgs`. The normal variance-aware type-argument comparison then rejects incompatible `type[...]` specializations. Classes with defaultless parameters, or defaults that resolve to top-level `Any`, `Unknown`, a gradual callable, or a gradual unpacked tuple, retain their existing unspecialized inference behavior. Structurally concrete defaults such as `tuple[Any, ...]` remain enforceable.

## Coverage

Add the `generics_defaults_specialization.py` conformance case: bare `Child` is accepted as `type[Child[str]]` and rejected as `type[Child[int]]` when the retained base type parameter defaults to `str`.

Add realistic framework and library coverage for:

- registries and factories storing `type[Subclass[...]]`
- partial specialization with a leading non-defaulted parameter
- multi-level and multiple-inheritance subclass forwarding
- aliases, explicit specialization, runtime class subscription, and constructor inference
- invariant and covariant defaulted parameters
- direct and transitive `Any`, unions containing `Any`, gradual `ParamSpec` and `TypeVarTuple` defaults, and defaultless/`Unknown` inference
- nested structural defaults containing `Any`, including a psycopg-style `Cursor[tuple[Any, ...]]`
- the real-world gradual `ctypes.py_object[Any]` counterexample

## mypy_primer audit

All eight shards completed successfully on `d72de0181`.

- **SymPy:** 38,394 -> 38,220 errors (net -174). The artifact contains 228 removed and 54 added diagnostic headers: 47 same-location text replacements, 181 pure removals, and 7 pure additions. Temporary class-name tracing found `ctypes.py_object` as the only eligible bare defaulted class reached in SymPy; its direct `Any` default is skipped by the final guard, so these sites have no semantic path through the new specialization. Classified as unrelated primer/run diagnostic churn, not a regression.
- **psycopg:** four added assignments for sync/async cursor and server-cursor factories. Their retained default is the structurally concrete `TupleRow = tuple[Any, ...]`; these are expected corrections because a bare `Cursor[TupleRow]` is not valid for every connection `Row`.
- **scikit-learn:** two same-diagnostic ndarray rendering replacements with no count change; benign text churn.
- **All other projects:** no deltas; shards 2-7 were empty.

## Precision regression report

- Failure classification: **B - Pyright limitation**
- Relevant typeshed commit: **N/A**
- Precision regressions: **none**
- Mitigation: concrete defaults are enforced only for exact bare class values whose retained parameters all have explicit, non-gradual defaults. Defaultless and gradual defaults preserve prior inference rather than materializing `Unknown` or `Any`.

## Tests

- All 1,184 type-evaluator tests
- `npx jest typeEvaluator5.test typeEvaluator8.test --runInBand --forceExit -t "TypeVarDefault|TypeForm"` (25 tests)
- `npm run build:cli:dev`
- `npm run typecheck`
- `npm run check:eslint`
- `npm run check:prettier`
- Focused CLI checks for the conformance sample, realistic registry/factory sample, `ctypes.py_object`, mixed/defaultless inference, and nested structural `Any` defaults

### DIFF
--- a/packages/pyright-internal/src/analyzer/typeEvaluator.ts
+++ b/packages/pyright-internal/src/analyzer/typeEvaluator.ts
@@ -24320,9 +24320,19 @@ export function createTypeEvaluator(
             }
         }
 
-        // A class value remains unspecialized so it can be subscripted. Apply its
-        // default type arguments when comparing it against a type specialization.
-        if (TypeBase.isInstantiable(srcType) && srcType.props?.typeForm) {
+        // A class value normally remains unspecialized so it can be subscripted.
+        // If every type parameter has an explicit default, apply those defaults
+        // when comparing it against a type specialization. Classes with one or
+        // more defaultless parameters retain their unspecialized behavior, so
+        // this cannot introduce Unknown.
+        if (
+            TypeBase.isInstantiable(srcType) &&
+            srcType.props?.typeForm &&
+            !srcType.priv.typeArgs &&
+            !srcType.priv.includeSubclasses &&
+            srcType.shared.typeParams.length > 0 &&
+            srcType.shared.typeParams.every((typeParam) => typeParam.shared.isDefaultExplicit)
+        ) {
             srcType = specializeWithDefaultTypeArgs(srcType);
         }
 
--- a/packages/pyright-internal/src/analyzer/typeEvaluator.ts
+++ b/packages/pyright-internal/src/analyzer/typeEvaluator.ts
@@ -24322,9 +24322,9 @@ export function createTypeEvaluator(
 
         // A class value normally remains unspecialized so it can be subscripted.
         // If every type parameter has an explicit default, apply those defaults
-        // when comparing it against a type specialization. Classes with one or
-        // more defaultless parameters retain their unspecialized behavior, so
-        // this cannot introduce Unknown.
+        // when comparing it against a type specialization. Classes with defaultless
+        // parameters or defaults that resolve directly to Any or Unknown retain
+        // their unspecialized behavior, so they don't degrade inference.
         if (
             TypeBase.isInstantiable(srcType) &&
             srcType.props?.typeForm &&
@@ -24333,7 +24333,19 @@ export function createTypeEvaluator(
             srcType.shared.typeParams.length > 0 &&
             srcType.shared.typeParams.every((typeParam) => typeParam.shared.isDefaultExplicit)
         ) {
-            srcType = specializeWithDefaultTypeArgs(srcType);
+            const specializedSrcType = specializeWithDefaultTypeArgs(srcType);
+            const specializedTypeArgs =
+                specializedSrcType.priv.typeArgs ??
+                specializedSrcType.priv.tupleTypeArgs?.map((tupleTypeArg) => tupleTypeArg.type);
+            const hasGradualTypeArg = specializedTypeArgs?.some(
+                (typeArg) =>
+                    !!containsAnyOrUnknown(typeArg, /* recurse */ false) ||
+                    (isFunction(typeArg) && FunctionType.isGradualCallableForm(typeArg)) ||
+                    (isUnpackedClass(typeArg) && !!containsAnyOrUnknown(typeArg, /* recurse */ true))
+            );
+            if (hasGradualTypeArg === false) {
+                srcType = specializedSrcType;
+            }
         }
 
         // Is it a structural type (i.e. a protocol)? If so, we need to
--- a/packages/pyright-internal/src/analyzer/typeEvaluator.ts
+++ b/packages/pyright-internal/src/analyzer/typeEvaluator.ts
@@ -24320,6 +24320,12 @@ export function createTypeEvaluator(
             }
         }
 
+        // A class value remains unspecialized so it can be subscripted. Apply its
+        // default type arguments when comparing it against a type specialization.
+        if (TypeBase.isInstantiable(srcType) && srcType.props?.typeForm) {
+            srcType = specializeWithDefaultTypeArgs(srcType);
+        }
+
         // Is it a structural type (i.e. a protocol)? If so, we need to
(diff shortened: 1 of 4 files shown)


StellaHuang95:

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -24320,6 +24320,12 @@ export function createTypeEvaluator(
             }
         }
 
+        // A class value remains unspecialized so it can be subscripted. Apply its
+        // default type arguments when comparing it against a type specialization.
+        if (TypeBase.isInstantiable(srcType) && srcType.props?.typeForm) {
+            srcType = specializeWithDefaultTypeArgs(srcType);
```
StellaHuang95: > **Issue** · Please address or respond

The matching async evaluator in `packages/pylance-internal/src/analyzer/typeEvaluator.ts` still builds its inheritance chain from an unspecialized `srcType`. Async program mode uses that path, so `Child` can still be accepted as `type[Child[int]]` there. Please mirror this normalization in the async `assignClass` path and cover the scenario with `enableAsyncProgram=true`.

packages/pyright-internal/src/tests/samples/typeVarDefaultClass5.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class Child(Parent[int, DefaultT]):
+    pass
+
+
+valid: type[Child[str]] = Child
+
+# This should generate an error because Child uses str for DefaultT.
+invalid: type[Child[int]] = Child
```
StellaHuang95, line 39: > **Warning** · Non-blocking recommendation

`specializeWithDefaultTypeArgs` also changes retained parameters without defaults to `Unknown`, but this sample covers only the defaulted-parameter case. Add a defaultless retained-TypeVar case that validates the intended behavior so that path cannot regress unnoticed.

rchiodo: Fixed in db658d94c -- added a defaultless retained-TypeVar case (ChildNoDefault(Parent[int, U]) where U has no default). The retained parameter specializes to Unknown, which is assignable to any specialization, so retained_int/retained_str correctly produce no errors. The targeted TypeVarDefaultClass5 test still passes (1 expected error).

packages/pyright-internal/src/analyzer/typeEvaluator.ts
rchiodo: Not applicable to this repo. This PR targets the standalone microsoft/pyright repository, which does not contain packages/pylance-internal or any async program mode (enableAsyncProgram / IAsyncProgram) -- `grep` finds zero references and no such package exists here. There is a single sync assignClass path, and the normalization added at typeEvaluator.ts:24325 covers it. The sync/async parity concern applies to the pyrx monorepo's vendored copy of Pyright; it does not apply to this upstream repo.

StellaHuang95 (approved):
Approved via Review Center.

StellaHuang95 (approved):
Approved via Review Center.

packages/pyright-internal/src/tests/samples/typeVarDefaultClass5.py
```diff
@@ -0,0 +1,165 @@
...
```
StellaHuang95, line 39: > **Info** · Optional note

The comment misstates the mechanism: `U` has no explicit default, so the `every(isDefaultExplicit)` guard skips `specializeWithDefaultTypeArgs` entirely. Please describe this as retaining the unspecialized class behavior rather than specializing `U` to `Unknown`, so the test documents the path it actually exercises.

StellaHuang95 (approved):
Approved via Review Center.

packages/pyright-internal/src/tests/samples/typeVarDefaultClass5.py
rchiodo: Fixed in 3bfb36f9f — the comment now states that the class remains unspecialized when a retained type parameter has no default. The `every(isDefaultExplicit)` guard skips `specializeWithDefaultTypeArgs` for this case.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -24320,6 +24320,34 @@ export function createTypeEvaluator(
             }
```
StellaHuang95, line 24320: > **Issue** · Please address or respond

This changes only the sync evaluator. `packages/pylance-internal/src/analyzer/typeEvaluator.ts` has the corresponding async `assignClass` decision point, which remains unchanged, so defaulted subclass `type[...]` assignability will differ in async Pylance mode. Please mirror this guard there and add async-mode coverage, or document a concrete reason that path is exempt.

rchiodo: Not taking this — this PR is in the standalone `microsoft/pyright` repository, which has exactly one evaluator implementation (`packages/pyright-internal/src/analyzer/typeEvaluator.ts`). There is no `packages/pylance-internal` tree here, and repository-wide searches find no `enableAsyncProgram`, `IAsyncProgram`, `runAsync`, or `runEditModeAsync` symbols. The normalization at `typeEvaluator.ts:24328-24349` therefore covers the only `assignClass` path available in this repository; async Pylance parity must be handled in the separate pyrx repository, not in this PR.

PR #11597, opened 2026-08-06
Support complete enum literal union equivalence

rchiodo:
## Summary

- Treat non-`Flag` enum classes and complete unions of their literal members as contextually equivalent for assignment and `assert_type`
- Support the equivalence in both directions, including invariant `list`/`dict` and enum-containing union positions
- Exclude `Flag`, `IntFlag`, and indirect `Flag` subclasses using full-MRO checks
- Conservatively disable expansion when the statically known enum member set may be incomplete

## Design

Enum identity remains nominal by default. The evaluator expands a bare enum only after the destination shape establishes that an enum-literal comparison is needed. A cached union bit makes this destination guard O(1), so unrelated assignments do not enumerate enum symbol tables.

Invariant comparisons normalize both sides locally, which supports reverse substitution such as `list[Literal[Color.RED, Color.GREEN]]` to `list[Color]` without globally rewriting enum types. Free TypeVar solving runs before expansion, preserving broad enum inference; bound source TypeVars are concretized only within the known literal-union assignment context.

`assert_type` uses the same expansion primitive and recursively normalizes nested generic, tuple, and union shapes only after ordinary type identity fails.

Enum expansion is disabled for full-MRO `Flag` ancestry, custom or unknown enum metaclasses, and class bodies with assignment targets that cannot be represented in the static class symbol table. This avoids treating a known subset as complete for detectable dynamic member creation while keeping aliases, `auto()`, and statically named members with unknown values supported.

## Relation to #11584

This implementation was developed independently from current `main` and does not modify or supersede #11584. It corrects edge cases identified during independent review of that approach:

- full-MRO `Flag`/`IntFlag` exclusion rather than direct-base-only checks
- reverse invariant equivalence
- destination guarding before member enumeration
- conservative handling of detectable incomplete member sets and custom metaclasses
- preservation of TypeVar inference and enum identity
- recursive `assert_type` equivalence for required nested shapes

## Conformance

The exact current upstream `python/typing` `conformance/tests/enums_expansion.py` was run with a rebuilt CLI. Before this change, Pyright reported the intentional `Flag` diagnostic at line 56 plus failures at lines 85-86. After this change, output contains exactly the intentional line 56 diagnostic: `1 error, 0 warnings, 0 informations`.

## Tests

- `cd packages/pyright-internal && npm run build`
- `cd packages/pyright-internal && npx jest typeEvaluator3.test -t Enum --forceExit --runInBand`
- all `typeEvaluator1.test` through `typeEvaluator8.test`: 1183 tests passed
- type-server test bundle and `src/tests/typeServer`: 24 tests passed
- `npm run typecheck`
- `npm run check:eslint`
- `npm run check:prettier`
- `git diff --check`
- rebuilt CLI against the exact upstream enum expansion conformance file

### DIFF
--- a/packages/pyright-internal/src/analyzer/typeEvaluator.ts
+++ b/packages/pyright-internal/src/analyzer/typeEvaluator.ts
@@ -8989,6 +8989,144 @@ export function createTypeEvaluator(
         return typeFormResult;
     }
 
+    function expandEnumTypeForLiteralComparison(typeToExpand: Type, comparisonType: Type): Type {
+        if (!containsTopLevelLiteralEnum(comparisonType)) {
+            return typeToExpand;
+        }
+
+        typeToExpand = makeTopLevelTypeVarsConcrete(typeToExpand);
+        const literalEnumClasses = collectLiteralEnumClasses(comparisonType);
+        return expandEnumTypeForLiteralClasses(typeToExpand, literalEnumClasses);
+    }
+
+    function containsTopLevelLiteralEnum(type: Type): boolean {
+        if (isClassInstance(type) && ClassType.isEnumClass(type) && type.priv.literalValue instanceof EnumLiteral) {
+            return true;
+        }
+
+        if (!isUnion(type)) {
+            return false;
+        }
+
+        return type.priv.includesEnumLiteral;
+    }
+
+    function expandEnumTypeForLiteralClasses(typeToExpand: Type, literalEnumClasses: ClassType[]): Type {
+        if (literalEnumClasses.length === 0) {
+            return typeToExpand;
+        }
+
+        return mapSubtypes(typeToExpand, (subtype) => {
+            if (
+                isClassInstance(subtype) &&
+                ClassType.isEnumClass(subtype) &&
+                subtype.priv.literalValue === undefined &&
+                literalEnumClasses.some((enumClass) => ClassType.isSameGenericClass(enumClass, subtype))
+            ) {
+                const literalTypes = enumerateLiteralsForType(evaluatorInterface, subtype);
+                if (literalTypes && literalTypes.length > 0) {
+                    return combineTypes(literalTypes);
+                }
+            }
+
+            return subtype;
+        });
+    }
+
+    function collectLiteralEnumClasses(
+        type: Type,
+        literalEnumClasses: ClassType[] = [],
+        recursively = false,
+        recursionCount = 0
+    ): ClassType[] {
+        if (recursionCount > maxTypeRecursionCount) {
+            return literalEnumClasses;
+        }
+        recursionCount++;
+
+        doForEachSubtype(type, (subtype) => {
+            if (!isClass(subtype)) {
+                return;
+            }
+
+            if (
+                isClassInstance(subtype) &&
+                ClassType.isEnumClass(subtype) &&
+                subtype.priv.literalValue instanceof EnumLiteral &&
+                !literalEnumClasses.some((enumClass) => ClassType.isSameGenericClass(enumClass, subtype))
+            ) {
+                literalEnumClasses.push(subtype);
+            }
+
+            if (recursively && subtype.priv.typeArgs) {
+                if (subtype.priv.tupleTypeArgs) {
+                    subtype.priv.tupleTypeArgs.forEach((typeArg) =>
+                        collectLiteralEnumClasses(typeArg.type, literalEnumClasses, recursively, recursionCount)
+                    );
+                } else {
+                    subtype.priv.typeArgs.forEach((typeArg) =>
+                        collectLiteralEnumClasses(typeArg, literalEnumClasses, recursively, recursionCount)
+                    );
+                }
+            }
+        });
+
+        return literalEnumClasses;
+    }
+
+    function normalizeEnumTypes(typeToNormalize: Type, literalEnumClasses: ClassType[], recursionCount = 0): Type {
+        if (recursionCount > maxTypeRecursionCount) {
+            return typeToNormalize;
+        }
+        recursionCount++;
+
+        typeToNormalize = expandEnumTypeForLiteralClasses(typeToNormalize, literalEnumClasses);
+
+        if (isUnion(typeToNormalize)) {
+            return mapSubtypes(typeToNormalize, (subtype) =>
+                normalizeEnumTypes(subtype, literalEnumClasses, recursionCount)
+            );
+        }
+
+        if (!isClass(typeToNormalize) || !typeToNormalize.priv.typeArgs) {
+            return typeToNormalize;
+        }
+
+        if (typeToNormalize.priv.tupleTypeArgs) {
+            let typeChanged = false;
+            const tupleTypeArgs = typeToNormalize.priv.tupleTypeArgs.map((typeArg) => {
+                const normalizedType = normalizeEnumTypes(typeArg.type, literalEnumClasses, recursionCount);
+                if (normalizedType !== typeArg.type) {
+                    typeChanged = true;
+                }
+
+                return { ...typeArg, type: normalizedType };
+            });
+
+            return typeChanged
+                ? specializeTupleClass(
+                      typeToNormalize,
+                      tupleTypeArgs,
+                      !!typeToNormalize.priv.isTypeArgExplicit,
+                      !!typeToNormalize.priv.isUnpacked
+                  )
+                : typeToNormalize;
+        }
+
+        let typeChanged = false;
+        const typeArgs = typeToNormalize.priv.typeArgs.map((typeArg) => {
+            const normalizedType = normalizeEnumTypes(typeArg, literalEnumClasses, recursionCount);
+            if (normalizedType !== typeArg) {
+                typeChanged = true;
+            }
+            return normalizedType;
+        });
+
+        return typeChanged
+            ? ClassType.specialize(typeToNormalize, typeArgs, typeToNormalize.priv.isTypeArgExplicit)
+            : typeToNormalize;
+    }
+
     function getTypeOfAssertType(node: CallNode, inferenceContext: InferenceContext | undefined): TypeResult {
         if (
             node.d.args.length !== 2 ||
@@ -9017,13 +9155,22 @@ export function createTypeEvaluator(
         // what mypy does -- and what various library authors expect.
         const arg0Type = stripTypeGuard(arg0TypeResult.type);
 
-        if (
-            !isTypeSame(assertedType, arg0Type, {
-                treatAnySameAsUnknown: true,
-                ignorePseudoGeneric: true,
-                ignoreConditions: true,
-            })
-        ) {
+        const typeSameOptions = {
+            treatAnySameAsUnknown: true,
+            ignorePseudoGeneric: true,
+            ignoreConditions: true,
+        };
+        let typesMatch = isTypeSame(assertedType, arg0Type, typeSameOptions);
+
+        if (!typesMatch) {
+            const literalEnumClasses = collectLiteralEnumClasses(assertedType, [], /* recursively */ true);
+            collectLiteralEnumClasses(arg0Type, literalEnumClasses, /* recursively */ true);
+            const normalizedAssertedType = normalizeEnumTypes(assertedType, literalEnumClasses);
+            const normalizedArg0Type = normalizeEnumTypes(arg0Type, literalEnumClasses);
+            typesMatch = isTypeSame(normalizedAssertedType, normalizedArg0Type, typeSameOptions);
+        }
+
+        if (!typesMatch) {
             const srcDestTypes = printSrcDestTypes(arg0TypeResult.type, assertedType, { expandTypeAlias: true });
 
             addDiagnostic(
@@ -18305,6 +18452,18 @@ export function createTypeEvaluator(
 
             const effectiveMetaclass = computeEffectiveMetaclass(classType, node.d.name);
 
+            if (
+                ClassType.isEnumClass(classType) &&
+                (AnalyzerNodeInfo.getScope(node)?.hasPotentiallyDynamicSymbolTable ||
+                    !isInstantiableClass(effectiveMetaclass) ||
+                    !ClassType.isBuiltIn(effectiveMetaclass, ['EnumMeta', 'EnumType']) ||
+                    classType.shared.mro.some(
+                        (mroClass) => isClass(mroClass) && ClassType.areEnumMembersUnknown(mroClass)
+                    ))
+            ) {
+                classType.shared.flags |= ClassTypeFlags.EnumMembersUnknown;
+            }
+
             // Clear the "partially constructed" flag.
             classType.shared.flags &= ~ClassTypeFlags.PartiallyEvaluated;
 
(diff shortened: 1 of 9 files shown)


StellaHuang95:

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -8989,6 +8989,144 @@ export function createTypeEvaluator(
         return typeFormResult;
     }
 
+    function expandEnumTypeForLiteralComparison(typeToExpand: Type, comparisonType: Type): Type {
+        if (!containsTopLevelLiteralEnum(comparisonType)) {
+            return typeToExpand;
+        }
+
+        typeToExpand = makeTopLevelTypeVarsConcrete(typeToExpand);
```
StellaHuang95: > **Warning** · Non-blocking recommendation

`makeTopLevelTypeVarsConcrete` runs before checking whether the source contains one of the destination's literal-enum classes. For an unrelated source TypeVar, this can replace the TypeVar with its bound even though no enum expansion occurs (for example, while comparing invariant `list[T]` with `list[ColorLiterals]`). Collect the literal classes first and concretize only a matching enum subtype; please add a regression for TypeVar sources in both invariant directions.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -18305,6 +18452,18 @@ export function createTypeEvaluator(
...
 
+            if (
+                ClassType.isEnumClass(classType) &&
+                (AnalyzerNodeInfo.getScope(node)?.hasPotentiallyDynamicSymbolTable ||
+                    !isInstantiableClass(effectiveMetaclass) ||
+                    !ClassType.isBuiltIn(effectiveMetaclass, ['EnumMeta', 'EnumType']) ||
+                    classType.shared.mro.some(
+                        (mroClass) => isClass(mroClass) && ClassType.areEnumMembersUnknown(mroClass)
+                    ))
+            ) {
```
StellaHuang95, line 18566: > **Warning** · Non-blocking recommendation

Could literal expansion reach `enumerateLiteralsForType` while this enum is still partially evaluated? `EnumMembersUnknown` is set only during class completion, so a dynamic-body or custom-metaclass enum could otherwise be expanded from an incomplete member table before this guard runs. Please make the pre-completion state conservative or add regression coverage demonstrating expansion cannot occur before completion.

StellaHuang95:
Separately, the downstream Pylance async evaluator still has the direct-base-only `Flag` exclusion and lacks this expansion/completeness behavior. Track the parity port when this Pyright change is imported so indirect `Flag` subclasses remain excluded.

StellaHuang95 (approved):
Approved via Review Center.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
rchiodo: Fixed in b5320e130. `expandEnumTypeForLiteralComparison` now collects the comparison's literal-enum classes first and bails early when there are none. It concretizes top-level TypeVars only to *probe* for a matching enum subtype, and if `expandEnumTypeForLiteralClasses` returns the same reference (i.e. no enum expansion actually occurred), it returns the original `typeToExpand` — so an unrelated source TypeVar is no longer replaced by its bound. Added regression coverage in `enum15.py` (`test_type_var_invariant_source`) for TypeVar sources in both invariant directions: a `T: Color`-bound source still expands and matches `list[ColorLiterals]`, while an unrelated `U: int` source stays intact and the invariant comparison correctly fails in both directions.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
rchiodo: This case is already handled conservatively. For a literal `EnumLiteral` of class `C` to appear in a comparison type at all, a member of `C` must have been evaluated, which only happens after the class type exists and this completion block (which sets `EnumMembersUnknown`) has run. For a standard, statically-bodied enum the member symbol table is populated by the binder before type evaluation, so `enumerateLiteralsForType` reads a complete table even mid-evaluation and the flag is correctly left unset. The problematic cases the comment calls out — dynamic-body and custom-metaclass enums — are exactly what this new block flags via `hasPotentiallyDynamicSymbolTable`, a non-`EnumMeta`/`EnumType` metaclass, or an MRO member with unknown members, so `enumerateLiteralsForType` returns `undefined` and no expansion occurs.

Regression coverage already demonstrates this: `enum15.py` has `DynamicBody` (uses `locals()[...] = 2`) and `CustomMetaEnum` (custom `EnumType` subclass), and both assert that assigning them as a complete literal union produces an error (they are *not* expanded). Those passing tests are empirical proof that by the time expansion is attempted the enum is completed and, for the dynamic/custom cases, conservatively marked as having unknown members. No additional change needed.

StellaHuang95:
Please mirror the enum expansion and full-MRO `Flag`/unknown-member exclusion in Pylance's async evaluator before merging; the current change leaves the shipped async path without this behavior.

rchiodo:
Thanks — agreed the Pylance async evaluator needs the same enum-literal expansion plus the full-MRO `Flag`/unknown-member exclusion. That change belongs in the downstream `pyrx` monorepo (`packages/pylance-internal`), which is a separate repository and not part of this `microsoft/pyright` checkout, so it can't be made in this PR.

This Pyright PR is complete on its own: the expansion + full-MRO/unknown-member exclusion live in `typeEvaluator.ts` and are covered by `enum15.py` (including the `Flag`, `DynamicBody`, and `CustomMetaEnum` cases). The Pylance async-evaluator parity port will be handled downstream when this change is imported into `pyrx`; I'll track it there so indirect `Flag` subclasses stay excluded and the async path gets the expansion/completeness behavior. It is a downstream follow-up rather than a blocker on the Pyright side.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -8989,6 +8989,156 @@ export function createTypeEvaluator(
...
+                return;
+            }
+
+            if (
+                isClassInstance(subtype) &&
+                ClassType.isEnumClass(subtype) &&
+                subtype.priv.literalValue instanceof EnumLiteral &&
+                !literalEnumClasses.some((enumClass) => ClassType.isSameGenericClass(enumClass, subtype))
+            ) {
+                literalEnumClasses.push(subtype);
```
StellaHuang95, line 9102: > **Warning** · Non-blocking recommendation

`assignType` and `assert_type` currently implement enum equivalence through different traversal strategies. Please document that these paths must preserve the same equivalence relation, or consolidate them, so nested generic and tuple behavior cannot silently drift.

StellaHuang95 (approved):
Approved via Review Center.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
rchiodo: Fixed in 4ff0bf2d5 — documented that `assignType` applies enum equivalence during recursive variance/constraint traversal while `assert_type` performs recursive structural normalization, and that the two paths must remain aligned. I also closed the drift gaps found while validating this: `assert_type` now normalizes nested generic, tuple, callable, and overloaded signatures; invariant comparisons preserve bound TypeVars to avoid alias widening; and dynamic enum namespaces are conservatively detected through direct, qualified, imported, and simple assignment aliases of `locals`/`vars`/`exec` without false positives for shadowed names. `enum15.py` covers all of these cases.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -8989,6 +8989,229 @@ export function createTypeEvaluator(
...
+        if (recursively && isFunction(type)) {
+            type.shared.parameters.forEach((_, index) =>
+                collectLiteralEnumClasses(
+                    FunctionType.getParamType(type, index),
+                    literalEnumClasses,
+                    recursively,
+                    recursionCount
+                )
+            );
+
```
StellaHuang95, line 9075: > **Warning** · Non-blocking recommendation

`assignType` and `assert_type` now implement the enum-equivalence relation through separate shape traversals. They agree today, but a future container or callable-shape change could make assignment support a case that `assert_type` silently misses. Please share the shape-descent logic or add focused parity coverage that exercises the same nested shapes through both entry points.

[verified]

packages/pyright-internal/src/analyzer/typeGuards.ts
```diff
@@ -2772,7 +2772,7 @@ export function enumerateLiteralsForType(evaluator: TypeEvaluator, type: ClassTy
     if (ClassType.isEnumClass(type)) {
         // Enum expansion doesn't apply to enum classes that derive
         // from enum.Flag.
-        if (type.shared.baseClasses.some((baseClass) => isClass(baseClass) && ClassType.isBuiltIn(baseClass, 'Flag'))) {
+        if (type.shared.mro.some((mroClass) => isClass(mroClass) && ClassType.isBuiltIn(mroClass, 'Flag'))) {
```
StellaHuang95: > **Warning** · Non-blocking recommendation

This shared helper also affects pattern matching, narrowing, and exhaustiveness. Please add an indirect-`Flag` match/exhaustiveness scenario, not only the existing `is`-narrowing coverage, to lock in the intended behavior for all consumers of this change.

[verified]

packages/pyright-internal/src/analyzer/types.ts
```diff
@@ -2602,6 +2610,7 @@ export interface UnionDetailsPriv {
     literalClasses: LiteralTypes;
     typeAliasSources: Set<UnionType> | undefined;
     includesRecursiveTypeAlias: boolean;
+    includesEnumLiteral: boolean;
```
StellaHuang95, line 2622: > **Info** · Optional note

`includesEnumLiteral` is correct only while every union-construction path routes through `UnionType.addType`. Please document that invariant here so a future union-construction refactor does not silently skip valid enum-equivalence expansion.

[verified]

StellaHuang95 (approved):
Approved via Review Center.

packages/pyright-internal/src/analyzer/types.ts
rchiodo: Fixed — documented that the `includesEnumLiteral` cache relies on every union-construction path adding subtypes through `UnionType.addType`.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -18305,6 +18540,23 @@ export function createTypeEvaluator(
...
+                    !isInstantiableClass(effectiveMetaclass) ||
+                    !ClassType.isBuiltIn(effectiveMetaclass, ['EnumMeta', 'EnumType']) ||
+                    classType.shared.mro.some(
+                        (mroClass) =>
+                            isClass(mroClass) &&
+                            (ClassType.isEnumMemberSetMayBeIncomplete(mroClass) ||
+                                (!ClassType.isBuiltIn(mroClass, 'Enum') &&
+                                    ClassType.getSymbolTable(mroClass).has('_missing_')))
+                    ))
+            ) {
```
StellaHuang95, line 18566: > **Warning** · Non-blocking recommendation

Marking every decorated enum as potentially incomplete disables literal-union equivalence for member-preserving standard decorators such as `@unique`. Consider recognizing known member-preserving decorators, or add a regression test that documents this intentional conservative limitation.

StellaHuang95 (approved):
Approved via Review Center.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
rchiodo: Fixed in bfce8c670 — added focused `@unique` regression coverage documenting the intentional conservative limitation. Even member-preserving decorators remain excluded from complete literal-union equivalence because the general class-decorator path can add or remove enum members at runtime; the test now locks in that both assignment and `assert_type` reject treating the decorated enum as a complete literal union.

StellaHuang95 (approved):
Approved via Review Center.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -8989,6 +8989,229 @@ export function createTypeEvaluator(
...
+            OverloadedType.getOverloads(type).forEach((overload) =>
+                collectLiteralEnumClasses(overload, literalEnumClasses, recursively, recursionCount)
+            );
+
+            const implementation = OverloadedType.getImplementation(type);
+            if (implementation) {
+                collectLiteralEnumClasses(implementation, literalEnumClasses, recursively, recursionCount);
+            }
+        }
+
```
StellaHuang95, line 9090: > **Warning** · Non-blocking recommendation

`collectLiteralEnumClasses` and `normalizeEnumTypes` must remain aligned with `assignType`'s recursive shape handling. Add a parity regression using a deeper mixed shape (for example, `dict[Color, list[Callable[[Color], Color]]]`) through both assignment and `assert_type` so future walker drift is detected.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -8989,6 +8989,229 @@ export function createTypeEvaluator(
...
+        const literalEnumClasses = collectLiteralEnumClasses(comparisonType);
+        if (literalEnumClasses.length === 0) {
+            return typeToExpand;
+        }
+
+        // Concretize top-level TypeVars only to probe for a matching enum subtype.
+        // If no enum expansion actually occurs (for example, an unrelated source
+        // TypeVar whose bound is not one of the comparison's enum classes), preserve
```
StellaHuang95, line 9011: > **Warning** · Non-blocking recommendation

For a source such as `T | Color`, concretizing the whole source also concretizes the unrelated `T`; once `Color` expands, the returned union retains that concretized sibling. Preserve unrelated union members' TypeVars, or add a focused regression test that locks down the intended inference behavior.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -18304,6 +18539,33 @@ export function createTypeEvaluator(
...
+                    ));
+
+            if (enumMemberSetMayBeDynamicallyModified) {
+                classType.shared.flags |= ClassTypeFlags.EnumMemberSetMayBeDynamicallyModified;
+            }
+
+            if (
+                ClassType.isEnumClass(classType) &&
+                (node.d.decorators.length > 0 ||
+                    enumMemberSetMayBeDynamicallyModified ||
```
StellaHuang95, line 18556: > **Warning** · Non-blocking recommendation

This `_missing_` check depends on the current typeshed layout: if a future stub declares `_missing_` on `IntEnum` or `StrEnum`, complete-union equivalence will silently become unavailable for those enums. Please scope the check to user-defined overrides, or document and test the typeshed assumption explicitly.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -25184,6 +25446,12 @@ export function createTypeEvaluator(
             return true;
         }
 
+        const isInvariant = (flags & AssignTypeFlags.Invariant) !== 0;
+        srcType = expandEnumTypeForLiteralComparison(srcType, destType, /* concretizeTypeVars */ !isInvariant);
+        if (isInvariant) {
+            destType = expandEnumTypeForLiteralComparison(destType, srcType, /* concretizeTypeVars */ false);
```
StellaHuang95, line 25452: > **Warning** · Non-blocking recommendation

This normalization applies to every `assignType` caller, including overload overlap and redundancy analysis. Add a regression test that documents the intended diagnostics for overloads distinguishing a nominal enum from its complete literal union, so this broad semantic change does not cause an accidental future regression.

StellaHuang95 (approved):
Approved via Review Center.

PR #11599, opened 2026-08-06
Complete TypeForm conformance

rchiodo:
## Summary

Merged #11412 safely enabled PEP 747 `TypeForm` support by default. This follow-up completes the remaining behavior required by the current typing specification on current `main`, without reusing the stale #11302 branch.

- Isolate TypeForm-context evaluation in an expected-type-keyed cache so it cannot overwrite ordinary runtime-expression results.
- Preserve contextual results for evaluator and language-service clients, including explicit `TypeForm(...)` calls.
- Treat bare `TypeForm` as `TypeForm[Any]` and retain TypeForm metadata through nested `Self` and experimental inline `TypedDict` expressions.
- Reject invalid root and nested special forms, empty `TypeForm[()]`, annotation-only qualifiers, `InitVar`, bare `ParamSpec`, and bare `TypeVarTuple` forms.
- Preserve valid nested `Callable`, `Concatenate`, unpack, forward-reference, and mixed-union behavior.
- Recognize subscripted runtime builtins such as `list[int]` as `types.GenericAlias` conservatively, without accepting typing special forms.

## String parsing and recursion safety

String literals are interpreted as quoted type expressions only when the active context specifically wants a TypeForm. Ordinary string-list evaluation remains unchanged. In particular, NamedTuple field-name strings are not speculatively evaluated as types, preserving the existing `Call5` regression for the recursion/cycle reported against #11302.

## Conformance

Using the current upstream `python/typing` `conformance/tests/typeforms_typeform.py` with the locally built CLI configured for Python 3.12 produces exactly 18 errors and no other diagnostics at:

`23, 24, 59, 67, 68, 69, 70, 71, 72, 73, 74, 75, 86, 86, 88, 88, 98, 108`

The current-main baseline produced an unwanted diagnostic at line 49 and missed required diagnostics at lines 70-72 and 74.

## Validation

- All eight type evaluator suites: 1,184 tests passed
- Focused TypeForm, NamedTuple `Call5`, Sentinel, TypeVarTuple, and inline TypedDict regressions
- `packages/pyright-internal`: `npm run build`
- `npm run check:eslint`
- `npm run check:prettier`
- `npm run build:cli:dev`
- Exact upstream conformance diagnostic-multiset check with `--pythonversion 3.12`

### DIFF
--- a/packages/pyright-internal/src/analyzer/typeEvaluator.ts
+++ b/packages/pyright-internal/src/analyzer/typeEvaluator.ts
@@ -760,7 +760,7 @@ export function createTypeEvaluator(
             ?.find((entry) => entryMatchesExpectedType(entry, expectedType));
     }
 
-    function readTypeCacheEntryForNode(node: ParseNode) {
+    function readContextualTypeCacheEntryForNode(node: ParseNode) {
         const expectedType = expectedTypeCache.get(node.id)?.type;
         if (expectedType && expectedTypeWantsTypeForm(expectedType)) {
             return (
@@ -770,7 +770,7 @@ export function createTypeEvaluator(
             );
         }
 
-        return readTypeFormTypeCacheEntry(node, /* expectedType */ undefined) ?? readTypeCacheEntry(node);
+        return readTypeCacheEntry(node) ?? readTypeFormTypeCacheEntry(node, /* expectedType */ undefined);
     }
 
     // Bumps the incomplete generation count using the same rules for both the
@@ -994,7 +994,7 @@ export function createTypeEvaluator(
     function getType(node: ExpressionNode): Type | undefined {
         initializePrefetchedTypes(node);
 
-        let type = evaluateTypeForSubnode(node, () => {
+        let type = evaluateContextualTypeForSubnode(node, () => {
             evaluateTypesForExpressionInContext(node);
         })?.type;
 
@@ -1046,13 +1046,13 @@ export function createTypeEvaluator(
     }
 
     function getTypeResult(node: ExpressionNode): TypeResult | undefined {
-        return evaluateTypeForSubnode(node, () => {
+        return evaluateContextualTypeForSubnode(node, () => {
             evaluateTypesForExpressionInContext(node);
         });
     }
 
     function getTypeResultForDecorator(node: DecoratorNode): TypeResult | undefined {
-        return evaluateTypeForSubnode(node, () => {
+        return evaluateContextualTypeForSubnode(node, () => {
             evaluateTypesForExpressionInContext(node.d.expr);
         });
     }
@@ -1061,7 +1061,7 @@ export function createTypeEvaluator(
     function getCachedType(node: ExpressionNode | DecoratorNode): Type | undefined {
         // Prefer the ordinary runtime type when both caches contain an entry for this node.
         // Fall back to the contextual cache so TypeForm-only evaluations remain discoverable.
-        const cacheEntry = readTypeCacheEntry(node) ?? readTypeCacheEntryForNode(node);
+        const cacheEntry = readTypeCacheEntry(node) ?? readContextualTypeCacheEntryForNode(node);
         if (!cacheEntry || cacheEntry.typeResult.isIncomplete) {
             return undefined;
         }
@@ -21474,9 +21474,21 @@ export function createTypeEvaluator(
     // within that tree. If the type cannot be determined (because it's part
     // of a cyclical dependency), the function returns undefined.
     function evaluateTypeForSubnode(subnode: ParseNode, callback: () => void): TypeResult | undefined {
+        return evaluateTypeForSubnodeWithCache(subnode, callback, readTypeCacheEntry);
+    }
+
+    function evaluateContextualTypeForSubnode(subnode: ParseNode, callback: () => void): TypeResult | undefined {
+        return evaluateTypeForSubnodeWithCache(subnode, callback, readContextualTypeCacheEntryForNode);
+    }
+
+    function evaluateTypeForSubnodeWithCache(
+        subnode: ParseNode,
+        callback: () => void,
+        readCacheEntry: (node: ParseNode) => TypeCacheEntry | undefined
+    ): TypeResult | undefined {
         // If the type cache is already populated with a complete type,
         // don't bother doing additional work.
-        let cacheEntry = readTypeCacheEntryForNode(subnode);
+        let cacheEntry = readCacheEntry(subnode);
         if (cacheEntry && !cacheEntry.typeResult.isIncomplete) {
             const typeResult = cacheEntry.typeResult;
 
@@ -21494,7 +21506,7 @@ export function createTypeEvaluator(
         }
 
         callback();
-        cacheEntry = readTypeCacheEntryForNode(subnode);
+        cacheEntry = readCacheEntry(subnode);
         if (cacheEntry) {
             return cacheEntry.typeResult;
         }
--- a/packages/pyright-internal/src/analyzer/typeEvaluator.ts
+++ b/packages/pyright-internal/src/analyzer/typeEvaluator.ts
@@ -745,14 +745,17 @@ export function createTypeEvaluator(
         return typeFormTypeCache;
     }
 
+    // Determines whether a TypeForm cache entry corresponds to the given expected type.
+    // This is the single source of truth for TypeForm-entry identity; both the lookup
+    // and eviction paths derive from it so they cannot drift.
+    function entryMatchesExpectedType(entry: TypeFormTypeCacheEntry, expectedType: Type | undefined) {
+        return expectedType ? !!entry.expectedType && isTypeSame(expectedType, entry.expectedType) : !entry.expectedType;
+    }
+
     function readTypeFormTypeCacheEntry(node: ParseNode, expectedType: Type | undefined) {
         return getTypeFormTypeCache(node)
             .get(node.id)
-            ?.find((entry) =>
-                expectedType
-                    ? !!entry.expectedType && isTypeSame(expectedType, entry.expectedType)
-                    : !entry.expectedType
-            );
+            ?.find((entry) => entryMatchesExpectedType(entry, expectedType));
     }
 
     function readTypeCacheEntryForNode(node: ParseNode) {
@@ -836,11 +839,7 @@ export function createTypeEvaluator(
             const expectedType = inferenceContext?.expectedType;
             const typeFormCache = getTypeFormTypeCache(node);
             let cacheEntries = typeFormCache.get(node.id) ?? [];
-            const oldEntry = cacheEntries.find((entry) =>
-                expectedType
-                    ? !!entry.expectedType && isTypeSame(expectedType, entry.expectedType)
-                    : !entry.expectedType
-            );
+            const oldEntry = cacheEntries.find((entry) => entryMatchesExpectedType(entry, expectedType));
 
             updateIncompleteGenerationCount(typeResult, oldEntry?.typeResult);
 
@@ -848,11 +847,7 @@ export function createTypeEvaluator(
                 return;
             }
 
-            cacheEntries = cacheEntries.filter((entry) =>
-                expectedType
-                    ? !entry.expectedType || !isTypeSame(expectedType, entry.expectedType)
-                    : !!entry.expectedType
-            );
+            cacheEntries = cacheEntries.filter((entry) => !entryMatchesExpectedType(entry, expectedType));
             cacheEntries.push({ typeResult, flags, incompleteGenCount, expectedType });
 
             const maxCacheEntriesPerNode = 8;
--- a/packages/pyright-internal/src/analyzer/typeEvaluator.ts
+++ b/packages/pyright-internal/src/analyzer/typeEvaluator.ts
@@ -1057,11 +1057,9 @@ export function createTypeEvaluator(
 
     // Reads the type of the node from the cache.
     function getCachedType(node: ExpressionNode | DecoratorNode): Type | undefined {
-        // Use readTypeCacheEntryForNode so TypeForm-cached entries are found, but
-        // preserve the incomplete-result safeguard from readTypeCache (don't expose
-        // partially-evaluated types). The evaluator-flag verification isn't applied
-        // here because TypeForm entries are cached with different flags than EvalFlags.None.
-        const cacheEntry = readTypeCacheEntryForNode(node);
+        // Prefer the ordinary runtime type when both caches contain an entry for this node.
+        // Fall back to the contextual cache so TypeForm-only evaluations remain discoverable.
+        const cacheEntry = readTypeCacheEntry(node) ?? readTypeCacheEntryForNode(node);
         if (!cacheEntry || cacheEntry.typeResult.isIncomplete) {
             return undefined;
         }
--- a/packages/pyright-internal/src/analyzer/typeEvaluator.ts
+++ b/packages/pyright-internal/src/analyzer/typeEvaluator.ts
@@ -760,6 +760,9 @@ export function createTypeEvaluator(
             ?.find((entry) => entryMatchesExpectedType(entry, expectedType));
     }
 
+    // Contextual reads consult expectedTypeCache and prefer a matching TypeForm result.
+    // Runtime-only consumers must use readTypeCacheEntry so this precedence is not
+    // accidentally applied where an ordinary runtime type is required.
     function readContextualTypeCacheEntryForNode(node: ParseNode) {
         const expectedType = expectedTypeCache.get(node.id)?.type;
         if (expectedType && expectedTypeWantsTypeForm(expectedType)) {
@@ -21731,7 +21734,7 @@ export function createTypeEvaluator(
                 }
 
                 case 'TypeAlias': {
-                    if ((flags & (EvalFlags.TypeFormArg | EvalFlags.NoConvertSpecialForm)) !== 0) {
+                    if ((flags & EvalFlags.TypeFormArg) !== 0) {
                         addDiagnostic(
                             DiagnosticRule.reportInvalidTypeForm,
                             LocMessage.typeAnnotationVariable(),
--- a/packages/pyright-internal/src/analyzer/typeEvaluator.ts
+++ b/packages/pyright-internal/src/analyzer/typeEvaluator.ts
@@ -625,6 +625,10 @@ interface TypeCacheEntry {
     flags: EvalFlags | undefined;
 }
 
+interface TypeFormTypeCacheEntry extends TypeCacheEntry {
+    expectedType: Type | undefined;
+}
+
 interface CodeFlowAnalyzerCacheEntry {
     typeAtStart: TypeResult | undefined;
     codeFlowAnalyzer: CodeFlowAnalyzer;
@@ -659,6 +663,7 @@ export function createTypeEvaluator(
     let functionRecursionMap = new Map<number, FunctionRecursionInfo[]>();
     let codeFlowAnalyzerCache = new Map<number, CodeFlowAnalyzerCacheEntry[]>();
     let typeCache = new Map<number, TypeCacheEntry>();
+    let typeFormTypeCache = new Map<number, TypeFormTypeCacheEntry[]>();
     let effectiveTypeCache = new Map<number, Map<string, EffectiveTypeResult>>();
     let expectedTypeCache = new Map<number, ExpectedTypeCacheEntry>();
     let asymmetricAccessorAssignmentCache = new Set<number>();
@@ -668,6 +673,7 @@ export function createTypeEvaluator(
     let incompleteGenCount = 0;
     const returnTypeInferenceContextStack: ReturnTypeInferenceContext[] = [];
     let returnTypeInferenceTypeCache: Map<number, TypeCacheEntry> | undefined;
+    let returnTypeInferenceTypeFormTypeCache: Map<number, TypeFormTypeCacheEntry[]> | undefined;
     const signatureTrackerStack: SignatureTrackerStackEntry[] = [];
     let prefetched: Partial<PrefetchedTypes> | undefined;
 
@@ -715,6 +721,7 @@ export function createTypeEvaluator(
         functionRecursionMap = new Map<number, FunctionRecursionInfo[]>();
         codeFlowAnalyzerCache = new Map<number, CodeFlowAnalyzerCacheEntry[]>();
         typeCache = new Map<number, TypeCacheEntry>();
+        typeFormTypeCache = new Map<number, TypeFormTypeCacheEntry[]>();
         effectiveTypeCache = new Map<number, Map<string, EffectiveTypeResult>>();
         expectedTypeCache = new Map<number, ExpectedTypeCacheEntry>();
         asymmetricAccessorAssignmentCache = new Set<number>();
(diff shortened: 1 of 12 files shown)


StellaHuang95:

StellaHuang95:
`getCachedType` now bypasses required cache-read safeguards and can expose incomplete or flag-incompatible results. The TypeForm changes also lack the required Pylance async-path implementation and product-harness coverage.

StellaHuang95 (approved):
Approved via Review Center.

rchiodo:
Thanks for the review.

**getCachedType safeguards** — Partially addressed in 2bbb136. Restored the incomplete-result guard so the cache-only type dump no longer exposes partially-evaluated types (matching the old `readTypeCache` behavior), while still routing through `readTypeCacheEntryForNode` so TypeForm-cached entries are found. The evaluator-flag verification is intentionally not reintroduced on this path: TypeForm entries are cached with `TypeFormArg` flags, so verifying against `EvalFlags.None` (as the old call did) would spuriously fail. Note this API is only consumed by `languageInfoUtils.ts` in `cacheOnly` debug-dump mode.

**Pylance async-path / product-harness coverage** — Not applicable here. This PR targets upstream `microsoft/pyright` (packages: pyright, pyright-internal, pyright-typeserver, vscode-pyright); there is no `pylance-internal` package or Pylance async program model in this repo, so there is no async path to mirror. Coverage follows Pyright's sample + `validateResults` convention, which the PR extends (typeForm2/3/4/6/8, typedDictInline1, and typeEvaluator8.test.ts). All TypeForm tests pass.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -730,8 +737,39 @@ export function createTypeEvaluator(
...
+                    ? !!entry.expectedType && isTypeSame(expectedType, entry.expectedType)
+                    : !entry.expectedType
+            );
+    }
+
+    function readTypeCacheEntryForNode(node: ParseNode) {
+        const expectedType = expectedTypeCache.get(node.id)?.type;
+        if (expectedType && expectedTypeWantsTypeForm(expectedType)) {
+            return (
+                readTypeFormTypeCacheEntry(node, expectedType) ??
```
StellaHuang95, line 765: > **Warning** · Non-blocking recommendation

This sync evaluator change has no corresponding update in Pylance's async evaluator, which uses snapshot-owned caches and different incomplete-generation handling. Please mirror the TypeForm cache behavior in the async path and add Pylance-harness coverage so `enableAsyncProgram` does not retain different TypeForm semantics.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -777,6 +815,52 @@ export function createTypeEvaluator(
...
+            }
+
+            cacheEntries = cacheEntries.filter((entry) =>
+                expectedType
+                    ? !entry.expectedType || !isTypeSame(expectedType, entry.expectedType)
+                    : !!entry.expectedType
+            );
+            cacheEntries.push({ typeResult, flags, incompleteGenCount, expectedType });
+
+            const maxCacheEntriesPerNode = 8;
```
StellaHuang95: > **Warning** · Non-blocking recommendation

The new TypeForm cache branch duplicates the subtle incomplete-generation bookkeeping from the regular cache path. Please factor the shared generation-count logic into a helper where possible, so future changes cannot cause the two cache invalidation paths to drift.

StellaHuang95 (approved):
Approved via Review Center.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
rchiodo: Not applicable in this repo. This PR targets upstream `microsoft/pyright` (packages: pyright, pyright-internal, pyright-typeserver, vscode-pyright). There is no `pylance-internal` package or Pylance async program model (`IAsyncProgram`/snapshot-owned caches) here, so there is no separate async evaluator path to mirror. The TypeForm cache behavior lives entirely in `typeEvaluator.ts` and is exercised by Pyright's sample + `validateResults` convention that this PR extends (typeForm2/3/4/6/8, typedDictInline1 in typeEvaluator8.test.ts). Any Pylance async-path mirroring would happen downstream in the pyrx monorepo, not in this PR.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
rchiodo: Fixed in abe7abae9. Extracted the shared incomplete-generation bookkeeping into a single `updateIncompleteGenerationCount(typeResult, oldTypeResult)` helper that both the regular type cache path and the TypeForm cache path now call, so the two invalidation paths can no longer drift. As a side benefit this aligns the TypeForm path's incomplete-result generation bump with the regular path (previously the TypeForm path returned early in speculative mode before the incomplete-diff bump). Verified with the full typeEvaluator8 suite (162 passed).

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -730,8 +737,52 @@ export function createTypeEvaluator(
...
+        if (returnTypeInferenceTypeFormTypeCache && isNodeInReturnTypeInferenceContext(node)) {
+            return returnTypeInferenceTypeFormTypeCache;
+        }
+
+        return typeFormTypeCache;
+    }
+
+    function readTypeFormTypeCacheEntry(node: ParseNode, expectedType: Type | undefined) {
+        return getTypeFormTypeCache(node)
+            .get(node.id)
```
StellaHuang95, line 754: > **Warning** · Non-blocking recommendation

📍 packages/pyright-internal/src/analyzer/typeEvaluator.ts:741
The expected-type match predicate is hand-written three times: `readTypeFormTypeCacheEntry` (`.find`), the `writeTypeCache` `oldEntry` lookup (`.find`), and the `writeTypeCache` eviction (`.filter`, logically inverted). The L858 fix consolidated the generation-count logic but left this matcher triplicated with one negated copy — a classic desync foot-gun where a future change to TypeForm-entry identity must edit three sites and get the inversion right. Extract `entryMatchesExpectedType(entry, expectedType)` and derive the eviction filter from its negation.

[verified]

StellaHuang95 (approved):
Approved via Review Center.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
rchiodo: Fixed in eaf3ce449. Extracted `entryMatchesExpectedType(entry, expectedType)` as the single source of truth for TypeForm-entry identity. `readTypeFormTypeCacheEntry` and the `writeTypeCache` `oldEntry` lookup now call it directly, and the eviction `.filter` derives from its negation (`!entryMatchesExpectedType(...)`), so the matcher and its inverse can no longer drift across the three sites. Verified with typeEvaluator8 (162 passed).

StellaHuang95 (approved):
Approved via Review Center.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -730,8 +737,57 @@ export function createTypeEvaluator(
...
+            ? !!entry.expectedType && isTypeSame(expectedType, entry.expectedType)
+            : !entry.expectedType;
+    }
+
+    function readTypeFormTypeCacheEntry(node: ParseNode, expectedType: Type | undefined) {
+        return getTypeFormTypeCache(node)
+            .get(node.id)
+            ?.find((entry) => entryMatchesExpectedType(entry, expectedType));
+    }
+
```
StellaHuang95, line 757: > **Warning** · Non-blocking recommendation

Could we add a regression that evaluates the same node first in a `TypeForm` context and then in an ordinary context, asserting that `getCachedType` preserves the ordinary runtime type? The undefined-expected-type fallback is intentionally used for TypeForm consumers, but this test would protect against it being returned to generic cache consumers after a dual-context evaluation.

StellaHuang95 (approved):
Approved via Review Center.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
rchiodo: Added a regression that evaluates the node first through a `TypeForm` assignment, then ordinarily, and asserts `getCachedType` returns the runtime `int` type. `getCachedType` now prefers an existing ordinary cache entry while retaining the contextual fallback for TypeForm-only evaluations.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -730,8 +737,57 @@ export function createTypeEvaluator(
...
+    // Determines whether a TypeForm cache entry corresponds to the given expected type.
+    // This is the single source of truth for TypeForm-entry identity; both the lookup
+    // and eviction paths derive from it so they cannot drift.
+    function entryMatchesExpectedType(entry: TypeFormTypeCacheEntry, expectedType: Type | undefined) {
+        return expectedType
+            ? !!entry.expectedType && isTypeSame(expectedType, entry.expectedType)
+            : !entry.expectedType;
+    }
+
```
StellaHuang95: > **Warning** · Non-blocking recommendation

Could treating a TypeForm-only result as cached here cause guard sites to skip the ordinary runtime evaluation? `readTypeCacheEntryForNode` can return a contextual TypeForm entry based on `expectedTypeCache`; please add a regression that evaluates a node first in a TypeForm context and then ordinarily, or restrict cache-presence checks to the ordinary cache.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -979,7 +1057,14 @@ export function createTypeEvaluator(
 
     // Reads the type of the node from the cache.
     function getCachedType(node: ExpressionNode | DecoratorNode): Type | undefined {
-        return readTypeCache(node, EvalFlags.None);
+        // Prefer the ordinary runtime type when both caches contain an entry for this node.
+        // Fall back to the contextual cache so TypeForm-only evaluations remain discoverable.
+        const cacheEntry = readTypeCacheEntry(node) ?? readTypeCacheEntryForNode(node);
+        if (!cacheEntry || cacheEntry.typeResult.isIncomplete) {
+            return undefined;
```
StellaHuang95, line 1062: > **Warning** · Non-blocking recommendation

This evaluator-visible TypeForm cache behavior is added only to the synchronous Pyright evaluator. Please add async-mode coverage showing the Pylance/TSP path returns equivalent TypeForm results, or document why it routes through this evaluator; otherwise IDE hover/getType behavior can diverge from the CLI path.

StellaHuang95 (approved):
Approved via Review Center.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
rchiodo: Not taking this — this PR is in upstream `microsoft/pyright`, which has no Pylance async program/evaluator (`IAsyncProgram`, snapshots, or `enableAsyncProgram`) to update or exercise. The Type Server packages in this repository route through this same Pyright evaluator; any separate Pylance async-evaluator parity work belongs downstream in `microsoft/pyrx`, not this PR.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
rchiodo: Fixed in 07bf9f828 — `isTypeCached` now checks only the ordinary runtime cache, so a contextual TypeForm entry cannot cause runtime touch guards to skip evaluation. Added a call-argument regression that fails with the old contextual presence check and passes with the fix.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -21254,7 +21493,7 @@ export function createTypeEvaluator(
         }
```
StellaHuang95, line 21506: > **Warning** · Non-blocking recommendation

`evaluateTypeForSubnode` now uses `readTypeCacheEntryForNode`, whose non-TypeForm branch prefers a bare TypeForm entry over the ordinary runtime entry. This differs from `getCachedType`, which deliberately prefers the runtime cache. Could this cause code-flow evaluation to reuse a contextual `TypeForm[...]` result when both entries exist? Please either mirror the runtime-first ordering here or add a regression demonstrating why contextual precedence is required.

[verified]

StellaHuang95 (approved):
Approved via Review Center.

StellaHuang95 (approved):
Approved via Review Center.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
rchiodo: Fixed in 5f38d96ae — split cache lookup intent explicitly. The public `evaluateTypeForSubnode` now reads only the ordinary runtime cache, while `getType`/`getTypeResult` use a separate contextual helper that can select TypeForm entries. Added regressions for both annotated `TypeForm[int] = int` and explicit `TypeForm(...)` nodes after runtime evaluation; the targeted TypeForm cache tests pass.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -21459,13 +21719,29 @@ export function createTypeEvaluator(
...
+                    ) {
                         addDiagnostic(DiagnosticRule.reportInvalidTypeForm, LocMessage.literalNotAllowed(), errorNode);
                     }
                     isValidTypeForm = false;
                     break;
                 }
 
+                case 'TypeAlias': {
+                    if ((flags & (EvalFlags.TypeFormArg | EvalFlags.NoConvertSpecialForm)) !== 0) {
+                        addDiagnostic(
```
StellaHuang95, line 21735: > **Warning** · Non-blocking recommendation

`getTypeArgs` applies `NoConvertSpecialForm` to ordinary nested type arguments too, so this new guard can emit `typeAnnotationVariable` for non-TypeForm code such as `list[TypeAlias]`. Is that intentional? If not, gate this on `TypeFormArg`; if it is, add an ordinary-context regression that locks the intended diagnostic behavior.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -730,7 +737,58 @@ export function createTypeEvaluator(
...
+    function readContextualTypeCacheEntryForNode(node: ParseNode) {
+        const expectedType = expectedTypeCache.get(node.id)?.type;
+        if (expectedType && expectedTypeWantsTypeForm(expectedType)) {
+            return (
+                readTypeFormTypeCacheEntry(node, expectedType) ??
+                readTypeFormTypeCacheEntry(node, /* expectedType */ undefined) ??
+                readTypeCacheEntry(node)
+            );
+        }
+
```
StellaHuang95, line 770: > **Warning** · Non-blocking recommendation

`readContextualTypeCacheEntryForNode` intentionally prefers a TypeForm result when the expected type requests one, while `getCachedType` intentionally prefers the runtime result. Please document this precedence difference and its dependency on `expectedTypeCache`, so a future cleanup does not incorrectly unify the cache reads.

StellaHuang95 (approved):
Approved via Review Center.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
rchiodo: Fixed in 94d1435d3 — restricted the `TypeAlias` diagnostic to `TypeFormArg`. `getTypeArgs` always adds `NoConvertSpecialForm`, so including that flag incorrectly diagnosed ordinary nested annotations such as `list[TypeAlias]`. Added an ordinary-context regression that failed before the change and now passes while the existing TypeForm-context case remains covered.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
rchiodo: Fixed in 94d1435d3 — documented that `readContextualTypeCacheEntryForNode` consults `expectedTypeCache` and intentionally prefers a matching TypeForm entry, while runtime-only consumers must use `readTypeCacheEntry`. This makes the precedence difference explicit and prevents the two lookup domains from being incorrectly unified.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -5163,6 +5276,99 @@ export function createTypeEvaluator(
     }
```
StellaHuang95, line 5269: > **Warning** · Non-blocking recommendation

📍 packages/pyright-internal/src/analyzer/typeEvaluator.ts:5276
The Architect and Skeptic both flag that `addTypeFormForSymbol`'s ~15-branch special-form ladder and the ~20 `(flags & TypeFormArg) ? UnknownType : classType/AnyType` return sites duplicate one decision across the file. The `createClassVarType` handler is the lone case that returns `AnyType` (not the class) on the non-TypeForm side — exactly the kind of variant a future copy-paste edit will get wrong. Consider a name→`LocMessage` table or a small `rejectSpecialFormInTypeForm(flags, fallback)` helper so the diagnose-and-return decision lives in one place. Non-blocking maintainability.

[verified]

StellaHuang95 (approved):
Approved via Review Center.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
rchiodo: Not taking this — the proposed consolidation would erase intentional policy differences rather than remove true duplication. The current file has eight TypeForm fallback ternaries, not ~20, and their surrounding branches diagnose under different flag combinations and preserve different non-TypeForm results (`AnyType` for the pre-existing `NoClassVar` recovery, the original special-form class in ordinary runtime contexts, or `UnknownType` after a hard type-expression error). A shared diagnose-and-return helper would therefore need caller-specific condition, diagnostic, rule, node, and fallback parameters. The explicit `addTypeFormForSymbol` ladder runs only for `TypeFormArg`, uses a distinct localized diagnostic for each special form, and is covered by the bare-special-form cases in `typeForm4.py` (Protocol through Final, including nested invalid forms). Keeping these branches explicit is safer and more readable than introducing a parameterized policy helper or message table.

StellaHuang95, line 5269: > **Warning** · Non-blocking recommendation

TypeForm special-form rejection logic is now distributed across this branch ladder, the special-form switch, and multiple `create*Type` helpers. Because the fallback behavior differs for some forms, future changes could easily update only one path. Please consider consolidating the shared TypeForm rejection decision behind a helper or table to keep these paths aligned.

StellaHuang95 (approved):
Approved via Review Center.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
rchiodo: Fixed in aca120e9e — the bare special-form rejection inventory now lives in one table/helper instead of a 16-branch ladder. Indexed forms remain in their existing handlers because their diagnostics and non-TypeForm fallbacks differ (for example, ClassVar intentionally falls back to Any outside TypeForm), so merging those paths would erase required semantics.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -16111,9 +16293,9 @@ export function createTypeEvaluator(
         typeArgs: TypeResultWithNode[] | undefined,
         flags: EvalFlags
     ): Type {
-        if (flags & EvalFlags.NoClassVar) {
+        if (flags & (EvalFlags.NoClassVar | EvalFlags.TypeFormArg)) {
```
StellaHuang95, line 16283: > **Warning** · Non-blocking recommendation

The TypeForm special-form rejection policy is now spread across the bare-form table, special-form switch cases, and many individual `create*Type` handlers. Consolidating the TypeForm rejection/fallback behavior into a shared helper would reduce the risk of future handlers diverging, particularly around the `ClassVar` fallback behavior.

[verified]

StellaHuang95 (approved):
Approved via Review Center.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
rchiodo: Not taking the broader consolidation: indexed special forms intentionally remain in dedicated handlers because their flag conditions, diagnostics, and non-TypeForm recovery differ; `ClassVar` returns `Any` only for the existing ordinary `NoClassVar` recovery, but returns `Unknown` for an invalid TypeForm. The genuinely shared bare-form decision is centralized in `aca120e9e`, with exact matching preserved in `797965f10`; parameterizing all handler-specific policy would obscure rather than align these semantics.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -777,21 +838,43 @@ export function createTypeEvaluator(
...
+                return;
+            }
+
+            cacheEntries = cacheEntries.filter((entry) => !entryMatchesExpectedType(entry, expectedType));
+            cacheEntries.push({ typeResult, flags, incompleteGenCount, expectedType });
+
+            const maxCacheEntriesPerNode = 8;
+            if (cacheEntries.length > maxCacheEntriesPerNode) {
+                cacheEntries = cacheEntries.slice(cacheEntries.length - maxCacheEntriesPerNode);
+            }
```
StellaHuang95: > **Warning** · Non-blocking recommendation

The TypeForm cache increments the global incomplete-generation counter before returning in speculative mode, but does not retain a speculative TypeForm entry. Repeated speculative reads therefore re-evaluate and re-invalidate dependent incomplete results. Please either defer the generation update until after the speculative early return or add scoped speculative caching for TypeForm results.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -17256,6 +17461,16 @@ export function createTypeEvaluator(
         let isIncomplete = false;
         let expectedTypeDiagAddendum: DiagnosticAddendum | undefined;
 
+        // A runtime-first query may have cached the RHS without its assignment
+        // context. Re-evaluate it when the annotation expects a TypeForm so the
+        // ordinary cache cannot suppress contextual validation and conversion.
+        if (rightHandType) {
+            const declaredType = getDeclaredTypeForExpression(node.d.leftExpr, { method: 'set' });
+            if (declaredType && expectedTypeWantsTypeForm(declaredType)) {
+                rightHandType = undefined;
```
StellaHuang95, line 17493: > **Warning** · Non-blocking recommendation

This adds a declared-type lookup for every assignment whose RHS was already cached, even though nearly all assignments are unrelated to TypeForm. Please avoid resolving the declared type on this hot path unless there is a cheaper indication that TypeForm contextual reevaluation may be required.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -730,7 +737,61 @@ export function createTypeEvaluator(
...
+    }
+
+    // Contextual reads consult expectedTypeCache and prefer a matching TypeForm result.
+    // Runtime-only consumers must use readTypeCacheEntry so this precedence is not
+    // accidentally applied where an ordinary runtime type is required.
+    function readContextualTypeCacheEntryForNode(node: ParseNode) {
+        const expectedType = expectedTypeCache.get(node.id)?.type;
+        if (expectedType && expectedTypeWantsTypeForm(expectedType)) {
+            return (
+                readTypeFormTypeCacheEntry(node, expectedType) ??
```
StellaHuang95, line 765: > **Warning** · Non-blocking recommendation

This introduces another expected-type-keyed, bounded per-node cache with its own matching and eviction rules alongside the similar speculative-cache implementation. Please consider sharing the entry matching, capacity, and eviction behavior so these caches cannot silently drift over time.

StellaHuang95 (approved):
Approved via Review Center.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
rchiodo: Fixed in `5fa57f821`: TypeForm cache writes now return from speculative mode before updating `incompleteGenCount`, so a discarded result cannot invalidate persistent incomplete entries. The new regression primes an incomplete entry, performs speculative TypeForm evaluation, and verifies that the generation remains unchanged.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
rchiodo: Fixed in `5fa57f821`: cached assignment handling now performs a cheap target-shape and symbol precheck before resolving a declared type, so ordinary inferred locals remain on the existing fast path. When contextual reevaluation is needed, the resolved declaration is reused; focused coverage includes inline and reassigned names, class names, member access, and TypedDict index targets.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
rchiodo: Refactored in `5fa57f821`: both TypeForm and speculative expected-type caches now share matching, replacement, stale-entry filtering, and newest-eight eviction logic. Direct helper tests cover all four policies, while speculative dependent-type matching remains in the tracker lookup where its active context is available.

StellaHuang95 (approved):
Approved via Review Center.

PR #11601, opened 2026-08-07
Apply overload materialization to nested Any

rchiodo:
## Summary

- apply overload Step 5 when `Any` or `Unknown` appears in an invariant nominal type argument
- keep all compatible overloads until materialization filtering determines which later overloads can be eliminated
- propagate ambiguous inferred `__init__` self types while preserving explicitly specialized constructor results
- add the `overloads_evaluation.py` conformance regression and realistic primer-derived coverage

## Why

Overload matching previously marked only a top-level `Any` or `Unknown` argument as non-definitive. An argument such as `list[Any]` was therefore treated as a definitive match for the first compatible overload, so overloads for `list[int]` and `list[str]` incorrectly selected `list[int]`.

The overload specification requires materialization filtering in this case. The updated logic detects nested gradual forms in invariant nominal positions, evaluates whether a candidate accepts every supported materialization, and eliminates only subsequent overloads. If multiple candidates with non-equivalent results remain, nested `Any` produces `Any` and nested `Unknown` produces `Unknown`.

## Precision boundaries

The proof is intentionally limited to nominal generic relationships, including covariant base adaptation through the MRO. Unsupported proofs involving structural protocols, callable signatures, unions, contravariant generic arguments, or TypeVar-dependent relationships retain the existing overload-selection behavior rather than making an unsound elimination.

Normal assignability cannot answer the Step 5 question because it checks one gradual source type and accepts `Any` or `Unknown` without proving that every materialization is covered. The dedicated traversal is therefore conservative and tri-state: `true` means all materializations are covered, `false` identifies a counterexample, and an unproven shape preserves legacy overload ordering.

For overloaded constructors, the effective specialized `self` types participate in ambiguity handling. Explicit class specialization remains authoritative and is not degraded to `Any` or `Unknown`.

Filtering intentionally keeps `overloadsUsedForCall` aligned with the retained candidates for both the existing top-level gradual path and the new nested-materialization path. Deprecation diagnostics therefore report only overloads that remain possible; an exact diagnostic-range assertion pins this behavior.

The broadened test matrix covers primer-derived `Series`/`Index` operators, Prefect-style callable receivers, Home Assistant optional overloads, scipy shape-specialized concatenation, `re.Pattern[Any | Unknown]`, inferred and explicit constructors, multi-argument filtering, and mixed keyword/default calls. It also pins unchanged behavior for covariance, top-level gradual arguments, equivalent returns, concrete/unambiguous calls, defaults and literal context, bounded/constrained TypeVars, unsupported proof shapes, contravariance, deprecation, and invalid calls. Concrete downstream operations and expected errors guard against accidental diagnostic suppression.

## Conformance case

For:

```python
@overload
def overloaded(value: list[int]) -> list[int]: ...

@overload
def overloaded(value: list[str]) -> list[str]: ...

def check(value: list[Any]) -> None:
    reveal_type(overloaded(value), expected_text="Any")
```

both invariant overloads remain possible, so the result is now `Any` rather than `list[int]`.

## Tests

- `npx jest typeEvaluator6.test --runInBand --forceExit` (150 tests; all overload and constructor evaluator tests)
- focused solver, TypedDict, and deprecation Jest selection (116 tests)
- `npm run typecheck`
- `npm run check:syncpack`
- `npm run check:eslint`
- `npm run check:prettier`
- `git diff --check`


StellaHuang95:

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -9622,22 +9630,58 @@ export function createTypeEvaluator(
...
                 possibleMatchResults = filterOverloadMatchesForUnpackedArgs(possibleMatchResults);
                 possibleMatchResults = filterOverloadMatchesForAnyArgs(possibleMatchResults);
+                overloadsUsedForCall.splice(
+                    overloadsUsedStartIndex,
+                    overloadsUsedForCall.length - overloadsUsedStartIndex,
+                    ...possibleMatchResults.map((result) => result.overload)
+                );
 
                 // Did the filtering produce a single result? If so, we're done.
                 if (possibleMatchResults.length === 1) {
```
StellaHuang95, line 9644: > **Warning** · Non-blocking recommendation

Please add a location/range assertion for the deprecated-overload diagnostics. The new `overloadsUsedForCall` filtering affects which overloads feed deprecation reporting, but the sample currently asserts only an aggregate deprecation count; a false positive in one deprecated group could mask a missing diagnostic in the other.

StellaHuang95 (approved):
Approved via Review Center.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
rchiodo: Fixed — added an exact range assertion for the deprecated-overload diagnostic, confirming it is emitted on the intended `overload5_deprecated_union` call rather than merely asserting the aggregate count.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -9743,38 +9802,405 @@ export function createTypeEvaluator(
         return unpackedArgsOverloads;
     }
 
+    function getAnyOrUnknownInInvariantPosition(type: Type, recursionCount = 0): AnyType | UnknownType | undefined {
+        if (recursionCount > maxTypeRecursionCount) {
+            return undefined;
+        }
+        recursionCount++;
+
```
StellaHuang95, line 9821: > **Warning** · Non-blocking recommendation

The materialization helpers duplicate variance, MRO-specialization, and tuple-matching logic already embodied in the normal assignability machinery. Please document why that machinery cannot be reused, or extract a shared traversal primitive, so later changes do not silently make overload elimination unsound.

[verified]

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -9743,38 +9802,405 @@ export function createTypeEvaluator(
...
+        destType: Type,
+        srcType: Type,
+        recursionCount = 0
+    ): MaterializationCoverage {
+        if (recursionCount > maxTypeRecursionCount) {
+            return undefined;
+        }
+        recursionCount++;
+
+        if (!containsAnyOrUnknown(srcType, /* recurse */ true)) {
```
StellaHuang95, line 9925: > **Info** · Optional note

The `isTypeVar(destType)` check inside the `isAnyOrUnknown(srcType)` branch is unreachable because TypeVars return `undefined` in the preceding guard. Please remove it, or make the intended distinct case reachable and cover it with a test.

[verified]

StellaHuang95 (approved):
Approved via Review Center.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -9622,22 +9630,58 @@ export function createTypeEvaluator(
             if (!isDefinitiveMatchFound && possibleMatchResults.length > 0) {
                 possibleMatchResults = filterOverloadMatchesForUnpackedArgs(possibleMatchResults);
                 possibleMatchResults = filterOverloadMatchesForAnyArgs(possibleMatchResults);
+                overloadsUsedForCall.splice(
```
StellaHuang95, line 9637: > **Warning** · Non-blocking recommendation

Filtering `overloadsUsedForCall` also changes deprecation reporting for the existing top-level-`Any` ambiguity path, not only the new nested-materialization case. Please confirm this narrowing is intentional and document it in the PR description.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
StellaHuang95, line 9925: > **Warning** · Non-blocking recommendation

These materialization walkers duplicate variance, MRO, and tuple logic that normally lives in assignability machinery. Please document the conservatism invariant—return `undefined` whenever coverage cannot be proven, and return `true` only when every materialization is covered—and why normal `assignType` cannot answer this question.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -9622,22 +9630,58 @@ export function createTypeEvaluator(
...
+                            const argParamPairs = getOverloadArgParamPairs(match);
+                            return index < argParamPairs.length ? argParamPairs[index].paramType : UnknownType.create();
+                        });
+
+                        if (!areTypesSame(paramTypes, { treatAnySameAsUnknown: true })) {
+                            if (isAnyOrUnknown(pair.argType)) {
+                                ambiguousMatchIncludesTopLevelAnyOrUnknown = true;
+                            } else {
+                                const anyOrUnknown = getAnyOrUnknownInInvariantPosition(pair.argType);
+                                if (anyOrUnknown && isAny(anyOrUnknown)) {
```
StellaHuang95, line 9664: > **Warning** · Non-blocking recommendation

`getOverloadArgParamPairs` is compared by index across candidates, so its caller-order invariant is load-bearing for keyword and default-argument calls. Please document that invariant and add a mixed keyword/default regression case.

packages/pyright-internal/src/tests/typeEvaluator6.test.ts
```diff
@@ -67,7 +67,11 @@ test('OverloadCall5', () => {
 
 test('OverloadCall6', () => {
     const analysisResults = TestUtils.typeAnalyzeSampleFiles(['overloadCall6.py']);
-    TestUtils.validateResults(analysisResults, 2);
+    TestUtils.validateResults(analysisResults, 2, 0, undefined, undefined, undefined, 1);
```
StellaHuang95, line 70: > **Warning** · Non-blocking recommendation

The aggregate deprecation count can mask a warning moved from `func6_deprecated_union` to `func6_deprecated` (or vice versa). Since this change alters the overload set used for deprecation reporting, split these scenarios into independently counted samples or assert each diagnostic location.

StellaHuang95 (approved):
Approved via Review Center.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
StellaHuang95, line 9821: > **Warning** · Non-blocking recommendation

Please document the tri-state materialization invariant here: `true` must mean every materialization is proven covered, while `undefined` must conservatively fall back to legacy overload selection. This is load-bearing for sound overload elimination and explains why ordinary `assignType` checks are insufficient.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -9684,7 +9735,12 @@ export function createTypeEvaluator(
...
 
-                    returnTypes.push(returnType);
+                    if (isInitSelfMaterializationAmbiguity) {
+                        specializedInitSelfTypes.push(returnType);
+                        returnTypes.push(possibleMatchResults[0].returnType);
+                    } else {
+                        returnTypes.push(returnType);
+                    }
                 }
             }
```
StellaHuang95, line 9754: > **Warning** · Non-blocking recommendation

The constructor ambiguity result is intentionally carried through `specializedInitSelfTypes` while `returnTypes` receives the first overload's result, then reconstructed in `constructors.ts`. Please document this cross-module placeholder-return contract and add coverage for union expansion combined with an ambiguous overloaded `__init__`.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -9743,38 +9802,405 @@ export function createTypeEvaluator(
...
+            });
+            return result;
+        }
+
+        if (!isClass(type)) {
+            return undefined;
+        }
+
+        // Tuple entries are covariant, but they can contain an invariant type.
+        if (type.priv.tupleTypeArgs) {
```
StellaHuang95, line 9841: > **Warning** · Non-blocking recommendation

`getOverloadArgParamPairs` is compared by array index across overloads. The current caller-order construction makes that valid, but this is a load-bearing invariant. Please document it and add a mixed keyword/default-argument regression so future changes cannot silently misalign same-length overload signatures.

StellaHuang95 (approved):
Approved via Review Center.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
rchiodo: Added comments explaining why `assignType` cannot expose invariant gradual dependence and documenting the conservative tri-state proof contract. This keeps unsupported shapes on the legacy path rather than treating incomplete coverage as proven.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
rchiodo: Removed the unreachable TypeVar check. TypeVar destinations already return the unproven state in the preceding guard, and the existing TypeVar controls continue to pass.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
rchiodo: Confirmed this narrowing is intentional: bookkeeping now mirrors candidates retained after both top-level gradual and nested-materialization filtering, so deprecations come only from still-possible overloads. I documented this in code and the PR description, and the exact diagnostic-range assertion pins the behavior.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
rchiodo: Documented why normal assignability cannot prove universal materialization coverage and made the tri-state conservatism invariant explicit. `undefined` remains the result whenever the nominal/tuple traversal cannot prove coverage.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
rchiodo: Documented the shared caller-order invariant for `argResults` and `argParams`. I also added a mixed keyword/default regression with reordered keyword-only declarations, an omitted default, and an explicitly supplied default.

packages/pyright-internal/src/tests/typeEvaluator6.test.ts
rchiodo: Addressed by pinning the single deprecated diagnostic to the exact `func6_deprecated_union` call range, while both deprecated scenarios now assert their result types. This prevents the aggregate count from passing if the warning moves between calls.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
rchiodo: The shared tri-state contract is now documented immediately above `MaterializationCoverage`: `true` proves universal coverage, `false` identifies a counterexample, and `undefined` preserves legacy selection. The detector comment also explains why `assignType` cannot expose invariant gradual dependence.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
rchiodo: Added both sides of the cross-module contract in `typeEvaluator.ts` and `constructors.ts`. The new constructor test combines union expansion with ambiguous `list[Any]` and `list[Unknown]` overloads and pins the results to `Any` and `Unknown`.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
rchiodo: This invariant is now documented directly above `getOverloadArgParamPairs`. The new `keyword_default_case` regression uses reordered keyword-only declarations and covers both an omitted default and an explicitly supplied default.

StellaHuang95 (approved):
Approved via Review Center.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -9684,7 +9739,17 @@ export function createTypeEvaluator(
                         }
                     }
 
-                    returnTypes.push(returnType);
+                    if (isInitSelfMaterializationAmbiguity) {
+                        // Overloaded __init__ methods normally return None. Preserve that
+                        // ordinary return as a placeholder while union-expanded calls are
```
StellaHuang95, line 9744: > **Warning** · Non-blocking recommendation

When at least one union-expanded constructor branch is materialization-ambiguous, this accumulated list takes precedence over `finalCallResult.specializedInitSelfType`. A later definitive branch's specialized self type is therefore omitted from the aggregate. Please add a regression covering mixed ambiguous and definitive constructor branches, and preserve all branch self types if their combined result is intended to be observable.

StellaHuang95 (approved):
Approved via Review Center.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
rchiodo: Addressed in follow-up PR #11621. It preserves every union-expanded overloaded `__init__` branch's specialized self type and adds coverage for mixed ambiguous/definitive branches, definitive-only branches, all-ambiguous branches, explicit specialization, and the `None` placeholder return.

Dr-Irv:
Given the number of reports of the effects of this change at https://github.com/microsoft/pyright/pull/11601#issuecomment-5245725566 (which seems to be missing tests on `pandas-stubs` that used to be there??), I am surprised that this was approved and merged.  @rchiodo can you explain?

rchiodo:
> Given the number of reports of the effects of this change at [#11601 (comment)](https://github.com/microsoft/pyright/pull/11601#issuecomment-5245725566) (which seems to be missing tests on `pandas-stubs` that used to be there??), I am surprised that this was approved and merged. Rich Chiodo (@rchiodo) can you explain?

Simple. There's no test that prevents the PR from merging. I set it to auto merge. When I fix this regression, I'll add a test to prevent it in the future and see why there's no panda-stubs testing.

rchiodo:
The pandas-stubs output was there in the mypy_primer differences, it was just cut off because there were so many

rchiodo:
So yes it's still being tested.

Dr-Irv:
> So yes it's still being tested.

May I suggest that if that many errors are reported, a new version of `pyright` is not released?  We've had to pin `pyright` to 1.1.411 in `pandas-stubs` due to this regression.

At least in the past, @erictraut used to let us know when there was a change that was made that would affect `pandas-stubs`, and then we could address it (and discuss whether what we were doing was considered correct).  But in this case, it seems the change was incorrect and hopefully #11732 and other PRs will address that.

Note that we are testing `pandas-stubs` with `mypy`, `pyright`, `ty` and `pyrefly` so new errors reported on `pandas-stubs` via the `mypy_primer` test in `pyright` should be considered to most likely be a bug introduced by the PR that includes the change.

Thanks for your consideration and support.

rchiodo:
> Note that we are testing pandas-stubs with mypy, pyright, ty and pyrefly so new errors reported on pandas-stubs via the mypy_primer test in pyright should be considered to most likely be a bug introduced by the PR that includes the change.

This would be nice, but it's not always obvious if the mypy_primer differences are expected or not. I'm working on something that will hopefully make this more obvious to the reviewer. 

At least in the future, we won't ignore mypy_primer differences (minus sympy, which seems to change on every run)

PR #11605, opened 2026-08-08
Fix inherited asymmetric descriptor detection

### LINKED ISSUE #11401: Incorrect type when descriptors are at different levels in the class hierarchy
When `__get__` and `__set__` are at different levels in the class hierarchy, they do not seem to be behaving properly. Example:
```py
from typing import assert_type, cast, Any

class AttributeGet[ T ]:
   def __get__( self, instance: Any, owner: Any = None ) -> T:
      return cast( Any, None )

class Attribute[ T, U ]( AttributeGet[ T ] ):
   def __set__( self, instance: Any, object: T | U ) -> None:
      pass

class StrInt( Attribute[ int, str] ):
    pass

class MyContainer:
   si: StrInt

c = MyContainer()
c.si = "2"
assert_type( c.si, int ) # ERROR HERE
```
Playground [link](https://pyright-play.net/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMoCGAzkQKZ4D68CpANFAMbEz0CCKcAsAFA8MA2xIlFYwYIJACMArjFIBxUjADaUACpQAugC4eUfQBNSwKBQpolZgBRQy-YPVREYBFA1LaRHemADuKck92OCgAXigAOTAAqABKKABaAD51XW59DKgQJWkQFEZmG2D6KJjYnj5BEhExCRk5VTV6AFUtItqpWQUlRq04tIyjEzMyGGtbUntHFGdXdyDvKDBJACtSBhhPDQAfKFb45Mjojz1MqAQhCu4BISgAZXEASRQYdvFOhswX%2BmcQTX7TvoLiQrjdqgBZOAAYWiLlQgUBRCQngeIGeMFBYSgkJhLwI8JAVnK1wAdEisQAiABMFJ4QnIY2opBsDDJSGm%2BGJQA)

patrickswedish:
## Summary

- detect asymmetric descriptors using the effective inherited `__get__` and `__set__` methods
- preserve assignment/read asymmetry when accessors live at different levels of the descriptor MRO
- add regression coverage for an inherited generic getter plus a wider setter, with a property control

## Root cause

`isAsymmetricDescriptorClass` looked up both accessors with `SkipBaseClasses`. A descriptor that inherited `__get__` but declared `__set__` locally therefore appeared symmetric. After assignment, flow analysis incorrectly narrowed subsequent reads to the setter input literal instead of the getter return type.

Fixes #11401

## Validation

- `npx jest packages/pyright-internal/src/tests/typeEvaluator8.test.ts --runInBand` — 161/161 passed
- `npm run typecheck` — passed all three packages
- `npm run check` — syncpack, ESLint, and Prettier passed
- `git diff --check` — passed

### DIFF
--- a/packages/pyright-internal/src/analyzer/typeEvaluator.ts
+++ b/packages/pyright-internal/src/analyzer/typeEvaluator.ts
@@ -7022,8 +7022,10 @@ export function createTypeEvaluator(
 
         let isAsymmetric = false;
 
-        const getterSymbolResult = lookUpClassMember(classType, '__get__', MemberAccessFlags.SkipBaseClasses);
-        const setterSymbolResult = lookUpClassMember(classType, '__set__', MemberAccessFlags.SkipBaseClasses);
+        // Accessors can be defined at different levels of the descriptor's MRO.
+        // Compare the effective inherited getter and setter types.
+        const getterSymbolResult = lookUpClassMember(classType, '__get__');
+        const setterSymbolResult = lookUpClassMember(classType, '__set__');
 
         if (!getterSymbolResult || !setterSymbolResult) {
             isAsymmetric = false;
--- a/packages/pyright-internal/src/analyzer/typeEvaluator.ts
+++ b/packages/pyright-internal/src/analyzer/typeEvaluator.ts
@@ -6843,8 +6843,8 @@ export function createTypeEvaluator(
 
         // Determine if we're calling __set__ on an asymmetric descriptor or property.
         let isAsymmetricAccessor = false;
-        if (usage.method === 'set' && isClass(methodClassType)) {
-            if (isAsymmetricDescriptorClass(methodClassType)) {
+        if (usage.method === 'set') {
+            if (isAsymmetricDescriptorClass(concreteMemberType)) {
                 isAsymmetricAccessor = true;
             }
         }
--- a/packages/pyright-internal/src/tests/samples/descriptor5.py
+++ b/packages/pyright-internal/src/tests/samples/descriptor5.py
@@ -1,0 +1,42 @@
+# This sample tests that assignment through an inherited asymmetric descriptor
+# does not narrow subsequent reads to the setter's input type.
+
+from typing import Any, assert_type, cast
+
+
+class Getter[T]:
+    def __get__(self, instance: Any, owner: Any = None) -> T:
+        return cast(Any, None)
+
+
+class Descriptor[T, U](Getter[T]):
+    def __set__(self, instance: Any, value: T | U) -> None:
+        pass
+
+
+class IntFromStr(Descriptor[int, str]):
+    pass
+
+
+class Container:
+    value: IntFromStr
+
+
+container = Container()
+container.value = "1"
+assert_type(container.value, int)
+
+
+class AsymmetricProperty:
+    @property
+    def value(self) -> int:
+        return 0
+
+    @value.setter
+    def value(self, new_value: str) -> None:
+        pass
+
+
+prop = AsymmetricProperty()
+prop.value = "1"
+assert_type(prop.value, int)
--- a/packages/pyright-internal/src/tests/samples/descriptor5.py
+++ b/packages/pyright-internal/src/tests/samples/descriptor5.py
@@ -27,6 +27,28 @@ container.value = "1"
 assert_type(container.value, int)
 
 
+class SetterBase:
+    def __get__(self, instance: Any, owner: Any = None) -> int | str:
+        return cast(Any, None)
+
+    def __set__(self, instance: Any, value: int | str) -> None:
+        pass
+
+
+class GetterOverride(SetterBase):
+    def __get__(self, instance: Any, owner: Any = None) -> int:
+        return 0
+
+
+class OverrideContainer:
+    value: GetterOverride
+
+
+override_container = OverrideContainer()
+override_container.value = "1"
+assert_type(override_container.value, int)
+
+
 class AsymmetricProperty:
     @property
     def value(self) -> int:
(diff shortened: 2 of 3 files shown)


StellaHuang95:

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -7022,8 +7022,10 @@ export function createTypeEvaluator(
 
         let isAsymmetric = false;
 
-        const getterSymbolResult = lookUpClassMember(classType, '__get__', MemberAccessFlags.SkipBaseClasses);
-        const setterSymbolResult = lookUpClassMember(classType, '__set__', MemberAccessFlags.SkipBaseClasses);
+        // Accessors can be defined at different levels of the descriptor's MRO.
+        // Compare the effective inherited getter and setter types.
+        const getterSymbolResult = lookUpClassMember(classType, '__get__');
+        const setterSymbolResult = lookUpClassMember(classType, '__set__');
```
StellaHuang95, line 7028: > **Warning** · Non-blocking recommendation

This removes `SkipBaseClasses`, but the lookup is still anchored to `methodClassType` (the class that declares `__set__`), rather than the concrete descriptor class. That handles the new linear case, but sibling-mixin/diamond descriptors and getter overrides below the setter can still evaluate asymmetry against the wrong accessor. If the intended behavior is to compare the effective accessors of the concrete descriptor, please thread that class into this check. [verified]

StellaHuang95 (approved):
Approved via Review Center.

rchiodo (approved):
Approved via Review Center.

StellaHuang95 (approved):
Approved via Review Center.

rchiodo (approved):
Approved via Review Center.

StellaHuang95 (approved):
Approved via Review Center.

rchiodo:

rchiodo (approved):
Approved via Review Center.

heejaechang (approved):
Approved via Review Center.

rchiodo:

bschnurr (approved):
Approved via Review Center.

heejaechang:
### 🔴 Pyright CLI QA — `9bc8e9a6` — verdict: red

11 focused check(s) completed.

**`main.py`**

```python
from typing import assert_type, reveal_type

class Meta(type):
    def __get__(cls, instance: object, owner: object = None) -> int:
        return 42

    def __set__(cls, instance: object, value: str) -> None:
        pass

class Descriptor(metaclass=Meta):
    pass

class Model:
    value = Descriptor

def test(model: Model) -> None:
    model.value = "42"
    assert_type(model.value, int)
    reveal_type(model.value)
```

**`pyrightconfig.json`**

```json
{
  "include": [
    "main.py"
  ],
  "pythonVersion": "3.13",
}
```

```text
```

**Base `e8af042e988f9b06b0c6da77b2269a8bd73e5d8d`**

```json
{
  "exit_code": 0,
  "diagnostics": [
    {
      "file": "main.py",
      "severity": "information",
      "rule": null,
      "message": "Type of \"model.value\" is \"int\"",
      "range": {
        "start": [
          18,
          16
        ],
        "end": [
          18,
          27
        ]
      }
    }
  ]
}
```

**PR `9bc8e9a6f840866ecb437a75b7a4fb26d8f2e875`**

```json
{
  "exit_code": 1,
  "diagnostics": [
    {
      "file": "main.py",
      "severity": "information",
      "rule": null,
      "message": "Type of \"model.value\" is \"Literal['42']\"",
      "range": {
        "start": [
          18,
          16
        ],
        "end": [
          18,
          27
        ]
      }
    },
    {
      "file": "main.py",
      "severity": "error",
      "message": "\"assert_type\" mismatch: expected \"int\" but received \"Literal['42']\"",
      "range": {
        "start": [
          17,
          16
        ],
        "end": [
          17,
          27
        ]
      }
    }
  ]
}
```

</details>

patrickswedish:
Thanks. I pushed a fix for both regressions.

The descriptor accessor lookup now resolves __get__ / __set__ against the concrete descriptor class, while specializing both accessors in that concrete class context. This avoids the instance-level __get__ shadowing case and keeps the inherited/overridden Self descriptor case in the correct generic context.

I also added regression coverage for both cases. Pushed in 526a4b6.

rchiodo:
The implementation looks reasonable to me. I checked head `526a4b6` and didn't find a correctness regression in additional focused cases, but the new `Descriptor5` test needs a couple of fixes:

- `descriptor5.py` uses `Self` without importing it. Running the test as committed produces four errors.
- After adding that import, `SelfSetterOverride.__set__` produces a `reportIncompatibleMethodOverride` diagnostic: the parameter narrows from `SelfDescriptor` to `SelfSetterOverride`. The test currently expects zero errors. Could you explicitly account for that expected diagnostic, or adjust the fixture, while keeping the `assert_type(..., SelfValue)` and `.extra()` checks?

In an isolated snapshot, adding the import and suppressing only that specific override diagnostic lets all 161 tests in `typeEvaluator8.test.ts` pass.

Could you also add regression coverage for sibling-mixin/diamond inheritance, preservation of assignment narrowing for symmetric inherited descriptors, and multiple specializations of the same generic descriptor (including through a generic container)? These cases passed in my additional probes; committing them would help protect both MRO selection and type precision. Inherited overloaded accessors and rejection of invalid setter inputs would also be useful coverage.

patrickswedish:
Fixed.

Added the missing Self import.
Handled the expected reportIncompatibleMethodOverride only in the descriptor override fixture while preserving the assert_type(..., SelfValue) and .extra() checks.
Added targeted regression coverage for diamond/sibling inheritance, symmetric inherited descriptors, multiple specializations, generic-container usage, inherited overloads, and invalid setter inputs.

typeEvaluator8.test.ts: all tests passing.
npm run typecheck: passing.
npm run check: passing.

Pushed in 9bc8e9a.

PR #11610, opened 2026-08-09
Fix TypedDict in narrowing wrongly eliminating subtype for extra_items keys

nileshpatil6:
For a TypedDict with `extra_items` (PEP 728), a membership check on a key that is not a known item wrongly makes the negative branch unreachable:

```python
class Movie(TypedDict, extra_items=bool):
    name: str

def func(movie: Movie) -> None:
    if "novel_adaptation" in movie:
        return
    reveal_type(movie)  # branch is wrongly marked unreachable today
```

The cause is in `narrowTypeForTypedDictKey`: the negative branch used `knownItems.get(key) ?? extraItems`, so a key that only matched the extra_items pseudo-entry was treated like a known provided item and eliminated the subtype. An extra item may or may not be present at runtime, so its absence can never rule the type out. Only a known item that is required or provided can do that.

The positive branch keeps its current behavior, it still uses the extra_items entry to mark the key as provided after the check.

Testing: added a case to `typedDictClosed1.py` where code after a negative `in` check on an extra key must stay reachable (it asserts an unrelated type error is still reported there, which the unreachability bug swallows). Without the fix that sample reports 7 errors instead of 8, with it all 10 TypedDictClosed tests pass and the full typeEvaluator suite is green: 1184 tests across 8 suites, 0 failures.

### DIFF
--- a/packages/pyright-internal/src/tests/samples/typedDictClosed1.py
+++ b/packages/pyright-internal/src/tests/samples/typedDictClosed1.py
@@ -48,6 +48,17 @@ def func2(movie: Movie) -> None:
     movie["other3"] = 1
 
 
+def func3(movie: Movie) -> None:
+    # "name" is a required known item, so it is always present and the
+    # negative branch of the "in" check must remain unreachable.
+    if "name" in movie:
+        return
+
+    # This would generate a type incompatibility error if the statement
+    # above were incorrectly considered reachable.
+    movie["other4"] = 1
+
+
 class MovieBase(TypedDict, extra_items=ReadOnly[str | None]):
     name: str
 
--- a/packages/pyright-internal/src/tests/samples/typedDictClosed1.py
+++ b/packages/pyright-internal/src/tests/samples/typedDictClosed1.py
@@ -35,6 +35,19 @@ def func1(movie: Movie) -> None:
     movie["other2"] = 1
 
 
+def func2(movie: Movie) -> None:
+    # An extra item is not guaranteed to be present, so the negative
+    # branch of the "in" check must remain reachable.
+    if "novel_adaptation" in movie:
+        return
+
+    reveal_type(movie, expected_text="Movie")
+
+    # This should generate a type incompatibility error. If the statements
+    # above are incorrectly marked unreachable, no error is reported here.
+    movie["other3"] = 1
+
+
 class MovieBase(TypedDict, extra_items=ReadOnly[str | None]):
     name: str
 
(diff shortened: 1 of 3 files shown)


StellaHuang95:

packages/pyright-internal/src/tests/samples/typedDictClosed1.py
```diff
@@ -35,6 +35,19 @@ def func1(movie: Movie) -> None:
...
+    # An extra item is not guaranteed to be present, so the negative
+    # branch of the "in" check must remain reachable.
+    if "novel_adaptation" in movie:
+        return
+
+    reveal_type(movie, expected_text="Movie")
+
+    # This should generate a type incompatibility error. If the statements
+    # above are incorrectly marked unreachable, no error is reported here.
+    movie["other3"] = 1
```
StellaHuang95, line 48: > **Warning** · Non-blocking recommendation

Please also cover a negative `in` check on a known required key (such as `name`) and assert that its following code remains unreachable. This would protect the existing narrowing behavior against an accidental over-relaxation of this change. [verified]

StellaHuang95 (approved):
Approved via Review Center.

nileshpatil6:
Good call. Added in f379f89: a negative `in` check on `name` (required known item) with an assignment in the branch that would be a type error if the branch were reachable. The sample's expected error count is unchanged, so the test fails if that branch ever becomes reachable.

Also sanity checked it guards what you had in mind: if I over-relax the fix to return the subtype unconditionally in the negative branch, TypedDictClosed1 fails on exactly this case; with the fix as written, all 10 TypedDictClosed tests pass.

StellaHuang95 (approved):
Approved via Review Center.

rchiodo (approved):
Approved via Review Center.

rchiodo (approved):
Approved via Review Center.

rchiodo:

rchiodo (approved):
Approved via Review Center.

heejaechang (approved):
Approved via Review Center.

rchiodo:

rchiodo:
@copilot resolve the merge conflicts in this pull request

nileshpatil6:
Closing this in favour of #11663, which landed the same fix while this was in review.

The two changes are equivalent: both separate the known-items lookup from the `extraItems` fallback and stop the negative branch eliminating a subtype when the key is only covered by `extra_items`. #11663 also carries a broader sample in `typedDictClosed11.py`, including the required-key guard case @StellaHuang95 asked me to add here (its `func6`, where the negative branch of a required key narrows to `Never`), so there is nothing from this PR worth rebasing on top.

@rchiodo no need to resolve the conflict here, this is now redundant. Thanks all for the reviews.

PR #11612, opened 2026-08-10
Fix negative type narrowing for mapping patterns with multiple key entries

hsusul:
### Minimal Reproduction Case

```python
from typing import Literal, TypedDict

class MsgA(TypedDict):
    v: Literal[1]
    kind: Literal["a"]
    data_a: int

class MsgB(TypedDict):
    v: Literal[1]
    kind: Literal["b"]
    data_b: str

def test_negative_narrowing(msg: MsgA | MsgB) -> None:
    match msg:
        case {"v": 1, "kind": "a"}:
            reveal_type(msg)  # MsgA
        case _:
            reveal_type(msg)  # Expected: MsgB, Actual pre-fix: MsgA | MsgB
```

### Current vs Expected Behavior

- **Current Behavior**: When a mapping pattern contains two or more key entries (e.g. `case {"v": 1, "kind": "a"}:`), Pyright does not eliminate matching `TypedDict` subtypes from the subject in the fallthrough (`case _:`) branch. Consequently, `reveal_type(msg)` in `case _:` reports `MsgA | MsgB`.
- **Expected Behavior**: Any subject instance matching `case {"v": 1, "kind": "a"}:` is guaranteed to be `MsgA`. The fallthrough `case _:` branch cannot receive `MsgA`, so `msg` must be narrowed to `MsgB`.

### Rationale

In Python 3.10+ pattern matching (PEP 634), matching a subject against a mapping pattern checks whether the subject is a mapping and contains matching values for all specified key entries. When a subject is a union of `TypedDict` types, a `TypedDict` subtype whose literal field values match all key entries of a mapping pattern will always match that `case` clause. In negative type narrowing (`!isPositiveTest`), such matching `TypedDict` subtypes must be eliminated from the remaining subject type in fallthrough branches.

### Root Cause

In `packages/pyright-internal/src/analyzer/patternMatching.ts`, `narrowTypeBasedOnMappingPattern` previously hardcoded a single-key check (`pattern.d.entries.length === 1`) when `!isPositiveTest`. Mapping patterns with two or more key entries were skipped, preventing negative type narrowing for multi-key patterns.

### Proposed Fix

Generalize `narrowTypeBasedOnMappingPattern` under `!isPositiveTest` to parse all key entries in `pattern.d.entries`. If a `TypedDict` subtype matches all key entries in the mapping pattern (matching string literal key names, literal value types, or wildcard/capture variables), eliminate that `TypedDict` subtype from the remaining type.

### Test Coverage

Added regression test `test_negative_narrowing3` to `packages/pyright-internal/src/tests/samples/matchMapping1.py`.

### Validation Results

- `npx jest src/tests/typeEvaluator6.test.ts` passed (148/148 tests passed).
- `npx tsc --noEmit` passed with 0 errors.
- `git diff --check` passed with 0 formatting or whitespace errors.

### DIFF
--- a/packages/pyright-internal/src/analyzer/patternMatching.ts
+++ b/packages/pyright-internal/src/analyzer/patternMatching.ts
@@ -469,58 +469,91 @@ function narrowTypeBasedOnMappingPattern(
             return combineTypes(mappingInfo.filter((m) => !m.isDefinitelyMapping).map((m) => m.subtype));
         }
 
-        if (pattern.d.entries.length !== 1 || pattern.d.entries[0].nodeType !== ParseNodeType.PatternMappingKeyEntry) {
-            return type;
-        }
-
         // Handle the case where the type is a union that includes a TypedDict with
-        // a field discriminated by a literal.
-        const keyPattern = pattern.d.entries[0].d.keyPattern;
-        const valuePattern = pattern.d.entries[0].d.valuePattern;
-        if (
-            keyPattern.nodeType !== ParseNodeType.PatternLiteral ||
-            valuePattern.nodeType !== ParseNodeType.PatternAs ||
-            !valuePattern.d.orPatterns.every((orPattern) => orPattern.nodeType === ParseNodeType.PatternLiteral)
-        ) {
-            return type;
+        // fields discriminated by literals or match patterns.
+        const keyEntryInfos: { keyValue: string; valueTypes?: Type[] }[] = [];
+        let hasUnsupportedKeyPattern = false;
+
+        for (const entry of pattern.d.entries) {
+            if (entry.nodeType === ParseNodeType.PatternMappingExpandEntry) {
+                continue;
+            }
+            if (entry.nodeType !== ParseNodeType.PatternMappingKeyEntry) {
+                hasUnsupportedKeyPattern = true;
+                break;
+            }
+
+            const keyPattern = entry.d.keyPattern;
+            if (keyPattern.nodeType !== ParseNodeType.PatternLiteral) {
+                hasUnsupportedKeyPattern = true;
+                break;
+            }
+
+            const keyType = evaluator.getTypeOfExpression(keyPattern.d.expr).type;
+            if (
+                !isClassInstance(keyType) ||
+                !ClassType.isBuiltIn(keyType, 'str') ||
+                keyType.priv.literalValue === undefined
+            ) {
+                hasUnsupportedKeyPattern = true;
+                break;
+            }
+            const keyValue = keyType.priv.literalValue as string;
+
+            const valuePattern = entry.d.valuePattern;
+            let valueTypes: Type[] | undefined;
+
+            if (
+                valuePattern.nodeType === ParseNodeType.PatternAs &&
+                valuePattern.d.orPatterns.every((orPattern) => orPattern.nodeType === ParseNodeType.PatternLiteral)
+            ) {
+                valueTypes = valuePattern.d.orPatterns.map(
+                    (orPattern) => evaluator.getTypeOfExpression((orPattern as PatternLiteralNode).d.expr).type
+                );
+            } else if (
+                valuePattern.nodeType === ParseNodeType.PatternAs &&
+                valuePattern.d.orPatterns.length === 1 &&
+                (valuePattern.d.orPatterns[0].nodeType === ParseNodeType.PatternCapture ||
+                    valuePattern.d.orPatterns[0].nodeType === ParseNodeType.PatternValue)
+            ) {
+                valueTypes = undefined;
+            } else {
+                hasUnsupportedKeyPattern = true;
+                break;
+            }
+
+            keyEntryInfos.push({ keyValue, valueTypes });
         }
 
-        const keyType = evaluator.getTypeOfExpression(keyPattern.d.expr).type;
-
-        // The key type must be a str literal.
-        if (
-            !isClassInstance(keyType) ||
-            !ClassType.isBuiltIn(keyType, 'str') ||
-            keyType.priv.literalValue === undefined
-        ) {
+        if (hasUnsupportedKeyPattern || keyEntryInfos.length === 0) {
             return type;
         }
-        const keyValue = keyType.priv.literalValue as string;
-
-        const valueTypes = valuePattern.d.orPatterns.map(
-            (orPattern) => evaluator.getTypeOfExpression((orPattern as PatternLiteralNode).d.expr).type
-        );
 
         return mapSubtypes(type, (subtype) => {
             if (isClassInstance(subtype) && ClassType.isTypedDictClass(subtype)) {
                 const typedDictMembers = getTypedDictMembersForClass(evaluator, subtype, /* allowNarrowed */ true);
-                const member = typedDictMembers.knownItems.get(keyValue);
 
-                if (member && (member.isRequired || member.isProvided) && isClassInstance(member.valueType)) {
-                    const memberValueType = member.valueType;
-
-                    // If there's at least one literal value pattern that matches
-                    // the literal type of the member, we can eliminate this type.
-                    if (
-                        valueTypes.some(
-                            (valueType) =>
-                                isClassInstance(valueType) &&
-                                ClassType.isSameGenericClass(valueType, memberValueType) &&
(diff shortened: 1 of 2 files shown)


StellaHuang95:

packages/pyright-internal/src/analyzer/patternMatching.ts
```diff
@@ -469,58 +469,91 @@ function narrowTypeBasedOnMappingPattern(
...
+            ) {
+                valueTypes = valuePattern.d.orPatterns.map(
+                    (orPattern) => evaluator.getTypeOfExpression((orPattern as PatternLiteralNode).d.expr).type
+                );
+            } else if (
+                valuePattern.nodeType === ParseNodeType.PatternAs &&
+                valuePattern.d.orPatterns.length === 1 &&
+                (valuePattern.d.orPatterns[0].nodeType === ParseNodeType.PatternCapture ||
+                    valuePattern.d.orPatterns[0].nodeType === ParseNodeType.PatternValue)
+            ) {
```
StellaHuang95, line 517: > **Issue** · Please address or respond

`PatternValue` is grouped with `PatternCapture` and therefore treated as an unconditional match (`valueTypes = undefined`). A value pattern such as `Color.RED` is an equality check, so a TypedDict whose required field contains `Color.BLUE` can fall through but is incorrectly eliminated here. Only treat captures as presence-only, or conservatively leave value patterns unsupported until their value can be compared soundly.

packages/pyright-internal/src/tests/samples/matchMapping1.py
```diff
@@ -150,5 +150,26 @@ def test_not_required_narrowing(subj: TD1) -> None:
...
+    kind: Literal["b"]
+    data_b: str
+
+
+def test_negative_narrowing3(msg: MsgA | MsgB) -> None:
+    match msg:
+        case {"v": 1, "kind": "a"}:
+            reveal_type(msg, expected_text="MsgA")
+        case _:
+            reveal_type(msg, expected_text="MsgB")
```
StellaHuang95, line 176: > **Warning** · Non-blocking recommendation

This test covers only multi-key literal matching. Add an enum/value-pattern regression that asserts the TypedDict is retained in the fallthrough branch; it would catch the unsound `PatternValue` narrowing above. Coverage for the new capture-value path would also lock in its intended behavior.

hsusul:
Thank you @StellaHuang95 for the review!

I have updated the PR (commit `08e9923d2`):
1. **Excluded PatternValue from presence-only matching**: Removed `ParseNodeType.PatternValue` from the wildcard/capture check in `patternMatching.ts`. Only `ParseNodeType.PatternCapture` is treated as a presence-only match. Value patterns (such as `Color.RED`) fall through to `hasUnsupportedKeyPattern = true`, preventing unsound elimination.
2. **Added value & capture pattern regression tests**: Added tests in `matchMapping1.py` covering:
   - Enum/value patterns (`case {"color": Color.RED}:`), verifying that `BlueMsg` is retained in the fallthrough branch.
   - Capture variable patterns (`case {"v": 1, "kind": x}:`), verifying capture narrowing and fallthrough exhaustiveness.

StellaHuang95 (approved):
Approved via Review Center.

rchiodo (approved):
Approved via Review Center.

rchiodo (approved):
Approved via Review Center.

rchiodo:

packages/pyright-internal/src/analyzer/patternMatching.ts
```diff
@@ -469,58 +469,90 @@ function narrowTypeBasedOnMappingPattern(
...
-                    ) {
-                        return undefined;
+                    if (entryInfo.valueTypes === undefined) {
+                        return true;
                     }
+
+                    const memberValueType = member.valueType;
+                    return entryInfo.valueTypes.some(
+                        (valueType) =>
+                            isClassInstance(valueType) &&
```
rchiodo, line 548: > **Warning** · Non-blocking recommendation

Capture entries are checked only after `isClassInstance(member.valueType)`, so a required field typed as `int | str`, `Any`, or another non-class type may incorrectly prevent elimination even though a capture accepts every value. Check `valueTypes === undefined` before requiring a class instance, and add a reveal-type regression using a union-typed captured field.

[verified]

packages/pyright-internal/src/analyzer/patternMatching.ts
```diff
@@ -469,58 +469,90 @@ function narrowTypeBasedOnMappingPattern(
...
-        }
+        // fields discriminated by literals or match patterns.
+        const keyEntryInfos: { keyValue: string; valueTypes?: Type[] }[] = [];
+        let hasUnsupportedKeyPattern = false;
 
-        const keyType = evaluator.getTypeOfExpression(keyPattern.d.expr).type;
+        for (const entry of pattern.d.entries) {
+            if (entry.nodeType === ParseNodeType.PatternMappingExpandEntry) {
+                continue;
+            }
```
rchiodo, line 480: > **Info** · Optional note

The new `PatternMappingExpandEntry` path is not directly covered. Add a multi-key case containing `**rest` to demonstrate that ignoring the expansion preserves correct negative narrowing.

[verified]

rchiodo (approved):
Approved via Review Center.

heejaechang (approved):
Approved via Review Center.

packages/pyright-internal/src/analyzer/patternMatching.ts
```diff
@@ -469,58 +469,90 @@ function narrowTypeBasedOnMappingPattern(
...
+        }
 
         return mapSubtypes(type, (subtype) => {
             if (isClassInstance(subtype) && ClassType.isTypedDictClass(subtype)) {
                 const typedDictMembers = getTypedDictMembersForClass(evaluator, subtype, /* allowNarrowed */ true);
-                const member = typedDictMembers.knownItems.get(keyValue);
 
-                if (member && (member.isRequired || member.isProvided) && isClassInstance(member.valueType)) {
-                    const memberValueType = member.valueType;
+                const matchesAllEntries = keyEntryInfos.every((entryInfo) => {
```
rchiodo, line 535: > **Warning** · Non-blocking recommendation

Capture entries should match any value once the key is known to be present, but the class-instance guard runs before `valueTypes === undefined`, potentially retaining `TypedDict`s with union, `Any`, or other non-class member types. Check captures before requiring `isClassInstance`, and add a regression case using a required `str | int` field.

rchiodo:

heejaechang:
### 🔴 Pyright CLI QA — `08e9923d` — verdict: red

10 focused check(s) completed.

**`main.py`**

```python
from typing import Literal, TypedDict, assert_type, reveal_type

class Reading(TypedDict):
    kind: Literal["reading"]
    value: float

def check(value: Reading | None) -> None:
    match value:
        case {"kind": "reading", "value": 1.5}:
            pass
        case _:
            reveal_type(value)
            assert_type(value, Reading | None)

check({"kind": "reading", "value": 2.5})
```

**`pyrightconfig.json`**

```json
{
  "include": [
    "."
  ],
  "pythonVersion": "3.13",
  "reportUnnecessaryComparison": "error",
  "reportMatchNotExhaustive": "error"
}
```

```text
```

**Base `e8af042e988f9b06b0c6da77b2269a8bd73e5d8d`**

```json
{
  "exit_code": 0,
  "diagnostics": [
    {
      "file": "main.py",
      "severity": "information",
      "rule": null,
      "message": "Type of \"value\" is \"Reading | None\"",
      "range": {
        "start": [
          11,
          24
        ],
        "end": [
          11,
          29
        ]
      }
    }
  ]
}
```

**PR `08e9923d23adad5f8f833c0931ac9853e3af4084`**

```json
{
  "exit_code": 1,
  "diagnostics": [
    {
      "file": "main.py",
      "severity": "information",
      "rule": null,
      "message": "Type of \"value\" is \"None\"",
      "range": {
        "start": [
          11,
          24
        ],
        "end": [
          11,
          29
        ]
      }
    },
    {
      "file": "main.py",
      "severity": "error",
      "message": "\"assert_type\" mismatch: expected \"Reading | None\" but received \"None\"",
      "range": {
        "start": [
          12,
          24
        ],
        "end": [
          12,
          29
        ]
      }
    }
  ]
}
```

</details>

PR #11614, opened 2026-08-10
Disallow assignment expressions in function default parameter values

hsusul:
### Summary

PEP 572 explicitly disallows assignment expressions (walrus operator `:=`) inside function default argument values:

> *"An assignment expression is not allowed in a function default argument value."*

At Python runtime (Python 3.8+), using an assignment expression in a default parameter value (e.g. `def foo(a=(b := 1)): pass`) raises a `SyntaxError`:
```
SyntaxError: cannot use assignment expressions with function default value
```

Previously, Pyright failed to emit a syntax error diagnostic when assignment expressions were used in parameter default values, even when parenthesized (e.g. `a=(b := 1)`), and mistakenly bound the target symbol `b` into the enclosing outer scope.

### Behavior Change

Pyright now correctly disallows assignment expressions inside function default parameter values during parsing and emits a `reportSyntaxError` diagnostic.

### Implementation Details

In `packages/pyright-internal/src/parser/parser.ts`, the parameter default argument value parser (`paramNode.d.defaultValue = this._parseTestExpression(...)`) is now wrapped in `this._disallowAssignmentExpression(...)`. This ensures `this._assignmentExpressionsAllowed` is set to `false` during default value evaluation, correctly catching both unparenthesized and parenthesized assignment expressions.

### Validation

- Added new sample regression test `assignmentExpr10.py` and test case `AssignmentExpr10` in `typeEvaluator1.test.ts`.
- Ran full test suite in `packages/pyright-internal` (`npm run test`): 24 test suites passed (798 passed, 0 failed).
- Ran repository check script (`npm run check`): 4 packages typechecked cleanly with 0 errors.
- Ran `git diff --check`: 0 issues found.

### DIFF
--- a/packages/pyright-internal/src/parser/parser.ts
+++ b/packages/pyright-internal/src/parser/parser.ts
@@ -2325,12 +2325,14 @@ export class Parser {
         }
 
         if (this._consumeTokenIfOperator(OperatorType.Assign)) {
-            paramNode.d.defaultValue = this._parseTestExpression(/* allowAssignmentExpression */ false);
-            paramNode.d.defaultValue.parent = paramNode;
-            extendRange(paramNode, paramNode.d.defaultValue);
+            this._disallowAssignmentExpression(() => {
+                paramNode.d.defaultValue = this._parseTestExpression(/* allowAssignmentExpression */ false);
+            });
+            paramNode.d.defaultValue!.parent = paramNode;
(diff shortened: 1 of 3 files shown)


hsusul:
Closing invalid PR after verifying python runtime behavior for parenthesized walrus in default parameter values.

PR #11594, opened 2026-08-06
Support version-conditional TypedDict fields

rchiodo:
## Summary
- allow statically evaluable `if`/`elif`/`else` statements in class-syntax TypedDict bodies as required by the typing spec
- reuse the binder's static-condition result so only the reachable branch contributes fields and is validated
- continue rejecting dynamically evaluated conditions and non-`if` control flow in TypedDict bodies

## Tests
- `npx jest typeEvaluator7.test -t "TypedDict2[67]" --forceExit`
- `npx jest typeEvaluator5.test typeEvaluator7.test typeEvaluator8.test -t "TypedDict|Conditional1|StaticExpression" --forceExit`
- `npm run build` (`packages/pyright-internal`)
- `npm run check:eslint -- --quiet`
- `npm run check:prettier`

PR #11617, opened 2026-08-10
Expand conditional TypedDict test coverage

rchiodo:
## Summary

Follow-up to merged #11594.

- add real-world positive coverage for Python version, platform, `os.name`, `TYPE_CHECKING`, and configured-constant field gates
- cover mutually exclusive and nested branches, aliases/indexing, forward references, generic fields, inheritance, totality, `Required`, `NotRequired`, and `ReadOnly`
- add counterexamples for runtime-unknown conditions, unsupported control flow, active assignments/methods, duplicate active fields, and ignored statically unreachable invalid code
- reanalyze equivalent conditions across classes and configurations to guard branch-order and cache contamination

No production bug was exposed; this is a tests-only follow-up.

## Validation

- `npm run test:norebuild -- --runInBand typeEvaluator7.test -t "TypedDict(26|27|28|29)"`
- `npm run test:norebuild -- --runInBand typeEvaluator7.test typeEvaluator8.test -t "TypedDict|StaticExpression"`
- `npm run build` (`packages/pyright-internal`)
- `npm run typecheck`
- ESLint on `typeEvaluator7.test.ts`
- Prettier and `git diff --check`

### DIFF
--- a/packages/pyright-internal/src/tests/samples/typedDict28.py
+++ b/packages/pyright-internal/src/tests/samples/typedDict28.py
@@ -8,6 +8,7 @@ from typing import Generic, NotRequired, Required, TYPE_CHECKING, TypeVar, Typed
 
 from typing_extensions import ReadOnly  # pyright: ignore[reportMissingModuleSource]
 
+# The test configuration overrides these values to select the enabled branches below.
 ENABLE_FAST_PATH = False
 FEATURE_SET = "legacy"
 _T = TypeVar("_T")
@@ -115,8 +116,10 @@ class ReverseOrder(TypedDict):
         value: str
 
 
-ForwardOrder(value="")
-ReverseOrder(value="")
+forward_order = ForwardOrder(value="")
+reverse_order = ReverseOrder(value="")
+assert_type(forward_order["value"], str)
+assert_type(reverse_order["value"], str)
 
 # These should generate errors. Equivalent conditions in different orders must select the same type.
 ForwardOrder(value=b"")
--- a/packages/pyright-internal/src/tests/samples/typedDict28.py
+++ b/packages/pyright-internal/src/tests/samples/typedDict28.py
@@ -1,0 +1,190 @@
+# This sample tests realistic static conditions and counterexamples in TypedDict definitions.
+
+import contextlib
+import os
+import sys as _sys
+import typing as _typing
+from typing import Generic, NotRequired, Required, TYPE_CHECKING, TypeVar, TypedDict, assert_type
+
+from typing_extensions import ReadOnly  # pyright: ignore[reportMissingModuleSource]
+
+ENABLE_FAST_PATH = False
+FEATURE_SET = "legacy"
+_T = TypeVar("_T")
+
+
+class CompatibilityBase(TypedDict, total=False):
+    inherited: int
+
+
+class CompatibilityMetadata(CompatibilityBase, total=False):
+    optional: NotRequired[str]
+
+    if _sys.version_info >= (3, 13):
+        minor_version: Required[str]
+        payload: Required["Payload"]
+
+    if _sys.version_info[0] == 3:
+        major_version: Required[int]
+
+    if _sys.platform == "linux":
+        platform_name: Required[str]
+
+    if os.name == "posix":
+        os_name: Required[str]
+
+    if TYPE_CHECKING:
+        checker_only: Required[bool]
+
+    if _typing.TYPE_CHECKING:
+        qualified_checker_only: Required[bytes]
+
+    if ENABLE_FAST_PATH:
+        fast_path: Required[float]
+
+    if FEATURE_SET == "modern":
+        feature_set: Required[str]
+
+    if _sys.version_info >= (3, 12):
+        if _sys.platform == "linux":
+            nested: Required[tuple[int, str]]
+
+
+class Payload:
+    pass
+
+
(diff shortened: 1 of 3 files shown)


StellaHuang95:

packages/pyright-internal/src/tests/typeEvaluator7.test.ts
```diff
@@ -808,6 +808,39 @@ test('TypedDict27', () => {
...
+test('TypedDict28', () => {
+    const configOptions = new ConfigOptions(Uri.empty());
+    configOptions.defaultPythonVersion = pythonVersion3_13;
+    configOptions.defaultPythonPlatform = 'Linux';
+    configOptions.defineConstant.set('ENABLE_FAST_PATH', true);
+    configOptions.defineConstant.set('FEATURE_SET', 'modern');
+
+    const analysisResults = TestUtils.typeAnalyzeSampleFiles(['typedDict28.py'], configOptions);
+
+    TestUtils.validateResults(analysisResults, 13);
```
StellaHuang95, line 820: > **Warning** · Non-blocking recommendation

`validateResults` checks only the aggregate diagnostic count, so the Forward/Reverse branch-order cases can cancel out: a wrong selected type can add an error for the string call while removing one for the bytes call. Add positive `assert_type` assertions on valid instances to directly verify the selected field type.

packages/pyright-internal/src/tests/samples/typedDict28.py
```diff
@@ -0,0 +1,190 @@
...
+
+import contextlib
+import os
+import sys as _sys
+import typing as _typing
+from typing import Generic, NotRequired, Required, TYPE_CHECKING, TypeVar, TypedDict, assert_type
+
+from typing_extensions import ReadOnly  # pyright: ignore[reportMissingModuleSource]
+
+ENABLE_FAST_PATH = False
```
StellaHuang95, line 12: > **Info** · Optional note

These source values are intentionally overridden by `defineConstant` in the test configuration. Add a brief comment so the `fast_path` and `feature_set` branches are not mistaken for inactive sample code.

StellaHuang95 (approved):
Approved via Review Center.

packages/pyright-internal/src/tests/typeEvaluator7.test.ts
rchiodo: Fixed — the valid ForwardOrder and ReverseOrder instances are now named and each field is checked with assert_type(..., str), so incorrect branch selection can no longer be hidden by aggregate diagnostic-count cancellation.

packages/pyright-internal/src/tests/samples/typedDict28.py
rchiodo: Fixed — added a comment explaining that the test configuration overrides ENABLE_FAST_PATH and FEATURE_SET to select the enabled branches below.

packages/pyright-internal/src/tests/samples/typedDict28.py
```diff
@@ -0,0 +1,193 @@
...
+
+    with contextlib.nullcontext():
+        with_field: int
+
+    match 1:
+        case 1:
+            match_field: int
+
+
+class InvalidActiveStatements(TypedDict):
```
StellaHuang95, line 165: > **Warning** · Non-blocking recommendation

The five control-flow cases are validated only through the aggregate error count. A changed diagnostic distribution (for example, `try` producing two diagnostics while another case produces none) could still total 13. Please isolate or otherwise document the expected diagnostic for each compound statement so these cases remain independently covered.

StellaHuang95 (approved):
Approved via Review Center.

packages/pyright-internal/src/tests/samples/typedDict28.py
rchiodo: Fixed — each invalid compound statement now has its own expected-error comment naming the prohibited construct, so the aggregate count is explicitly tied to independent while, for, try, with, and match coverage.

packages/pyright-internal/src/tests/samples/typedDict28.py
```diff
@@ -0,0 +1,197 @@
...
+
+forward_order = ForwardOrder(value="")
+reverse_order = ReverseOrder(value="")
+assert_type(forward_order["value"], str)
+assert_type(reverse_order["value"], str)
+
+# These should generate errors. Equivalent conditions in different orders must select the same type.
+ForwardOrder(value=b"")
+ReverseOrder(value=b"")
+
```
StellaHuang95, line 128: > **Warning** · Non-blocking recommendation

The aggregate error count can still mask an incorrect selected branch here: choosing `int` under the 3.13 configuration would make the string call fail and the integer call pass, leaving the total unchanged. Bind the valid construction and assert its field type (for example, `assert_type(selected["value"], str)`) to pin the selected branch.

StellaHuang95 (approved):
Approved via Review Center.

packages/pyright-internal/src/tests/samples/typedDict28.py
rchiodo: Added a positive `assert_type` for `SelectedWireValue["value"]`, so the test now pins the Python 3.13 branch to `str` independently of the aggregate diagnostic count.

StellaHuang95 (approved):
Approved via Review Center.

PR #11619, opened 2026-08-11
Fix type narrowing for type(x) expressions with type[T] parameters and reverse comparisons

hsusul:
## Summary

Fixes a type narrowing issue where binary comparison type guards involving `type(x)` expressions failed to narrow types when:
1. `type(x)` appeared on the right-hand side of the comparison (e.g. `if cls is type(x):` or `if int is type(x):`).
2. The comparison operand was a variable or parameter of type `type[T]` (e.g. `cls: type[int]`).

## Root Cause
1. Pyright's binary type guard logic for `type(x)` previously checked only `testExpression.d.leftExpr.nodeType === ParseNodeType.Call`. When `type(x)` was on the RHS, Pyright skipped generating a type guard callback.
2. When comparing `type(x)` against a variable `cls` of type `type[T]`, `getTypeOfExpression(cls)` returns a `ClassType` instance of `type`. Pyright checked `isInstantiableClass(expandedSubtype)`, which returned `false` for `type[T]` instances, preventing `classTypes` from being populated.
3. Positive narrowing in `narrowTypeForTypeIs` did not eliminate disjoint subtypes when comparing against class types derived from `type[T]`.

## Changes
- Supported `type(x)` on either `leftExpr` or `rightExpr` of `is`, `is not`, `==`, and `!=` binary comparison operators.
- Unwrapped `type[T]` instances (`ClassType` built-in `type` with `typeArgs`) to instantiable class types (`convertToInstantiable(typeArgs[0], false)`).
- Updated positive narrowing logic in `narrowTypeForTypeIs` to check subclass overlap and eliminate disjoint subtypes.
- Excluded `type[object]` operands to preserve correct behavior when comparing against generic `object` instances.
- Added regression tests in `typeIs6.py` covering all `type(x)` comparison forms, `type[T]` parameters, `@final` classes, and union types.

### DIFF
--- a/packages/pyright-internal/src/analyzer/typeGuards.ts
+++ b/packages/pyright-internal/src/analyzer/typeGuards.ts
@@ -13,6 +13,7 @@ import { assert } from '../common/debug';
 import {
     ArgCategory,
     AssignmentExpressionNode,
+    CallNode,
     ExpressionNode,
     isExpressionNode,
     NameNode,
@@ -209,45 +210,76 @@ export function getTypeNarrowingCallback(
                 }
             }
 
-            // Look for "type(X) is Y", "type(X) is not Y", "type(X) == Y" or "type(X) != Y".
+            // Look for "type(X) is Y", "type(X) is not Y", "type(X) == Y" or "type(X) != Y",
+            // as well as reverse forms "Y is type(X)", "Y is not type(X)", "Y == type(X)", "Y != type(X)".
+            let typeCallNode: CallNode | undefined;
+            let otherExpr: ExpressionNode | undefined;
+
             if (testExpression.d.leftExpr.nodeType === ParseNodeType.Call) {
-                if (
-                    testExpression.d.leftExpr.d.args.length === 1 &&
-                    testExpression.d.leftExpr.d.args[0].d.argCategory === ArgCategory.Simple
-                ) {
-                    const arg0Expr = testExpression.d.leftExpr.d.args[0].d.valueExpr;
-                    if (isMatchingExpressionOrWalrusRhs(evaluator, reference, arg0Expr)) {
-                        const callType = evaluator.getTypeOfExpression(
-                            testExpression.d.leftExpr.d.leftExpr,
-                            EvalFlags.CallBaseDefaults
-                        ).type;
+                typeCallNode = testExpression.d.leftExpr;
+                otherExpr = testExpression.d.rightExpr;
+            } else if (testExpression.d.rightExpr.nodeType === ParseNodeType.Call) {
+                typeCallNode = testExpression.d.rightExpr;
+                otherExpr = testExpression.d.leftExpr;
+            }
 
-                        if (isInstantiableClass(callType) && ClassType.isBuiltIn(callType, 'type')) {
-                            const rhsResult = evaluator.getTypeOfExpression(testExpression.d.rightExpr);
-                            const classTypes: ClassType[] = [];
-                            let isClassType = true;
+            if (
+                typeCallNode &&
+                otherExpr &&
+                typeCallNode.d.args.length === 1 &&
+                typeCallNode.d.args[0].d.argCategory === ArgCategory.Simple
+            ) {
+                const arg0Expr = typeCallNode.d.args[0].d.valueExpr;
+                if (isMatchingExpressionOrWalrusRhs(evaluator, reference, arg0Expr)) {
+                    const callType = evaluator.getTypeOfExpression(
+                        typeCallNode.d.leftExpr,
+                        EvalFlags.CallBaseDefaults
+                    ).type;
 
-                            evaluator.mapSubtypesExpandTypeVars(
-                                rhsResult.type,
-                                /* options */ undefined,
-                                (expandedSubtype) => {
-                                    if (isInstantiableClass(expandedSubtype)) {
-                                        classTypes.push(expandedSubtype);
-                                    } else {
-                                        isClassType = false;
+                    if (isInstantiableClass(callType) && ClassType.isBuiltIn(callType, 'type')) {
+                        const otherResult = evaluator.getTypeOfExpression(otherExpr);
+                        const classTypes: ClassType[] = [];
+                        let isClassType = true;
+
+                        evaluator.mapSubtypesExpandTypeVars(
+                            otherResult.type,
+                            /* options */ undefined,
+                            (expandedSubtype) => {
+                                let instantiable: ClassType | undefined;
+                                if (isClass(expandedSubtype)) {
+                                    if (
+                                        ClassType.isBuiltIn(expandedSubtype, 'type') &&
+                                        expandedSubtype.priv.typeArgs &&
+                                        expandedSubtype.priv.typeArgs.length > 0
+                                    ) {
+                                        const extracted = convertToInstantiable(
+                                            expandedSubtype.priv.typeArgs[0],
+                                            /* includeSubclasses */ false
+                                        );
+                                        if (isInstantiableClass(extracted)) {
+                                            instantiable = extracted;
+                                        }
+                                    } else if (isInstantiableClass(expandedSubtype)) {
+                                        instantiable = expandedSubtype;
                                     }
-                                    return undefined;
                                 }
-                            );
 
-                            if (isClassType && classTypes.length > 0) {
-                                return (type: Type) => {
-                                    return {
-                                        type: narrowTypeForTypeIs(evaluator, type, classTypes, adjIsPositiveTest),
-                                        isIncomplete: !!rhsResult.isIncomplete,
-                                    };
-                                };
+                                if (instantiable && !ClassType.isBuiltIn(instantiable, 'object')) {
+                                    classTypes.push(ClassType.cloneIncludeSubclasses(instantiable, false));
+                                } else {
+                                    isClassType = false;
+                                }
+                                return undefined;
                             }
+                        );
+
+                        if (isClassType && classTypes.length > 0) {
+                            return (type: Type) => {
+                                return {
+                                    type: narrowTypeForTypeIs(evaluator, type, classTypes, adjIsPositiveTest),
+                                    isIncomplete: !!otherResult.isIncomplete,
+                                };
+                            };
                         }
                     }
                 }
@@ -2538,17 +2570,20 @@ function narrowTypeForTypeIs(evaluator: TypeEvaluator, type: Type, classTypes: C
             /* options */ undefined,
             (subtype: Type, unexpandedSubtype: Type) => {
                 if (isClassInstance(subtype)) {
-                    const matches = ClassType.isDerivedFrom(classType, ClassType.cloneAsInstantiable(subtype));
+                    const instantiableSubtype = ClassType.cloneAsInstantiable(subtype);
+                    const matches = ClassType.isDerivedFrom(classType, instantiableSubtype);
+                    const isSubclass = ClassType.isDerivedFrom(instantiableSubtype, classType);
+
                     if (isPositiveTest) {
                         if (matches) {
-                            if (ClassType.isSameGenericClass(ClassType.cloneAsInstantiable(subtype), classType)) {
+                            if (ClassType.isSameGenericClass(instantiableSubtype, classType)) {
                                 return addConditionToType(subtype, getTypeCondition(classType));
                             }
 
                             return addConditionToType(ClassType.cloneAsInstance(classType), subtype.props?.condition);
                         }
 
-                        if (!classType.priv.includeSubclasses) {
+                        if (!classType.priv.includeSubclasses || !isSubclass) {
                             return undefined;
                         }
 
@@ -2560,7 +2595,7 @@ function narrowTypeForTypeIs(evaluator: TypeEvaluator, type: Type, classTypes: C
                     if (!classType.priv.includeSubclasses) {
                         // If the class if marked final and it matches, then
                         // we can eliminate it in the negative case.
-                        if (matches && ClassType.isFinal(subtype)) {
+                        if (matches && (ClassType.isFinal(classType) || ClassType.isFinal(instantiableSubtype))) {
                             return undefined;
                         }
 
--- a/packages/pyright-internal/src/analyzer/typeGuards.ts
+++ b/packages/pyright-internal/src/analyzer/typeGuards.ts
@@ -246,15 +246,18 @@ export function getTypeNarrowingCallback(
                             /* options */ undefined,
                             (expandedSubtype) => {
                                 let instantiable: ClassType | undefined;
+                                let isTypeParam = false;
+
                                 if (isClass(expandedSubtype)) {
                                     if (
                                         ClassType.isBuiltIn(expandedSubtype, 'type') &&
                                         expandedSubtype.priv.typeArgs &&
                                         expandedSubtype.priv.typeArgs.length > 0
                                     ) {
+                                        isTypeParam = true;
                                         const extracted = convertToInstantiable(
                                             expandedSubtype.priv.typeArgs[0],
-                                            /* includeSubclasses */ false
+                                            /* includeSubclasses */ true
                                         );
                                         if (isInstantiableClass(extracted)) {
                                             instantiable = extracted;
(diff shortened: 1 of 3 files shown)


StellaHuang95:

StellaHuang95:
Negative `type(...) is not ...` narrowing can eliminate valid supertypes, producing unsound results. The PR needs this corrected and covered by a regression test.

StellaHuang95:
The negative narrowing change can incorrectly eliminate a valid superclass subtype when the compared class is a final subclass, producing unsound results.

StellaHuang95:
The negative narrowing change can incorrectly eliminate a non-final base class when the compared class is a final subclass, producing an unsound narrowed type.

hsusul:
Thank you for the review and sharp catch @StellaHuang95!

I have updated the negative narrowing logic in `narrowTypeForTypeIs` to check `ClassType.isFinal(instantiableSubtype)` rather than `ClassType.isFinal(classType)`. This ensures that a non-final base class (such as `Base`) is preserved when comparing `type(x) is not FinalSub` against a `@final` subclass (`FinalSub`), preserving sound type narrowing when `x` is an instance of `Base`.

I have pushed the fix and added a regression test (`func8`) in `typeIs6.py` covering this exact pattern:
```python
class Base: pass

@final
class FinalSub(Base): pass

def func8(x: Base):
    if type(x) is not FinalSub:
        reveal_type(x, expected_text="Base")
    else:
        reveal_type(x, expected_text="FinalSub")
```

StellaHuang95:
GitHub cannot anchor PR review comments to unchanged lines in the diff. Falling back to a general PR comment for [packages/pyright-internal/src/analyzer/typeGuards.ts:L2626](https://github.com/microsoft/pyright/blob/0d980d43ea441d7dfd5eba19e798f601fb858ea6/packages/pyright-internal/src/analyzer/typeGuards.ts#L2626).

> **Info** · Optional note

`classType` is always created with `includeSubclasses` set to `false` at its sole call site, so the first disjunct always short-circuits and `isSubclass` cannot affect narrowing. Please remove this inert condition and computation, or add a reachable subclass-inclusive case with a comment explaining the intended future behavior.

packages/pyright-internal/src/analyzer/typeGuards.ts
```diff
@@ -209,45 +210,76 @@ export function getTypeNarrowingCallback(
...
+                        if (isClassType && classTypes.length > 0) {
+                            return (type: Type) => {
+                                return {
+                                    type: narrowTypeForTypeIs(evaluator, type, classTypes, adjIsPositiveTest),
+                                    isIncomplete: !!otherResult.isIncomplete,
+                                };
+                            };
                         }
                     }
                 }
```
StellaHuang95, line 292: > **Warning** · Non-blocking recommendation

The blanket `object` exclusion also changes direct `type(x) is object` checks from narrowing the positive branch to `Never` to performing no narrowing. It also disables the entire callback for `type[int] | type[object]`, losing the useful `int` narrowing. Please scope the exclusion to the unwrapped `type[object]` case, or add tests documenting that this broader precision loss is intentional.

packages/pyright-internal/src/tests/samples/typeIs6.py
```diff
@@ -0,0 +1,60 @@
...
+class Base: pass
+
+@final
+class FinalSub(Base): pass
+
+def func8(x: Base):
+    if type(x) is not FinalSub:
+        reveal_type(x, expected_text="Base")
+    else:
+        reveal_type(x, expected_text="FinalSub")
```
StellaHuang95, line 60: > **Warning** · Non-blocking recommendation

Please add product-path regression coverage under `packages/pylance-internal/src/tests` for this shared narrowing behavior. Also add the negative case `x: int | str; if type(x) is bool: ... else: reveal_type(x, expected_text="int | str")` to guard against incorrectly eliminating a valid supertype when the compared class is a final subclass.

StellaHuang95 (dismissed):
Approved via Review Center.

rchiodo:
I took a closer look at the `mypy_primer` differences. I don't think all of them are expected, and I would not merge this as-is.

The main issue is that the new logic converts the argument of `type[T]` with `includeSubclasses=false`. A value annotated `type[T]` can be a class object for a subclass of `T` unless `T` is final, so treating it as exactly `T` is unsound.

* **Spark:** This looks like a real regression. In `MultiIndex.symmetric_difference`, `other` is annotated as `Index`, and the code checks `if type(self) is not type(other): raise`. The continuation is reachable when `other` is a `MultiIndex`, but this PR narrows `self` to `Never`, producing the two new errors.
* **koda-validate:** The three new errors are false positives. The reverse comparison is now recognized, but narrowing loses the correlation between the constrained `ExactMatchT` and `val` after `type(self.match) == type(val)`.
* **yarl:** The removed error is an expected and desirable improvement. `if type(val) is cls: return val` should narrow `val` sufficiently for the `URL` return type.
* **SymPy:** The large cascade appears to be primer/inference noise rather than a reproducible semantic difference from this PR. I would rerun primer rather than accept or update expectations based on that block.

I recommend preserving subclass inclusion for operands typed `type[T]` (using exact-class treatment only for direct class expressions or final classes), and adding a Spark-shaped regression test that verifies the continuation does not become `Never`. It would also be useful to distinguish `Base` from `cls: type[Base]` in the tests.

hsusul:
Thank you for the detailed `mypy_primer` analysis and architectural guidance!

I have updated the binary type guard unwrapping logic in `getTypeNarrowingCallback`:
1. **Operands typed `type[T]`** (e.g. `cls: type[Index]` or `type(other)` where `other: Index`): We preserve subclass inclusion (`includeSubclasses = true`) when converting `type[T]`. This ensures that for `type(self) is not type(other)` (where `self: MultiIndex` and `other: Index`), `type(other)` can represent `MultiIndex` or any subclass of `Index`, so `self` is not unsoundly narrowed to `Never` on continuation (resolving the Spark `MultiIndex.symmetric_difference` regression).
2. **Direct Class Expressions** (e.g. `if type(x) is Base:` or `if int is type(x):`): We retain exact-class treatment (`includeSubclasses = false`).

### Added Tests in `typeIs6.py`
- **Spark Regression Test** (`test_spark_regression`): Verifies `self` remains `MultiIndex` (does not become `Never`) after `if type(self) is not type(other): pass`.
- **Direct Class vs `type[Base]` Test** (`test_direct_class_vs_type_param`): Verifies `Base` handling for direct class expressions vs `cls: type[Base]` parameters.

The update has been pushed in commit `94afc46b3`.

packages/pyright-internal/src/analyzer/typeGuards.ts
```diff
@@ -209,45 +210,83 @@ export function getTypeNarrowingCallback(
...
```
StellaHuang95, line 292: > **Warning** · Non-blocking recommendation

The `object` exclusion now also rejects the pre-existing direct `type(x) is object` form. Previously it produced an impossible positive branch (`Never`); now `isClassType` becomes false and no narrowing callback is returned. It also discards useful alternatives in unions such as `type[int] | type[object]`. Please exclude only the parameterized `type[object]` case, or skip that subtype rather than disabling the entire callback.

packages/pyright-internal/src/analyzer/typeGuards.ts
```diff
@@ -209,45 +210,83 @@ export function getTypeNarrowingCallback(
...
+                    const callType = evaluator.getTypeOfExpression(
+                        typeCallNode.d.leftExpr,
+                        EvalFlags.CallBaseDefaults
+                    ).type;
+
+                    if (isInstantiableClass(callType) && ClassType.isBuiltIn(callType, 'type')) {
+                        const otherResult = evaluator.getTypeOfExpression(otherExpr);
+                        const classTypes: ClassType[] = [];
+                        let isClassType = true;
+
```
StellaHuang95, line 243: > **Info** · Optional note

When both operands are calls, the left call is selected before checking whether its argument matches the reference. As a result, querying narrowing for `x` in `type(y) is type(x)` stops after `type(y)` fails the match and never considers the RHS `type(x)`. Please fall through to the RHS call when the selected call does not match the reference.

StellaHuang95 (approved):
Approved via Review Center.

rchiodo (approved):
Approved via Review Center.

rchiodo:

packages/pyright-internal/src/analyzer/typeGuards.ts
```diff
@@ -209,45 +210,83 @@ export function getTypeNarrowingCallback(
...
-                                    } else {
-                                        isClassType = false;
+                typeCallNode = testExpression.d.leftExpr;
+                otherExpr = testExpression.d.rightExpr;
+            } else if (testExpression.d.rightExpr.nodeType === ParseNodeType.Call) {
+                typeCallNode = testExpression.d.rightExpr;
+                otherExpr = testExpression.d.leftExpr;
+            }
+
+            if (
```
rchiodo, line 226: > **Warning** · Non-blocking recommendation

The left-first `if/else if` selection skips a valid RHS `type(reference)` when the LHS is also a call, such as `factory() is type(x)` or `type(y) is type(x)`. Check both operands independently and add a reverse call/call regression test.

packages/pyright-internal/src/analyzer/typeGuards.ts
```diff
@@ -209,45 +210,83 @@ export function getTypeNarrowingCallback(
...
+                                    ) {
+                                        isTypeParam = true;
+                                        const extracted = convertToInstantiable(
+                                            expandedSubtype.priv.typeArgs[0],
+                                            /* includeSubclasses */ true
+                                        );
+                                        if (isInstantiableClass(extracted)) {
+                                            instantiable = extracted;
+                                        }
+                                    } else if (isInstantiableClass(expandedSubtype)) {
```
rchiodo, line 265: > **Warning** · Non-blocking recommendation

Direct classes are cloned with `includeSubclasses=false`, while extracted `type[T]` classes retain subclass inclusion. Add focused tests combining both operand representations to demonstrate that their differing downstream elimination behavior is sound.

rchiodo (approved):
Approved via Review Center.

heejaechang (approved):
Approved via Review Center.

packages/pyright-internal/src/analyzer/typeGuards.ts
```diff
@@ -209,45 +210,83 @@ export function getTypeNarrowingCallback(
...
-                                        classTypes.push(expandedSubtype);
-                                    } else {
-                                        isClassType = false;
+                typeCallNode = testExpression.d.leftExpr;
+                otherExpr = testExpression.d.rightExpr;
+            } else if (testExpression.d.rightExpr.nodeType === ParseNodeType.Call) {
+                typeCallNode = testExpression.d.rightExpr;
+                otherExpr = testExpression.d.leftExpr;
+            }
+
```
rchiodo, line 225: > **Issue** · Please address or respond

When the left operand is any call, it is selected before verifying that it is `type(reference)`. As a result, `get_class() is type(x)` and `type(y) is type(x)` cannot create a guard for `x`, despite `type(x)` being on the RHS. If the selected call does not match the reference, inspect the other operand as well.

packages/pyright-internal/src/analyzer/typeGuards.ts
```diff
@@ -2538,17 +2577,20 @@ function narrowTypeForTypeIs(evaluator: TypeEvaluator, type: Type, classTypes: C
...
+                            if (ClassType.isSameGenericClass(instantiableSubtype, classType)) {
                                 return addConditionToType(subtype, getTypeCondition(classType));
                             }
 
                             return addConditionToType(ClassType.cloneAsInstance(classType), subtype.props?.condition);
                         }
 
-                        if (!classType.priv.includeSubclasses) {
+                        if (!classType.priv.includeSubclasses || !isSubclass) {
                             return undefined;
```
rchiodo, line 2594: > **Issue** · Please address or respond

This removes an `A` subtype for a positive comparison against `type[B]` whenever `A` is not a subclass of `B`, but that does not establish disjointness. A non-final `C(A, B)` can satisfy the comparison when `cls` is `C`, so this narrowing makes a valid runtime type impossible. Preserve such bases unless overlap is impossible, and add a multiple-inheritance regression test.

packages/pyright-internal/src/tests/samples/typeIs6.py
```diff
@@ -0,0 +1,76 @@
...
+    if type(self_val) is not type(other):
+        pass
+    reveal_type(self_val, expected_text="MultiIndex")
+
+def test_direct_class_vs_type_param(x: Base, cls: type[Base]):
+    if type(x) is not Base:
+        reveal_type(x, expected_text="Base")
+    
+    if type(x) is not cls:
+        reveal_type(x, expected_text="Base")
```
rchiodo, line 76: > **Warning** · Non-blocking recommendation

The regression coverage does not exercise all newly supported operand/operator paths: reversed `is not` and `==`, both `!=` orientations, two competing `type(...)` calls, literal `object`, and multiple-inheritance overlap are missing. Add these cases so the new branches and edge conditions remain protected.

rchiodo:

heejaechang:
### 🔴 Pyright CLI QA — `94afc46b` — verdict: red

13 focused check(s) completed.

**`main.py`**

```python

def check(value: bool, kind: type[int]) -> int:
    if type(value) is kind:
        reveal_type(value)
        return "not an int"
    return 0

check(True, bool)
```

**`pyrightconfig.json`**

```json
{"include": ["main.py"], "pythonVersion": "3.13", "typeCheckingMode": "standard", "reportUnreachable": "warning"}
```

```text
```

**Base `eefd22541da12c4d5e7703c4ab9d1561131cacd9`**

```json
{
  "exit_code": 1,
  "diagnostics": [
    {
      "file": "main.py",
      "severity": "error",
      "rule": "reportReturnType",
      "message": "Type \"Literal['not an int']\" is not assignable to return type \"int\"\n  \"Literal['not an int']\" is not assignable to \"int\"",
      "range": {
        "start": [
          5,
          15
        ],
        "end": [
          5,
          27
        ]
      }
    },
    {
      "file": "main.py",
      "severity": "information",
      "rule": null,
      "message": "Type of \"value\" is \"bool\"",
      "range": {
        "start": [
          4,
          20
        ],
        "end": [
          4,
          25
        ]
      }
    }
  ]
}
```

**PR `94afc46b3c36cd7eb40c6723e15b9e90a6168cd6`**

```json
{
  "exit_code": 0,
  "diagnostics": [
    {
      "file": "main.py",
      "severity": "warning",
      "rule": "reportUnreachable",
      "message": "Type analysis indicates code is unreachable",
      "range": {
        "start": [
          4,
          8
        ],
        "end": [
          4,
          19
        ]
      }
    }
  ]
}
```

</details>

#### Needs review: `reportReturnType` newly reported

The PR reports `reportReturnType` for this example; the base does not. Is this change intended?

**`main.py`**

```python

def require_exact[T](value: object, kind: type[T]) -> T:
    if type(value) is kind:
        reveal_type(value)
        return value
    raise TypeError

reveal_type(require_exact(1, int))
```

**`pyrightconfig.json`**

```json
{"include": ["main.py"], "pythonVersion": "3.13", "typeCheckingMode": "standard", "reportUnreachable": "warning"}
```

```text
```

**Base `eefd22541da12c4d5e7703c4ab9d1561131cacd9`**

```json
{
  "exit_code": 0,
  "diagnostics": [
    {
      "file": "main.py",
      "severity": "information",
      "rule": null,
      "message": "Type of \"require_exact(1, int)\" is \"int\"",
      "range": {
        "start": [
          8,
          12
        ],
        "end": [
          8,
          33
        ]
      }
    },
    {
      "file": "main.py",
      "severity": "information",
      "rule": null,
      "message": "Type of \"value\" is \"object*\"",
      "range": {
        "start": [
          4,
          20
        ],
        "end": [
          4,
          25
        ]
      }
    }
  ]
}
```

**PR `94afc46b3c36cd7eb40c6723e15b9e90a6168cd6`**

```json
{
  "exit_code": 1,
  "diagnostics": [
    {
      "file": "main.py",
      "severity": "error",
      "rule": "reportReturnType",
      "message": "Type \"object\" is not assignable to return type \"T@require_exact\"\n  Type \"object\" is not assignable to type \"T@require_exact\"",
      "range": {
        "start": [
          5,
          15
        ],
        "end": [
          5,
          20
        ]
      }
    },
    {
      "file": "main.py",
      "severity": "information",
      "rule": null,
      "message": "Type of \"require_exact(1, int)\" is \"int\"",
      "range": {
        "start": [
          8,
          12
        ],
        "end": [
          8,
          33
        ]
      }
    },
    {
      "file": "main.py",
      "severity": "information",
      "rule": null,
      "message": "Type of \"value\" is \"object\"",
      "range": {
        "start": [
          4,
          20
        ],
        "end": [
          4,
          25
        ]
      }
    }
  ]
}
```

</details>

Issue #10805, opened 2025-08-15
Remove unsafe type narrowing for non-closed TypedDicts


erictraut:
[This issue](https://github.com/microsoft/pyright/issues/10517) attempted to fix a type checking hole related to TypedDict discrimination, but I reverted it because there was no clean way to implement this pattern without PEP 728. Now that PEP 728 is accepted, this unsafe behavior should be reverted.

alicederyn:
I'd like to request the old behaviour not be reverted until mypy has stably supported the new syntax for a period of time, as otherwise we're going to be unable to create type hints in libraries that need to support both tools (basically, all libraries using this idiom).

erictraut:
I'm willing to hold off a little bit longer, but I'm not willing to commit to mypy's timeline, which I don't control. If you want to ensure that mypy supports the new syntax before I make this change in pyright, please be proactive about making the change in mypy and testing it for stability.

alicederyn:
Can pyright at least support the community-wide legacy `@final` decorator approach until this new solution has proper community support, as an alias for `closed=True`? PEP 728 won't even be in Python until 3.15.

erictraut:
PEP 728 was recently incorporated into the typing spec. When the conformance tests are updated for PEP 728, the correct (safe) behavior will likely be added to the conformance tests. At that point, I will be compelled to make the change in pyright so it conforms to the spec and the conformance tests.

PEP 728 is supported by `typing_expensions` for backward compatibility, so I don't see that as a blocker.

Is there a particular library you are concerned about here? I'd like to get a sense for the impact to users.

alicederyn:
My company has many hundreds of libraries and repos using this idiom, and we use both pyright (generally via vscode) and mypy. They all use the `@final` decorator to mark TypedDicts used in this way.

erictraut:
OK, thanks. Just to clarify, these are libraries that your company develops (as opposed to externally-developed libraries that you consume)? Are they published and used externally? Just wondering if this impacts anyone outside of your organization. If it's limited to just your org, then you could use version pinning as a short-term workaround.

alicederyn:
They are not used externally. Given the number of times this issue of type discrimination has come up, though, I think a lot of people will now be relying on the current solution.

I'm not sure how we could practically pin pyright, it's integrated into vscode, including in online platforms with disposable environments.

alicederyn:
I had a quick check this morning and there's many thousands of files using this pattern in our repositories.

PR #11620, opened 2026-08-11
Narrow closed TypedDicts on a key membership check

### LINKED ISSUE #11518: Closed TypedDicts losing type discrimination
**Describe the bug**
Adding `closed=True` to a TypedDict appears to be disabling type discrimination.

This is related to, but kind of the opposite of, #10805: in that case, unsafe discrimination for open `TypedDict`s is being allowed as it is idiomatic; in this case, *safe* discrimination for closed `TypedDict`s is being disallowed.

In the example below, `Bar` is closed so cannot contain a `"foo"` key; `"foo" in u` should therefore narrow the type of `u` to `Foo`, but this is not happening.

**Code or Screenshots**
```python
from typing import TypedDict

class Foo(TypedDict, closed=True):
    foo: int

class Bar(TypedDict, closed=True):  # Removing "closed" here causes the error to go away
    bar: int

U = Foo | Bar

def get_field(u: U) -> int:
    if "foo" in u:
        return u["foo"]  # E: "foo" is not a defined key in "Bar"
    else:
        return u["bar"]
```

[Playground link](https://pyright-play.net/?pythonVersion=3.15&code=GYJw9gtgBALgngBwJYDsDmUkQWEMoAqiApgCYAiSAxjALABQDVANgIYDO7UAYmGABREEZSjQA0UFmHZkAvARABXYgEoAXAyhaowPmswo6jeiw5cAQqxCCSFajAlSZpeUtUb62qACMr%2B1EYMAKpQsjx8UAA%2BUJYgDAykxMBQaMQwAPrASMTMpPyK%2BkEqUAC0AHwGMB5eSMkARLpgdQZQBZpe2iBpiiAorQDaDXx1ALpaAMRQAKL6Q02YXChg%2BKxQiVkoZFAA1sRwLXWxde1aOTLVHVpdMD19ioO%2BIKMMQA)

CLI, v1.1.410

nileshpatil6:
Fixes #11518.

A union of closed TypedDicts can't be discriminated with an `in` check today:

```python
class Foo(TypedDict, closed=True):
    foo: int

class Bar(TypedDict, closed=True):
    bar: int

def get_field(u: Foo | Bar) -> int:
    if "foo" in u:
        return u["foo"]  # error: "foo" is not a defined key in "Bar"
    else:
        return u["bar"]
```

`Bar` is closed, so it can't hold a `"foo"` key and should be eliminated in the positive branch.

The cause is in `narrowTypeForTypedDictKey`. `getTypedDictMembersForClass` represents `closed=True` by synthesizing an "extra items" entry whose value type is `Never`, and the positive branch looks up `knownItems.get(key) ?? extraItems`. For a closed TypedDict and an unknown key that lookup returns the synthesized `Never` entry, which is not `undefined`, so the code treats it as a real entry and keeps the subtype.

The check for this was already written in that function, commented out. I enabled just the `isNever` half. The other commented block (returning `subtype` when there's no entry at all) stays as it is, because that's the one that would disable the deliberately-unsound narrowing of open TypedDicts discussed in #10805, and the comment above it explains why it's being kept for now.

Added `typedDictClosed11.py` with seven cases: the union from the issue, the same thing through `not in`, a `Foo | Baz` union where `Baz` declares `extra_items=int` and so must not be eliminated, a `Foo | Open` union where `Open` is an ordinary TypedDict, pinning that the existing idiomatic narrowing still happens, a TypedDict with an item declared as `Never`, a union of two closed TypedDicts sharing a required key, and a closed TypedDict with a `NotRequired` key that takes the mark-as-provided path. Put it in a new sample rather than extending `typedDictClosed1.py` so it doesn't collide with #11610, which is still open against that file.

The new sample reports five errors before the change and none after: the four from the issue (two failed `reveal_type` narrowings and the two key-access errors that follow) plus the declared-`Never` case. All eleven `TypedDictClosed` tests pass.

Ran the narrowing tests across all suites (54 passing) and the full typeEvaluator5 suite (68 passing) locally on Windows.

### DIFF
--- a/packages/pyright-internal/src/analyzer/typeGuards.ts
+++ b/packages/pyright-internal/src/analyzer/typeGuards.ts
@@ -2277,11 +2277,13 @@ function narrowTypeForTypedDictKey(
                     //     return subtype;
                     // }
 
-                    // if (isNever(tdEntry.valueType)) {
-                    //     // If the entry is typed as Never or the "extra items" is typed as Never,
-                    //     // then this key cannot be present in the TypedDict, and we can eliminate it.
-                    //     return undefined;
-                    // }
+                    if (isNever(tdEntry.valueType)) {
+                        // If the entry is typed as Never or the "extra items" is typed as Never,
+                        // then this key cannot be present in the TypedDict, and we can eliminate it.
+                        // A closed TypedDict has an "extra items" of Never, so this is what allows
+                        // a key check to discriminate between closed TypedDicts.
+                        return undefined;
+                    }
 
                     // If the entry is currently not required and not marked provided, we can mark
                     // it as provided after this guard expression confirms it is.
--- a/packages/pyright-internal/src/tests/samples/typedDictClosed11.py
+++ b/packages/pyright-internal/src/tests/samples/typedDictClosed11.py
@@ -3,7 +3,7 @@
 # is not one of its known items, so such a check can discriminate.
 
 
-from typing import TypedDict
+from typing import Never, NotRequired, TypedDict
 
 
 class Foo(TypedDict, closed=True):
@@ -56,3 +56,44 @@ def func4(u: Foo | Open) -> None:
         reveal_type(u, expected_text="Foo")
     else:
         reveal_type(u, expected_text="Open")
+
+
+class NeverItem(TypedDict):
+    always: int
+    never: Never
+
+
+def func5(td: NeverItem) -> None:
+    # A declared item typed as Never can never be present either, so the
+    # same elimination applies to it and not only to the "extra items"
+    # entry synthesized for a closed TypedDict.
+    if "never" in td:
+        reveal_type(td, expected_text="Never")
+    else:
+        reveal_type(td, expected_text="NeverItem")
+
+
+class Left(TypedDict, closed=True):
+    common: int
+    left: int
+
+
+class Right(TypedDict, closed=True):
+    common: int
+    right: NotRequired[int]
+
+
+def func6(u: Left | Right) -> None:
+    # "common" is a required known item of both, so neither is eliminated.
+    if "common" in u:
+        reveal_type(u, expected_text="Left | Right")
+    else:
+        reveal_type(u, expected_text="Never")
+
+
+def func7(td: Right) -> None:
+    # "right" is a known item that is not required, so the subtype is kept
+    # and the key is marked as provided rather than eliminated.
+    if "right" in td:
+        reveal_type(td, expected_text="Right")
+        reveal_type(td["right"], expected_text="int")
--- a/packages/pyright-internal/src/tests/samples/typedDictClosed11.py
+++ b/packages/pyright-internal/src/tests/samples/typedDictClosed11.py
@@ -1,0 +1,58 @@
+# This sample tests narrowing of a union of closed TypedDicts based on
+# an "in" check for a key. A closed TypedDict cannot contain a key that
+# is not one of its known items, so such a check can discriminate.
+
+
+from typing import TypedDict
+
+
+class Foo(TypedDict, closed=True):
+    foo: int
+
+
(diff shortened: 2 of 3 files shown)


StellaHuang95:

packages/pyright-internal/src/analyzer/typeGuards.ts
```diff
@@ -2277,11 +2277,13 @@ function narrowTypeForTypedDictKey(
...
-                    //     // then this key cannot be present in the TypedDict, and we can eliminate it.
-                    //     return undefined;
-                    // }
+                    if (isNever(tdEntry.valueType)) {
+                        // If the entry is typed as Never or the "extra items" is typed as Never,
+                        // then this key cannot be present in the TypedDict, and we can eliminate it.
+                        // A closed TypedDict has an "extra items" of Never, so this is what allows
+                        // a key check to discriminate between closed TypedDicts.
+                        return undefined;
+                    }
```
StellaHuang95, line 2286: > **Info** · Optional note

The new guard also changes positive membership checks for declared keys typed `Never`, not only synthesized closed-TypedDict extra-items. Please add a small regression case to pin that intended behavior.

packages/pyright-internal/src/tests/samples/typedDictClosed11.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def func4(u: Foo | Open) -> None:
+    # An open TypedDict without "extra_items" is narrowed on a key check
+    # even though it is not sound to do so; this is idiomatic and is
+    # relied upon in practice.
+    if "foo" in u:
+        reveal_type(u, expected_text="Foo")
+    else:
+        reveal_type(u, expected_text="Open")
```
StellaHuang95, line 58: > **Info** · Optional note

Consider adding a closed-TypedDict union with an overlapping key to exercise the existing required-key retain/clone-and-mark-provided path alongside the new elimination path.

StellaHuang95:
The negative branch implicitly relies on closed TypedDicts' synthesized `Never` extra-items entry being non-required and non-provided. A brief comment there could help preserve this subtle symmetry in future edits.

StellaHuang95 (approved):
Approved via Review Center.

nileshpatil6:
Thanks, all three were worth doing. Added the two test cases in a7b094b.

On the declared-`Never` item: you're right that this isn't limited to the synthesized closed-TypedDict entry, and it was worth pinning rather than leaving implied. `func5` covers a TypedDict with an item declared as `Never`, and a positive check on that key now eliminates the subtype. I checked it fails against the previous commit ("expected Never but received NeverItem"), so it's holding the new behavior rather than passing either way. I think eliminating there is right for the same reason as the closed case, since a key whose value type is uninhabited can't be present, but it's now explicit in the tests either way.

On the overlapping key: added `Left | Right`, both closed and both declaring `common`. A check on `common` keeps both subtypes, and the negative branch eliminates both since it's required in each. `func7` covers the other half with `right: NotRequired[int]`, which takes the clone-and-mark-provided path rather than being eliminated. Both of those pass before and after the change, which is the point of having them.

On the comment in the negative branch: that branch is what #11610 rewrites, and it adds the note you're describing as part of the change, distinguishing the known item from the "extra items" fallback so the asymmetry is stated outright. I've left the negative branch untouched here so the two PRs don't collide on the same lines. If #11610 ends up not landing, say the word and I'll bring the comment over to this one.

StellaHuang95:
Please update the PR description: the committed sample contains seven cases, not four, so its stated pre-change error count is stale.

StellaHuang95 (approved):
Approved via Review Center.

nileshpatil6:
Updated the description, thanks for catching it. It now says seven cases and five pre-change errors, the extra one being the declared-`Never` case from `func5`. Missed that when I added the two cases from your review.

PR #11621, opened 2026-08-11
Preserve union-expanded constructor types

rchiodo:
## Summary

- accumulate each union-expanded overloaded `__init__` branch's effective specialized self type, including definitive and single-possible branches
- reconstruct union-valued constructor handoffs subtype-by-subtype so gradual members and concrete constructed types are all preserved
- add focused coverage for mixed ambiguous/definitive branches, all-ambiguous branches, definitive constructor unions, explicit specialization, and the `None` `__init__` placeholder

Follow-up to #11601. This addresses the unresolved review finding in https://github.com/microsoft/pyright/pull/11601#discussion_r3754688682, where a later definitive branch such as `Table[bytes]` was omitted once an earlier materialization-ambiguous branch populated the separate constructor handoff.

## Tests

- `npx jest typeEvaluator6.test -t "OverloadCall12" --runInBand --forceExit`
- `npx jest typeEvaluator6.test -t "OverloadCall(6|11|12)|Constructor" --runInBand --forceExit`
- `npm run build` (`packages/pyright-internal`)
- `npm run typecheck`
- focused ESLint and Prettier checks for changed TypeScript files

### DIFF
--- a/packages/pyright-internal/src/analyzer/typeEvaluator.ts
+++ b/packages/pyright-internal/src/analyzer/typeEvaluator.ts
@@ -9948,12 +9948,13 @@ export function createTypeEvaluator(
         const overloadsUsedForCall: FunctionType[] = [];
         let isDefinitiveMatchFound = false;
         let hasInitSelfMaterializationAmbiguity = false;
+        let hasSpecializedInitSelfType = false;
         const speculativeNode = getSpeculativeNodeForCall(errorNode);
 
         for (let expandedTypesIndex = 0; expandedTypesIndex < expandedArgTypes.length; expandedTypesIndex++) {
             const overloadsUsedStartIndex = overloadsUsedForCall.length;
             let matchedOverload: FunctionType | undefined;
-            let specializedInitSelfType: Type | undefined;
+            let effectiveInitSelfType: Type | undefined;
             const argTypeOverride = expandedArgTypes[expandedTypesIndex];
             const hasArgTypeOverride = argTypeOverride.some((a) => a !== undefined);
             let possibleMatchResults: MatchedOverloadInfo[] = [];
@@ -10023,7 +10024,8 @@ export function createTypeEvaluator(
                         }
                     } else {
                         returnTypes.push(callResult.returnType);
-                        specializedInitSelfType = callResult.specializedInitSelfType;
+                        effectiveInitSelfType = getEffectiveOverloadReturnType(matchedOverloadInfo);
+                        hasSpecializedInitSelfType ||= !!callResult.specializedInitSelfType;
                         isDefinitiveMatchFound = true;
                         break;
                     }
@@ -10051,7 +10053,8 @@ export function createTypeEvaluator(
                 // Did the filtering produce a single result? If so, we're done.
                 if (possibleMatchResults.length === 1) {
                     returnTypes.push(possibleMatchResults[0].returnType);
-                    specializedInitSelfType = possibleMatchResults[0].specializedInitSelfType;
+                    effectiveInitSelfType = getEffectiveOverloadReturnType(possibleMatchResults[0]);
+                    hasSpecializedInitSelfType ||= !!possibleMatchResults[0].specializedInitSelfType;
                     matchedOverloads = [possibleMatchResults[0]];
                 } else {
                     const firstArgParamPairs = getOverloadArgParamPairs(possibleMatchResults[0]);
@@ -10148,13 +10151,14 @@ export function createTypeEvaluator(
                         }
                     }
 
+                    effectiveInitSelfType = returnType;
                     if (isInitSelfMaterializationAmbiguity) {
                         // Overloaded __init__ methods normally return None. Preserve that
                         // ordinary return as a placeholder while union-expanded calls are
                         // combined, and carry the effective constructed types separately.
                         // validateInitMethod consumes specializedInitSelfType to reconstruct
                         // the constructor result after call validation is complete.
-                        specializedInitSelfType = returnType;
+                        hasSpecializedInitSelfType = true;
                         hasInitSelfMaterializationAmbiguity = true;
                         returnTypes.push(possibleMatchResults[0].returnType);
                     } else {
@@ -10163,8 +10167,8 @@ export function createTypeEvaluator(
                 }
             }
 
-            if (specializedInitSelfType) {
-                specializedInitSelfTypes.push(specializedInitSelfType);
+            if (effectiveInitSelfType) {
+                specializedInitSelfTypes.push(effectiveInitSelfType);
             }
 
             if (!matchedOverload) {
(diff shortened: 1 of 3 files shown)


StellaHuang95:

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -10192,7 +10201,8 @@ export function createTypeEvaluator(
             returnType: combineTypes(returnTypes),
             isTypeIncomplete,
             specializedInitSelfType:
-                specializedInitSelfTypes.length > 0
+                specializedInitSelfTypes.length > 0 &&
+                (hasInitSelfMaterializationAmbiguity || expandedArgTypes.length > 1)
                     ? combineTypes(specializedInitSelfTypes)
                     : finalCallResult.specializedInitSelfType,
             overloadsUsedForCall,
```
StellaHuang95, line 10217: > **Issue** · Please address or respond

When one expanded branch supplies `specializedInitSelfType` but another matched branch does not, this nonempty list causes the final constructor handoff to contain only the specialized branches. `validateInitMethod` then reconstructs the result solely from that handoff, dropping the other branch's constructed type. Preserve a per-expanded-branch handoff for branches without a specialized self type as well.

packages/pyright-internal/src/tests/samples/overloadCall12.py
```diff
@@ -171,9 +188,18 @@ def check_constructors(values_any: list[Any], values_unknown: list, values_int:
...
+    values_any: list[Any] | bytes,
+    values_unknown: list | bytes,
+    values_int: list[int] | bytes,
+    values_gradual: list[Any] | set,
+) -> None:
+    reveal_type(Table(values_any), expected_text="Any | Table[bytes]")
+    reveal_type(Table(values_unknown), expected_text="Unknown | Table[bytes]")
+    reveal_type(Table(values_int), expected_text="Table[int] | Table[bytes]")
+    reveal_type(AmbiguousTable(values_gradual), expected_text="Any | Unknown")
+    reveal_type(table.__init__(values_any), expected_text="None")
```
StellaHuang95, line 233: > **Warning** · Non-blocking recommendation

Add regression coverage for a union that combines a self-specializing branch with a matched branch that has no `specializedInitSelfType` (or takes the multi-possible path). The existing cases do not exercise the mixed case where the latter branch can be omitted from the reconstructed constructor result.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -10192,7 +10201,8 @@ export function createTypeEvaluator(
             returnType: combineTypes(returnTypes),
             isTypeIncomplete,
             specializedInitSelfType:
-                specializedInitSelfTypes.length > 0
+                specializedInitSelfTypes.length > 0 &&
+                (hasInitSelfMaterializationAmbiguity || expandedArgTypes.length > 1)
                     ? combineTypes(specializedInitSelfTypes)
                     : finalCallResult.specializedInitSelfType,
```
StellaHuang95, line 10216: > **Info** · Optional note

Please document why both conditions are required here: union expansion requires aggregating per-expanded results, whereas materialization ambiguity requires aggregation even for a single expanded argument list. This makes the fallback to `finalCallResult.specializedInitSelfType` safe to preserve.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
rchiodo: Fixed in 548b3447f. Each successful expanded branch now contributes an effective constructor handoff; ordinary `__init__` branches materialize their bound type with branch-local constraints, so their constructed type is preserved without inheriting constraints from the final replay.

packages/pyright-internal/src/tests/samples/overloadCall12.py
rchiodo: Added focused coverage with `PartiallySpecializedTable`, combining ambiguous explicit-`self` overloads with an ordinary `set[_T]` branch. The assertion pins the complete result as `Any | PartiallySpecializedTable[float]`.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
rchiodo: Documented the two aggregation triggers beside `shouldCombineInitSelfTypes`: union expansion contributes one handoff per expanded argument list, while materialization ambiguity can require a combined handoff within a single list.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -10186,13 +10199,19 @@ export function createTypeEvaluator(
             isTypeIncomplete = true;
         }
 
+        // Union expansion requires one constructor handoff per expanded argument list.
+        // Materialization ambiguity requires a combined handoff even when there is only
+        // one argument list because multiple overload candidates contribute to its result.
+        const shouldCombineInitSelfTypes =
+            hasInitSelfMaterializationAmbiguity || (expandedArgTypes.length > 1 && hasSpecializedInitSelfType);
+
         return {
```
StellaHuang95, line 10208: > **Warning** · Non-blocking recommendation

`shouldCombineInitSelfTypes` is gated by `hasSpecializedInitSelfType`, but that flag is not set when `effectiveInitSelfType` comes solely from the new `boundToType` fallback. A union with two definitive `__init__` branches using only that fallback will accumulate their effective types, then discard them by returning `finalCallResult.specializedInitSelfType`. Track whether an effective self type was accumulated, or set the flag at the push site.

StellaHuang95:
GitHub cannot anchor PR review comments to unchanged lines in the diff. Falling back to a general PR comment for [packages/pyright-internal/src/tests/samples/overloadCall12.py:L227](https://github.com/microsoft/pyright/blob/548b3447f80160518438338797f48e86c509da8d/packages/pyright-internal/src/tests/samples/overloadCall12.py#L227).

> **Warning** · Non-blocking recommendation

Add coverage for a union of two definitive branches that both rely only on the `boundToType` fallback, such as `set[int] | set[float]` through the non-`self:`-annotated `set[_T]` overload. The current cases trigger either ambiguity or a raw `specializedInitSelfType`, so they do not exercise the combine-gate gap.

StellaHuang95 (approved):
Approved via Review Center.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
rchiodo: Fixed in 225318d91. The aggregation gate now tracks effective `__init__` self types at the push site, including bound-type fallbacks, and the regression pins two definitive ordinary branches as `InferredTable[int] | InferredTable[str]`; the final replay also uses the first branch's constraints to avoid cross-branch leakage.

StellaHuang95 (approved):
Approved via Review Center.

PR #11627, opened 2026-08-12
Add npm-only release recovery mode

heejaechang:
## Summary

- replaces the removed ESRP npm approver with `brcan@microsoft.com`
- adds an `npmOnly` release mode that republishes the original retained npm artifact
- keeps tagged releases on the existing `full` mode by default

For npm-only recovery, the operator supplies the release tag and original source build ID. The pipeline verifies that the source build belongs to this release pipeline and exact tag, downloads its retained `pyright` artifact, installs it, verifies the embedded package version, and runs only ESRP npm publication. It does not rebuild or republish the VSIX or create another GitHub release.

## Validation

- `npm run check`
- `git diff --check`
- verified `brcan@microsoft.com` resolves to a valid Microsoft Entra identity
- rehearsed recovery against build `14944137`: pipeline `22838`, tag `1.1.412`, commit `fceca4d1`, retained `pyright.tgz` version `1.1.412`

Azure local-YAML preview could not be run because the caller lacks the `EditBuild` permission required for modified-YAML previews.

### DIFF
--- a/build/azuredevops/azure-pipelines-release.yml
+++ b/build/azuredevops/azure-pipelines-release.yml
@@ -10,6 +10,10 @@ parameters:
     values:
       - full
       - npmOnly
+  - name: releaseTag
+    displayName: Release tag (required for npmOnly)
+    type: string
+    default: ''
 resources:
   repositories:
     - repository: MicroBuildTemplate
@@ -53,7 +57,7 @@ extends:
       - ES365AIMigrationTooling
     stages:
       - stage: BuildVsix
-        condition: and(succeeded(), eq('${{ parameters.releaseMode }}', 'full'))
+        condition: and(not(canceled()), succeeded(), eq('${{ parameters.releaseMode }}', 'full'))
         jobs:
           - job: build_vsix
             displayName: Build VSIX
@@ -179,10 +183,25 @@ extends:
                 submodules: true
                 fetchTags: false
                 persistCredentials: True
+              - ${{ if eq(parameters.releaseMode, 'npmOnly') }}:
+                  - bash: |
+                      set -euo pipefail
+                      tag='${{ parameters.releaseTag }}'
+                      if [[ ! "$tag" =~ ^1\.1\.[0-9]+$ ]]; then
+                        echo "##vso[task.logissue type=error]releaseTag must match 1.1.<patch>"
+                        exit 1
+                      fi
+                      git fetch origin "refs/tags/$tag:refs/tags/$tag"
+                      git checkout --detach "$tag"
+                    displayName: Check out release tag
               - task: NodeTool@0
                 displayName: Use Node 24.15.0
                 inputs:
                   versionSpec: 24.15.0
+              - ${{ if eq(parameters.releaseMode, 'npmOnly') }}:
+                  - script: |
+                      node -e "const actual=require('./packages/pyright/package.json').version; const expected='${{ parameters.releaseTag }}'; if(actual!==expected){console.error('Expected package version '+expected+', got '+actual); process.exit(1)}"
+                    displayName: Validate release tag version
               - template: /build/templates/npmAuthenticate.yml@self
               - task: CmdLine@2
                 displayName: npm install
@@ -206,21 +225,16 @@ extends:
                   Contents: '${{ variables.PACKAGE_NAME }}'
                   TargetFolder: '$(Build.ArtifactStagingDirectory)/${{ variables.ARTIFACT_NAME_PACKAGE }}'
 
-      - stage: CreateRelease
-        condition: and(succeeded(), eq('${{ parameters.releaseMode }}', 'full'))
+      - stage: ValidateNpm
         dependsOn:
-          - BuildVsix
           - BuildNpm
         jobs:
-          - job: create_release
-            displayName: Create GitHub Release
+          - job: validate_npm
+            displayName: Validate NPM package
             templateContext:
               type: releaseJob
-              isProduction: true
+              isProduction: false
               inputs:
-                - input: pipelineArtifact
-                  artifactName: $(ARTIFACT_NAME_VSIX)
-                  targetPath: $(Pipeline.Workspace)/$(ARTIFACT_NAME_VSIX)
                 - input: pipelineArtifact
                   artifactName: $(ARTIFACT_NAME_PACKAGE)
                   targetPath: $(Pipeline.Workspace)/$(ARTIFACT_NAME_PACKAGE)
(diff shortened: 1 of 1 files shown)


rchiodo:

build/azuredevops/azure-pipelines-release.yml
```diff
@@ -299,7 +311,19 @@ extends:
...
+          and(
+            in(dependencies.BuildNpm.result, 'Succeeded', 'SucceededWithIssues'),
+            or(
+              eq('${{ parameters.releaseMode }}', 'npmOnly'),
+              and(
+                in(dependencies.WaitForValidation.result, 'Succeeded', 'SucceededWithIssues'),
+                in(dependencies.PublishToMarketplace.result, 'Succeeded', 'SucceededWithIssues')
+              )
+            )
+          )
```
rchiodo, line 429: > **Issue** · Please address or respond

The custom condition replaces Azure's cancellation-aware default and remains true for a successful npm-only build without checking cancellation. Add `not(canceled())` to the outer `and(...)` so canceling a production recovery cannot start npm publication.

[verified]

build/azuredevops/azure-pipelines-release.yml
```diff
@@ -198,6 +207,7 @@ extends:
                   TargetFolder: '$(Build.ArtifactStagingDirectory)/${{ variables.ARTIFACT_NAME_PACKAGE }}'
```
rchiodo, line 223: > **Warning** · Non-blocking recommendation

`npmOnly` structurally skips the existing package installation and `pyright --version` smoke test because validation remains inside full-only `CreateRelease`. Preserve this package safeguard in a stage shared by both modes.

[verified]

build/azuredevops/azure-pipelines-release.yml
rchiodo, line 429: > **Warning** · Non-blocking recommendation

A manually queued `npmOnly` run is not constrained to the release tag whose Marketplace publication it is recovering. Require a tag source, or accept and validate an explicit release tag, before allowing npm publication.

[verified]

build/azuredevops/azure-pipelines-release.yml
heejaechang: Fixed in `c52546b95`. I moved the tarball installation and `pyright --version` smoke test into a shared `ValidateNpm` stage that depends on `BuildNpm`. Both full and npm-only flows must pass it before continuing; `CreateRelease` no longer owns that safeguard.

build/azuredevops/azure-pipelines-release.yml
heejaechang: Fixed in `c52546b95`. The `PublishToNpm` condition now starts with `not(canceled())`; the other custom full-only stage conditions also preserve cancellation. `npm run check` and the pipeline-structure assertions pass.

heejaechang: Fixed in `c52546b95`. `npmOnly` now requires an explicit `releaseTag` matching `1.1.<patch>`, fetches and checks out that tag in `BuildNpm`, and verifies `packages/pyright/package.json` exactly matches the requested tag before packaging.

build/azuredevops/azure-pipelines-release.yml
```diff
@@ -170,10 +183,25 @@ extends:
...
+                      git fetch origin "refs/tags/$tag:refs/tags/$tag"
+                      git checkout --detach "$tag"
+                    displayName: Check out release tag
               - task: NodeTool@0
                 displayName: Use Node 24.15.0
                 inputs:
                   versionSpec: 24.15.0
+              - ${{ if eq(parameters.releaseMode, 'npmOnly') }}:
+                  - script: |
+                      node -e "const actual=require('./packages/pyright/package.json').version; const expected='${{ parameters.releaseTag }}'; if(actual!==expected){console.error('Expected package version '+expected+', got '+actual); process.exit(1)}"
```
rchiodo: > **Issue** · Please address or respond

A crafted `releaseTag` can escape the quoted Bash or JavaScript string and execute commands before validation. Pass the parameter through an environment variable, validate `$RELEASE_TAG`, and read `process.env.RELEASE_TAG` in Node so the value is always treated as data.

[verified]

build/azuredevops/azure-pipelines-release.yml
```diff
@@ -170,10 +183,25 @@ extends:
...
+                    displayName: Validate release tag version
               - template: /build/templates/npmAuthenticate.yml@self
               - task: CmdLine@2
                 displayName: npm install
```
rchiodo, line 203: > **Warning** · Non-blocking recommendation

Recovery rebuilds the package from the selected source instead of consuming the artifact from the failed release run, so source or dependency drift could produce npm contents that differ from the published extension. Reuse the original artifact or verify its commit and package identity before publication.

[verified]

build/azuredevops/azure-pipelines-release.yml
heejaechang: Fixed in `715ad6460`. `releaseTag` is passed only through `RELEASE_TAG` environment variables; PowerShell validates `$env:RELEASE_TAG`, and Node reads `process.env.RELEASE_TAG`. No user-supplied parameter is interpolated into shell or JavaScript source.

build/azuredevops/azure-pipelines-release.yml
```diff
@@ -197,20 +221,71 @@ extends:
...
+                  }
+                  if ([int]$env:SOURCE_BUILD_ID -le 0) {
+                      Write-Error 'sourceBuildId must identify the original release run'
+                  }
+                  $headers = @{ Authorization = "Bearer $env:SYSTEM_ACCESSTOKEN" }
+                  $sourceBuild = Invoke-RestMethod `
+                    -Uri "$(System.CollectionUri)$(System.TeamProjectId)/_apis/build/builds/$env:SOURCE_BUILD_ID?api-version=7.1" `
+                    -Headers $headers
+                  if ($sourceBuild.definition.id -ne [int]"$(System.DefinitionId)") {
+                    Write-Error 'sourceBuildId belongs to a different pipeline'
```
rchiodo, line 253: > **Issue** · Please address or respond

The `sourceBranch` check alone does not prove the retained artifact was built from the requested tag: a build can be queued with `refs/tags/<tag>` and a different `sourceVersion`. Resolve `refs/tags/$RELEASE_TAG` and require its commit SHA to match `sourceBuild.sourceVersion` before downloading and publishing the artifact.

heejaechang: Fixed in `3f5601045`. Before artifact download, recovery now fetches the explicit release tag, peels it with `git rev-list -n 1`, and requires that commit to equal `sourceBuild.sourceVersion`. Rehearsal against the real release passed: tag `1.1.412` and build `14944137` both resolve to `fceca4d133a83034fc195dbba8b3814992b0b26b`.

rchiodo (approved):
Approved via Review Center.

StellaHuang95 (approved):
Approved via Review Center.

PR #11630, opened 2026-08-12
Fix npm recovery build lookup URL

heejaechang:
## Summary

- fixes PowerShell parsing of `SOURCE_BUILD_ID` before the Builds API query string
- fixes PowerShell parsing of `RELEASE_TAG` inside the Git fetch refspec
- adds `npmValidateOnly` to exercise artifact recovery and package smoke validation without entering ESRP npm publication

Recovery run `14951826` exposed the Builds API interpolation issue. Pre-merge validation run `14953223` then exposed the same scoped-variable boundary problem in the tag refspec, where PowerShell produced `refs/tags//tags/1.1.412`.

## Validation

- `npm run check`
- `git diff --check`
- verified the Builds API URI contains `builds/14944137?api-version=7.1`
- executed the corrected refspec against tag `1.1.412`; it resolves to `fceca4d1`
- `npmValidateOnly` reaches recovery/validation and cannot enter npm publication

### DIFF
--- a/build/azuredevops/azure-pipelines-release.yml
+++ b/build/azuredevops/azure-pipelines-release.yml
@@ -10,6 +10,7 @@ parameters:
     values:
       - full
       - npmOnly
+      - npmValidateOnly
   - name: releaseTag
     displayName: Release tag (required for npmOnly)
     type: string
@@ -40,7 +41,7 @@ variables:
   # Route npm child processes through the runtime-generated authenticated CFS config.
   - name: NPM_CONFIG_USERCONFIG
     value: $(Agent.TempDirectory)/cfs.npmrc
-  - ${{ if eq(parameters.releaseMode, 'npmOnly') }}:
+  - ${{ if ne(parameters.releaseMode, 'full') }}:
       - name: EXPECTED_RELEASE_TAG
         value: ${{ parameters.releaseTag }}
   - ${{ else }}:
@@ -222,7 +223,7 @@ extends:
                   TargetFolder: '$(Build.ArtifactStagingDirectory)/${{ variables.ARTIFACT_NAME_PACKAGE }}'
 
       - stage: RecoverNpmArtifact
-        condition: and(not(canceled()), succeeded(), eq('${{ parameters.releaseMode }}', 'npmOnly'))
+        condition: and(not(canceled()), succeeded(), ne('${{ parameters.releaseMode }}', 'full'))
         dependsOn: []
         jobs:
           - job: recover_npm_artifact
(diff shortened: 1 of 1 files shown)


StellaHuang95:

StellaHuang95:
Please also brace `RELEASE_TAG` in the recovery `git fetch` refspec: `refs/tags/${env:RELEASE_TAG}:refs/tags/${env:RELEASE_TAG}`. The colon is otherwise absorbed into the environment-variable reference.

build/azuredevops/azure-pipelines-release.yml
```diff
@@ -222,7 +223,7 @@ extends:
                   TargetFolder: '$(Build.ArtifactStagingDirectory)/${{ variables.ARTIFACT_NAME_PACKAGE }}'
```
StellaHuang95, line 224: > **Issue** · Please address or respond

This new condition sends `npmValidateOnly` through `RecoverNpmArtifact`, but the existing `git fetch` refspec in that job still embeds `$env:RELEASE_TAG` directly before `:refs`. PowerShell consumes `:refs` as part of the environment-variable reference, producing a malformed refspec and failing recovery. Brace both references as `${env:RELEASE_TAG}` so validate-only runs can reach validation.

StellaHuang95, line 224: > **Warning** · Non-blocking recommendation

`ne('full')` makes every future non-`full` release mode opt into recovery and validation automatically. Use an explicit allow-list for `npmOnly` and `npmValidateOnly` here and in the matching validation condition so future modes fail closed unless deliberately added to this path.

[verified]

build/azuredevops/azure-pipelines-release.yml
```diff
@@ -10,6 +10,7 @@ parameters:
     values:
       - full
```
StellaHuang95, line 11: > **Info** · Optional note

The `releaseTag` and `sourceBuildId` display names still say they are required only for `npmOnly`, but `npmValidateOnly` uses both inputs in the same recovery flow. Update the help text so operators know to provide them for the new mode.

[verified]

StellaHuang95 (approved):
Approved via Review Center.

rchiodo (approved):
Approved via Review Center.

build/azuredevops/azure-pipelines-release.yml
heejaechang: Fixed in `faa563202`: both parameter display names now state that `releaseTag` and `sourceBuildId` are required for `npmOnly` and `npmValidateOnly`.

StellaHuang95 (approved):
Approved via Review Center.

PR #11632, opened 2026-08-12
Pull Pylance with Pyright 1.1.411

bschnurr:


rchiodo:

.github/actions/standard-npm-install/action.yml
```diff
@@ -1,15 +1,20 @@
...
-    - uses: ./.github/actions/npm-cache-dir
-
-    - name: Standard npm install
+    - name: Standard pnpm install
       shell: bash
-      run: npm run install:all
+      working-directory: ${{ inputs.working-directory }}
+      env:
+        NPM_CONFIG_REGISTRY: https://registry.npmjs.org/
+      run: pnpm run install:all
```
rchiodo, line 20: > **Info** · Optional note

The public path delegates to `pnpm run install:all`, while the CFS path directly encodes `pnpm install --frozen-lockfile --prefer-offline`, creating two installation contracts that can drift. Route both paths through one shared script or composite step after registry authentication. [verified]

rchiodo:
The new bounded-type-parameter specialization is missing from Pylance's async pattern-matching path, causing observable sync/async behavior divergence.

rchiodo:
The Pyright sync path now handles bounded generic class patterns differently from Pylance's async path, creating user-visible mode divergence.

PR #11638, opened 2026-08-14
Fix full release stage conditions

heejaechang:
## Summary

Fix full Pyright release runs skipping all production stages after npm artifact validation.

`RecoverNpmArtifact` is intentionally skipped in `full` mode. The downstream stages used `succeeded()`, which evaluated false because that skipped stage was a transitive dependency through `ValidateNpm`. Check each stage's explicit prerequisite results instead.

## Validation

- YAML parses successfully and passes Prettier and Syncpack checks.
- Official `Pyright-Release` condition-only run passed in `full` mode: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=14977419&view=results
- The test run reproduced the intended graph: `RecoverNpmArtifact` skipped, while `CreateRelease`, `WaitForValidation`, `PublishToMarketplace`, and `PublishToNpm` all succeeded.

This unblocks recovery of the tagged but unpublished `1.1.413` release.

### DIFF
--- a/build/azuredevops/azure-pipelines-release.yml
+++ b/build/azuredevops/azure-pipelines-release.yml
@@ -324,7 +324,13 @@ extends:
                   RELEASE_TAG: $(EXPECTED_RELEASE_TAG)
 
       - stage: CreateRelease
-        condition: and(not(canceled()), succeeded(), eq('${{ parameters.releaseMode }}', 'full'))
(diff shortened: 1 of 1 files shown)


rchiodo:

rchiodo (approved):
Approved via Review Center.

PR #11640, opened 2026-08-14
Pull Pylance with Pyright 1.1.412

bschnurr:


rchiodo:

.github/actions/standard-npm-install/action.yml
```diff
@@ -1,15 +1,20 @@
...
-    - uses: ./.github/actions/npm-cache-dir
-
-    - name: Standard npm install
+    - name: Standard pnpm install
       shell: bash
-      run: npm run install:all
+      working-directory: ${{ inputs.working-directory }}
+      env:
+        NPM_CONFIG_REGISTRY: https://registry.npmjs.org/
+      run: pnpm run install:all
```
rchiodo, line 20: > **Info** · Optional note

`working-directory` may point at a package that does not define `install:all`, while the CFS path runs `pnpm install` directly. Use the same frozen install command in both paths, or centralize installation after conditional authentication, so both actions support non-root directories consistently.

[verified]

rchiodo (approved):
Approved via Review Center.

rchiodo (approved):
Approved via Review Center.

heejaechang (approved):
Approved via Review Center.

.github/actions/standard-npm-install/action.yml
rchiodo, line 20: > **Issue** · Please address or respond

This action accepts a package `working-directory`, but `pnpm run install:all` requires that package to define the root-only script. Fork workflows targeting a package directory will fail while the CFS path succeeds. Run `pnpm install --frozen-lockfile --prefer-offline` here as well.

package.json
```diff
@@ -1,24 +1,24 @@
 {
     "name": "pyright-root",
     "private": true,
+    "packageManager": "pnpm@10.12.2",
```
rchiodo, line 4: > **Issue** · Please address or respond

The repository now makes pnpm authoritative but still commits `package-lock.json`, leaving two dependency lockfiles as competing sources of truth. Remove the obsolete npm lockfile as part of this migration.

rchiodo:

PR #11643, opened 2026-08-15
Report illegal zero-argument super() in nested functions and lambdas

hsusul:
## Summary
- Zero-argument `super()` is valid only in a method (and in comprehensions that run in that method). Nested `def`/`async def` and lambdas raise `RuntimeError: super(): no arguments` because they do not get the compiler-inserted `__class__` cell.
- Pyright previously treated `super()` inside a nested function as if it were in the enclosing method (`super1.py` even contained this case with no expected error).
- Reuses the existing `superCallZeroArgForm` diagnostic. Nested class *bases* evaluated in a method remain allowed; nested class *bodies* still error.

## Test plan
- [x] `npx jest typeEvaluator2.test.ts -t Super --forceExit` (from `packages/pyright-internal`)
- [x] Runtime check: nested `def`/`lambda` raise; `[super() for _ in [1]]` inside a method succeeds
- [ ] Confirm nested `super().method()` in an editor shows the zero-argument diagnostic

### DIFF
--- a/packages/pyright-internal/src/analyzer/parseTreeUtils.ts
+++ b/packages/pyright-internal/src/analyzer/parseTreeUtils.ts
@@ -704,6 +704,65 @@ export function getEnclosingFunction(node: ParseNode): FunctionNode | undefined 
     return undefined;
 }
 
+// Zero-argument super() is valid only in a method (or a comprehension
+// nested in that method). Nested functions and lambdas raise
+// RuntimeError at runtime because they do not receive the compiler-inserted
+// __class__ cell used by the zero-argument form.
+export function isZeroArgumentSuperCallAllowed(node: ParseNode): boolean {
+    let prevNode: ParseNode | undefined;
+    let curNode = node.parent;
+
+    while (curNode) {
+        if (curNode.nodeType === ParseNodeType.Lambda) {
+            return false;
+        }
+
+        if (curNode.nodeType === ParseNodeType.Function) {
+            // Don't treat a decorator as being "enclosed" in the function.
+            if (curNode.d.decorators.some((decorator) => decorator === prevNode)) {
+                prevNode = curNode;
+                curNode = curNode.parent;
+                continue;
+            }
+
+            let parent: ParseNode | undefined = curNode.parent;
+            while (parent) {
+                if (parent.nodeType === ParseNodeType.Class) {
+                    return true;
+                }
+                if (parent.nodeType === ParseNodeType.Function) {
+                    return false;
+                }
+                parent = parent.parent;
+            }
+
+            return false;
+        }
+
+        if (curNode.nodeType === ParseNodeType.Class) {
+            // super() in a class decorator, type parameter, or base-class
+            // argument is still evaluated in the enclosing scope, not the
+            // class body.
+            if (
+                curNode.d.decorators.some((decorator) => decorator === prevNode) ||
+                curNode.d.arguments.some((arg) => arg === prevNode) ||
+                curNode.d.typeParams === prevNode
+            ) {
+                prevNode = curNode;
+                curNode = curNode.parent;
+                continue;
+            }
+
+            return false;
+        }
+
+        prevNode = curNode;
+        curNode = curNode.parent;
+    }
+
+    return false;
+}
+
 // This is similar to getEnclosingFunction except that it uses evaluation
 // scopes rather than the parse tree to determine whether the specified node
 // is within the scope. That means if the node is within a class decorator
--- a/packages/pyright-internal/src/analyzer/parseTreeUtils.ts
+++ b/packages/pyright-internal/src/analyzer/parseTreeUtils.ts
@@ -11,6 +11,7 @@ import * as AnalyzerNodeInfo from '../analyzer/analyzerNodeInfo';
 import { containsOnlyWhitespace } from '../common/core';
 import { assert, assertNever, fail } from '../common/debug';
 import { convertPositionToOffset, convertTextRangeToRange } from '../common/positionUtils';
+import { PythonVersion, pythonVersion3_12 } from '../common/pythonVersion';
 import { Position, Range, TextRange } from '../common/textRange';
 import { TextRangeCollection, getIndexContaining } from '../common/textRangeCollection';
 import {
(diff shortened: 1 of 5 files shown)


rchiodo:

packages/pyright-internal/src/analyzer/parseTreeUtils.ts
```diff
@@ -704,6 +704,65 @@ export function getEnclosingFunction(node: ParseNode): FunctionNode | undefined
...
+                if (parent.nodeType === ParseNodeType.Class) {
+                    return true;
+                }
+                if (parent.nodeType === ParseNodeType.Function) {
+                    return false;
+                }
+                parent = parent.parent;
+            }
+
+            return false;
```
rchiodo: > **Issue** · Please address or respond

A nested callable can validly use zero-argument `super()` when its own first argument is the instance and it closes over `__class__` (for example, `def inner(self): return super()`). Rejecting every nested function or lambda creates a false diagnostic; classify this based on the executing frame's arguments and `__class__` cell rather than nesting alone.

rchiodo: > **Issue** · Please address or respond

Defaults of a nested callable are evaluated in the enclosing method frame. For example, `def inner(value=super()): ...` can validly use the enclosing method's instance and `__class__` cell, but this returns `false` as soon as it reaches `inner`. Handle parameter defaults as expressions evaluated in the enclosing scope.

packages/pyright-internal/src/analyzer/parseTreeUtils.ts
```diff
@@ -704,6 +704,65 @@ export function getEnclosingFunction(node: ParseNode): FunctionNode | undefined
...
+        }
+
+        prevNode = curNode;
+        curNode = curNode.parent;
+    }
+
+    return false;
+}
+
```
rchiodo, line 758: > **Issue** · Please address or respond

This accepts all comprehensions, but generator expressions retain a separate frame whose implicit iterator argument is not the method instance. List, set, and dict comprehensions likewise had separate frames before Python 3.12. Distinguish comprehension kinds and the configured Python version so runtime-failing forms receive the diagnostic.

rchiodo, line 758: > **Warning** · Non-blocking recommendation

This duplicates evaluation-scope boundary handling already provided by `getEvaluationScopeNode`, including decorators, defaults, class headers, lambdas, and comprehensions. Derive this predicate from that machinery with only the needed frame-specific logic so the two scope definitions do not drift.

packages/pyright-internal/src/tests/typeEvaluator2.test.ts
```diff
@@ -177,7 +177,7 @@ test('AugmentedAssignment3', () => {
 test('Super1', () => {
     const analysisResults = TestUtils.typeAnalyzeSampleFiles(['super1.py']);
 
-    TestUtils.validateResults(analysisResults, 6);
+    TestUtils.validateResults(analysisResults, 8);
```
rchiodo: > **Info** · Optional note

The aggregate diagnostic count cannot show that errors are reported on the nested calls and remain absent from valid expressions. Add focused assertions that distinguish each positive and negative scenario.

packages/pyright-internal/src/tests/samples/super1.py
```diff
@@ -47,8 +47,17 @@ def __init__(self):
...
         def inner():
+            # This should generate an error because the zero-arg form
+            # of super is illegal in a nested function.
             super().method1()
 
+        # This should generate an error because the zero-arg form
+        # of super is illegal in a lambda.
+        lambda: super().method1()
+
+        # Comprehensions execute in the enclosing method and are valid.
```
rchiodo: > **Info** · Optional note

The described behavior includes nested `async def`, but this regression sample covers only `def` and lambda. Add an `async def` case to protect that behavior.

packages/pyright-internal/src/analyzer/parseTreeUtils.ts
```diff
@@ -704,6 +704,65 @@ export function getEnclosingFunction(node: ParseNode): FunctionNode | undefined
...
+        if (curNode.nodeType === ParseNodeType.Lambda) {
+            return false;
+        }
+
+        if (curNode.nodeType === ParseNodeType.Function) {
+            // Don't treat a decorator as being "enclosed" in the function.
+            if (curNode.d.decorators.some((decorator) => decorator === prevNode)) {
+                prevNode = curNode;
+                curNode = curNode.parent;
+                continue;
```
rchiodo: > **Issue** · Please address or respond

Can this preserve nested callables that have an explicit first parameter? `super()` uses the current frame's first argument plus the inherited `__class__` cell, so a nested `def inner(obj): return super()` invoked as `inner(self)` can succeed. Treating every nested function or lambda as illegal produces a false diagnostic for this valid form; please model that case and add coverage.

packages/pyright-internal/src/analyzer/parseTreeUtils.ts
```diff
@@ -704,6 +704,65 @@ export function getEnclosingFunction(node: ParseNode): FunctionNode | undefined
...
+            }
+
+            let parent: ParseNode | undefined = curNode.parent;
+            while (parent) {
```
rchiodo: > **Issue** · Please address or respond

Does this distinguish a nested function's default-value evaluation from its body? A default such as `def inner(value=super()): ...` declared inside a method is evaluated in the enclosing method's frame, but this walk reaches `inner` and returns `false`. Please classify parameter defaults by their enclosing evaluation scope so this valid use is not diagnosed.

packages/pyright-internal/src/tests/typeEvaluator2.test.ts
```diff
@@ -177,7 +183,7 @@ test('AugmentedAssignment3', () => {
 test('Super1', () => {
     const analysisResults = TestUtils.typeAnalyzeSampleFiles(['super1.py']);
 
-    TestUtils.validateResults(analysisResults, 6);
+    TestUtils.validateResults(analysisResults, 10);
```
rchiodo, line 186: > **Warning** · Non-blocking recommendation

📍 packages/pyright-internal/src/tests/typeEvaluator2.test.ts:183
The prior test-precision feedback remains unresolved: an aggregate count can pass if a missing diagnostic is offset by one on a valid expression. Add range-specific assertions, or split valid and invalid scenarios into separately analyzed samples, so each new case is independently protected.

[verified]

rchiodo (approved):
Approved via Review Center.

heejaechang (approved):
Approved via Review Center.

rchiodo:

bschnurr (approved):
Approved via Review Center.

heejaechang:
### 🔴 Pyright CLI QA — `01c4ed9c` — verdict: red

14 focused check(s) completed.

#### Needs review: `reportGeneralTypeIssues` newly reported

The PR reports `reportGeneralTypeIssues` for this example; the base does not. Is this change intended?

**`main.py`**

```python
from typing import assert_type, reveal_type

class Base:
    def value(self) -> int:
        return 42

class Child(Base):
    def __iter__(self) -> "Child":
        return self

    def __next__(self) -> int:
        return 0

    def value(self) -> int:
        # The generator frame receives this Child instance as its iterator.
        return next(super().value() for _ in self)

assert_type(Child().value(), int)
assert Child().value() == 42
reveal_type(Child().value())
```

**`pyrightconfig.json`**

```json
{
  "include": [
    "main.py"
  ],
  "typeCheckingMode": "basic",
  "pythonVersion": "3.13"
}
```

```text
```

**Base `789d8275fef25f347ffef7b847305fefd8a3e363`**

```json
{
  "exit_code": 0,
  "diagnostics": [
    {
      "file": "main.py",
      "severity": "information",
      "rule": null,
      "message": "Type of \"Child().value()\" is \"int\"",
      "range": {
        "start": [
          19,
          12
        ],
        "end": [
          19,
          27
        ]
      }
    }
  ]
}
```

**PR `01c4ed9cf2c7ba8d784ccc1a448804689b6229cc`**

```json
{
  "exit_code": 1,
  "diagnostics": [
    {
      "file": "main.py",
      "severity": "information",
      "rule": null,
      "message": "Type of \"Child().value()\" is \"int\"",
      "range": {
        "start": [
          19,
          12
        ],
        "end": [
          19,
          27
        ]
      }
    },
    {
      "file": "main.py",
      "severity": "error",
      "rule": "reportGeneralTypeIssues",
      "message": "Zero-argument form of \"super\" call is valid only within a method",
      "range": {
        "start": [
          15,
          20
        ],
        "end": [
          15,
          25
        ]
      }
    }
  ]
}
```

</details>

PR #11645, opened 2026-08-16
Do not iterate via __getitem__ when __iter__ is explicitly set

### LINKED ISSUE #11220: Instances of classes that define `__iter__ = None` and `__getitem__` should not be iterable.
**Describe the bug**
Pyright still allows instances of a class to be iterable if the class defines `__getitem__` even if the class also defines `__iter__ = None`.

Attempting to iterate over an instance of such a class will lead to a runtime `TypeError`.

Both mypy and ty report this as an error:
- https://play.ty.dev/f5d9047c-0928-4560-a52e-8a8f8348e7ac
- https://mypy-play.net/?mypy=latest&python=3.12&gist=3102320bfc49a1c80cee68d587b5f6c1

**Code or Screenshots**

```python
class Foo:
    def __getitem__(self, item: int) -> None:
        return

    __iter__ = None

for _ in Foo():
    pass
```

As a Node binary - pyright@1.1.391

hsusul:
## Summary
- Fixes #11220.
- CPython treats an object as not iterable when `__iter__` is present but not callable (`__iter__ = None`), even if `__getitem__` exists. Pyright still used the legacy sequence protocol and accepted `for x in Foo()`.
- `__getitem__` fallback now applies only when `__iter__` is missing.

## Test plan
- [x] `npx jest typeEvaluator3.test.ts -t ForLoop --forceExit` (from `packages/pyright-internal`)
- [x] Runtime: `for _ in Foo()` with `__iter__ = None` raises `TypeError`; a `__getitem__`-only class still iterates
- [ ] Confirm a `for` loop over such a class reports a not-iterable error in the language server

### DIFF
--- a/packages/pyright-internal/src/analyzer/typeEvaluator.ts
+++ b/packages/pyright-internal/src/analyzer/typeEvaluator.ts
@@ -3402,9 +3402,22 @@ export function createTypeEvaluator(
                 const iterReturnType = getTypeOfMagicMethodCall(subtype, iterMethodName, [], errorNode)?.type;
 
                 if (!iterReturnType) {
-                    // There was no __iter__. See if we can fall back to
-                    // the __getitem__ method instead.
+                    // There was no callable __iter__. The legacy sequence protocol
+                    // falls back to __getitem__, but CPython does that only when
+                    // __iter__ is missing. An explicit `__iter__ = None` (or any
+                    // non-callable __iter__) makes the object not iterable.
+                    let hasExplicitIterMember = false;
                     if (!isAsync && isClassInstance(subtype)) {
+                        hasExplicitIterMember = !!lookUpObjectMember(
+                            subtype,
+                            iterMethodName,
+                            MemberAccessFlags.SkipInstanceMembers |
+                                MemberAccessFlags.SkipAttributeAccessOverride |
+                                MemberAccessFlags.SkipObjectBaseClass
+                        );
+                    }
+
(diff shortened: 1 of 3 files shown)


rchiodo:

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -3402,9 +3402,22 @@ export function createTypeEvaluator(
...
+                    // non-callable __iter__) makes the object not iterable.
+                    let hasExplicitIterMember = false;
                     if (!isAsync && isClassInstance(subtype)) {
+                        hasExplicitIterMember = !!lookUpObjectMember(
+                            subtype,
+                            iterMethodName,
+                            MemberAccessFlags.SkipInstanceMembers |
+                                MemberAccessFlags.SkipAttributeAccessOverride |
+                                MemberAccessFlags.SkipObjectBaseClass
+                        );
```
rchiodo, line 3417: > **Warning** · Non-blocking recommendation

The parallel Pylance async evaluator still unconditionally falls back to `__getitem__`, which can produce mode-dependent diagnostics. Mirror this discriminator in `packages/pylance-internal/src/analyzer/typeEvaluator.ts` and add async-mode diagnostic coverage.

rchiodo (approved):
Approved via Review Center.

rchiodo (approved):
Approved via Review Center.

heejaechang (approved):
Approved via Review Center.

rchiodo:

PR #11652, opened 2026-08-18
Migrate CI package installation to pnpm

heejaechang:
Migrate package installation and CI workflows to pnpm.

Validation:
- `pnpm install --frozen-lockfile --ignore-scripts`
- `pnpm run check`
- `pnpm run typecheck`
- `npm pack --dry-run` for `pyright` and `pyright-typeserver`
- `pnpm run package` for `vscode-pyright`

REVIEWER: Review only the pnpm package and CI migration required by this PR; do not leave unrelated feedback.


rchiodo:

build/azuredevops/azure-pipelines.yml
```diff
@@ -43,20 +47,21 @@ extends:
                 submodules: true
                 fetchTags: true
                 persistCredentials: True
-              - task: NodeTool@0
+              # Generate + authenticate the CFS feed .npmrc at runtime (ICM 839806358).
+              - template: /build/templates/npmAuthenticate.yml@self
+              - task: UseNode@1
                 displayName: Use Node 24.15.0
```
rchiodo, line 53: > **Warning** · Non-blocking recommendation

`UseNode@1` does not explicitly provision pnpm, unlike the GitHub workflows. Add a pinned Corepack/pnpm setup step before the first `pnpm` command in both Azure pipelines so clean agent pools do not depend on a preinstalled pnpm.

.github/actions/standard-npm-install/action.yml
```diff
@@ -1,15 +1,20 @@
...
 runs:
   using: composite
   steps:
-    - uses: ./.github/actions/npm-cache-dir
-
-    - name: Standard npm install
+    - name: Standard pnpm install
       shell: bash
-      run: npm run install:all
+      working-directory: ${{ inputs.working-directory }}
```
rchiodo, line 17: > **Warning** · Non-blocking recommendation

This action accepts `working-directory` but runs the root-only `install:all` script. A non-root caller will therefore behave differently from the CFS path, which runs `pnpm install` directly. Use the same install command in both actions so the input contract works consistently.

rchiodo (approved):
Approved via Review Center.

build/azuredevops/azure-pipelines.yml
heejaechang: this PR just takes what is in pyright, we don't make change in this PR that make things to diverge from pylance.

.github/actions/standard-npm-install/action.yml
heejaechang: this PR just takes what is in pyright, we don't make change in this PR that make things to diverge from pylance.

PR #11654, opened 2026-08-18
Reject callable assignment when dest keyword form is incompatible

### LINKED ISSUE #10237: False negative for callable assignability
**Describe the bug**

Pyright reports a false negative when checking assignability between two callable types when the RHS has variadic and keyword-variadic parameters while the LHS has only standard parameter. Refer to the code example below.

**Code or Screenshots**

```python
from typing import Protocol


class A(Protocol):
    def __call__(self, *args: int, **kwargs: bool) -> None: ...


class B(Protocol):
    def __call__(self, a: int) -> None: ...


def _(a: A):
    # This should be not assignable
    x: B = a
```

The reason is that the positional form of the standard parameter is matched against the variadic parameter and is assignable but the keyword form of the standard parameter is not assignable to kwargs because `int` is not assignable to `bool`.

Here, `x(a=10)` is not a valid call as the type of keyword-variadic is a `bool`.

mypy does report this as an error:
```
mypy: Incompatible types in assignment (expression has type "A", variable has type "B") 
    Following member(s) of "A" have conflicts:
        Expected:
            def __call__(self, a: int) -> None
        Got:
            def __call__(self, *args: int, **kwargs: bool) -> None [assignment]
```

Command-line with version `1.1.398`

hsusul:
## Summary
- A dest positional-or-keyword parameter can be passed by name, so callable assignability must check that form against the source `**kwargs` (or a same-named keyword-only parameter), not only against `*args`.
- This rejects cases like `(*args: int, **kwargs: bool)` assigned to `(a: int)`, which is invalid because the dest may be called as `f(a=10)`.

Fixes #10237

## Test plan
- [x] `pnpm exec jest typeEvaluator2.test.ts -t "CallbackProtocol" --forceExit` in `packages/pyright-internal`
- [x] `pnpm exec jest typeEvaluator4.test.ts -t "FunctionAssignability" --forceExit`

### DIFF
--- a/packages/pyright-internal/src/analyzer/typeEvaluator.ts
+++ b/packages/pyright-internal/src/analyzer/typeEvaluator.ts
@@ -28393,7 +28393,14 @@ export function createTypeEvaluator(
                         return;
                     }
 
-                    const destParamInfo = destParamMap.get(srcParamInfo.param.name);
+                    const destParamInfo =
+                        destParamMap.get(srcParamInfo.param.name) ??
+                        destParamDetails.params.find(
+                            (paramInfo) =>
+                                paramInfo.param.name === srcParamInfo.param.name &&
+                                paramInfo.kind === ParamKind.Standard &&
+                                paramInfo.param.category === ParamCategory.Simple
+                        );
                     const paramDiag = diag?.createAddendum();
                     const srcParamType = srcParamInfo.type;
 
@@ -28519,6 +28526,43 @@ export function createTypeEvaluator(
                     diag?.createAddendum().addMessage(
                         LocAddendum.namedParamMissingInSource().format({ name: paramName })
                     );
+                    canAssign = false;
+                }
+            });
+
+            // Positional-or-keyword dest parameters can also be passed by name.
+            // If the source has no matching named parameter, the keyword form must
+            // be compatible with the source **kwargs (when present).
+            destParamDetails.params.forEach((destParamInfo) => {
+                if (
+                    destParamInfo.kind !== ParamKind.Standard ||
+                    !destParamInfo.param.name ||
+                    destParamInfo.param.category !== ParamCategory.Simple
+                ) {
+                    return;
+                }
+
+                const srcHasNamed = srcParamDetails.params.some(
+                    (srcParamInfo) =>
+                        srcParamInfo.param.name === destParamInfo.param.name &&
+                        srcParamInfo.param.category === ParamCategory.Simple &&
+                        srcParamInfo.kind !== ParamKind.Positional
+                );
+                if (srcHasNamed || srcParamDetails.kwargsIndex === undefined) {
+                    return;
+                }
+
+                if (
+                    !assignParam(
+                        destParamInfo.type,
+                        srcParamDetails.params[srcParamDetails.kwargsIndex].type,
+                        destParamInfo.index,
+                        diag?.createAddendum(),
+                        constraints,
+                        flags,
+                        recursionCount
+                    )
+                ) {
                     canAssign = false;
                 }
             });
--- a/packages/pyright-internal/src/tests/samples/callbackProtocol12.py
+++ b/packages/pyright-internal/src/tests/samples/callbackProtocol12.py
@@ -1,0 +1,37 @@
+# This sample tests that a callable with *args/**kwargs is not assignable
+# to a callable whose positional-or-keyword parameter cannot be passed
+# by keyword to the source (typing spec callable assignability).
+
+from typing import Protocol
+
+
+class AcceptsArgsKwargs(Protocol):
+    def __call__(self, *args: int, **kwargs: bool) -> None: ...
+
+
+class AcceptsKeywordOrPositional(Protocol):
+    def __call__(self, a: int) -> None: ...
+
+
+def func1(cb: AcceptsArgsKwargs):
+    # This should generate an error because AcceptsKeywordOrPositional
+    # can be called as cb(a=10), which is not valid for **kwargs: bool.
+    x: AcceptsKeywordOrPositional = cb
+
+
+class AcceptsKeywordOnly(Protocol):
+    def __call__(self, *args: int, a: bool = False) -> None: ...
+
+
+def func2(cb: AcceptsKeywordOnly):
+    # This should generate an error because the keyword form of
+    # parameter "a" has type int on the dest and bool on the source.
+    y: AcceptsKeywordOrPositional = cb
+
+
+def ok_cb(a: int) -> None:
+    pass
+
+
+def func3(cb: AcceptsKeywordOrPositional):
+    z: AcceptsKeywordOrPositional = ok_cb
--- a/packages/pyright-internal/src/tests/typeEvaluator2.test.ts
+++ b/packages/pyright-internal/src/tests/typeEvaluator2.test.ts
@@ -79,6 +79,12 @@ test('CallbackProtocol11', () => {
     TestUtils.validateResults(analysisResults, 0);
 });
 
+test('CallbackProtocol12', () => {
+    const analysisResults = TestUtils.typeAnalyzeSampleFiles(['callbackProtocol12.py']);
+
+    TestUtils.validateResults(analysisResults, 2);
+});
+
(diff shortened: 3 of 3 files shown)


heejaechang:

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -28523,6 +28530,43 @@ export function createTypeEvaluator(
...
+                        destParamInfo.type,
+                        srcParamDetails.params[srcParamDetails.kwargsIndex].type,
+                        destParamInfo.index,
+                        diag?.createAddendum(),
+                        constraints,
+                        flags,
+                        recursionCount
+                    )
+                ) {
+                    canAssign = false;
```
heejaechang, line 28566: > **Issue** · Please address or respond

The corresponding Pylance async evaluator still contains the pre-fix matching logic. Mirror this change and add async-mode coverage, or demonstrate that the async TSP path owns this behavior.

[verified]

packages/pyright-internal/src/tests/samples/callbackProtocol12.py
```diff
@@ -0,0 +1,37 @@
...
+    # parameter "a" has type int on the dest and bool on the source.
+    y: AcceptsKeywordOrPositional = cb
+
+
+def ok_cb(a: int) -> None:
+    pass
+
+
+def func3(cb: AcceptsKeywordOrPositional):
+    z: AcceptsKeywordOrPositional = ok_cb
```
rchiodo, line 37: Could we add positive cases that exercise each newly accepted path? Assigning `(*args: int, **kwargs: int)` to `(a: int)` covers the new `**kwargs` fallback, and assigning `(*args: int, a: int)` to `(a: int)` covers the named keyword-only lookup. The current `ok_cb` case is an exact-signature assignment and would pass before this change.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -28523,6 +28530,43 @@ export function createTypeEvaluator(
                 }
```
rchiodo, line 28530: This check needs to account for overload-overlap mode. For example, `(a: int) -> int` and `(*args: int, **kwargs: bool) -> str` still overlap for the positional call `f(1)`, even though their keyword forms do not. During `PartialOverloadOverlap`, this unconditional failure makes `assignFunction` return false and suppresses the incompatible-return diagnostic. Could we model the positional and keyword paths disjunctively (or otherwise preserve an overlap once a shared positional call exists) and add a regression case to `overloadOverlap1.py`?

crypt0lith:
I opened microsoft#11670 for this same issue; it keeps the `destParamInfo` fallback from this PR (@hsusul is co-authored on the commit), but the rest of microsoft#11670's implementation is **fundamentally in conflict with this PR**. It repairs the two existing sites where `**kwargs`-assignment previously bailed rather than using a separate sweep over dest params. And it gates the new failures on `PartialOverloadOverlap` per @rchiodo's comment.

Opening another PR and flagging here seemed more straightforward than framing it as an addendum. Thank you for your time.

heejaechang:
### 🔴 Pyright CLI QA — `c7cd518b` — verdict: red

16 focused check(s) completed.

#### Needs review: `reportArgumentType` no longer reported

The base reports `reportArgumentType` for this example; the PR does not. Is this change intended?

**`main.py`**

```python
from typing import Protocol

class Callback(Protocol):
    def __call__(self, a: int) -> None: ...

def source(*args: int, a: int) -> None:
    pass

def consume(callback: Callback) -> None:
    callback(1)

# Must be rejected: source(1) is missing the required keyword-only argument a.
consume(source)
```

**`pyrightconfig.json`**

```json
{"include": ["*.py"], "typeCheckingMode": "strict", "pythonVersion": "3.13"}
```

```text
```

**Base `5cf4afa88b491795af0a2bfd1683ec24c60dcf03`**

```json
{
  "exit_code": 1,
  "diagnostics": [
    {
      "file": "main.py",
      "severity": "error",
      "rule": "reportArgumentType",
      "message": "Argument of type \"(*args: int, a: int) -> None\" cannot be assigned to parameter \"callback\" of type \"Callback\" in function \"consume\"\n  Type \"(*args: int, a: int) -> None\" is not assignable to type \"(a: int) -> None\"\n    Extra parameter \"a\"",
      "range": {
        "start": [
          12,
          8
        ],
        "end": [
          12,
          14
        ]
      }
    }
  ]
}
```

**PR `c7cd518bf5759a73477b96b15cb016bb0e446983`**

```json
{
  "exit_code": 0,
  "diagnostics": []
}
```

</details>

#### Needs review: `reportOverlappingOverload` no longer reported

The base reports `reportOverlappingOverload` for this example; the PR does not. Is this change intended?

**`main.py`**

```python
from typing import overload, reveal_type

# Keyword incompatibility does not make the positional call select(1) disjoint.
# These overloads must produce reportOverlappingOverload.
@overload
def select(a: int) -> str: ...
@overload
def select(*args: int, **kwargs: bool) -> int: ...
def select(*args: object, **kwargs: object) -> str | int:
    raise NotImplementedError

reveal_type(select(1))
reveal_type(select(a=1))
reveal_type(select(b=True))

# Unknown tuple length selects the second overload, although a one-item tuple
# also matches the first overload with its incompatible return type.
def through_tuple(values: tuple[int, ...]) -> int:
    reveal_type(select(*values))
    return select(*values)
```

**`pyrightconfig.json`**

```json
{"include": ["*.py"], "typeCheckingMode": "strict", "pythonVersion": "3.13"}
```

```text
```

**Base `5cf4afa88b491795af0a2bfd1683ec24c60dcf03`**

```json
{
  "exit_code": 1,
  "diagnostics": [
    {
      "file": "main.py",
      "severity": "error",
      "rule": "reportOverlappingOverload",
      "message": "Overload 1 for \"select\" overlaps overload 2 and returns an incompatible type",
      "range": {
        "start": [
          5,
          4
        ],
        "end": [
          5,
          10
        ]
      }
    },
    {
      "file": "main.py",
      "severity": "information",
      "rule": null,
      "message": "Type of \"select(*values)\" is \"int\"",
      "range": {
        "start": [
          18,
          16
        ],
        "end": [
          18,
          31
        ]
      }
    },
    {
      "file": "main.py",
      "severity": "information",
      "rule": null,
      "message": "Type of \"select(1)\" is \"str\"",
      "range": {
        "start": [
          11,
          12
        ],
        "end": [
          11,
          21
        ]
      }
    },
    {
      "file": "main.py",
      "severity": "information",
      "rule": null,
      "message": "Type of \"select(a=1)\" is \"str\"",
      "range": {
        "start": [
          12,
          12
        ],
        "end": [
          12,
          23
        ]
      }
    },
    {
      "file": "main.py",
      "severity": "information",
      "rule": null,
      "message": "Type of \"select(b=True)\" is \"int\"",
      "range": {
        "start": [
          13,
          12
        ],
        "end": [
          13,
          26
        ]
      }
    }
  ]
}
```

**PR `c7cd518bf5759a73477b96b15cb016bb0e446983`**

```json
{
  "exit_code": 0,
  "diagnostics": [
    {
      "file": "main.py",
      "severity": "information",
      "rule": null,
      "message": "Type of \"select(*values)\" is \"int\"",
      "range": {
        "start": [
          18,
          16
        ],
        "end": [
          18,
          31
        ]
      }
    },
    {
      "file": "main.py",
      "severity": "information",
      "rule": null,
      "message": "Type of \"select(1)\" is \"str\"",
      "range": {
        "start": [
          11,
          12
        ],
        "end": [
          11,
          21
        ]
      }
    },
    {
      "file": "main.py",
      "severity": "information",
      "rule": null,
      "message": "Type of \"select(a=1)\" is \"str\"",
      "range": {
        "start": [
          12,
          12
        ],
        "end": [
          12,
          23
        ]
      }
    },
    {
      "file": "main.py",
      "severity": "information",
      "rule": null,
      "message": "Type of \"select(b=True)\" is \"int\"",
      "range": {
        "start": [
          13,
          12
        ],
        "end": [
          13,
          26
        ]
      }
    }
  ]
}
```

</details>

PR #11655, opened 2026-08-18
Retain Sentinel types for non-CONSTANT_NAME bindings

### LINKED ISSUE #10744: Sentinel instances only work as constants
**Describe the bug**

There's some unexpected behavior in v1.1.402 regarding `Sentinel` naming. The desired type inference only works if the sentinel has an uppercased name, which is treated as a constant. As a variable that is not treated as a constant, we run into the following:

```
Cannot assign to attribute "email" for class "Address"
  Type "str | Empty" is not assignable to type "str"
    "Empty" is not assignable to "str"  (reportAttributeAccessIssue)
```

Code sample in [pyright playground](https://pyright-play.net/?enableExperimentalFeatures=true&reportMissingModuleSource=true&code=GYJw9gtgBAJghgFzgYwDZwM4YKYagSwgAcwQFZEV0sAoUSKBATyPwDsBzAfWwA8FsbDPjBCCxUuQDKghO2yoaSgLIBJKVNUA5AOJQAvFBls5bBQAoARGo3adlgJQ0AosWYGjs%2BaiuuizRyUaAAF4JDRMDBoIrCgAQRgYEFwMAC4aKEyobAg4fFRUqAwEEAystjgIbELikCgAHygtUWwgmGxgKABXIjDsLjYwBC4IfCx2DnMyzLhE5KxChKSUgBooaezc-JqShqgbTV01jYqqnbrG5rM9g7saBygAWgA%2BJpb0rIJOnLzUAjxBuRbroPp8srNllgAHQ-fIeWGKT74TqnbD-KCA-bqQ46UFgqAQ%2BYYKGojyotodbq9RD9QE8NxMKafQkpRZzVbrT4I857PzMY6fVE8y4tXkM%2B5PV5XaobZGbX7ozF8ph4sEs6EI%2BFbRFZOWksYYoZQZWq5ns6GkwzkoA)

> The reason it works for you when you use an all-caps name is that pyright treats variables with all-caps names as a constant and prevents them from being reassigned. This is generally important for `x is y` and `x is not y` narrowing. However, in the case of sentinels, this constant requirement isn't needed for soundness. 
> 
> Could you please file a bug report so this is tracked? 
> 
> In the meantime, you can work around the issue by using a `Final` type qualifier.
> 
> ```python
> Empty: Final = Sentinel("Empty")
> ``` 

 _Originally posted by @erictraut in [#10565](https://github.com/microsoft/pyright/issues/10565#issuecomment-3097011053)_

**Code or Screenshots**

```python
from dataclasses import dataclass
from typing_extensions import Sentinel


MISSING = Sentinel("MISSING")
Empty = Sentinel("Empty")


@dataclass
class Address:
    email: str
    name: str | None


def update_not_missing(
    address: Address, 
    email: str | MISSING, 
    name: str | None | MISSING
) -> None:
    if email is not MISSING:
        address.email = email
    if name is not MISSING:
        address.name = name


def update_not_empty(
    address: Address, 
    email: str | Empty, 
    name: str | None | Empty
) -> None:
    if email is not Empty:
        address.email = email
    if name is not Empty:
        address.name = name

```

In VS Code: Pylance language server 2025.6.2 (pyright version 1.1.402, commit 9496ae00)

hsusul:
## Summary
- Sentinel objects are unique singletons, so pyright can keep their literal types even when the binding is not a `CONSTANT_NAME` or `Final`.
- That makes `is` / `is not` narrowing work for names like `Empty = Sentinel("Empty")`, matching the behavior already used for all-caps sentinels.

Fixes #10744

## Test plan
- [x] `pnpm exec jest typeEvaluator2.test.ts -t "Sentinel" --forceExit` in `packages/pyright-internal`

### DIFF
--- a/packages/pyright-internal/src/analyzer/typeEvaluator.ts
+++ b/packages/pyright-internal/src/analyzer/typeEvaluator.ts
@@ -4011,7 +4011,8 @@ export function createTypeEvaluator(
                 if (
                     TypeBase.isInstance(destType) &&
                     !isConstantName(nameValue) &&
-                    !isFinalVariable(symbolWithScope.symbol)
+                    !isFinalVariable(symbolWithScope.symbol) &&
+                    !isSentinelLiteral(destType)
                 ) {
                     destType = stripTypeForm(stripLiteralValue(destType));
                 }
@@ -24654,6 +24655,12 @@ export function createTypeEvaluator(
                                 ClassType.isEnumClass(type) &&
                                 isDeclInEnumClass(evaluatorInterface, decl)
                             ) {
+                                isConstant = true;
+                            }
+
+                            // Sentinel objects are singletons; retaining the literal is sound
+                            // even when the binding is not a CONSTANT_NAME or Final.
+                            if (isSentinelLiteral(type)) {
                                 isConstant = true;
                             }
 
--- a/packages/pyright-internal/src/tests/samples/sentinel3.py
+++ b/packages/pyright-internal/src/tests/samples/sentinel3.py
@@ -1,0 +1,44 @@
+# This sample tests that Sentinel values retain their literal type
+# even when the binding is not a CONSTANT_NAME.
+
+from dataclasses import dataclass
+from typing_extensions import Sentinel  # pyright: ignore[reportMissingModuleSource]
+
+
+Empty = Sentinel("Empty")
+MISSING = Sentinel("MISSING")
+
+
+def func1(value: int | Empty) -> None:
+    if value is Empty:
+        reveal_type(value, expected_text="Empty")
+    else:
+        reveal_type(value, expected_text="int")
+
+
+def func2(value: int | Empty) -> None:
+    if value is not Empty:
+        reveal_type(value, expected_text="int")
+    else:
+        reveal_type(value, expected_text="Empty")
+
+
+@dataclass
+class Address:
+    email: str
+    name: str | None
+
+
+def update_not_empty(address: Address, email: str | Empty, name: str | None | Empty) -> None:
+    if email is not Empty:
+        address.email = email
+        reveal_type(email, expected_text="str")
+    if name is not Empty:
+        address.name = name
+        reveal_type(name, expected_text="str | None")
+
+
+def update_not_missing(address: Address, email: str | MISSING) -> None:
+    if email is not MISSING:
+        address.email = email
+        reveal_type(email, expected_text="str")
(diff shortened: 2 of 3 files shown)


heejaechang:

packages/pyright-internal/src/tests/samples/sentinel3.py
```diff
@@ -0,0 +1,44 @@
+# This sample tests that Sentinel values retain their literal type
+# even when the binding is not a CONSTANT_NAME.
+
+from dataclasses import dataclass
+from typing_extensions import Sentinel  # pyright: ignore[reportMissingModuleSource]
+
+
+Empty = Sentinel("Empty")
```
heejaechang, line 8: > **Warning** · Non-blocking recommendation

📍 packages/pyright-internal/src/tests/samples/sentinel3.py:8
[unverified] The new policy relies on singleton identity while this non-`Final` binding remains mutable. Add adversarial coverage for reassignment and aliasing—and preferably lowercase local and module-attribute bindings—to establish that narrowing remains sound across the affected flow paths.

[verified]

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -24657,6 +24658,12 @@ export function createTypeEvaluator(
                                 isConstant = true;
                             }
 
+                            // Sentinel objects are singletons; retaining the literal is sound
+                            // even when the binding is not a CONSTANT_NAME or Final.
+                            if (isSentinelLiteral(type)) {
```
heejaechang, line 24663: > **Warning** · Non-blocking recommendation

📍 packages/pyright-internal/src/analyzer/typeEvaluator.ts:24658
[unverified] The corresponding Pylance async-evaluator widening branches do not contain this exception, so IDE behavior may diverge when async evaluation is enabled. Verify the scenario with a downstream async-mode Pylance regression test and mirror the evaluator change only if that path owns the behavior rather than TSP.

[verified]

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -4011,7 +4011,8 @@ export function createTypeEvaluator(
                 if (
                     TypeBase.isInstance(destType) &&
                     !isConstantName(nameValue) &&
-                    !isFinalVariable(symbolWithScope.symbol)
+                    !isFinalVariable(symbolWithScope.symbol) &&
+                    !isSentinelLiteral(destType)
                 ) {
                     destType = stripTypeForm(stripLiteralValue(destType));
                 }
```
heejaechang, line 4018: > **Warning** · Non-blocking recommendation

The new sample exercises module-level names but not the member-access widening path modified here. Add a class or instance sentinel-member regression case so this distinct retention path is covered.

heejaechang (approved):
Approved via Review Center.

rchiodo (approved):
Approved via Review Center.

bschnurr (approved):
Approved via Review Center.

PR #11662, opened 2026-08-19
Fix reflected operator priority when the right operand is a subclass

### LINKED ISSUE #11515: Type narrowing not done correctly for bitwise operations with `int` and instance of subclass of `enum.IntFlag`
**Describe the bug**

When &-ing an `int` and a member of an `enum.IntFlag`, another member of the flags is returned, but Pyright reveals it as an `int`. I believe this isn't an issue with typeshed, since mypy [has the correct behaviour](https://gist.github.com/mypy-play/0c29d5580d780221622b7d8af6221996).

**Code or Screenshots**

Code sample in [pyright playground](https://pyright-play.net/?code=GYJw9gtgBApgdgV2gSwgBzCALlAhgrMAGigEk4sAxAG1wHMBYAKAGNaBndqG%2B9gCnJVadAJQAuZlClQAglAC8eAmD4jJ0gEIKlhVcxAwAbjFzUA%2BlgCeaGH2QVVUAGTdh7AHQyRQA)

```python
from enum import auto, IntFlag
class Flags(IntFlag):
    A = auto()
    B = auto()
reveal_type(int() & Flags.A)
```

nithin42:
Fixes #11515

### Root cause

There are two separate problems behind the reported symptom.

**1. The reflected form of the operator is never given priority.**

Python evaluates `x op y` by calling `type(x).__op__` first, except in one case: if `type(y)` is a proper subclass of `type(x)` and it overrides the reflected method `__rop__`, then `type(y).__rop__` is called first. `validateArithmeticOperation` in `operations.ts` always tried the left operand's method first and only fell back to the reflected form if that failed. Since `int.__and__` accepts any `int`, it always won, and the more precise result from `IntFlag.__rand__` was never considered.

**2. An undeclared class member does not override a declared method in a base class.**

`getTypeOfClassMemberName` looks for a member with a declared type first, walking the whole MRO before falling back to a lookup that also accepts inferred types. When a derived class defines a member through a plain assignment, that member has no declared type, so the lookup skipped it and returned a declared member from a base class instead.

typeshed defines `IntFlag`'s reflected methods with the alias idiom:

```python
class IntFlag(int, ReprEnum, Flag, boundary=KEEP):
    def __and__(self, other: int) -> Self: ...
    __rand__ = __and__
```

`int` already declares `__rand__`, so `Flags.A.__rand__` resolved to `int.__rand__` (`(value: int, /) -> int`) rather than the alias. Even with the operator ordering fixed, the reflected call would still have produced `int`.

### Fix

`operations.ts` now checks, before calling the left operand's magic method, whether the right operand's type is a proper subclass of the left operand's type and overrides the reflected magic method. If so, the reflected method is tried first, with the existing evaluation order used as a fallback so nothing regresses when the reflected call doesn't apply.

`typeEvaluator.ts` uses the existing but previously unused `skippedUndeclaredType` flag on `ClassMember`. When the declared-types-only lookup skipped an undeclared symbol in a class earlier in the MRO and the declared symbol it found is a method, the overriding symbol is used instead. This is deliberately limited to methods so that `reportIncompatibleVariableOverride` behavior for variable overrides is unchanged.

### Verification

New sample `operator13.py` registered in `typeEvaluator8.test.ts` covers:

- `int() & Flags.A`, `|` and `^` now evaluate to `Flags`
- `int() & int()` still evaluates to `int`
- a subclass that overrides `__radd__` takes priority over the base class `__add__`
- a subclass that does not override `__radd__` keeps the normal evaluation order
- a right operand that is not a subclass keeps the normal evaluation order
- a method alias assignment overriding a base class method of the same name

Reverting either source change makes the new test fail. With both applied, all 1285 tests across the `typeEvaluator*` and `checker` suites pass with no regressions.

### DIFF
--- a/packages/pyright-internal/src/analyzer/operations.ts
+++ b/packages/pyright-internal/src/analyzer/operations.ts
@@ -28,7 +28,9 @@ import { evaluateStaticBoolExpression } from './staticExpressions';
 import { EvalFlags, MagicMethodDeprecationInfo, TypeEvaluator, TypeResult } from './typeEvaluatorTypes';
 import {
     InferenceContext,
+    MemberAccessFlags,
     convertToInstantiable,
+    derivesFromClassRecursive,
     getLiteralTypeClassName,
     getTypeCondition,
     getUnionSubtypeCount,
@@ -38,6 +40,7 @@ import {
     isUnboundedTupleClass,
     isUnionableType,
     lookUpClassMember,
+    lookUpObjectMember,
     makeInferenceContext,
     mapSubtypes,
     preserveUnknown,
@@ -1245,6 +1248,60 @@ function validateContainmentOperation(
     return { type, magicMethodDeprecationInfo: deprecatedInfo };
 }
 
+// Python evaluates a binary operation by calling the left operand's magic
+// method first, but there is an exception to this rule: if the right operand's
+// type is a proper subclass of the left operand's type and it overrides the
+// reflected magic method, the reflected method is called first.
+function isReflectedOperatorPrioritized(
+    leftType: Type,
+    rightType: Type,
+    magicMethodName: string,
+    altMagicMethodName: string
+): boolean {
+    if (!isClassInstance(leftType) || !isClassInstance(rightType)) {
+        return false;
+    }
+
+    // The right operand's class must be a proper subclass of the left operand's class.
+    if (ClassType.isSameGenericClass(leftType, rightType)) {
+        return false;
+    }
+
+    if (
+        !derivesFromClassRecursive(
+            ClassType.cloneAsInstantiable(rightType),
+            ClassType.cloneAsInstantiable(leftType),
+            /* ignoreUnknown */ true
+        )
+    ) {
+        return false;
+    }
+
+    // The right operand's class must override the reflected magic method.
+    const rightAltMember = lookUpObjectMember(rightType, altMagicMethodName, MemberAccessFlags.SkipInstanceMembers);
+    if (!rightAltMember) {
+        return false;
+    }
+
+    const leftAltMember = lookUpObjectMember(leftType, altMagicMethodName, MemberAccessFlags.SkipInstanceMembers);
+    if (
+        leftAltMember &&
+        isClass(leftAltMember.classType) &&
+        isClass(rightAltMember.classType) &&
+        ClassType.isSameGenericClass(leftAltMember.classType, rightAltMember.classType)
+    ) {
+        return false;
+    }
+
+    // If the left operand's class doesn't implement the normal form of the
+    // operator, there's nothing to prioritize over.
+    if (!lookUpObjectMember(leftType, magicMethodName, MemberAccessFlags.SkipInstanceMembers)) {
+        return false;
+    }
+
+    return true;
+}
+
 function validateArithmeticOperation(
     evaluator: TypeEvaluator,
     operator: OperatorType,
@@ -1304,7 +1361,29 @@ function validateArithmeticOperation(
                     }
 
                     const magicMethodName = binaryOperatorMap[operator][0];
-                    let resultTypeResult = evaluator.getTypeOfMagicMethodCall(
+                    let resultTypeResult: TypeResult | undefined;
+
+                    // If the right operand's type is a proper subclass of the left
+                    // operand's type and overrides the reflected magic method, Python
+                    // gives the reflected form of the operator priority.
+                    if (
+                        isReflectedOperatorPrioritized(
+                            leftSubtypeExpanded,
+                            rightSubtypeExpanded,
+                            magicMethodName,
+                            binaryOperatorMap[operator][1]
+                        )
+                    ) {
+                        resultTypeResult = evaluator.getTypeOfMagicMethodCall(
+                            convertFunctionToObject(evaluator, rightSubtypeUnexpanded),
+                            binaryOperatorMap[operator][1],
+                            [{ type: leftSubtypeUnexpanded, isIncomplete: leftTypeResult.isIncomplete }],
+                            errorNode,
+                            inferenceContext
+                        );
+                    }
+
+                    resultTypeResult ??= evaluator.getTypeOfMagicMethodCall(
                         convertFunctionToObject(evaluator, leftSubtypeUnexpanded),
                         magicMethodName,
                         [{ type: rightSubtypeUnexpanded, isIncomplete: rightTypeResult.isIncomplete }],
--- a/packages/pyright-internal/src/analyzer/typeEvaluator.ts
+++ b/packages/pyright-internal/src/analyzer/typeEvaluator.ts
@@ -6579,6 +6579,12 @@ export function createTypeEvaluator(
         };
     }
 
+    // Determines whether all of the symbol's typed declarations are methods.
+    function isDeclaredAsMethod(symbol: Symbol): boolean {
+        const decls = symbol.getTypedDeclarations();
+        return decls.length > 0 && decls.every((decl) => decl.type === DeclarationType.Function);
+    }
+
     function getTypeOfClassMemberName(
         errorNode: ExpressionNode | undefined,
         classType: ClassType,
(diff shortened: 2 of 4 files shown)


heejaechang:

packages/pyright-internal/src/analyzer/operations.ts
```diff
@@ -1304,7 +1361,29 @@ function validateArithmeticOperation(
...
+                            magicMethodName,
+                            binaryOperatorMap[operator][1]
+                        )
+                    ) {
+                        resultTypeResult = evaluator.getTypeOfMagicMethodCall(
+                            convertFunctionToObject(evaluator, rightSubtypeUnexpanded),
+                            binaryOperatorMap[operator][1],
+                            [{ type: leftSubtypeUnexpanded, isIncomplete: leftTypeResult.isIncomplete }],
+                            errorNode,
+                            inferenceContext
```
heejaechang, line 1382: > **Issue** · Please address or respond

📍 packages/pyright-internal/src/analyzer/operations.ts:1361
This changes operator dispatch without mirroring the active Pylance async implementation, which remains left-first. The canonical sync/async parity rule makes this non-waivable: update `packages/pylance-internal/src/analyzer/operations.ts` and add equivalent async-mode coverage before integrating this change into pyrx.

[verified]

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -6598,6 +6604,13 @@ export function createTypeEvaluator(
         // a symbol with an inferred type.
         if (!memberInfo) {
             memberInfo = lookUpClassMember(classType, memberName, flags);
+        } else if (memberInfo.skippedUndeclaredType && isDeclaredAsMethod(memberInfo.symbol)) {
+            // A symbol without a declared type was skipped in a class that appears
+            // earlier in the MRO, and the declared symbol we found is a method.
+            // A class-scoped assignment that aliases a method (a common idiom for
+            // defining reflected dunder methods) overrides the base class method,
+            // so use the overriding symbol instead.
```
heejaechang, line 6612: > **Issue** · Please address or respond

📍 packages/pyright-internal/src/analyzer/typeEvaluator.ts:6604
The Pylance async `getTypeOfClassMemberName` counterpart still skips this undeclared-method override correction. Mirror the semantic change there or demonstrate TSP ownership through an async-mode regression test; the documented evaluator-parity requirement cannot be waived per PR.

[verified]

packages/pyright-internal/src/analyzer/operations.ts
```diff
@@ -1245,6 +1248,60 @@ function validateContainmentOperation(
...
+// Python evaluates a binary operation by calling the left operand's magic
+// method first, but there is an exception to this rule: if the right operand's
+// type is a proper subclass of the left operand's type and it overrides the
+// reflected magic method, the reflected method is called first.
+function isReflectedOperatorPrioritized(
+    leftType: Type,
+    rightType: Type,
+    magicMethodName: string,
+    altMagicMethodName: string
+): boolean {
```
heejaechang, line 1260: > **Warning** · Non-blocking recommendation

📍 packages/pyright-internal/src/analyzer/operations.ts:1258
[unverified] Restricting priority detection to class instances may omit class-object operations whose metaclasses have the relevant subclass relationship and reflected override. Add a metaclass `A | B` regression test and, if it fails, perform the comparison using the effective metaclass types.

[verified]

heejaechang:
### 🔴 Pyright CLI QA — `c0913e20` — verdict: red

25 focused check(s) completed.

**`main.py`**

```python

class Base:
    def __add__(self, other: Base) -> int:
        return 1
    def __radd__(self, other: Base) -> str:
        return "reflected"

# Rebinding the same function does not change reflected-operator priority.
class Derived(Base):
    __radd__ = Base.__radd__

def check(left: Base, right: Derived):
    reveal_type(left + right)
    assert_type(left + right, int)
```

**`pyrightconfig.json`**

```json
{
  "include": [
    "main.py"
  ],
  "pythonVersion": "3.13",
}
```

```text
```

**Base `d5e77a7f4856f8e7caee1fa788b25bf67901a3c9`**

```json
{
  "exit_code": 0,
  "diagnostics": [
    {
      "file": "main.py",
      "severity": "information",
      "rule": null,
      "message": "Type of \"left + right\" is \"int\"",
      "range": {
        "start": [
          14,
          16
        ],
        "end": [
          14,
          28
        ]
      }
    }
  ]
}
```

**PR `c0913e2030e2c303e8d45e501f9533a4c79e013e`**

```json
{
  "exit_code": 1,
  "diagnostics": [
    {
      "file": "main.py",
      "severity": "information",
      "rule": null,
      "message": "Type of \"left + right\" is \"str\"",
      "range": {
        "start": [
          14,
          16
        ],
        "end": [
          14,
          28
        ]
      }
    },
    {
      "file": "main.py",
      "severity": "error",
      "message": "\"assert_type\" mismatch: expected \"int\" but received \"str\"",
      "range": {
        "start": [
          15,
          16
        ],
        "end": [
          15,
          28
        ]
      }
    }
  ]
}
```

</details>

**`main.py`**

```python
from types import NotImplementedType

class Base:
    def __add__(self, other: Base) -> int:
        return 1

# NotImplemented must fall back to Base.__add__, which returns int.
class Derived(Base):
    def __radd__(self, other: Base) -> NotImplementedType:
        return NotImplemented

def check(left: Base, right: Derived):
    reveal_type(left + right)
    assert_type(left + right, int)
```

**`pyrightconfig.json`**

```json
{
  "include": [
    "main.py"
  ],
  "pythonVersion": "3.13",
}
```

```text
```

**Base `d5e77a7f4856f8e7caee1fa788b25bf67901a3c9`**

```json
{
  "exit_code": 0,
  "diagnostics": [
    {
      "file": "main.py",
      "severity": "information",
      "rule": null,
      "message": "Type of \"left + right\" is \"int\"",
      "range": {
        "start": [
          14,
          16
        ],
        "end": [
          14,
          28
        ]
      }
    }
  ]
}
```

**PR `c0913e2030e2c303e8d45e501f9533a4c79e013e`**

```json
{
  "exit_code": 1,
  "diagnostics": [
    {
      "file": "main.py",
      "severity": "information",
      "rule": null,
      "message": "Type of \"left + right\" is \"NotImplementedType\"",
      "range": {
        "start": [
          14,
          16
        ],
        "end": [
          14,
          28
        ]
      }
    },
    {
      "file": "main.py",
      "severity": "error",
      "message": "\"assert_type\" mismatch: expected \"int\" but received \"NotImplementedType\"",
      "range": {
        "start": [
          15,
          16
        ],
        "end": [
          15,
          28
        ]
      }
    }
  ]
}
```

</details>

#### Needs review: `reportAttributeAccessIssue` no longer reported

The base reports `reportAttributeAccessIssue` for this example; the PR does not. Is this change intended?

**`main.py`**

```python
# Replacing an inherited method must preserve its parameter and return contract.

class Base:
    def method(self, x: int) -> int:
        return x

class Derived(Base):
    def __init__(self):
        self.method = self.text_method

    def text_method(self, x: str) -> str:
        return x.upper()

def use(value: Base) -> int:
    return value.method(1)

use(Derived())
```

**`pyrightconfig.json`**

```json
{
  "include": [
    "main.py"
  ],
  "pythonVersion": "3.13",
}
```

```text
```

**Base `d5e77a7f4856f8e7caee1fa788b25bf67901a3c9`**

```json
{
  "exit_code": 1,
  "diagnostics": [
    {
      "file": "main.py",
      "severity": "error",
      "rule": "reportAttributeAccessIssue",
      "message": "Cannot assign to attribute \"method\" for class \"Derived*\"\n  Type \"(x: str) -> str\" is not assignable to type \"(x: int) -> int\"\n    Parameter 1: type \"int\" is incompatible with type \"str\"\n      \"int\" is not assignable to \"str\"\n    Function return type \"str\" is incompatible with type \"int\"\n      \"str\" is not assignable to \"int\"",
      "range": {
        "start": [
          9,
          22
        ],
        "end": [
          9,
          38
        ]
      }
    }
  ]
}
```

**PR `c0913e2030e2c303e8d45e501f9533a4c79e013e`**

```json
{
  "exit_code": 0,
  "diagnostics": []
}
```

</details>

#### Needs review: `reportAttributeAccessIssue` no longer reported

The base reports `reportAttributeAccessIssue` for this example; the PR does not. Is this change intended?

**`main.py`**

```python
# An inherited property without a setter remains read-only in a subclass.
class Base:
    @property
    def value(self) -> int:
        return 1

class Derived(Base):
    def __init__(self):
        self.value = 2

Derived()
```

**`pyrightconfig.json`**

```json
{
  "include": [
    "main.py"
  ],
  "pythonVersion": "3.13",
}
```

```text
```

**Base `d5e77a7f4856f8e7caee1fa788b25bf67901a3c9`**

```json
{
  "exit_code": 1,
  "diagnostics": [
    {
      "file": "main.py",
      "severity": "error",
      "rule": "reportAttributeAccessIssue",
      "message": "Cannot assign to attribute \"value\" for class \"Derived*\"\n  \"Literal[2]\" is not assignable to \"property\"",
      "range": {
        "start": [
          8,
          21
        ],
        "end": [
          8,
          22
        ]
      }
    }
  ]
}
```

**PR `c0913e2030e2c303e8d45e501f9533a4c79e013e`**

```json
{
  "exit_code": 0,
  "diagnostics": []
}
```

</details>

PR #11663, opened 2026-08-19
Do not eliminate a TypedDict with extra items on a negative key check

### LINKED ISSUE #11438: TypedDict with extra_items breaks control flow analyis of `in` operator
**Describe the bug**
When checking the existence of a key in a `TypedDict` which has `extra_items` defined, the unreachability analysis seems to assume every key is present.
Testing a normally defined key in the same dict behaves as expected.

**Code or Screenshots**

In the code below, pyright claims the `print` statement is never reached.

```python
from typing import TypedDict

class Typed(TypedDict, extra_items=str):
    a: str

def test(d: Typed) -> None:
    if "b" in d:
        return
    print("here")
```

https://pyright-play.net/?strict=true&code=GYJw9gtgBALgngBwJYDsDmUkQWEMoAqiApgCYAiSAxjAFC1UA2AhgM6uEmkAURCZlGgBooxAB4wQzAPpIYxCKwC8rSQEoAXLSg6ozDVFUh6pYsFjFV3Ugb5k1UALQA%2BKADkwKYlt2ZzAIgAjf0wUKBttX10QYhgAVxAUSJ0EEFQYbn8AC2IY-zVaIA

pyright 1.1409

luantaraschi:
Fixes #11438.

`narrowTypeForTypedDictKey` looks the tested key up in the known items and falls back to the "extra items" entry when it is not there. That entry is built in `typedDicts.ts` with `isProvided` set to true, so on the negative side of an `in` check the key is treated as definitely present and the type is eliminated. Everything after the check becomes unreachable:

```python
class Typed(TypedDict, extra_items=str):
    a: str

def test(d: Typed) -> None:
    if "b" in d:
        return
    print("here")  # not analyzed
```

The same function without `extra_items` keeps the last line reachable, which is what pointed at the fallback.

An `extra_items` declaration constrains the type of the extra keys, not which of them are present, so the absence of `"b"` is consistent with `Typed` and the subtype has to survive the negative branch. A known item behaves as before, and a closed TypedDict is unaffected because its synthesized entry has `isProvided` false.

`typedDictClosed11.py` already states the rule in a comment on `func3`, "Baz allows extra items, so it cannot be eliminated here", with `reveal_type(u, expected_text="Baz")` in the else branch. That assertion is vacuous today: I swapped the expected text for a string that cannot match and the test still passed, because the branch is unreachable and `reveal_type` is never evaluated. With this change the same probe fails with `expected "..." but received "Baz"`, so the branch runs and reveals what the comment says it should.

The two functions added to that sample each carry one intentional error in the branch that has to be reachable, so the test fails on main with `Expected 2 errors, got 0` and passes with the change. No existing expected type was weakened.

Full jest run here: 2696 passed. The 28 failures are all in `languageServer.test.ts` and fail the same way on a clean checkout on this machine, so they are my setup and not the change. `check:prettier`, `check:eslint` and `tsc --noEmit` are clean.

### DIFF
--- a/packages/pyright-internal/src/analyzer/typeGuards.ts
+++ b/packages/pyright-internal/src/analyzer/typeGuards.ts
@@ -2281,7 +2281,8 @@ function narrowTypeForTypedDictKey(
 
             if (isClassInstance(subtype) && ClassType.isTypedDictClass(subtype)) {
                 const entries = getTypedDictMembersForClass(evaluator, subtype, /* allowNarrowed */ true);
-                const tdEntry = entries.knownItems.get(literalKey.priv.literalValue as string) ?? entries.extraItems;
+                const knownEntry = entries.knownItems.get(literalKey.priv.literalValue as string);
+                const tdEntry = knownEntry ?? entries.extraItems;
 
                 if (isPositiveTest) {
                     // The code that is commented out below implements the behavior that is technically
@@ -2335,7 +2336,15 @@ function narrowTypeForTypedDictKey(
                         )
                     );
                 } else {
-                    return tdEntry !== undefined && (tdEntry.isRequired || tdEntry.isProvided) ? undefined : subtype;
+                    if (!knownEntry) {
+                        // The key is not one of the known items, so it can be supplied
+                        // only through an "extra items" entry, which says nothing about
+                        // whether this particular key is present. Its absence is therefore
+                        // consistent with the type, and the subtype cannot be eliminated.
+                        return subtype;
+                    }
+
+                    return knownEntry.isRequired || knownEntry.isProvided ? undefined : subtype;
                 }
             }
 
--- a/packages/pyright-internal/src/tests/samples/typedDictClosed11.py
+++ b/packages/pyright-internal/src/tests/samples/typedDictClosed11.py
@@ -97,3 +97,26 @@ def func7(td: Right) -> None:
     if "right" in td:
         reveal_type(td, expected_text="Right")
         reveal_type(td["right"], expected_text="int")
+
+
+def func8(td: Baz) -> None:
+    # "extra_items" constrains the type of the extra keys, not which of them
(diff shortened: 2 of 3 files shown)


luantaraschi:

rchiodo:

rchiodo:

rchiodo (approved):
Approved via Review Center.

rchiodo:

luantaraschi:
The review note says the tests could not be fully run there, so here is the run from my machine, against current main.

I rebased b91d350 onto bacb814 and it applied cleanly; none of the three files has an upstream change since I opened the PR. In `packages/pyright-internal`, with `npx jest -w 3 --testPathIgnorePatterns languageServer.test.ts`:

```
Test Suites: 1 skipped, 71 passed, 71 of 72 total
Tests:       8 skipped, 2696 passed, 2704 total
Time:        205.966 s
```

I leave `languageServer.test.ts` out because plain jest does not build the test server bundle it needs, and its 28 failures reproduce on a clean checkout here. `check:prettier` and `check:eslint` are clean, and `tsc --noEmit` in `pyright-internal` passes.

For the behavior, I built the CLI at both commits and ran the code from #11438 with a probe after the check and `reportUnreachable` on:

```python
class Typed(TypedDict, extra_items=str):
    a: str

def test(d: Typed) -> None:
    if "b" in d:
        return
    reveal_type(d)
    probe: int = "here"
```

On main the `reveal_type` line comes back as unreachable, and neither it nor the bad assignment is evaluated. With the change there is no unreachable report, `reveal_type(d)` gives `Typed`, and the assignment error is reported.

Reverting only `typeGuards.ts` while keeping the new sample makes `TypedDictClosed11` fail with `Expected 2 errors, got 0`, so the two functions I added do cover the fix.

I have not pushed the rebase, the branch is still at b91d350.

PR #11664, opened 2026-08-20
Drop inherited dataclass default when an override assigns a field specifier

### LINKED ISSUE #11660: Signature of dataclass __init__ inferred incorrectly, when overriding a field with a default without specifying a new default
**Bug description:**
When overriding a dataclass field (say `x`) with a default, with a field without a default as follows: `x: int = field()`, the `__init__` signature doesn't contain a default value for `x` at runtime. But pyright's inferred signature does contain a default.

```python
from dataclasses import dataclass, field

@dataclass
class Base:
    x: int = 1  # Field with a default

@dataclass
class Foo(Base):
    x: int = field()  # This line shouldn't be marked as an error (unless it causes a default arg to appear after a non-default arg). Pyright does mark it as an error.

Foo()  # This line should be marked as an error. Pyright doesn't
```

**More info:**

The runtime behavior of stdlib dataclasses when field with defaults are overridden without specifying a new default is kinda complicated and depends on the specific form of the override.
1. When the form is `x: int = field()` (as in the code snippet above):
    (a) Remove default value of `x` from `Foo.__init__`.
    (b) Subclasses of `Foo` can still inherit the default value of `x` defined in `Base`.
2. When the form is `x: int`:
    (a) Keep the default value of `x` from `Base` in `Foo.__init__`.
    (b) Subclasses of `Foo` can still inherit  the default value of `x` defined in `Base`.
3. When the form is `x: int = field(init=False)`:
    (a) Remove the parameter `x` from `Foo.__init__`.
    (b) Subclasses of `Foo` can still inherit the default value of `x` defined in `Base`.

Pyright handles the `x: int = field()` and `x: int` forms identically, matching the runtime behavior of the `x: int` form. OTOH, mypy, ty, and pyrefly handle the `x: int = field()` and `x: int` forms identically, matching the runtime behavior of the `x: int = field()` form.

For my purposes (which involves a custom dataclass mechanism), I only care that the `x: int = field()` form is handled properly (I forbid the `x: int` form via runtime checks). I'd really appreciate it if the narrower bug described near the top is fixed.

**Playground links:**
These contain the example above, along with some other cases from the **"more info"** section:

* pyright: [https://pyright-play.net/?code=...](https://pyright-play.net/?code=MTAECEFcHNQJQKYAcD2AnALgLgLACgRQBaUAGQEsA7BUAJgE5QBnACxUgBsATUAIxoC2AQzQBrBDyFNQQyqARo06ADR8EAYyGQmNAB6hy0oaDQIAjpHKmeSEUIEIMCmRhORKGcg4B0%2BQiQpqUABGRlZ2bkoAcld%2BUGExCRkjOQUlNFAACncOBCZpcldNbTyZUEoUSiIuBAAzLQ5XEVgMFFBalA4OFAB3Mpr6zia0aABKXzx8WqUBUC4hDCF1DikdAoFUTDmFpZX81VryBG58U7wAAXnF5dX8G-yIKQRcPFA30F0sAw9QAF4Qt5gABiR24oB6hRY-TqDQwZz8YAA8gA3NLkLhUWC6VQQjBsSCuJhIDTkWoAT0xZWofQGsJeZxAjKZzJZrLZ%2BEuO3uTDue2kQJQKEy4Ceoxe7w%2BXyorn%2Bh2OXEyo3hBFAADkUE4vgAVFiGUDcuZ1SjecHkLomISGGiyeSKdAuNweLw0SocMkGWomBACFDIyl4hB%2BQ2DRqgZFCDiQGjFNZUyrVGFDGQjUCtdqdbp9Yy0pPNb6gAUobwAfWLVEKpYmQbg7k8DmY5GglAWkFMoBQnsLEu7PeIb0yn2%2BGFGxAAfGrKoGVQBlRvNjCtmhUWppJK8d1IMloRssVy994kLKD6V-EIjojj9XUIOzpsttvL1c8dfxMmb1QYMmqTemWpuvtHlKHjnpek5nIWiqAlkNZOg4XxpOgI5gJkP47tgoCIgA0khWQCG%2BX6pgRP51G68F2mgSoEGy1E0ayyq0QxDEclcuy3HgBoimgwqiuK7zHh49HwLWzoNneC5th2jwZPuvaHgOQEymeY4TteM5zveS6UCuihrhuW5oTJbxyfxinBCBKlTmAt7zou3zadYfDunh76Ed%2BW4ke6xkKeZV5TvgnGQaAyEwXWzwYdhQVZKh0C7l8WE4ZkzkEZ%2Bbm-qRtrpJRjHZeykx5TlBVMsxXJ8ryqyPAAXtxOhivgEomaecrcJk5YYL8QIRjVgkhSJTDqeJLqeiKFWGT2ck%2BWBalibZj46c%2BenbjFe4yeNym%2BTe-UzVpT6Oa%2BLkpaAxF-p5-YTap-lCFVI6RZkPVweFCXRbFD03UlH5Ee5x1xdhCKFQVDJWZAvA6BYCA-L6aI1Ew3gw-Rf3-RcLHcmVDyFrQmSFrVrx8Qp3XCfWfXTRJnaCrQo0Sl5Q6nmZa2TVZm0Pttc27U9y37pTJ7-DTF4WRtROafZul7clH1pSdgFDmdfl4GjgXBfjYXxTdrPfQlb2uYdn3pQhFG-fDjFw-rBuIyVbEcSI6OcVj9W43l1YK6JNnE1JZPk0Z-YNVzUt807As7S%2BrOjRzPxe7Tqn0-zdn%2B05%2BHvalHkAfJkth9LnHo9d8uwYrEXISrL3IerB1Hdr5FZUbxsA%2BXtHFdcpXsXylXo5jvFvCZeNZ47GntkNl2u27wemd7U2%2B1HzMB-pS1Bx7CnU0PEcj7NDkvoXosJwPQ8XRV6dQbdCuq8rE-PUrBexxrxdkmRmV61XuVAA)
* mypy: https://mypy-play.net/?gist=7f0c3c8f62305650598f0184b2d2fc7a
* ty: https://play.ty.dev/b5568978-3390-4385-ba3f-a51acbdafa7c
* pyrefly: [https://pyrefly.org/sandbox/?project=...](https://pyrefly.org/sandbox/?project=v2.1VfLbsIwEPyVlXKASi1R2xtSLz1w6YGq9MjFgCGRwEaJw-vrO-tHKiDQABJqIgWJxGvvemY241MtM4qGKtyXNYEqnb6LkjibLmAzqOWZKEK3RRStU5OQYNqIYm7cJFFEfexklk64a28e7SBdoBks5TidbvmpICXXIQzzu7g_r6oMe1q3OcsggDJNW5fjHSYHmQ0k-Z1AZS4Qq6sOkpvPvR6EcvwhYSgrlEkXkrSab1mhmVzoFWduEnAqCql75Y5FwXuLqsCi8AqNgYymKXqoXv9ukn0O_iPvTom9S_HLL5qDUQISxvJTHhdkG64norYrs5Qphw_KsFRNUYmc0GhLoHo6S0xlOINZY4rFdolvrsHtdVOdwVAxFg_MjrYvpet29IEf-US61P-w_49mDYNrUuEGzryLrIIzgSmVMCDkNAw19_GeUHCJB1D4ff9XOOzOaZfF8dYT-HaVKj6Bza4Km6sRuRqHAwR21yOAN7W3Hw13UIxya_3gxlzvlXmnY9vKLY31pY2fC0TCIU0SiS2xASrJXiCT0uLV61aNAsKW2IR2dakiOKRZQKDE-wBxbMMj-vTZeUeLAyj8-6zAMRe-HO0M9UlJiTHLvBvHIVxns1iqeKLHebwXEcN9wc6tRGoPz_vTkbbHhhzTVvv_11r-v-Xsf-uc-2-x-ceIs4b-Bw)

**VS Code extension or command-line:** VS Code extension: 1.1.411

nileshpatil6:
Fixes #11660.

## The bug

When a dataclass field overrides an inherited field that has a default, `dataClasses.ts` unconditionally re-inherits that default:

```ts
// While this isn't documented behavior, it appears that the dataclass implementation
// causes overridden variables to "inherit" default values from parent classes.
if (!dataClassEntry.hasDefault && oldEntry.hasDefault && oldEntry.includeInInit) {
    dataClassEntry.hasDefault = true;
    dataClassEntry.defaultExpr = oldEntry.defaultExpr;
    ...
}
```

That is correct for a bare annotation, but not when the override assigns a field specifier that supplies no default. I checked the three forms against CPython 3.11 rather than relying on the issue report:

```
Base (self, x: int = 1) -> None

x: int = field()            -> Foo (self, x: int) -> None        Foo() TypeError: missing 1 required positional argument: 'x'
x: int                      -> Bar (self, x: int = 1) -> None    Bar() Bar(x=1)
x: int = field(init=False)  -> Baz (self) -> None                Baz() Baz(x=1)
```

So `x: int = field()` replaces the inherited entry and the parameter becomes required. pyright instead synthesized `(self: Foo, a: int = 0) -> None`, which produced two wrong results at once:

- **False negative:** `Foo()` was accepted, though it raises `TypeError` at runtime.
- **False positive:** the override line was reported with `"x" overrides a field of the same name but is missing a default value`, even though in this form nothing is inherited.

There is a knock-on ordering case too. When a defaulted field precedes the override, dropping the default makes a non-default parameter follow a default one, which CPython rejects at class creation:

```python
@dataclass
class B1:
    a: int = 0
    x: int = 1

@dataclass
class F1(B1):
    x: int = field()   # TypeError: non-default argument 'x' follows default argument
```

pyright did not report that either. With this change the existing "Fields without default values cannot appear after fields with default values" check fires on that line.

## The change

Track whether the entry had an assigned value, and only inherit the base default for bare annotations. Behavior for `x: int` is unchanged: it still inherits the default and still reports the existing diagnostic.

## Tests

Added to `dataclass4.py`, the sample for inherited dataclasses. The new cases pin the synthesized signature so they actually discriminate:

```python
reveal_type(DC13.__init__, expected_text="(self: DC13, a: int) -> None")
```

Without the fix that reports `(self: DC13, a: int = 0) -> None` and the suite fails; with it, it passes. The expected error count for `DataClass4` goes from 6 to 8, covering the now-reported `DC13()` call and the ordering case.

`npx jest typeEvaluator` passes 1212/1212 across all 8 suites, and Prettier reports the changed files clean.

### DIFF
--- a/packages/pyright-internal/src/tests/samples/dataclass4.py
+++ b/packages/pyright-internal/src/tests/samples/dataclass4.py
@@ -141,3 +141,15 @@ class DC15(DC14):
     # This should generate an error because "b" drops its inherited default
     # and would then follow "a", which still has one.
     b: int = field()
+
+
+@dataclass
+class DC16(DC12):
+    # An assigned field specifier with init=False removes the parameter from
+    # __init__ entirely rather than dropping the inherited default, so the
+    # attribute keeps the base class value.
+    a: int = field(init=False)
+
+
+reveal_type(DC16.__init__, expected_text="(self: DC16) -> None")
+reveal_type(DC16().a, expected_text="int")
--- a/packages/pyright-internal/src/tests/samples/dataclass4.py
+++ b/packages/pyright-internal/src/tests/samples/dataclass4.py
@@ -108,3 +108,36 @@ class DC10:
 class DC11(DC10):
     a: str = field()
     b: bool = field()
+
+
+@dataclass
+class DC12:
+    a: int = 0
+
+
+@dataclass
+class DC13(DC12):
+    # Unlike a bare annotation, an assigned field specifier that supplies no
+    # default replaces the inherited entry rather than inheriting its default,
+    # so "a" becomes a required parameter.
+    a: int = field()
+
+
+reveal_type(DC13.__init__, expected_text="(self: DC13, a: int) -> None")
+
+# This should generate an error because "a" no longer has an
+# inherited default value, so it must be provided.
+DC13()
+
+
+@dataclass
+class DC14:
+    a: int = 0
+    b: int = 1
+
+
+@dataclass
+class DC15(DC14):
+    # This should generate an error because "b" drops its inherited default
+    # and would then follow "a", which still has one.
+    b: int = field()
--- a/packages/pyright-internal/src/analyzer/dataClasses.ts
+++ b/packages/pyright-internal/src/analyzer/dataClasses.ts
@@ -331,6 +331,7 @@ export function synthesizeDataClassMethods(
             let aliasName: string | undefined;
             let variableTypeEvaluator: EntryTypeEvaluator | undefined;
             let hasDefault = false;
+            let hasAssignedValue = false;
             let isDefaultFactory = false;
             let isKeywordOnly = ClassType.isDataClassKeywordOnly(classType) || sawKeywordOnlySeparator;
             let defaultExpr: ExpressionNode | undefined;
@@ -362,6 +363,7 @@ export function synthesizeDataClassMethods(
                 }
 
                 hasDefault = true;
+                hasAssignedValue = true;
                 defaultExpr = statement.d.rightExpr;
 
                 // If the RHS of the assignment is assigning a field instance where the
@@ -578,7 +580,16 @@ export function synthesizeDataClassMethods(
 
                         // While this isn't documented behavior, it appears that the dataclass implementation
                         // causes overridden variables to "inherit" default values from parent classes.
-                        if (!dataClassEntry.hasDefault && oldEntry.hasDefault && oldEntry.includeInInit) {
+                        // This applies only when the override is a bare annotation (`x: int`). If the
+                        // override assigns a field specifier that supplies no default (`x: int = field()`),
+                        // the runtime replaces the entry outright, so the inherited default is dropped and
+                        // the parameter becomes required.
+                        if (
+                            !dataClassEntry.hasDefault &&
+                            !hasAssignedValue &&
+                            oldEntry.hasDefault &&
+                            oldEntry.includeInInit
+                        ) {
                             dataClassEntry.hasDefault = true;
                             dataClassEntry.defaultExpr = oldEntry.defaultExpr;
                             hasDefault = true;
--- a/packages/pyright-internal/src/tests/typeEvaluator4.test.ts
+++ b/packages/pyright-internal/src/tests/typeEvaluator4.test.ts
@@ -326,7 +326,7 @@ test('DataClass3', () => {
 test('DataClass4', () => {
     const analysisResults = TestUtils.typeAnalyzeSampleFiles(['dataclass4.py']);
 
-    TestUtils.validateResults(analysisResults, 6);
+    TestUtils.validateResults(analysisResults, 8);
 });
 
 test('DataClass5', () => {


rchiodo:

rchiodo:
The sample disables `reportIncompatibleVariableOverride`, so the new cases do not verify removal of the erroneous override diagnostic. Please add focused coverage with that rule enabled.

rchiodo (approved):
Approved via Review Center.

packages/pyright-internal/src/tests/samples/dataclass4.py
```diff
@@ -108,3 +108,36 @@ class DC10:
...
+
+@dataclass
+class DC14:
+    a: int = 0
+    b: int = 1
+
+
+@dataclass
+class DC15(DC14):
+    # This should generate an error because "b" drops its inherited default
```
rchiodo, line 141: > **Info** · Optional note

📍 packages/pyright-internal/src/tests/samples/dataclass4.py:141
[unverified] Add the documented `field(init=False)` override case and pin its no-parameter constructor signature and inherited-member behavior. The current tests cover only `field()` with `init=True`, leaving this distinct branch unprotected.

[verified]

rchiodo:

nileshpatil6:
Thanks for the review.

Added the `field(init=False)` branch as `DC16`, pinning both halves of its documented behaviour:

```python
@dataclass
class DC16(DC12):
    a: int = field(init=False)

reveal_type(DC16.__init__, expected_text="(self: DC16) -> None")
reveal_type(DC16().a, expected_text="int")
```

That matches CPython, where `init=False` removes the parameter from `__init__` entirely rather than dropping the inherited default, so the attribute keeps the base class value. `typeEvaluator4` passes 156/156.

On the other point, I think the rule named there is not the one this PR affects. The erroneous diagnostic being removed is `dataClassFieldInheritedDefault` ("x" overrides a field of the same name but is missing a default value), which `dataClasses.ts` reports under `reportGeneralTypeIssues`:

```ts
evaluator.addDiagnostic(
    DiagnosticRule.reportGeneralTypeIssues,
    LocMessage.dataClassFieldInheritedDefault().format({ fieldName: variableName }),
    variableNameNode
);
```

`dataclass4.py` disables `reportIncompatibleVariableOverride`, which is a different check and is unaffected by this change. `reportGeneralTypeIssues` is enabled in the sample, so the removal is already covered: `validateResults` asserts an exact error count, and `DC13`'s `a: int = field()` line carries no expected-error comment. If that diagnostic still fired the file would produce 9 errors against the expected 8 and the test would fail. I verified that directly by reverting only the `dataClasses.ts` change, which fails with the extra diagnostic at that line.

Happy to add a separate sample with `reportIncompatibleVariableOverride` enabled if you still want it, but as far as I can tell it would not exercise this code path.

nileshpatil6:
Closing this in favour of #11665, which landed the same fix while this was in review.

For the record the two changes are equivalent: #11665 tracks `isFieldSpecifierWithoutDefault` where this tracked whether the entry had an assigned value, and both gate the same inherit branch so a bare annotation still inherits and `field()` without a default does not. No complaint, and thanks @rchiodo for the review.

Two cases from this PR's tests are not covered by the merged sample, including the `field(init=False)` coverage you asked for above, so I have rebased just those onto the merged fix and opened them as a tests-only follow-up: #11679.

PR #11670, opened 2026-08-21
Fix keyword-form assignability for standard dest parameters

### LINKED ISSUE #10237: False negative for callable assignability

crypt0lith:
## Summary

Fixes microsoft#10237, where a function assignment was not rejected because the keyword form's type was never checked against the source. The destination's positional-or-keyword parameter can be passed by a name the source only accepts through an incompatible `**kwargs` or keyword-only parameter.

## Root cause

In `assignFunction`, the main loop over `positionalsToMatch` and the surplus-dest-positional branch both only checked for the presence of a source parameter's `kwargsIndex` to determine if the keyword home was satisfied. It did not check if the source keyword parameter was assignable to the dest positional-or-keyword parameter.

`assignFunction` also did not fallback to dest positional-or-keyword parameters whenever a dest keyword-only parameter was not found. This caused control to fall into the `!destParamInfo` branch where, if the source parameter had a `defaultArgType`, assignability between src and dest was never checked, resulting in a false negative.

## Fix

Move the source `kwargsIndex` condition into the inner branch scope, and add an `assignParam` check for when source `kwargsIndex` is present.

Add a dest positional-or-keyword parameter fallback to allow the assignability check between source keyword parameter and dest positional-or-keyword parameter.

Gate the new `assignParam` checks on `PartialOverloadOverlap`.

## Tests

- Adds a test sample `callbackProtocol12` which contains positive and negative assignments to a positional-or-keyword callback protocol for both kwargs and keyword-only source callback protocol forms.
- Adds regression test cases `func35` and `func36` in `overloadOverlap1` to exercise the changes that involve `PartialOverloadOverlap` gating.


rchiodo:

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -28586,7 +28616,14 @@ export function createTypeEvaluator(
...
                     }
 
-                    const destParamInfo = destParamMap.get(srcParamInfo.param.name);
+                    const destParamInfo =
+                        destParamMap.get(srcParamInfo.param.name) ??
+                        destParamDetails.params.find(
+                            (paramInfo) =>
+                                paramInfo.param.name === srcParamInfo.param.name &&
+                                paramInfo.kind === ParamKind.Standard &&
+                                paramInfo.param.category === ParamCategory.Simple
```
rchiodo, line 28641: > **Issue** · Please address or respond

This fallback can match destination `a: int` with a required source keyword-only `a: int`, even though a valid positional call such as `dest(1)` leaves the source's `a` unsatisfied. Reject this match when the source keyword-only parameter has no default, and add a regression case.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -28441,15 +28456,30 @@ export function createTypeEvaluator(
...
+                                    recursionCount
+                                )
+                            ) {
+                                if ((flags & AssignTypeFlags.PartialOverloadOverlap) === 0) {
+                                    canAssign = false;
+                                }
+                            }
                         }
                     }
                 }
```
rchiodo, line 28501: > **Info** · Optional note

The `kwargsIndex` and `assignParam` policy is duplicated across positional-matching branches, which makes future changes prone to divergence. Extract the shared keyword-form validation into a narrowly scoped helper.

packages/pyright-internal/src/tests/typeEvaluator2.test.ts
```diff
@@ -79,6 +79,12 @@ test('CallbackProtocol11', () => {
     TestUtils.validateResults(analysisResults, 0);
 });
 
+test('CallbackProtocol12', () => {
+    const analysisResults = TestUtils.typeAnalyzeSampleFiles(['callbackProtocol12.py']);
+
+    TestUtils.validateResults(analysisResults, 2);
+});
```
rchiodo, line 86: > **Info** · Optional note

This test asserts only the aggregate diagnostic count, so an unrelated diagnostic could replace an intended one unnoticed. Isolate the negative cases or assert diagnostic locations/messages if the harness supports it.

rchiodo:

packages/pyright-internal/src/analyzer/typeEvaluator.ts
rchiodo, line 28641: > **Warning** · Non-blocking recommendation

Could this fallback let one destination standard parameter satisfy both the source `*args` channel and a same-name required keyword-only parameter? For example, `Dest(a: int) = Src(*args: int, a: int)` should be rejected because `Dest(1)` leaves `Src`'s required `a` unset. Please add this regression case and restrict the fallback if it reproduces.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -28291,20 +28291,35 @@ export function createTypeEvaluator(
...
+                        paramIndex,
+                        diag?.createAddendum(),
+                        constraints,
+                        flags,
+                        recursionCount
+                    )
+                ) {
+                    if ((flags & AssignTypeFlags.PartialOverloadOverlap) === 0) {
+                        canAssign = false;
+                    }
```
rchiodo: > **Warning** · Non-blocking recommendation

The same `kwargsIndex` assignability check and `PartialOverloadOverlap` exception are duplicated again near line 28456. Extract a narrowly scoped helper so these keyword-form compatibility rules cannot drift independently.

[verified]

rchiodo:

rchiodo (approved):
Approved via Review Center.

rchiodo (changes requested):
Please address the comments

rchiodo:

rchiodo (approved):
Approved via Review Center.

packages/pyright-internal/src/tests/typeEvaluator2.test.ts
crypt0lith: I use the same pattern as every other `CallbackProtocol` test sample and am already isolating the negative cases by providing the expected error count in the way "the harness supports". I am going to assume a bot authored this review comment, and mark it as resolved. Please provide further clarification if I am missing something.

rchiodo:

rchiodo (approved):

packages/pyright-internal/src/analyzer/typeEvaluator.ts
crypt0lith: Addressed in 7f08d1fe1

For the destination signature `(a: int)`, I was able to reproduce the following:

| Source signature | Parent (3f870712c) | 277816bcf | 7f08d1fe1 |
| ------------------------- | --------------------------- | ---------------- | -------------- |
|`(*args: int, a: int = 0)` | pass | pass | pass |
|`(*args: int, a: int)` | error | pass (**regression**) | error |
|`(*, a: int = 0)` | error | error | error |
|`(*, a: int)` | error | error | error |

PR #11672, opened 2026-08-23
Fix functools.partial type evaluation for callable class instances

hsusul:
### Description
This PR fixes type evaluation for `functools.partial` when applied to callable class instances (objects implementing `__call__`).

Previously, `applyPartialTransform` in `constructorTransform.ts` checked if the target callable was an instantiable class (to create a function from its constructor), a plain function, or an overloaded function. When an instance of a callable class was passed to `partial(callable_instance, ...)`, `applyPartialTransform` returned `undefined`, causing the resulting partial object to evaluate as unspecialized `partial[Unknown]`. This resulted in missing parameter type checking, missing argument count validation, and unspecialized return types.

### Changes
- Updated `applyPartialTransform` in `packages/pyright-internal/src/analyzer/constructorTransform.ts` to look up the bound `__call__` method when the target callable argument is a class instance.
- Added regression tests in `packages/pyright-internal/src/tests/samples/partial9.py` and `typeEvaluator8.test.ts` covering partial application with simple and overloaded `__call__` methods on callable objects.

### Validation
- Ran full test suite in `typeEvaluator8.test.ts` (174/174 passing).
- Ran repository checks (`pnpm run check` and `pnpm run typecheck` passing cleanly).

### DIFF
--- a/packages/pyright-internal/src/tests/samples/partial9.py
+++ b/packages/pyright-internal/src/tests/samples/partial9.py
@@ -1,0 +1,35 @@
+# This sample tests functools.partial with callable class instances (objects with __call__).
+
+from functools import partial
+from typing import overload
+
+class CallableClass:
+    def __call__(self, a: int, b: str, c: float = 1.0) -> bool:
+        return True
+
+c = CallableClass()
+p1 = partial(c, 1)
+reveal_type(p1("hello"), expected_text="bool")
+reveal_type(p1("hello", c=2.0), expected_text="bool")
+
+# This should generate an error because 'b' must be a str
+p1(123)
+
+# This should generate an error because 'c' must be a float
+p1("hello", c="invalid")
+
+
+class OverloadedCallable:
+    @overload
+    def __call__(self, a: int, b: int) -> int: ...
+    @overload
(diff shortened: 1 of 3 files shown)


rchiodo:

packages/pyright-internal/src/tests/samples/partial9.py
```diff
@@ -0,0 +1,35 @@
...
+c = CallableClass()
+p1 = partial(c, 1)
+reveal_type(p1("hello"), expected_text="bool")
+reveal_type(p1("hello", c=2.0), expected_text="bool")
+
+# This should generate an error because 'b' must be a str
+p1(123)
+
+# This should generate an error because 'c' must be a float
+p1("hello", c="invalid")
```
rchiodo, line 19: > **Warning** · Non-blocking recommendation

The PR states that argument-count validation is restored, but this sample only checks argument types. Add too-few or excess-argument calls so the regression directly covers arity specialization.

[verified]

rchiodo:

rchiodo (approved):
Approved via Review Center.

rchiodo:

PR #11420, opened 2026-05-07
Optimize type equality fast paths

bschnurr:
## Summary

This optimizes `isTypeSame` by avoiding recursive comparisons when two types share the same immutable backing structures. The fast paths cover shared class type arguments, shared function parameter metadata, shared overload arrays, shared union subtype arrays, and shared TypeVar bound/constraint structures.

## Details

The change keeps the existing structural comparison behavior for non-shared data, but skips recursive walks when reference identity proves the relevant backing structure is identical. This commonly applies to shallow-cloned or decorated types.

## Validation

- `npm run test:norebuild -- --runTestsByPath src/tests/typeEquality.test.ts --forceExit`
- `npm run build`

## Performance Notes

On a local full-project NumPy CPU profile, `isTypeSame` self time dropped from about `21.98s` to `19.02s` (`-13.5%`) after these fast paths. A sampled heap profile did not show the edited helper functions as allocation sites. Whole-program wall-clock timing remains noisy, so this PR is presented as a targeted hotspot reduction rather than a claimed end-to-end throughput win.

PR #11680, opened 2026-08-25
Fix benchmark comparison and reporting

bschnurr:
## Summary

- stop treating the informational uncounted-validation-pass count as a regression-gating environment field
- move trusted benchmark rendering and PR commenting into a separate job in the dispatched workflow
- remove the `workflow_run` consumer, which GitHub does not trigger after the `GITHUB_TOKEN` workflow dispatch chain
- preserve the security boundary: benchmark code runs read-only, while the comment job runs separately and validates head, base, and merge provenance

## Failures addressed

Run https://github.com/microsoft/pyright/actions/runs/32891326101 measured improvements for all nine packages but failed only because `uncounted_validation_runs_per_checker` differed (`1 != 0`). Replaying the artifact with this change produces **No performance regressions detected**.

The same run uploaded a valid artifact, but GitHub created no `workflow_run` for `.github/workflows/typecheck_benchmark_comment.yml`, so PR #11420 received no chart comment.

## Validation

- `python -m unittest test_compare_benchmarks test_typecheck_benchmark` (45 tests)
- replayed run `32891326101` artifact: all nine rows pass
- Prettier and editor diagnostics clean


rchiodo:

build/benchmark/test_compare_benchmarks.py
```diff
@@ -570,9 +561,22 @@ def test_pr_benchmark_requires_authorized_comment(self) -> None:
...
+            "pullRequest.data.merge_commit_sha !== expectedMergeSha",
+            benchmark_workflow,
+        )
+        self.assertFalse(
+            (
+                REPO_ROOT
+                / ".github"
+                / "workflows"
+                / "typecheck_benchmark_comment.yml"
+            ).exists()
```
rchiodo, line 616: > **Issue** · Please address or respond

These whole-file substring checks do not enforce the job-level security boundary: `pull-requests: write` could move to `benchmark` and the test would still pass. Parse the workflow and assert that only `comment` has write permission, that it depends on `benchmark`, and that `benchmark` remains read-only.

[verified]

rchiodo:

build/benchmark/test_compare_benchmarks.py
bschnurr: Addressed in 6d88d1848. The test now parses the workflow as YAML and asserts the exact permission map for each job: `benchmark` is `{contents: read}`, `comment` is `{actions: read, contents: read, pull-requests: write}`, `comment.needs` is `benchmark`, and `comment` is the only job with `pull-requests: write`. I also ran the full benchmark test suites with the workspace `.venv` (45 tests passing) and validated the frozen lockfile offline.

bschnurr:
Follow-up on verification: the sole inline issue was addressed in `6d88d1848` and its thread is resolved. I ran `test_compare_benchmarks` and `test_typecheck_benchmark` with the workspace `.venv` (45 tests passing), plus the frozen offline lockfile check. CodeQL, Style, and Typecheck are green on the updated head; the platform Build/Test jobs are still running.

rchiodo:

rchiodo (approved):
Approved via Review Center.

PR #11695, opened 2026-09-01
Optimize recursive sequence protocol rejection

bschnurr:
## Summary

Improve checking performance for lists of overloaded callables, including NumPy's `np.array([np.sum, np.mean])` and pandas' `test_series_apply.py`, without adding NumPy-specific logic.

### Recursive sequence rejection

- Add a negative-only shortcut for built-in `list` assignments to a recognized recursive sequence protocol with one covariant type parameter. Tuples use ordinary protocol matching.
- Require the specialized, non-overloaded `__getitem__` return type to be exactly `Leaf | Protocol[Leaf]`. Similar member names alone are not sufficient.
- Decline top-level source/destination element TypeVars, Any, and Unknown, and unsupported protocol shapes. This is not a recursive exclusion of all nested TypeVars or gradual types.
- Clone caller constraints for the reduced element check. A successful reduced check does not establish protocol compatibility; ordinary structural matching still runs.

### Constraint solving

- Skip `solveAndApplyConstraints` when the target type requires no specialization and has no alias metadata, avoiding full constraint solves whose results would not transform the target.
- Preserve the normal path for alias metadata because the transformer processes generic aliases before its ordinary no-op check.

### Protocol cache provenance

- Record whether a cached result came from fast rejection.
- Allow a subsequent full check to replace the matching fast-rejection entry and exclude fast entries from the entry-existence check used when considering universal incompatibility.
- This does not provide transactional cache isolation: cached negatives can still be reused before a diagnostic-bearing full check replaces them. Behavioral validation of inference and reuse remains an open gate below.

### Why checking is slow

For `np.array([np.sum, np.mean])`, Pyright must determine which NumPy overload accepts a list of overloaded functions. Some candidates describe recursive sequences: each element can be a supported leaf value or another sequence. Rejecting an unsuitable candidate can trigger expensive recursive protocol matching, overloaded-function comparison, and repeated constraint solving. Profiling also identified substantial work solving constraints even when the particular type being transformed could not change.

The goal is to eliminate unnecessary work without rejecting valid candidates, accepting invalid ones, or losing inferred type precision. There are two independent changes: skip solving when substitution cannot transform the target (excluding alias metadata), and reject narrowly recognized recursive-sequence candidates using an element-compatibility condition. A successful element check never establishes full protocol compatibility; ordinary structural matching still runs.

### Necessary-condition argument

Suppose the source is `list[E]` and protocol `P[L]` promises:

```python
def __getitem__(self, index: int, /) -> L | P[L]: ...
```

For integer indexing, the built-in list signature returns `E`. Implementing the protocol therefore requires that this return type be assignable to the promised return type. Writing $\preceq$ for assignability:

$$
\mathrm{list}[E] \preceq P[L] \implies E \preceq L \cup P[L]
$$

Taking the contrapositive gives the early-rejection rule:

$$
E \not\preceq L \cup P[L] \implies \mathrm{list}[E] \not\preceq P[L]
$$

This justifies rejection, not acceptance: other protocol members can still be incompatible.

**The integer-index premise is essential.** A slice returns another list, not an element. A slice-only recursive protocol supplied a real counterexample to the return-type-only guard: the shortcut selected fallback `str`, whereas full matching selected `Literal['slice']`. A local follow-up adds a built-in-integer parameter guard and a precise slice regression. As of this description update (2026-09-04), that fix is in local commit `ea13f4cc3`, not the published PR head `34179d476`. The published implementation therefore does not yet enforce this premise. The approximately 4.1-second post-fix pandas measurements and additional local tests must not be mistaken for validation of a published integer-index fix.

### What an exception-free proof still requires

The implication above is not a complete proof of the implementation. The remaining obligations are:

1. **Eligibility establishes every premise.** Binding, specialization, actual list signatures, parameter domains, and return types must have the meaning assumed by the argument.
2. **Reduced failure means genuine incompatibility.** A false result must not merely mean that inference could not solve the relation in this context. Nested generics, conditional types, recursive assumptions, and incomplete inference matter; excluding only top-level Any, Unknown, and TypeVars does not prove this.
3. **Inference assumptions are appropriate.** The reduced check must not reject a candidate that the full check could accept after solving relevant variables.
4. **The probe cannot corrupt later decisions.** Cloning constraints protects the caller tracker, not every nested cache effect. Fast-rejection provenance prevents direct promotion to universal incompatibility, but is not a complete state-isolation proof.
5. **Skipping constraint solving is observationally harmless.** Returning the same type is necessary; omitted solving must also have no required effects on subsequent inference. Alias metadata remains on the normal path because it can require transformation.

The proof target is that every early rejection is justified under the applicable typing rules and inference state, and subsequent decisions remain unaffected. Each obligation needs a code-backed argument or an explicit fallback when its premises cannot be established. Passing suites, adversarial cases, performance measurements, and no-shortcut comparisons provide evidence, not a universal guarantee. The discovered slice counterexample demonstrates the distinction. These obligations remain merge gates, not completed claims of correctness.

## Performance

Latest local measurements used Node 26.5.0, Python 3.12.3, NumPy 2.4.6, and pandas revision `82a712a52a55e9ac99c59a818e36c4d400bc5dbe`. The core was compiled before each CLI bundle. Base checker: `c77393240247e017db6538d35876e723c518e06c`; measured implementation revision: `9577c1663acf0751c858a8cd63d4b36c4145b911`. The subsequent test-only commit `34179d47639970e3d8be13ca21949bde2af2f6a6` does not change production code; benchmarks have not been rerun for it.

| Configuration | Combined NumPy check time | Combined NumPy wall time | pandas file wall time |
| --- | ---: | ---: | ---: |
| Base | 79.75 s | 80.69 s | Historical: about 4 minutes |
| Solver guard alone, without sequence shortcut | 24.68 s | 25.66 s | 78.53 s |
| Current integrated changes | 0.27 s | 1.22 s | 4.18 s |

The base singleton lists containing only `np.sum` or only `np.mean` checked in 0.32 and 0.34 seconds. The combined trigger revealed `ndarray[tuple[Any, ...], dtype[Any]]` with no errors or warnings across these measurements. The final pandas runs also reported no errors or warnings.

These are single-run local observations, not statistical benchmarks or Node 24 CI comparisons. The historical pandas base timing is not a fresh paired baseline. The independent solver guard provides a substantial partial improvement; it does not by itself retain the approximately four-second pandas result.

## Tests and Local Validation

- Existing sequence coverage includes positive nested sequences, incompatible callables, similar member names with different semantics, constrained generic results, tuple fallback, and subsequent valid list specializations after a rejection.
- Direct helper tests cover scoped/unification TypeVars, Any/Unknown, concrete and generic-callable sources, and fast-cache replacement after a full diagnostic-bearing check.
- Added `SolveAndApplyConstraintsConcreteType`: concrete types, Any, and Unknown avoid constraint-set traversal; bound TypeVars stay bound; free TypeVars specialize precisely through nested `list[tuple[T, int]]` and generic alias metadata.
- Added eight test cases in `34179d476`, plus expanded existing assertions:
  - Fresh evaluators exercise positive-first and negative-first protocol assignments, repeated rejection, diagnostic details, preserved caller constraints, and subsequent generic assignments inferring `int` with fresh constraint trackers.
  - A dependency-free overload sample asserts precise `Literal['array']` selection for valid array-like sequences and `str` fallback for overloaded callables, scoped `list[T]`, nested `list[tuple[T, int]]`, ParamSpec callables, and TypeVarTuple tuples.
  - Uncertain destination elements (scoped/unification TypeVars, Any, Unknown) decline the helper without changing constraints.
  - Concrete unions, tuples, callables, None, and Never avoid solver traversal; alias assertions distinguish bound preservation from free-TypeVar substitution.
- Full internal suite: **77 suites, 2,795 tests passed** for the implementation preceding this test-only commit. The full suite has not been rerun after the new tests.
- Latest pre-push targeted shard: **180 tests passed**, with core TypeScript build, targeted ESLint/Prettier, and `git diff --check` also passing on `34179d476`.
- The earlier implementation validation also included a CLI development build. Jest emitted worker-exit/listener warnings; the latest targeted run emitted the force-exit notice.
- No existing expected types or diagnostic expectations were weakened for performance.

## Coverage Progress and Remaining Before Merge

- [x] Add focused protocol cache-reuse tests in both assignment orders, asserting results, caller constraints, diagnostic element incompatibility, and subsequent generic inference.
- [ ] Validate broader inference-context and cross-candidate cache reuse beyond these focused scenarios; these tests do not establish general transactional cache isolation.
- [x] Add dependency-free recursive-protocol overload regressions for scoped `list[T]` and nested `list[tuple[T, int]]`, callable fallback, and precise positive overload selection. These model the relevant matching behavior rather than importing NumPy.
- [x] Cover ParamSpec callables and TypeVarTuple tuple inputs in overload selection, and uncertain top-level destination elements in direct helper tests.
- [ ] Add remaining nested Any/Unknown and incomplete-inference coverage at the shortcut boundary.
- [ ] Repeat pandas measurements at least three times, targeting median wall time below five seconds in the pinned environment, and add stable small-trigger work/scaling coverage.
- [ ] Validate the relevant hydra-zen, pandas-stubs, JAX, and frozen scikit-learn cases, then run the broader primer comparison with recorded revisions.
- [ ] Resolve the reported SymPy `dict[Dummy, Basic | Unknown] -> dict[Dummy, Basic]` change with a reproducible, semantically justified expectation. Repeated unchanged-base full-project runs produced different diagnostics, so those broad diffs do not yet establish causality.

Correct typing behavior is the acceptance criterion, not exact parity with every old diagnostic. Any reproducible semantic change needs an explanation and focused coverage; passing local tests and having pushed the code do not close the remaining cache/inference gates.

### DIFF
--- a/packages/pyright-internal/src/analyzer/protocols.ts
+++ b/packages/pyright-internal/src/analyzer/protocols.ts
@@ -921,10 +921,6 @@ export function tryFastRejectSequenceProtocol(
     flags: AssignTypeFlags,
     recursionCount: number
 ): '__getitem__' | undefined {
-    if ((flags & AssignTypeFlags.ArgAssignmentFirstPass) !== 0) {
-        return undefined;
-    }
-
     if (!isClassInstance(srcType) || !ClassType.isBuiltIn(srcType, 'list')) {
         return undefined;
     }
--- a/packages/pyright-internal/src/analyzer/protocols.ts
+++ b/packages/pyright-internal/src/analyzer/protocols.ts
@@ -30,8 +30,8 @@ import {
     isFunctionOrOverloaded,
     isInstantiableClass,
     isOverloaded,
-    isTypeVar,
     isTypeSame,
+    isTypeVar,
     ModuleType,
     OverloadedType,
     Type,
@@ -940,16 +940,12 @@ export function tryFastRejectSequenceProtocol(
 
     const srcElementType = srcType.priv.typeArgs?.[0];
 
-    if (
-        !srcElementType ||
-        isAnyOrUnknown(srcElementType) ||
-        (isTypeVar(srcElementType) && TypeVarType.isUnification(srcElementType))
-    ) {
+    if (!srcElementType || isAnyOrUnknown(srcElementType) || isTypeVar(srcElementType)) {
         return undefined;
     }
 
     const destElementType = destType.priv.typeArgs[0];
-    if (isAnyOrUnknown(destElementType)) {
+    if (isAnyOrUnknown(destElementType) || isTypeVar(destElementType)) {
         return undefined;
     }
 
--- a/packages/pyright-internal/src/analyzer/protocols.ts
+++ b/packages/pyright-internal/src/analyzer/protocols.ts
@@ -921,6 +921,10 @@ export function tryFastRejectSequenceProtocol(
     flags: AssignTypeFlags,
     recursionCount: number
 ): '__getitem__' | undefined {
+    if ((flags & AssignTypeFlags.ArgAssignmentFirstPass) !== 0) {
+        return undefined;
+    }
+
     if (!isClassInstance(srcType) || !ClassType.isBuiltIn(srcType, 'list')) {
         return undefined;
     }
--- a/packages/pyright-internal/src/analyzer/protocols.ts
+++ b/packages/pyright-internal/src/analyzer/protocols.ts
@@ -30,6 +30,7 @@ import {
     isFunctionOrOverloaded,
     isInstantiableClass,
     isOverloaded,
+    isTypeVar,
     isTypeSame,
     ModuleType,
     OverloadedType,
@@ -912,7 +913,7 @@ function assignToProtocolInternal(
 // return type is exactly `Leaf | Protocol[Leaf]`. This makes element compatibility a necessary
 // condition of the full protocol assignment. If that condition fails, the source sequence cannot
 // satisfy __getitem__. Every uncertain case falls back to the normal structural protocol walk.
-function tryFastRejectSequenceProtocol(
+export function tryFastRejectSequenceProtocol(
     evaluator: TypeEvaluator,
     destType: ClassType,
     srcType: ClassType | ModuleType,
@@ -939,7 +940,11 @@ function tryFastRejectSequenceProtocol(
 
     const srcElementType = srcType.priv.typeArgs?.[0];
 
-    if (!srcElementType || isAnyOrUnknown(srcElementType)) {
+    if (
+        !srcElementType ||
+        isAnyOrUnknown(srcElementType) ||
+        (isTypeVar(srcElementType) && TypeVarType.isUnification(srcElementType))
+    ) {
         return undefined;
     }
 
--- a/packages/pyright-internal/src/analyzer/protocols.ts
+++ b/packages/pyright-internal/src/analyzer/protocols.ts
@@ -21,7 +21,9 @@ import { getLastTypedDeclarationForSymbol, isEffectivelyClassVar } from './symbo
 import { AssignTypeFlags, TypeEvaluator } from './typeEvaluatorTypes';
 import {
     ClassType,
+    combineTypes,
     FunctionType,
+    isAnyOrUnknown,
     isClass,
     isClassInstance,
     isFunction,
@@ -57,6 +59,16 @@ interface ProtocolAssignmentStackEntry {
     destType: ClassType;
 }
 
+const sequenceProtocolMemberNames = new Set([
+    '__len__',
+    '__getitem__',
+    '__contains__',
+    '__iter__',
+    '__reversed__',
+    'count',
+    'index',
+]);
+
 interface ProtocolCompatibility {
     // Specialized source type or undefined if this entry applies
     // to all specializations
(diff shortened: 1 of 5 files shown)


heejaechang:
### ✅ Pyright CLI QA — `34179d47` — verdict: **green**

4 focused check(s) completed.

heejaechang:

packages/pyright-internal/src/analyzer/protocols.ts
```diff
@@ -850,6 +927,147 @@ function assignToProtocolInternal(
...
+    }
+
+    const returnType = FunctionType.getEffectiveReturnType(boundMemberType);
+    return !!returnType && isTypeSame(returnType, expectedReturnType);
+}
+
+// Restrict the semantic check above to the standard read-only sequence protocol surface. This is
+// a scope guard rather than the correctness proof: protocols with these names but different
+// __getitem__ semantics are rejected by hasRecursiveSequenceGetItem and use normal matching.
+function isSequenceLikeProtocol(classType: ClassType): boolean {
```
heejaechang, line 1043: > **Issue** · Please address or respond

The fast path verifies only `__getitem__`'s return type and does not ensure its index parameter accepts `int`. A slice-only recursive protocol can therefore be treated as element-indexed and rejected based on an unrelated element compatibility check, changing overload selection from `Literal['slice']` to `str`. Require the bound index parameter to accept built-in `int` before taking this shortcut; otherwise fall back to normal protocol matching, and add the slice regression.

heejaechang:

**Summary:** \[unavailable\] Container verification could not start and local execution was not authorized for this PR HEAD\: Podman CLI is not installed or is not on PATH.

**Test runs:** none recorded.
</details>

PR #11702, opened 2026-09-03
Report async comprehensions inside lambdas

### LINKED ISSUE #11158: Missing SyntaxError check for `async for` in lambda inside an async function
# Description and reproduce

pyright, (and pylance in vscode), does not warn against SyntaxError for the following code:

```python
import asyncio
import random


async def async_gen(limit: int = 10):
    while True:
        if limit <= 0:
            return
        await asyncio.sleep(rtn := random.random())
        yield rtn
        limit -= 1


async def run_with_callback[U, T](u: U, callback: Callable[[U], T]) -> T:
    await asyncio.sleep(0)  # some async ops
    return callback(u)


async def test_async_for_in_lambda():
    result = await run_with_callback(
        async_gen(5),
        lambda ag: [x async for x in ag],  # <- fails to detect syntax error, raises in runtime
    )
    print(result)


# def run_with_callback_sync[U, T](u: U, callback: Callable[[U], T]) -> T:
#     return callback(u)
# 
# 
# def test_async_for_in_lambda_sync():
#     result = run_with_callback_sync(
#         async_gen(5),
#         lambda ag: [x async for x in ag],  # <- correctly detect syntax error, raises in runtime
#     )
#     print(result)


if __name__ == "__main__":
    asyncio.run(test_async_for_in_lambda())
```

Runtime output:

```python
 ➜ python /tmp/test.py
  File "/tmp/test.py", line 23
    lambda ag: [x async for x in ag],  # <- fails to detect syntax error, raises in runtime
               ^^^^^^^^^^^^^^^^^^^^^
SyntaxError: asynchronous comprehension outside of an asynchronous function
```

pyright output:

```bash
 ➜ pyright /tmp/test.py
0 errors, 0 warnings, 0 informations
```

# Related versions

* `pyright`: `1.1.407`
* `pylance`: `2025.10.4`
* `python`: `Python 3.14.0`, conda distributed
* OS: should be irrelevant, as this is pure Python

# Note

May be this can be related to #9414

bschnurr:
## Description

Pyright did not report an error for eager asynchronous comprehensions inside a lambda body when the lambda was nested in an async function. A lambda body is a separate, non-async function boundary, so list, set, and dictionary async comprehensions there are invalid even when the enclosing `def` is async.

Lambda default expressions and class header expressions are evaluated in their enclosing context and must continue to inherit that context. Async generator expressions also remain allowed because their execution is deferred.

I reproduced the behavior for every comprehension form and compared the accepted syntax with CPython. I then traced comprehension binding through `Binder.visitComprehension` and found that it used `getEnclosingFunction`, which skipped lambda and class evaluation boundaries and could therefore select an outer async function.

I reviewed the existing evaluation-scope utilities and the binder's handling of comprehension scopes, lambda defaults, nested functions, class bodies, and deferred generator expressions. This identified `getEvaluationScopeNode` as the appropriate way to determine the context where each comprehension is evaluated.

## Implementation

`Binder.visitComprehension` now determines the nearest evaluation scope rather than only the nearest function declaration. It skips the comprehension's own synthetic scope and type-parameter scopes, then requires that the resulting scope be an async function before permitting eager async comprehensions.

This makes lambda and class bodies non-async boundaries while preserving the enclosing context for lambda defaults and class headers. Existing deferred-generator behavior is unchanged.

Screenshots or GIFs are not applicable because this change affects syntax diagnostics and has no visual UI.

## Testing

Added parser coverage for list, set, dictionary, and generator comprehensions in:

- async and synchronous functions
- direct and nested lambda bodies
- lambda default expressions
- nested synchronous functions
- class bodies

Added `comprehension12.py` evaluator coverage with expected diagnostic counts. The targeted parser and evaluator tests, Pyright build and typecheck, Prettier, and ESLint passed.

## Addresses

Fixes #11158

### DIFF
--- a/packages/pyright-internal/src/analyzer/binder.ts
+++ b/packages/pyright-internal/src/analyzer/binder.ts
@@ -2444,7 +2444,15 @@ export class Binder extends ParseTreeWalker {
     }
 
     override visitComprehension(node: ComprehensionNode): boolean {
-        const enclosingFunction = ParseTreeUtils.getEnclosingFunction(node);
+        // Use evaluation scopes so lambda and class bodies act as non-async boundaries,
+        // while lambda defaults and class headers retain their enclosing context.
+        let enclosingScopeNode = ParseTreeUtils.getEvaluationScopeNode(node, this._nodeInfo).node;
+        while (
+            enclosingScopeNode.nodeType === ParseNodeType.Comprehension ||
+            enclosingScopeNode.nodeType === ParseNodeType.TypeParameterList
+        ) {
+            enclosingScopeNode = ParseTreeUtils.getEvaluationScopeNode(enclosingScopeNode.parent!, this._nodeInfo).node;
+        }
 
         // The first iterable is executed outside of the comprehension scope.
         if (node.d.forIfNodes.length > 0 && node.d.forIfNodes[0].nodeType === ParseNodeType.ComprehensionFor) {
@@ -2474,7 +2482,10 @@ export class Binder extends ParseTreeWalker {
                         // Async for is not allowed outside of an async function
                         // unless we're in ipython mode.
                         if (compr.d.asyncToken && !this._fileInfo.ipythonMode) {
-                            if (!enclosingFunction || !enclosingFunction.d.isAsync) {
+                            if (
+                                enclosingScopeNode.nodeType !== ParseNodeType.Function ||
+                                !enclosingScopeNode.d.isAsync
+                            ) {
                                 // Allow if it's within a generator expression. Execution of
                                 // generator expressions is deferred and therefore can be
                                 // run within the context of an async function later.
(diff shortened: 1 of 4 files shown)


heejaechang:
### 🔴 Pyright CLI QA — `5c48b0b0` — verdict: red

6 focused check(s) completed.

**`main.py`**

```python
from collections.abc import AsyncIterable


async def nested_async_generator(values: AsyncIterable[int]) -> None:
    factory = lambda: (
        [inner async for inner in values]
        async for _outer in values
    )
    _ = factory
```

**`pyrightconfig.json`**

```json
{
    "pythonVersion": "3.13",
    "typeCheckingMode": "strict",
    "reportUnusedFunction": "none",
    "reportUnusedVariable": "none"
}
```

```text
```

**Base `fa9b200906fadc093e862ce62ec1afa985c7233d`**

```json
{
  "exit_code": 0,
  "diagnostics": []
}
```

**PR `5c48b0b0d526db4e350e572c87268bd86c560be2`**

```json
{
  "exit_code": 1,
  "diagnostics": [
    {
      "file": "main.py",
      "severity": "error",
      "rule": null,
      "message": "Use of \"async\" not allowed outside of async function",
      "range": {
        "start": [
          5,
          15
        ],
        "end": [
          5,
          20
        ]
      }
    }
  ]
}
```

</details>

PR #11704, opened 2026-09-03
Fix overload override matching order

### LINKED ISSUE #11549: Order of additional overloads matters in subclass where it shouldnt [reportIncompatibleMethodOverride]
**Describe the bug**

In some cases, a valid order of instance-method overloads will be (wrongly) rejected by pyright/pylance, but only if the implementing class inherits from a class that defines some of the overloads.

For example (see below for a complete source code including all imports etc) the following two classes are type hinted correctly (as expected):

```python
class ValidBase:
    @overload
    def map(self, x: str, wrap: Literal[True]) -> Tuple[str]: ...
    @overload
    def map(self, x: int, wrap: Literal[True]) -> Tuple[int]: ...

    def map(self, x: str|int, wrap: Literal[True]) -> Tuple[str]|Tuple[int]: ...

class ValidComplete:
    @overload
    def map(self, x: str, wrap: Literal[False]=False) -> str: ...
    @overload
    def map(self, x: int, wrap: Literal[False]=False) -> int: ...

    @overload
    def map(self, x: str, wrap: Literal[True]) -> Tuple[str]: ...
    @overload
    def map(self, x: int, wrap: Literal[True]) -> Tuple[int]: ...

    def map(self, x: str|int, wrap: bool=False) -> str|int|Tuple[str]|Tuple[int]: ...
```

Notice, that the two last overloads in `ValidComplete` are the same as in `ValidBase` and that every object inheriting from `ValidComplete` also completely matches the interface defined in `ValidBase`.

However, the following class, although being a 1:1 copy of `ValidComplete` (+ inheriting from `ValidBase`) and having the exact same (mutually exclusive) overloads on the `map(...)` method as `ValidBase` will show a type error:

```python
class InvalidDerived(ValidBase):
    @overload
    def map(self, x: str, wrap: Literal[False]=False) -> str: ...
    @overload
    def map(self, x: int, wrap: Literal[False]=False) -> int: ...

    @overload
    def map(self, x: str, wrap: Literal[True]) -> Tuple[str]: ...
    @overload
    def map(self, x: int, wrap: Literal[True]) -> Tuple[int]: ...

    def map(self, x: str|int, wrap: bool=False) -> str|int|Tuple[str]|Tuple[int]: ... # -> type error saying that "map(...)" overrides "ValidBase" in an incompatible manner
```

Finally, if the two overload blocks are switched in the derived class, the type error disappears:

```python
class ValidDerived(ValidBase):
    @overload
    def map(self, x: str, wrap: Literal[True]) -> Tuple[str]: ...
    @overload
    def map(self, x: int, wrap: Literal[True]) -> Tuple[int]: ...

    @overload
    def map(self, x: str, wrap: Literal[False]=False) -> str: ...
    @overload
    def map(self, x: int, wrap: Literal[False]=False) -> int: ...

    def map(self, x: str|int, wrap: bool=False) -> str|int|Tuple[str]|Tuple[int]: ... # -> no type error?
```

I think the following behaviours are wrong here:
1. `map(...)` on `InvalidDerived` should not be marked as invalid by pyright
2. The order between these mutually exclusive overloads should not matter, so if the behaviour of `InvalidDerived` is intended to be a type error, then `ValidDerived` should also be marked as invalid

**Code or Screenshots**

```python
from typing import Literal, Tuple, overload

# works as expected:
class ValidBase:
    @overload
    def map(self, x: str, wrap: Literal[True]) -> Tuple[str]: ...
    @overload
    def map(self, x: int, wrap: Literal[True]) -> Tuple[int]: ...

    def map(self, x: str|int, wrap: Literal[True]) -> Tuple[str]|Tuple[int]: ...

# works as expected
class ValidComplete:
    @overload
    def map(self, x: str, wrap: Literal[False]=False) -> str: ...
    @overload
    def map(self, x: int, wrap: Literal[False]=False) -> int: ...

    @overload
    def map(self, x: str, wrap: Literal[True]) -> Tuple[str]: ...
    @overload
    def map(self, x: int, wrap: Literal[True]) -> Tuple[int]: ...

    def map(self, x: str|int, wrap: bool=False) -> str|int|Tuple[str]|Tuple[int]: ...

# does not work, this is the issue i am reporting:
class InvalidDerived(ValidBase):
    @overload
    def map(self, x: str, wrap: Literal[False]=False) -> str: ...
    @overload
    def map(self, x: int, wrap: Literal[False]=False) -> int: ...

    @overload
    def map(self, x: str, wrap: Literal[True]) -> Tuple[str]: ...
    @overload
    def map(self, x: int, wrap: Literal[True]) -> Tuple[int]: ...

    def map(self, x: str|int, wrap: bool=False) -> str|int|Tuple[str]|Tuple[int]: ... # -> type error saying that "map(...)" overrides "ValidBase" in an incompatible manner

# works, but i dont understand why the order between the overloads would matter?
class ValidDerived(ValidBase):
    @overload
    def map(self, x: str, wrap: Literal[True]) -> Tuple[str]: ...
    @overload
    def map(self, x: int, wrap: Literal[True]) -> Tuple[int]: ...

    @overload
    def map(self, x: str, wrap: Literal[False]=False) -> str: ...
    @overload
    def map(self, x: int, wrap: Literal[False]=False) -> int: ...

    def map(self, x: str|int, wrap: bool=False) -> str|int|Tuple[str]|Tuple[int]: ... # -> no type error?
```

Image

(roughly translates to)

> The method "map" overrides the class "ValidBase" in an incompatible manner.
> The override does not handle all overloads of the base method.
> Pylance(reportIncompatibleMethodOverride)
> Untitled-6(10, 9): Overriden method

The screenshot was made in VSCode (1.127.0 Universal) with the latest pylance Plugin (2026.2.1) installed.

bschnurr:
## Description

Accept overloaded overrides that add disjoint extra overloads or cover multiple base domains without spurious ordering failures, while continuing to reject incompatible overlaps and genuinely reordered matches.

I traced overloaded override comparison in `typeEvaluator.ts` and added a focused matrix of positive and negative subclass cases. The prior implementation selected one best match, which made compatibility depend on the placement of otherwise valid additional overloads.

## Implementation

Compare override overloads against base call domains rather than selecting a single match. The new logic tracks newly covered base overload indices, preserves monotonic base ordering, and rejects incompatible overlapping extras unless an earlier compatible match already shadows the overlap.

## Testing

Added `overloadOverride2.py` and its evaluator test with positive cases for extras before and after base overloads and broad compatible coverage, plus negative cases for reordered matches and incompatible broad overlaps. The evaluator regression suite passed.

## Addresses

Fixes #11549

### DIFF
--- a/packages/pyright-internal/src/analyzer/typeEvaluator.ts
+++ b/packages/pyright-internal/src/analyzer/typeEvaluator.ts
@@ -29129,20 +29129,37 @@ export function createTypeEvaluator(
             });
         }
 
-        // For an overloaded method overriding an overloaded method, the overrides
-        // must all match and be in the correct order. It is OK if the base method
-        // has additional overloads that are not present in the override.
+        // For an overloaded method overriding an overloaded method, all base overloads
+        // must match in the correct order. The override can have additional overloads.
 
         let previousMatchIndex = -1;
         const baseOverloads = OverloadedType.getOverloads(baseMethod);
+        const matchedBaseOverloadIndices = new Set<number>();
+        const getCallDomainType = (overload: FunctionType) => {
+            // Remove the implicit receiver because it belongs to different classes
+            // and isn't part of the overload's call domain.
+            return FunctionType.clone(
+                overload,
+                FunctionType.isInstanceMethod(overload) ||
+                    FunctionType.isClassMethod(overload) ||
+                    FunctionType.isConstructorMethod(overload)
+            );
+        };
+        const baseCallDomainTypes = baseOverloads.map(getCallDomainType);
+        const overlapFlags =
+            AssignTypeFlags.SkipReturnTypeCheck |
+            AssignTypeFlags.OverloadOverlap |
+            AssignTypeFlags.DisallowExtraKwargsForTd;
 
         for (const overrideOverload of OverloadedType.getOverloads(overrideMethod)) {
-            let possibleMatchIndex: number | undefined;
+            const compatibleBaseOverloadIndices: number[] = [];
+            const incompatibleOverlappingBaseIndices: number[] = [];
+            const overrideCallDomainType = getCallDomainType(overrideOverload);
 
-            let matchIndex = baseOverloads.findIndex((baseOverload, index) => {
+            baseOverloads.forEach((baseOverload, index) => {
                 // If the override isn't applicable for this base class, skip the check.
                 if (baseClass && !isOverrideMethodApplicable(baseOverload, baseClass)) {
-                    return false;
+                    return;
                 }
 
                 const isCompatible = validateOverrideMethodInternal(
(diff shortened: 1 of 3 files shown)


heejaechang:
### 🟢 Pyright CLI QA — `96880e54` — verdict: green

9 focused check(s) completed.

PR #11707, opened 2026-09-03
Fix variadic overload implementation validation

### LINKED ISSUE #11533: Some incompatible overload implementations are only reported as incompatible in subclasses
**Describe the bug**

In the following minimal example (code including all imports below):

```python
class Base:
    @overload
    def get(self, key): ...
    @overload
    def get(self, *keys): ... # allows item.get()

    def get(self, key, *keys): ... # prohibits item.get(), so should be marked invalid but is not
```

the implementation of `get(self, key, *keys)` should be marked as invalid, as it does not conform to the second overload (`get(self, *keys)` allows calling `item.get()` with no arguments, whereas `get(self, key, *keys)` does not)

The method itself is not marked as invalid by pylance, but subclasses overriding the method with the exact same signature will be (correctly?) marked as invalid:

```python
class Derived(Base):
    def get(self, key, *keys): ... # exact same signature as in Base, but now marked invalid -> reportIncompatibleMethodOverride
```

Additional info:
- the same issue occurs when using `typing_extension.overload` instead of `typing.overload`
- the same issue occurs when decorating the implementations with `@abc.abstractmethod`

**Code or Screenshots**

```python
from typing import overload

class Base:
    @overload
    def get(self, key): ...
    @overload
    def get(self, *keys): ...

    def get(self, key, *keys): ...

class Derived(Base):
    def get(self, key, *keys): ...
```

Image

(translates into)
> The method `get` overrides the class "Base" in an incompatible manner.
> Number of positional parameters is not the same; the base method has 2, but the override method has 3 `Pylance(reportIncompatibleMethodOverride)`
> `base.py(22, 9)`: overridden method

The screenshot was made in VSCode (1.127.0 Universal) with the latest pylance Plugin (2026.2.1) installed.

bschnurr:
## Description

Reject overload implementations whose required leading parameter is masked by an unbounded `*args`, while preserving valid variadic patterns and existing override diagnostics.

I traced overload implementation validation through `Checker.validateOverloadImplementation` and function assignment. An unbounded `*args` parameter was incorrectly treated as satisfying a required source parameter even though it can consume zero arguments.

## Implementation

Introduce an assignment flag for overload implementation call-domain validation. Under that flag, an unbounded `*args` cannot satisfy a required simple source parameter, so invalid base implementations are reported consistently with equivalent subclass overrides.

## Testing

Added `overloadImpl4.py` and its evaluator test with invalid base and derived implementations and valid controls using defaults, keyword-only variadics, `ParamSpec`, and `TypeVarTuple`. The evaluator regression suite passed.

## Addresses

Fixes #11533

### DIFF
--- a/packages/pyright-internal/src/analyzer/checker.ts
+++ b/packages/pyright-internal/src/analyzer/checker.ts
@@ -2819,7 +2819,8 @@ export class Checker extends ParseTreeWalker {
             AssignTypeFlags.SkipReturnTypeCheck |
                 AssignTypeFlags.Contravariant |
                 AssignTypeFlags.SkipSelfClsTypeCheck |
-                AssignTypeFlags.DisallowExtraKwargsForTd
+                AssignTypeFlags.DisallowExtraKwargsForTd |
+                AssignTypeFlags.EnforceOverloadImplCallDomain
         );
 
         // Now check the return types.
--- a/packages/pyright-internal/src/analyzer/typeEvaluator.ts
+++ b/packages/pyright-internal/src/analyzer/typeEvaluator.ts
@@ -28143,7 +28143,9 @@ export function createTypeEvaluator(
         let canAssign = true;
         const checkReturnType = (flags & AssignTypeFlags.SkipReturnTypeCheck) === 0;
         const isContra = (flags & AssignTypeFlags.Contravariant) !== 0;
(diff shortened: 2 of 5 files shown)


heejaechang:
### 🟢 Pyright CLI QA — `141efc45` — verdict: green

4 focused check(s) completed.

PR #11708, opened 2026-09-03
Fix protocol Self specialization for class properties

### LINKED ISSUE #11519: Incorrect type of the returned value of `map()` when applying to a class property

**Describe the bug**
A clear and concise description of the behavior you are seeing and the expected behavior along with steps to reproduce it.

Pyright reports an incorrect error when validating the type of the returned iterator from `map()`. The error is found only when the iteratable being mapped is a **class property**.

**Code or Screenshots**
If possible, provide a minimal, self-contained code sample (surrounded by triple back ticks) to demonstrate the issue. The code should define or import all referenced symbols.

```python
from typing import Iterator, Self

class MetaClass(type):
    meta_attr = list(range(10))

    @property
    def meta_prop(cls) -> list[int]:
        return list(range(10))

class MyClass(metaclass=MetaClass):
    val: int
    class_attr = list(range(10))

    def __init__(self, val: int):
        self.val = val

    @classmethod
    @property
    def class_prop(cls) -> list[int]:
        return list(range(10))

    @property
    def inst_prop(self) -> list[int]:
        return list(range(10))

    @classmethod
    def from_class_prop(cls) -> Iterator[Self]:
        prop = cls.meta_prop  # error in map(cls, prop): "Self@MyClass | _S@map" is not a subtype of "Self@MyClass"
        # prop = cls.class_prop  # same error
        # prop = cls.meta_attr  # no error
        # prop = cls.class_attr  # no error
        # prop: list[int] = cls.meta_prop  # no error
        return map(cls, prop)

    def from_inst_prop(self) -> Iterator[Self]:
        prop = self.inst_prop  # no error
        return map(type(self), prop)
```

Error message:
```
Type "map[Self@MyClass]*" is not assignable to return type "Iterator[Self@MyClass]"
  "map[Self@MyClass]*" is incompatible with protocol "Iterator[Self@MyClass]"
    Type parameter "_T_co@Iterator" is covariant, but "Self@MyClass | _S@map" is not a subtype of "Self@MyClass"
      Type "Self@MyClass | _S@map" is not assignable to type "Self@MyClass"  (reportReturnType)
```

No error is reported when using `mypy`.

Are you running pyright as a VS Code extension, a language server in another editor, integrated into Pylance, or the command-line tool? Which version?

Using Pylance `v2026.2.1` in VS Code. But the bug is also reproducible on [Pyright Playground](https://pyright-play.net/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMoCSMApiAIYy4A0UAysQDbACwAUGwMYNkDOPUAWWIwyAYW58AFPATEAlAC42UFVAjCyAfQowQUALxQGSHjEnl0xSQEYADHLltlqgAIJwsvHGcqAJsWA1DU13MARJLh45KABaAD4jExgAbVQYAF0lVlUcqBBhAFcQFETTczJLG3tHdlYuXn4BOHEGyXURer59IREWvkUfKAA3MgYFTBQYQc6ebRhdA1KzCzQrOwcnbNV-QM1NVCQYPckeRmAaEbGJmAGt3KhTpgA6S8XLzZyXGfaACzBfQZuDykeCDHZQGYhDwRBhRWIJYymVKTTKDHL5GBFEqI5YVVZVDa1T6hTygu5QcGoUxQsInM7ReJLZEZLL3FQYrFLcqVdY1QHfYR-AHk8GgSCaSEkmFwxlEUgUXDJehMVHknIkxaRJ7tLQaqAAYigpHAelQajI4UiNBJiigACJlcAXE0%2BvwAD5QTS0FwQC12zD8FBgfBkB4FABGMmIUDAgQdZ2dzQkPDtaNUho1hi1ko8KkNPDI6iNIBNabzUEzENh2uCOj0BqgQeLpbV6Yruaz1chdfLTeNuDLDZJ4xxzPSmurOppCF7YGbA9b7MKxXNlth1o8fJFASgYog%2BxQ1Klj2ADIScvIlBASrOqrZ7bCixPTypRz1hr7JYX945K994SjOkmDkDcwjkIA) with Pyright version `Latest (1.1.410)`.

bschnurr:
## Description

Keep `Self`-based constructor inference stable when `map(cls, values)` reads values through class or metaclass properties, preventing unrelated synthetic return types from entering the result.

I traced protocol assignment and descriptor-backed constructor inference through `assignToProtocolInternal`. A conditioned `Self` from an enclosing expression could be reused even when its bound class was unrelated to the current source-class hierarchy.

## Implementation

Reuse a conditioned `Self` only when its bound class belongs to the current source-class hierarchy. Otherwise fall back to the concrete source class so descriptor access cannot leak an unrelated `Self` into constructor callables.

## Testing

Added `self12.py` and its evaluator test covering metaclass properties, class properties, class attributes, annotated locals, instance properties, and custom iterators, with negative constructor type and arity cases. The evaluator regression suite and build passed.

## Addresses

Fixes #11519

### DIFF
--- a/packages/pyright-internal/src/analyzer/protocols.ts
+++ b/packages/pyright-internal/src/analyzer/protocols.ts
@@ -41,6 +41,7 @@ import {
     applySolvedTypeVars,
     ClassMember,
     containsLiteralType,
+    derivesFromClassRecursive,
     lookUpClassMember,
     makeFunctionTypeVarsBound,
     MemberAccessFlags,
@@ -378,9 +379,19 @@ function assignToProtocolInternal(
 
     let selfType: ClassType | TypeVarType | undefined;
     if (isClass(srcType)) {
-        // If the srcType is conditioned on "self", use "Self" as the selfType.
-        // Otherwise use the class type for selfType.
-        const synthCond = srcType.props?.condition?.find((c) => TypeVarType.isSelf(c.typeVar));
+        const srcClassType = srcType;
+
+        // If the srcType is conditioned on its own "self", use "Self" as the
+        // selfType. A condition from an enclosing expression's Self type should
+        // not affect the specialization of this class.
+        const synthCond = srcClassType.props?.condition?.find((c) => {
+            const boundType = c.typeVar.shared.boundType;
+            if (!TypeVarType.isSelf(c.typeVar) || !boundType || !isClass(boundType)) {
+                return false;
+            }
+
+            return derivesFromClassRecursive(srcClassType, boundType, /* ignoreUnknown */ true);
+        });
         if (synthCond) {
             selfType = synthesizeTypeVarForSelfCls(
                 TypeBase.cloneForCondition(srcType, undefined),
--- a/packages/pyright-internal/src/tests/samples/self12.py
+++ b/packages/pyright-internal/src/tests/samples/self12.py
@@ -1,0 +1,121 @@
+# This sample tests contextual type inference for constructors passed as
+# callables when the iterable is read through a descriptor.
+
+from collections.abc import Callable, Iterable, Iterator
+from typing import Self
+
+
+class Meta(type):
+    meta_values = [1, 2, 3]
+
+    @property
+    def values(cls) -> list[int]:
+        return [1, 2, 3]
+
+
+class ClassA(metaclass=Meta):
+    class_values = [1, 2, 3]
+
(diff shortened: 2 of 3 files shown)


heejaechang:
### 🟢 Pyright CLI QA — `4eec8b0b` — verdict: green

6 focused check(s) completed.

PR #11711, opened 2026-09-03
Bind slot attributes initialized in __new__

### LINKED ISSUE #11373: Setting attributes of class instance in `__new__` method included in `__slots__` causes those attributes to be inferred as `Unbound`
**Describe the bug**
When a class defines a `__new__` method rather than `__init__` and uses `__slots__` (e.g. for immutable objects with a fixed structure), the attributes set on the return object, even if known to be an instance of that class, in its body, are not recognized to have been 'bound' if not re-set elsewhere, and are revealed to be `Unbound` when accessed in other methods of the class.

**Code or Screenshots**

```python
from _thread import LockType
class Foo:
    __slots__ = '_spam', '_ham'
    def __new__(cls, tup: tuple[int, LockType]):
        self = super().__new__(cls)
        self._spam, self._ham = tup
        return self
    def bar(self): reveal_type(self._spam) # no diagnostic produced
    # when I hover over the attribute, it says 'Unbound'
    def baz(self):
        with self._ham: # ERROR!
             self.bar()
```

The error shown in the above case is twofold:

```md
Object of type "Unbound" cannot be used with "with" because it does not correctly implement __enter__ Pylance[reportGeneralTypeIssues](https://github.com/microsoft/pylance-release/blob/main/docs/diagnostics/reportGeneralTypeIssues.md)
Object of type "Unbound" cannot be used with "with" because it does not correctly implement __exit__ Pylance[reportGeneralTypeIssues](https://github.com/microsoft/pylance-release/blob/main/docs/diagnostics/reportGeneralTypeIssues.md)
```

VS Code extension; preview (v2026.2.100).
I have seen analogs of this issue in previous versions as well.

bschnurr:
## Description

Bind slot-backed instance attributes assigned inside `__new__`, including tuple unpacking and direct writes, while keeping conditional initialization and alternate-return paths conservative.

I traced member binding and uninitialized-variable checks for concrete instances created by `object.__new__(cls)` and `super().__new__(cls)`. Assignments to the returned instance were not treated like initialization, leaving declared slot members inferred as `Unbound`.

## Implementation

Recognize member writes in `__new__` when they target a concrete instance created once at top level and returned from the method. Cache the per-function analysis, support tuple unpacking and direct assignment, and treat declarations initialized there like `__init__` assignments for uninitialized-member checks.

## Testing

Added `slots5.py`, `uninitializedVariable4.py`, and checker/evaluator coverage for tuple and direct assignment, annotated and inherited slots, normal `__init__`, conditional writes, early returns, and returning a different object. The focused checker/evaluator regressions passed.

## Addresses

Fixes #11373

### DIFF
--- a/packages/pyright-internal/src/analyzer/binder.ts
+++ b/packages/pyright-internal/src/analyzer/binder.ts
@@ -121,7 +121,7 @@ import {
 import { ImplicitImport, ImportResult, ImportType } from './importResult';
 import { getWildcardImportNames } from './importStatementUtils';
 import * as ParseTreeUtils from './parseTreeUtils';
-import { ParseTreeWalker } from './parseTreeWalker';
+import { ParseTreeWalker, getChildNodes } from './parseTreeWalker';
 import { CellChainIndexProvider } from './cellChainIndex';
 import {
     ChainedModuleLevelLookupContext,
@@ -140,6 +140,12 @@ interface MemberAccessInfo {
     methodNode: FunctionNode;
     classScope: Scope;
     isInstanceMember: boolean;
+}
+
+interface NewMethodInstanceAnalysis {
+    instanceAssignment: AssignmentNode | undefined;
+    writeCount: number;
+    returnedInstanceNodes: ReturnNode[];
 }
 
 interface DeferredBindingTask {
@@ -318,6 +324,8 @@ export class Binder extends ParseTreeWalker {
 
     // List of string nodes associated with the "__all__" symbol.
     private _dunderAllStringNodes: StringNode[] = [];
+
+    private _newMethodInstanceAnalysisCache = new WeakMap<FunctionNode, Map<string, NewMethodInstanceAnalysis>>();
 
     // One or more statements are manipulating __all__ in a manner that a
     // static analyzer doesn't understand.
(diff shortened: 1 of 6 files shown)


heejaechang:
### 🔴 Pyright CLI QA — `7648f029` — verdict: red

5 focused check(s) completed.

#### Needs review: `reportUninitializedInstanceVariable` no longer reported

The base reports `reportUninitializedInstanceVariable` for this example; the PR does not. Is this change intended?

**`main.py`**

```python
import object


class Target:
    __slots__ = ("value",)

    def __new__(cls) -> object.Foreign:
        instance = object.__new__(cls)
        instance.value = "foreign"
        return instance


def inspect_target(target: Target) -> None:
    reveal_type(target.value)
    assert_type(target.value, str)
```

**`object.py`**

```python
import builtins


class Foreign:
    def __init__(self) -> None:
        self.value = ""


def __new__(cls: type[builtins.object]) -> Foreign:
    return Foreign()
```

**`pyrightconfig.json`**

```json
{
    "typeCheckingMode": "strict",
    "pythonVersion": "3.13",
    "reportUninitializedInstanceVariable": "error"
}
```

```text
```

**Base `fa9b200906fadc093e862ce62ec1afa985c7233d`**

```json
{
  "exit_code": 1,
  "diagnostics": [
    {
      "file": "main.py",
      "severity": "error",
      "rule": "reportUninitializedInstanceVariable",
      "message": "Instance variable \"value\" is not initialized in the class body or __init__ method",
      "range": {
        "start": [
          5,
          17
        ],
        "end": [
          5,
          24
        ]
      }
    }
  ]
}
```

**PR `7648f029e7d020084a951f3f5a79e19a34073f82`**

```json
{
  "exit_code": 0,
  "diagnostics": [
    {
      "file": "main.py",
      "severity": "information",
      "rule": null,
      "message": "Type of \"target.value\" is \"str\"",
      "range": {
        "start": [
          14,
          16
        ],
        "end": [
          14,
          28
        ]
      }
    }
  ]
}
```

</details>

PR #11718, opened 2026-09-08
Optimize recursive sequence protocol rejection with requirement checks

bschnurr:
## Summary

Reject provably incompatible builtin-list/recursive-sequence protocol assignments before reconciling expensive overloaded element bounds. This is a negative-only requirement check, not relation memoization, solver reuse, deferred inference, or COW.

Based directly on upstream c77393240247e017db6538d35876e723c518e06c. Production changes are limited to protocols.ts and a function/method-class accessor in the evaluator and its interface. No NumPy-specific member names or function-category absence assumptions. Recognizes actual list indexing contracts, guards int/slice overlap, and requires both leaf and recursive-sequence alternatives to fail. Populated unresolved caller contexts fall back to structural matching. Fast negatives are not universal incompatibility proofs, and diagnostic retries retain failed-call inference.

Related to #11695, but independently based and scoped. That PR and its branch are unchanged.

## Teaching Walkthrough: Why Two Function Objects Triggered Explosive Checking

```python
import numpy as np

result = np.array([np.mean, np.sum])
reveal_type(result)
```

**The surprising part:** neither `mean` nor `sum` is called. They are values in a list. Nevertheless, inferring whether that list satisfies a generic array-like protocol can make the checker compare their overloaded signatures and repeatedly expand its inference state.

### 1. Try the Generic Candidate Before the Broad Fallback

NumPy 2.4.6 declares five `array` overloads. For this argument shape, overloads 3 and 5 remain: the generic `_ArrayLike[S]` overload and a broad `Any`-input fallback. Two other overloads require `subok=True`, and another requires `dtype`, so those do not apply here.

The relevant generic contract can be abbreviated as follows (these are explanatory names, not complete NumPy stubs):

```text
ArrayLike[S] = SupportsArray[dtype[S]]
             | NestedSequence[SupportsArray[dtype[S]]]

SupportsArray[D]:
    __array__() -> ndarray[Any, D]

NestedSequence[E]:
    __len__() -> int
    __getitem__(int) -> E | NestedSequence[E]
    __iter__() -> Iterator[E | NestedSequence[E]]
```

The checker must investigate the generic candidate before settling on the fallback. A cheap final return type does not imply cheap failed candidates.

### 2. The Inline List Retains Expensive Callable Information

The candidate parameter supplies an expected type to the list literal. With that context, the list can retain information equivalent to `list[MeanOverloads | SumOverloads]` during matching. In this NumPy version, `mean` has 11 overload declarations and `sum` has 9.

This matters even under default settings. Moving the list into an unannotated variable first can instead produce `list[Unknown]` under default heterogeneous-list inference. That changes the matching problem; it is not an equivalent performance fix.

### 3. Recursive Protocol Matching Infers an Element Type

Ordinary structural matching compares the source list's members with the sequence protocol's members. Indexing and iteration expose the element type repeatedly. The matcher accumulates protocol-local bounds for `E`, then checks inferred protocol arguments against the requested specialization.

Three different variable families participate: NumPy's scalar `S`, the sequence's element `E`, and generic variables inside individual `mean`/`sum` signatures. Keeping them distinct explains how work on callable signatures can arise before the final `SupportsArray` requirement fails.

### 4. Adding a Lower Bound Can Compare `mean` Against `sum`

A lower bound says "the inferred type must accommodate this type". Adding another lower bound does not always immediately construct a union. `assignUnconstrainedTypeVar` first tests whether the existing bound accommodates the incoming one, or the reverse, by calling `assignType` with the current constraints.

In the earlier instrumented precise NumPy sequence probe, the first expansion occurred here:

```text
protocol element variable: E
existing lower bound:     overloaded mean
incoming lower bound:     overloaded sum
next operation:           callable assignability / subsumption checks
```

This is **callable compatibility nested inside outer call resolution**, not an attempt to execute either function. The checker is asking whether one bound can cover another while inferring a common element type.

### 5. Successful Overload Matches Multiply Constraint Alternatives

For a single destination signature and an overloaded source, Pyright collects successful source-signature inference alternatives rather than stopping at the first success. Each trial clones or selects a signature-specific constraint context; the successful sets are collected by `addConstraintSets`. An overloaded destination requires support for each destination signature.

A constraint set records a correlated alternative of type-variable bounds and signature scopes. If two source signatures work for each of several successive comparisons, a simplified growth pattern is:

```text
incoming alternatives       1
after a branching check      2
after another check          4
after another check          8
after another check         16
```

With two retained branches at each of k stages, the schematic count is 2^k. Actual growth depends on scopes, specialization, which signatures succeed, and implementation limits. It is not a claim that all overload checking is exponential.

We reproduced `1 -> 2 -> 4 -> 8 -> 16` without NumPy using a recursive-sequence fixture with three concrete `first` signatures (`int`, `str`, `bytes` inputs) and two generic `second[T]` signatures that can accept those inputs. The two-versus-two version took the equal-count, positional one-to-one shortcut; the three-versus-two version exposed the branching path.

**Why this is worse than simply 11 x 9 comparisons:** a signature comparison may process many already-existing alternatives. Subsequent protocol-member comparisons and bound reconciliation can revisit specialized overloads and the expanded state, causing repeated cloning, solving, and substitution. An earlier related precise/prebound profile recorded 4,136 solves, 1,017,385 constraint-set visits, and 4,456,285 clones. Those counts belong to that profiling fixture and revision, not to the final exact-call benchmark in this PR.

The existing 1,024 constraint-set limit does not undo this cost: incoming collections are installed only when their size is strictly below the limit, and expensive work can occur while building those collections and repeatedly processing large retained families. The separate 256-combination limit in outer call union expansion is not this mechanism.

### 6. The Expensive Generic Candidate Ultimately Fails

For the stock stubs in this example, the function objects cannot meet either the array-conversion leaf requirement or the recursive sequence alternative. The generic candidate fails, and the broad fallback accepts the input and produces:

```text
ndarray[tuple[Any, ...], dtype[Any]]
```

That broad result is NumPy's declared fallback result. It does not mean Pyright erased the callable signatures to make the generic candidate succeed.

### 7. Reject at the Necessary Requirement Before Creating Those Alternatives

The new missing-member path recognizes a guarded builtin-list/recursive-sequence relationship. Once the actual indexing contracts are verified, each source element must satisfy `Leaf | NestedSequence[Leaf]`. If every eligible element lacks both the leaf's required member and `__getitem__`, neither alternative can work. The matcher can reject this candidate before reconciling overloaded element bounds.

For NumPy the discovered leaf member is `__array__`; the implementation does not hardcode NumPy or that member name. It looks up actual function/method capabilities and falls back for uncertain or customized contracts. The separate reduced-element comparison path has its own guards and is not the same missing-member proof.

```text
Before: generic candidate -> infer element bounds -> reconcile overloads
        -> expand/clone/solve alternatives -> reject -> ordinary fallback

After:  generic candidate -> verify necessary requirements -> reject
        -> ordinary fallback
```

This changes where a proven negative is discovered, not overload order, solver branching rules, or result precision. Rejection is still stateful: failed comparisons can infer caller variables, and cached answers can affect later checks. Populated unresolved caller contexts therefore fall back, detailed non-speculative diagnostic checks retain structural inference, and fast negatives never become universal incompatibility proofs. The earlier idea of blindly unioning overloaded bounds was rejected because it lost enclosing-variable inference.

**Evidence boundary:** the causal mechanism is supported by the earlier first-expansion trace and NumPy-free reproduction. The diagram is a logical walkthrough, not a timestamped trace of every invocation in the exact call. The independent, rebuilt, default-settings exact-call benchmark below measures the retained change against unmodified upstream: 85.31s to 1.15s median, with identical diagnostics and dtype controls.

For source navigation, see the [full overload walkthrough](https://github.com/bschnurr/pyright/blob/447b76c5732d900b1726d1d58604c9cd0bc6184d/docs/overload-resolution-numpy.md) and [review evidence and limits](https://github.com/bschnurr/pyright/blob/447b76c5732d900b1726d1d58604c9cd0bc6184d/docs/requirement-first-validation.md). The key code path is `getTypeOfListOrSet` -> `assignToProtocolInternal` -> `assignUnconstrainedTypeVar` -> callable branches of `assignType` -> `cloneWithSignature` / `addConstraintSets`.

## Validation

- Full local suite: 89 suites / 3,067 tests pass, no extra exclusions; test server rebuilt. Existing listener/worker-teardown warnings remain.
- Expanded semantic review: 85 cases across speculative, populated-alternative, and recursive-cycle suites pass normally and with shortcut dispatch disabled. Includes exception unwinding and failed recursive ancestor cache checks. Existing precision assertions are retained.
- Portable state audit: 64 traces / 4,992 exact observations, including later assignments and caller state.
- Complexity gates cover overloaded-element width and recursive depth/member growth. New pinned NumPy canary validates exact result type plus positive float64 and negative int64 controls.
- Fresh core/CLI builds, targeted lint/format, editor diagnostics, and whitespace checks pass.

## Direct Upstream Benchmark

Node 24.15.0, Python 3.12.3, NumPy 2.4.6; exact default-settings `result = np.array([np.mean, np.sum])`. Both roots rebuilt, three alternating fresh-process pairs against unmodified upstream.

| Run | Candidate wall seconds | Upstream wall seconds |
| --- | ---: | ---: |
| 1 | 1.196166 | 86.557670 |
| 2 | 1.149528 | 85.309787 |
| 3 | 1.143793 | 83.754357 |
| Median | 1.149528 | 85.309787 |

Approximately 74.2x median improvement for this exact workload. All diagnostics match, including `ndarray[tuple[Any, ...], dtype[Any]]`; all six dtype-control runs pass. The historical ~16s comparator is an older protocol-control branch, NOT upstream. No general workload speedup claim.

## Review Status

Draft pending hosted CI and independent internal model review. Start with docs/requirement-first-validation.md (reviewer map, commands, comparator provenance, coverage limits) and docs/overload-resolution-numpy.md (algorithm and overload behavior).

The bounded tests do not prove arbitrary coinductive-cache rollback or deep mutable-type ownership. Constraint clones retain bound type references; this change does not introduce solver result sharing. Other flags and broader speculative re-entry remain review considerations. No independent model review is claimed yet. Local results are not hosted CI results.


bschnurr:
## Investigation of the mypy_primer diff

Follow-up to https://github.com/microsoft/pyright/pull/11718#issuecomment-5589237410: the scikit-learn and SymPy differences have different explanations. They should not all be dismissed as random.

### Scikit-learn: reproducible cache-related diagnostic change

Compared upstream parent `d72e2fe89` with PR merge `dcb222930` (PR head `b461d61eb`), using Node 24.15.0, Python 3.11.16, NumPy 2.4.6, and scikit-learn `ffbd0f8bbc3be496a9fb7a8ef26940af47723810`. Both checkers used the same project checkout and Python environment.

| Checker | Run 1 errors | Run 2 errors |
| --- | ---: | ---: |
| Upstream | 17,184 | 17,184 |
| PR | 17,185 | 17,185 |

Each checker's complete normalized diagnostic output was identical across its two fresh-process runs. The only difference between checkers was the reported `nan_to_num` argument error at `sklearn/model_selection/_search.py:1263`.

An isolated experiment restoring **only** the old `isConstraintTrackerSame` implementation (`context1.isSame(context2)`) in the PR checker removed that error and made the entire output match upstream exactly. The stricter comparison of constraint scopes and `retainLiterals` is therefore causal in this reproduction; this is not merely a direct rejection by the list fast path.

Importantly, **upstream and the PR both report the identical error when checking the affected file alone**, with identical full output (26 errors, 1 warning). The vendored `Array` protocol requires `__invert__`, whose NumPy self annotation excludes floating arrays. This supports the interpretation that stricter cache matching exposes an existing declared-type mismatch and makes whole-project checking agree with cold-file checking. It does not imply the runtime `nan_to_num` call is invalid. The exact cache interaction has not yet been reduced to a focused regression test.

### SymPy: variability also occurs without this PR

Using SymPy `e0e25c8eaab46155ee7c35c783e3f6f41e147b9b`, repeated fresh-process checks produced:

| Checker | Run 1 errors | Run 2 errors |
| --- | ---: | ---: |
| Upstream | 38,930 | 38,934 |
| PR | 38,930 | 38,934 |

The complete normalized diagnostic outputs matched across checkers for the corresponding runs, not just the totals. Both `Alg.subs` errors at `recurr.py:440` and `:442` were absent in the first run and present in the second run **on both upstream and the PR**. Baseline-only variability also included the hosted changes at `single.py:867`, `recurr.py:249`, and `solvers.py:956`. These differences are not reliable evidence of a PR regression. The underlying source of the variability has not been isolated.

### Assessment and limits

The evidence supports retaining the stricter cache comparison, documenting the scikit-learn diagnostic change, and adding a reduced regression test. It does not support claiming complete diagnostic equivalence or attributing all primer differences to noise.

Exact hosted dependency/project provenance could not be recovered because CI log downloads required authorization; local totals differ from CI, and not every hosted SymPy line has been accounted for. The earlier Python 3.12 / NumPy 2.5.2 local comparison did not reproduce the scikit-learn delta and was insufficient to dismiss it. Python 3.11 selected NumPy 2.4.6 and reproduced it.

All diagnostic comparisons above retained full output. The temporary cache-key experiment was restored afterward; no production changes or test expectation changes were made during this investigation.

bschnurr:
looks like home-assistant is no longer crashing

heejaechang:
### 🟢 Pyright CLI QA — `b461d61e` — verdict: green

4 focused check(s) completed.

heejaechang:

heejaechang:

**Summary:** \[unavailable\] Container verification could not start and local execution was not authorized for this PR HEAD\: Podman CLI is not installed or is not on PATH.

**Test runs:** none recorded.
</details>

heejaechang (approved):
Approved via Review Center.

rchiodo (approved):
Approved via Review Center.

PR #11745, opened 2026-09-13
Fix TypedDict branch union inference

### LINKED ISSUE #11743: False positive when using a TypedDict after multiple branches may assign to it: all types inside TypedDict turn into object

**Describe the bug**
When multiple code branches assign to a TypedDict-typed variable, Pyright incorrectly discards all information about the types inside the TypedDict and thinks they're all `object`s. This makes it impossible to use the dict variable in any way without a type cast.

**Code or Screenshots**
If possible, provide a minimal, self-contained code sample (surrounded by triple back ticks) to demonstrate the issue. The code should define or import all referenced symbols.

```python
from typing import TypedDict, Unpack, assert_type

class AwesomeDict(TypedDict, total=False):
    foo: int
    bar: int
    baz: int

def consume_a(arg: str, **kw: Unpack[AwesomeDict]): pass
def get_bool() -> bool: raise NotImplementedError

a: AwesomeDict = {}
if get_bool():
    a = {"foo": 2, "bar": 3}
else:
    a = {"foo": 3, "baz": 4}

assert_type(a, AwesomeDict)
# Expected: no error
# Actual:
#    "assert_type" mismatch: expected "AwesomeDict" but received "AwesomeDict"
consume_a("hi", **a)
# Expected: no error
# Actual:
#    Argument of type "object" cannot be assigned to parameter "foo" of type "int" in function "consume_a"
#      "object" is not assignable to "int"
#    Argument of type "object" cannot be assigned to parameter "bar" of type "int" in function "consume_a"
#      "object" is not assignable to "int"
#    Argument of type "object" cannot be assigned to parameter "baz" of type "int" in function "consume_a"
#      "object" is not assignable to "int"
```

Pyright 1.1.412, Pyright Playground also reproduces this bug.

GruffElixir:
### Summary

Fixes #11743.

When control flow combines two instances of the same TypedDict narrowed to incompatible optional-key sets, treat the result as the un-narrowed TypedDict. This preserves the declared optional-key availability and avoids object-valued union diagnostics.

### Validation

- pnpm --dir packages/pyright-internal exec jest typeEvaluator7.test.ts -t TypedDict30 --runInBand --forceExit
- pnpm --dir packages/pyright-internal exec jest typeEvaluator1.test.ts typeEvaluator7.test.ts -t TypeNarrowingTypedDict --runInBand --forceExit
- Full typeEvaluator1.test.ts suite: 160 passed
- Full typeEvaluator7.test.ts suite: 171 passed
- pnpm --dir packages/pyright-internal run build


GruffElixir:

bschnurr:

packages/pyright-internal/src/analyzer/types.ts
```diff
@@ -4087,6 +4087,15 @@ function _addTypeIfUnique(unionType: UnionType, typeToAdd: UnionableType, elideR
...
                         unionType.priv.subtypes[i] = typeToAdd;
                         return;
+                    } else {
+                        // If neither narrowed form subsumes the other, retain
+                        // the un-narrowed TypedDict shape rather than exposing
+                        // a union of incompatible presence states.
+                        unionType.priv.subtypes[i] = ClassType.cloneForNarrowedTypedDictEntries(
+                            type,
+                            undefined
+                        );
```
bschnurr, line 4097: > **Warning** · Non-blocking recommendation

Clearing the narrowed-entry map also loses that `foo` is provided in both branches, so a subsequent `a["foo"]` can regain the optional-key access diagnostic. Preserve entries marked `isProvided` in both forms, and extend `typedDict30.py` with an access to `a["foo"]` to cover the shared narrowing fact.

bschnurr (approved):
Approved via Review Center.

heejaechang (approved):
Approved via Review Center.

bschnurr (approved):

rchiodo (approved):

PR #11732, opened 2026-09-10
Preserve container types for ambiguous overload returns

rchiodo:
## Summary

Addresses #11731. Fixes the reported shared-container regression and the related Spark timedelta-operation regression. Also adds advisory explanations of mypy_primer differences for future PRs. The broader historical pandas-stubs differences are not all resolved.

- Preserve shared generic return structure when nested `Any` or `Unknown` makes overload resolution ambiguous, including specialized `self` parameters and inferred constructors. `Container[Any].copy()` retains its container type and nonexistent-member diagnostics.
- Preserve correlated protocol type-variable solutions when binding methods, rather than discarding every signature after the first. Spark's `(Timestamp - Series[Any]) // np.timedelta64(...) - 1` now resolves to `Series[int]` instead of reporting an unsupported operation.
- Keep ordinary protocol argument inference and existing first-signature fallbacks unchanged. Newly recognized bound alternatives cannot hide later declared overloads when receiver ambiguity is unresolved.
- Add a separate, advisory agentic workflow that explains the complete mypy_primer output without replacing the existing raw comment or automatically approving a PR.

## Precision rationale

Classification **B: Pyright inference limitation**; no typeshed changes.

[python/typing#2350](https://github.com/python/typing/pull/2350) permits more precise gradual overload results that satisfy both assignability and supported-operation requirements. The container merge recursively retains compatible generic arguments and tuple shapes, and verifies assignability to every retained return type. Unsupported shapes, differing non-invariant arguments, incompatible nominal families, and mixtures with `Never` retain conservative fallbacks.

The Spark follow-up is intentionally limited to protocol-annotated method binding. An experimental broader change affected existing precise expectations and was discarded. Existing Pyright expectations remain unchanged by the Spark follow-up; no assertions or diagnostic expectations were weakened to hide failures.

### Reviewed precision changes in pandas-stubs

On unchanged pandas-stubs commit [`507243e3c1c1f0a6d92b5c409684fd691e2d6980`](https://github.com/pandas-dev/pandas-stubs/commit/507243e3c1c1f0a6d92b5c409684fd691e2d6980), the Spark correction introduces these four assertion differences relative to the initial container fix:

| Test | Lines | Before | After |
| --- | --- | --- | --- |
| `tests/series/test_floordiv.py` | 47, 64, 107 | `Series[Timedelta]` | `Series[Any]` |
| `tests/indexes/test_floordiv.py` | 47 | `TimedeltaIndex` | `Any` |

[...]

Issue #11731, opened 2026-09-10
Regression: specialised self overloads erase container types and suppress attribute errors


cmp0xff:
## Describe the bug

Starting with Pyright 1.1.412, `Container[Any].copy()` becomes `Any` instead of `Container[Any]`, losing the outer container type and suppressing the nonexistent-member error. Expected: preserve `Container[Any]` and report the invalid attribute, since every overload returns a container.

## Code or Screenshots

```python
from typing import Any, Generic, Never, TypeVar, overload, reveal_type

T = TypeVar("T")

class Container(Generic[T]):
    @overload
    def copy(self: "Container[Never]") -> "Container[Any]": ...
    @overload
    def copy(self: "Container[int]") -> "Container[int]": ...
    @overload
    def copy(self: "Container[float]") -> "Container[float]": ...
    def copy(self) -> "Container[Any]":
        return self

def example(value: Container[Any], concrete: Container[int]) -> None:
    reveal_type(value.copy())
    reveal_type(concrete.copy())
    value.copy().nonexistent_member()
```

## VS Code extension or command-line

- https://pyright-play.net/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMoCCKcANFAOICmKlISAxmQHKUButZAKopQGoCGIMmHYgANmH4ATMiDaV%2BYgPrwElAFDrOUALxRuagSAAUAIk6mAlJvpj%2BAZ3tQAwmBQx%2BqWsao069AG1OAF1LAC51KCioAAERWglpSOipSmAoejAEOGN7SjFgMKhTV3dPPwCWUWCrKABaAD5i0o8vEACiOBqigDo%2B5Ki40USpAahU9Mzs3PzC5rdWitQYGst6ppKF8toA5e6oPp6xoYTJUejxtIysnLyCos2ytoDgRJXaxvmnitfJd97%2BhcJtdpndgGtPo9FjtOt0xhc5DAAK4gFBQMGaYGUAAe-GwYkoxlYiiRlCKLW27VhZEyKHoiLJLi2zz2EKaTDcZLGcnYihUPCJJMoPSmOUsaygAGIoNk6GgABb4NBIdhOTrc%2BR81SE2n0ygwYWi4ziqLS2VIBVKlWUJwUlnuYJjYliUkim7GnooTnYpD2A3uJQQSgQABG3glZrgcsV4zANqgXvwIE8eSAA
[...]

PR #11746, opened 2026-09-14
Added pandas-stubs primer gate

bschnurr:
This pull request introduces an additional blocking compatibility check for `pandas-stubs` diagnostics in the `mypy_primer_pr.yaml` workflow. This ensures that any changes affecting `pandas-stubs` are detected and will block the PR if diagnostics differ. Related regression https://github.com/microsoft/pyright/issues/11731

**Testing and compatibility enforcement:**

* Added a new workflow step to require stable `pandas-stubs` diagnostics using `mypy_primer` with `pyright` as the type checker, making any change in its diagnostics blocking for the PR.
* PR #11732 already added the right focused regression coverage. It verifies that ambiguous overload matching involving Any preserves the common outer container instead of collapsing the entire result to Any.
*


bschnurr:
apparently these already run in mypy_primer

PR #11634, opened 2026-08-13
Fix nested TypeForm alias evaluation

rchiodo:
## Summary
- consult the TypeForm-specific cache for nested generic arguments evaluated in a TypeForm context
- preserve incomplete-cache generation checks when selecting cached results
- add regression coverage for PEP 695 aliases used as bounded generic arguments

Fixes #11633

## Testing
- `npx jest typeEvaluator8.test --forceExit -t "TypeForm"`
- `npm run build` (packages/pyright-internal)

PR #11766, opened 2026-09-15
Optimize contextual TypeForm cache lookup

bschnurr:
## Summary

- add an identity-first lookup for contextual type cache entries
- search recent contextual entries before falling back to structural type comparison
- use the optimized lookup for the TypeForm cache introduced in #11634
- cover identity, structurally equivalent, context-free, and missing lookup cases

## Motivation

PR #11634 made nested TypeForm evaluation consult a separate contextual cache. A node can have up to eight entries in this cache, distinguished by expected type.

The existing lookup uses `Array.find` from oldest to newest. For each entry, `contextualTypeCacheEntryMatches` calls `isTypeSame` unless both expected types are absent. `isTypeSame` is a recursive structural comparison, so a cache hit can still require several type-tree comparisons before reaching the matching entry. This lookup occurs from the type evaluator's frequently used `getTypeOfExpression` path and from the additional TypeForm cache probes added by #11634.

In normal contextual evaluation, the caller commonly supplies the exact same expected-type object that was used to create the cache entry. Structural equality is therefore unnecessary for the common case.

## Optimization

`findContextualTypeCacheEntry` performs lookup in two stages:

1. Scan newest to oldest using object identity (`entry.expectedType === expectedType`). This handles the common repeated-context lookup without calling `isTypeSame`.
2. Only if identity lookup misses, scan again using `isTypeSame`. This preserves matches for separately allocated but structurally equivalent expected types.

Entries are appended when added or replaced, so checking newest entries first also favors recently evaluated contexts. The undefined-expected-type case is handled entirely by the identity pass.

`contextualTypeCacheEntryMatches` also receives the identity fast path because it remains in use when contextual entries are added or replaced.

The matching semantics and eight-entry eviction policy are unchanged. This only changes how an existing matching entry is found.

## Performance

A focused microbenchmark performs one million lookups in an eight-entry contextual cache where the most recent entry has the same expected-type object:

| Lookup | Time |
| --- | ---: |
| Existing linear structural lookup | ~190 ms |
| Identity-first lookup | ~8 ms |

Five repeated runs were stable within approximately 2 ms. This isolates lookup overhead rather than claiming an equivalent end-to-end type-checking improvement. The end-to-end benefit depends on the number and shape of contextual TypeForm cache lookups in the analyzed project.

## Testing

- `pnpm exec jest typeCacheUtils.test typeEvaluator8.test --runInBand --forceExit`
- `pnpm run typecheck`


rchiodo (approved):

heejaechang:
### ✅ Pyright CLI QA — `e240614c` — verdict: **green**

5 focused check(s) completed.

PR #11739, opened 2026-09-11
Revert nested overload materialization and dependent checker fixes

rchiodo:
## Summary

Selectively restore pre-materialization overload behavior after the pandas regressions introduced by nested overload materialization and unsuccessful incremental design experiments. This is a rollback to establish a baseline before a separate redesign, **not a conformance fix or a replacement algorithm**.

Based on `microsoft/pyright:main` at `e4967c79c05c7823f226893208d0324f0c1accd8`. The rollback was prepared and measured in isolation; its current-behavior test changes were reviewed before publication.

## Reverted and retained scope

- Reverse #11601's nested Any/Unknown materialization, overload candidate filtering, and inferred constructor-self handoff.
- Reverse the dependent constructor-union repair in #11621.
- Reverse only the checker portions of #11732: common gradual return merging, correlated protocol-self alternatives, and their supporting flags/signature APIs. Adapt the API-coupled `mapSignatures` test to the restored callback contract; this does not claim equivalent expansion coverage.
- **Preserve all primer-analysis tooling, workflow/configuration, documentation, and tests from #11732.** Retain the independent protocol negative-cache correctness fix #11649, defaulted constructor specialization fix #11596, and other subsequent changes, including newer typeshed, dependencies, actions, and synchronization work.

No downstream expectations are changed. Earlier experimental algorithms and local evidence artifacts are not included.

## Measured pandas compatibility

Using pandas-stubs commit `711cf2f7eedaee369812d293a9cc83a08432096b`, Python 3.12.13, and the same read-only source and installed-wheel snapshots:

| Snapshot | Files analyzed | Untouched starting main | Earlier #11732 merged baseline | Rollback | Retained Pyright 1.1.411 pin |
| --- | ---: | ---: | ---: | ---: | ---: |
| Source checkout | 356 | 47 errors | 47 errors | 0 errors | 0 errors |
| Installed wheel | 172 | 47 errors | 47 errors | 0 errors | 0 errors |

All runs had zero warnings. Each shape removes the same 47 diagnostics with **no additions**, compared by normalized file, range, rule, and message: 23 assertion mismatches and 24 unnecessary-ignore diagnostics. These are diagnostic differences, not a claim of 47 distinct language bugs.

[...]

PR #11780, opened 2026-09-18
Retain bounded alternatives for ambiguous overload calls

rchiodo:
## Why this change

Restore the bounded local overload-conformance case motivating #11601 without repeating the loss of type information and error checking reported in #11731. This is a new implementation following the #11739 rollback, not a reapplication of the original widening behavior.

The [overload specification](https://typing.python.org/en/latest/spec/overload.html#step-5) requires a permissive result when gradual arguments leave different return types ambiguous. For example, a `list[Any]` argument can match overloads returning `list[int]` and `list[str]`. Choosing only the first return incorrectly rejects the valid `list[str]` assignment and string append.

Making that result `Any` or `Unknown` is permitted by the specification, but it discards information we still know: both alternatives are lists, their element types are known, and neither has arbitrary members. Earlier attempts exposed the danger of gaining permissiveness by losing this information: real errors could disappear along with the unwanted conformance errors. An ordinary union is not a substitute either; its operations must be valid across the alternatives, which does not provide the required permissiveness for invariant containers.

The goal is therefore **to preserve as much information as possible while allowing the operations justified by the surviving overloads**, rather than treating fewer diagnostics as success by itself.

## How information and error checking are preserved

The internal `OverloadResult` retains complete ordinary return alternatives, their generic arguments, the ordinary baseline type and the uncertainty kind. It lives in `TypeResult.type`, so extracting a `Type` does not silently discard the alternatives. Supported aliases and generic projections carry the alternatives forward; public queries display `OverloadResult[list[int], list[str]]` honestly rather than presenting a misleading first match or erased container. This is an internal representation, not Python annotation syntax.

Each supported operation requires **one complete successful alternative for the entire operation**. Partial successes cannot be combined across alternatives. With the example above, both invariant assignments and either int/string append are accepted, but `append(3.14)`, assignment to `list[bytes]`, a nonexistent member, or a mixed-element `extend` accepted by neither complete alternative remain errors. Different uses may select different witnesses; this is not a claim about runtime mutation correlation.

Preserving a candidate's `Type` alone is insufficient. Evaluation also carries incomplete/error/contextual/binding and overload metadata, cached results and diagnostics. Candidate checking uses ordinary evaluator validation with isolated cache and diagnostic ownership. A discarded trial must not leak false errors, and diagnostics for ordinary work whose caches survive must not disappear with that trial. The cold-query, query-order, speculative evaluation, cancellation and edit/restore coverage addresses these failure modes. A plausible return type by itself is not proof that an operation succeeded.

Concrete calls retain ordinary precision. Unsupported producers, constructors/bound-method producers, escaping uses and unsupported gradual shapes retain ordinary behavior. Admission/proof/work limits remain bounded; exhausting them is not permission to fall back to a more permissive dynamic type. This does not claim full general overload conformance.

Automatic activation is the normal ConfigOptions/Program default, including CLI, language-server and background analysis. There is no new public setting; `experimentalOverloadResults = false` remains an internal baseline control, with explicit internal controller precedence. The separate checker live-cache handoff experiment is **not adopted**: production uses `checkerHandoff: false`, and handoff requires explicit internal opt-in.

The cumulative change is substantial: 37 files, +6,459/-72 lines, including 20 production files and 15 test/sample files. It includes the original checkpoint, subsequent reviewed activation/cache/diagnostic correctness work, public type printing/transport, default wiring and documentation, not merely the latest default switch.

## Validation

- Fresh integrated validation: core, CLI and LSP test-server builds pass. Targeted: 10 suites / 856 tests. Full: 84 suites / 3,122 tests pass, with 12 existing opt-in tests skipped. Repository-supported transpile-only Jest follows a separate full TypeScript build, including tests. Scoped ESLint/Prettier pass; existing listener/deprecation/force-exit warnings remain. No integration-specific test expectations changed.
- Fresh ordinary and `--threads 2` CLI checks show the exact carrier and successful conformance assignments/appends. Four invalid-operation controls remain errors, including a mixed-element operation with no complete witness; the concrete control remains `list[int]` with its intentional argument error. Module/function complexity behavior matches main at limits 768/4096.
- Fresh three-way comparisons use exact `a07bbdea` rollback, pinned current main, and integrated feature, with identical frozen project files, dependencies, interpreters, options and current-main stubs. Frozen pandas source/wheel retain zero errors and complete matching 935/931 native diagnostics across all arms and historical controls. These are frozen inputs, not latest pandas catalog checkouts.
- Broader pinned mypy_primer coverage remains partial: 12/153 catalog projects ran, 15 prior dependency setups were TLS-blocked and were not retried, and 126 projects were not attempted. Across 981 tracked files, all 1,954 native diagnostics match across the three arms, with zero CLI/native additions or removals; 659 baseline errors remain. attrs covers its actual one-file catalog command.
- Compared with historical runs, latest-main stubs add two `message_factory` errors in packaging and pip's vendored packaging. Both also reproduce using rollback code with current stubs; neither is introduced by this feature. Thirty native replacements change only related stub-location metadata. No historical primary error disappears; complete before/after records are retained.
- All external projects and pandas controls have zero carrier activations: this establishes ordinary-fallback compatibility only, not active external or all-corpus-clean coverage. A separate default-on local conformance control activates one producer/eight carrier outcomes with no handoffs.

## Performance and limits

This deliberately trades additional checking work for retained information. Collapsing an ambiguous result to a dynamic type is cheap, but prevents subsequent checks from using the original alternatives. Here, candidate operations can repeat argument matching, member lookup and code-flow evaluation, with cache isolation and diagnostic handling. The observed active-case cost is not merely the allocation cost of storing several types; that allocation cost was not independently quantified.

To limit ordinary-path cost, activation starts from a native overload match and performs bounded eligibility/proof work before building the local admission graph. Measured natural zero-use controls avoid module scans, owned candidate trials and private restoration. This is not a zero-cost claim: controller setup and request/retirement boundaries remain. Active work is bounded, and unsupported or budget-limited cases retain ordinary behavior rather than skipping required checks.

Historical pre-integration paired measurements against the same implementation with the feature explicitly off:

| Workload | Initial analysis | Complete fixed public-query batch |
| --- | --- | --- |
| Ordinary controls | Sensitivity-limited; no equivalence claim | +0.01175 to +0.17415 ms (+3.89% to +11.63%) |
| Active-alternatives fixture | About +37.06 ms (+23.34%) | About +17.00 ms (+742%) |
| Mixed active/ordinary fixture | About +36.10 ms (+19.46%) | About +25.63 ms (+727%) |

The large active-query percentages have small ordinary baselines, but the absolute added batch costs are real. These are **whole matched workload batches, not individual queries, hover latency or IDE-wide estimates**. Ordinary initial-analysis estimates were noisy, and an additional dataclass query-batch cost versus the preceding V3 implementation was inconclusive on confirmation. No claim of exact-zero overhead or an ecosystem-wide average is supported.

These costs were accepted provisionally for continuing default-on development because retaining useful error checking is the purpose of the change. Observation counters and timing runs were separate. **No fresh integrated-tree timing was performed**: these numbers are historical measurements, not measurements or performance certification of the updated-main implementation. External projects in the compatibility run did not activate alternatives and therefore do not establish real-world active-case performance.

## Integration status

The complete feature is integrated onto pinned canonical main `916d7dc6` in a fresh worktree. The original reviewed worktree and historical artifacts are preserved.

Three additive conflict resolutions retain both upstream and feature additions in Program options, evaluator options/closure and cache-test imports. Upstream complexity configuration/API, contextual TypeForm lookup/tests, TypedDict union handling and updated typeshed are retained. The cumulative current-main-relative feature footprint remains 37 files, +6,459/-72. No domain, limit, default or handoff-policy change was made during integration. Publication targets only `microsoft/pyright:main`.


bschnurr:

packages/pyright-internal/src/analyzer/typeCacheUtils.ts
```diff
@@ -135,6 +162,24 @@ export class SpeculativeTypeTracker {
     private _speculativeTypeCache = new Map<number, SpeculativeTypeEntry[]>();
     private _activeDependentTypes: DependentType[] = [];
 
+    testOnlyEvict(nodeIds: ReadonlySet<number>) {
```
bschnurr: > **Warning** · Non-blocking recommendation

📍 packages/pyright-internal/src/analyzer/typeCacheUtils.ts:122
The default-on production path now relies on `useTestOnlyCacheIsolation`, `testOnlyEvict`, and related test-only contracts. Rename these APIs for their production guarantees and formalize the evaluator/controller isolation interface so future maintainers do not treat required production behavior as disposable test infrastructure.

[verified]

bschnurr:
Please add runtime coverage for the new type-server protocol conversion path; the current targeted tests do not exercise it.

bschnurr (approved):
Approved via Review Center.

bschnurr:
**Result:** ✅ `verified`

**Verification:** Isolated verification observed failures that were not classified as caused by this PR\: Repository discovery preflight.

**Summary:** Targeted verification passed\: 164 tests across overload-result units, controller behavior, configuration, CLI analysis, and language-server paths. The LSP test-server build also succeeded, including foreground/background and push/pull diagnostic cases. Dependencies were restored successfully with the offline frozen lockfile. An initial repository-discovery probe used an invalid empty working directory, but this was unrelated to the PR and did not prevent test execution.

**Test runs:** 4 passed, 1 failed

- :x: **Failed** | unrelated to this PR | <code>Repository discovery preflight</code> | <code>printf &#x27;%s\n&#x27; &#x27;== sandbox ==&#x27; &quot;$AUTOMATION_SANDBOX_PROFILE&quot; &#x27;== head ==&#x27; &amp;&amp; git log -1 --oneline &amp;&amp; printf &#x27;%s\n&#x27; &#x27;== changed files ==&#x27; &amp;&amp; git diff --name-status HEAD^ &amp;&amp; printf &#x27;%s\n&#x27; &#x27;== test scripts ==&#x27; &amp;&amp; node -e &quot;const p=require(&#x27;./packages/pyright-internal/package.json&#x27;); console.log(JSON.stringify(p.scripts,null,2))&quot; &amp;&amp; printf &#x27;%s\n&#x27; &#x27;== dependencies ==&#x27; &amp;&amp; if [ -x packages/pyright-internal/node_modules/.bin/jest ]; then echo package-jest-present; else echo package-jest-missing; fi &amp;&amp; if [ -d node_modules ]; then echo root-node_modules-present; else echo root-node_modules-missing; fi</code>
- :white_check_mark: **Passed** | <code>Overload-result type, cache isolation, and utility tests</code> | <code>node ./node_modules/jest/bin/jest.js src/tests/overloadResult.test.ts src/tests/testOnlyCacheTransaction.test.ts src/tests/typeCacheUtils.test.ts src/tests/typeUtils.test.ts --runInBand</code>
- :white_check_mark: **Passed** | <code>OverloadResultController targeted suite</code> | <code>node ./node_modules/jest/bin/jest.js src/tests/overloadResultController.test.ts --runInBand</code>
- :white_check_mark: **Passed** | <code>OverloadCall11 and automatic-overload configuration tests</code> | <code>node ./node_modules/jest/bin/jest.js src/tests/typeEvaluator6.test.ts src/tests/config.test.ts --runInBand --testNamePattern=&#x27;OverloadCall11|automatic overload results&#x27;</code>
- :white_check_mark: **Passed** | <code>Automatic overload-result language-server integration</code> | <code>pnpm run webpack:testserver &amp;&amp; node ./node_modules/jest/bin/jest.js src/tests/languageServer.test.ts --runInBand --testNamePattern=&#x27;automatic overload results default&#x27;</code>

<details><summary>:x: <code>Repository discovery preflight</code> diagnostic output</summary>

<pre>fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary</pre>
</details>
</details>

bschnurr (approved):

packages/pyright-internal/src/analyzer/typeCacheUtils.ts
rchiodo: Fixed in [ea2db303](https://github.com/microsoft/pyright/commit/ea2db303b8ec8d502c0d9d54f60b81c7cde34180). Renamed the production APIs to `useNodeCacheIsolation`, `canUseNodeCacheIsolation`, and `evictCacheEntries`, and added explicit `EvaluationCacheIsolation` / `EvaluationOperationController` contracts.

The contract documents exact bounded node ownership, evaluator lifetime, rollback and successful retention, native speculation/return-inference preconditions, and separate diagnostic ownership. Isolation and eviction are now installed together once by the owning evaluator. Updated the callers, regression tests, documentation, and checked tracing hooks. Actual `testOnlyExpression` and `testOnlyMemberAccess` injection options remain explicitly test-only; the production isolation guards and behavior are unchanged.

rchiodo:
Fixed the [runtime protocol-coverage request](https://github.com/microsoft/pyright/pull/11780#issuecomment-5737117027) in [ea2db303](https://github.com/microsoft/pyright/commit/ea2db303b8ec8d502c0d9d54f60b81c7cde34180).

Added a real in-process JSON-RPC `GetComputedType` test that normally activates the overload-result carrier and checks both nested generic alternatives, instance/generic flags, `any` uncertainty, the baseline reference ID, and unchanged concrete-call output. A supplemental real `ProtocolTypeFactory` test covers a nested `unknown`-origin carrier and repeated reference identity without broadening automatic admission.

Both new tests failed when the `OverloadResult` conversion branch was temporarily disabled, and passed after restoring it. There is no persistent conversion implementation change. The final local full suite passed with 3,135 tests; CI for this pushed commit is still pending.

bschnurr (approved):

PR #11784, opened 2026-09-19
Don't pick the first __init__ overload for an ambiguous Any argument

### LINKED ISSUE #11304: nullcontext: untyped or Any enter_result type is ignored and inferred as None on __enter__
**Describe the bug**
The `nullcontext` context manager from stdlib lets to pass arbitrary value to be returned on `__enter__`. It's named `enter_result`.
If the value passed is untyped or `Any`, pyright infers return type of `__enter__` as `None`. But should preserve untyped/`Any` type.

**Code or Screenshots**
```python
from contextlib import nullcontext
from typing import Any, revela_type

# also with: def test(items) -> None:
def test(items: Any) -> None:
    with nullcontext(items) as _items:
        reveal_type(_items)
        for item in _items:
            print(item)


if __name__ == '__main__':
    test([1, 2, 3])
```

**Expected behavior**
Revealed type is Any, no errors

**Observed behavior**
```sh
$ pyright ./test.py
test.py
  test.py:9:21 - information: Type of "_items" is "None"
  test.py:10:21 - error: "None" is not iterable
    "__iter__" method not defined (reportGeneralTypeIssues)
1 error, 0 warnings, 1 information
```

CLI pyright 1.1.408

Shubham-Padkonde:
Fixes #11304

When an `Any` (or Unknown) argument matches several overloads, pyright doesn't just take the first match. It compares the overloads' results and evaluates to Unknown, or Any, if they differ. For an overloaded `__init__`, the results compared are the `__init__` return types, which are always `None`. So the check always saw identical results and picked the first overload, even when the overloads construct different specializations of the class.

`contextlib.nullcontext` in typeshed is an example:

```python
@overload
def __init__(self: nullcontext[None], enter_result: None = None) -> None: ...
@overload
def __init__(self: nullcontext[_T], enter_result: _T) -> None: ...
```

With `enter_result: Any`, the first overload was chosen, the result was `nullcontext[None]`, and `with nullcontext(items) as x` inferred `x: None` ("`None` is not iterable" in the issue).

### Change

- `validateOverloadsWithExpandedTypes` now records, for each matched `__init__` overload bound to a class instance, the specialized `self` type it constructs (`MatchedOverloadInfo.specializedInitSelfType`, using the same computation the union-expansion path already used).
- `filterOverloadMatchesForAnyArgs` compares those self types instead of the `None` return types.
- When the match stays ambiguous, the self types are combined with the existing Any/Unknown dedupe logic, which is extracted into `combineAmbiguousOverloadResults` and shared with the return-type path. `nullcontext[None]` is subsumed by `nullcontext[Any]`, so `nullcontext(items)` is `nullcontext[Any]` and the `with` target is `Any`, as the issue expects.
- `validateInitMethod` doesn't re-apply the caller's constraints when the constructed self type is Unknown, because those constraints only reflect the first overload. Genuinely conflicting specializations, like `ClassA[int]` vs `ClassA[str]`, now give `ClassA[Unknown]`.

Calls that aren't ambiguous are unchanged, e.g. `nullcontext()`, `nullcontext(1)`, and overloads that construct the same specialization.

### Tests

New sample `constructor35.py`:
- `nullcontext` with `Any`, with `int`, and with no argument;
- an overloaded `__init__` that constructs `ClassA[int]` / `ClassA[str]`, with `Any`, `int` and `str` arguments;
- two overloads that construct the same specialization, so the result isn't ambiguous.

Results:
- It fails on `main` and passes with this change.
- The full `pyright-internal` jest run passes: 79 suites, 2943 tests, including all 37 `Constructor*` tests.

This change was written with help from an AI coding assistant (Claude Code). I reviewed and tested it as described above.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


bschnurr:

packages/pyright-internal/src/tests/samples/constructor35.py
```diff
@@ -0,0 +1,56 @@
...
+    def __init__(self: "ClassA[int]", x: int) -> None: ...
+
+    @overload
+    def __init__(self: "ClassA[str]", x: str) -> None: ...
+
+    def __init__(self, x: Any) -> None:
+        pass
+
+
+def func2(a: Any, b: int, c: str) -> None:
```
bschnurr, line 36: > **Warning** · Non-blocking recommendation

Could we add a checked-in case with an actual `Unknown` argument as well? The change explicitly handles both `Any` and `Unknown`, but this regression sample only protects the `Any` path.

bschnurr (approved):
Approved via Review Center.

Shubham-Padkonde:
Good catch — added in the latest commit. `constructor35.py` now has:

```python
def func4(a):
    # An unannotated parameter is Unknown rather than an explicit Any. It
    # takes the same code path, so it must not pick the first overload either.
    reveal_type(ClassA(a), expected_text="ClassA[Unknown]")
    reveal_type(ClassB(a, 1), expected_text="ClassB[int]")
```

I used an unannotated parameter as the `Unknown` source, and included the `ClassB` case too so the non-ambiguous path (both overloads specializing the same way) is covered for `Unknown` as well, not just the ambiguous one.

It does protect that path rather than riding on the `Any` case. Reverting just `constructors.ts` and `typeEvaluator.ts` to `main`:

- with `func4`: `Expected 0 errors, got 2`
- without `func4`: `Expected 0 errors, got 1`

so the new case contributes its own failure. With the fix in place the whole file passes, and `typeEvaluator6.test.ts` is `159 passed, 159 total`.

---

Disclosure: this change was written by Claude Code (Claude Opus 5) working as my agent, at my direction. The test output quoted above comes from real runs in my local environment; I am accountable for this PR.

bschnurr:
**Result:** ⚠️ `partially-verified`

**Verification:** Isolated verification observed failures that were not classified as caused by this PR\: Dependency and test discovery preflight.

**Summary:** The targeted \`typeEvaluator6\` Jest suite passed all 159 tests, including the new \`Constructor35\` test and existing overload/constructor coverage. Offline dependency installation succeeded. The initial discovery command failed only because the copied sandbox lacks Git history; this was unrelated to the PR but makes the result conservatively partially verified.

**Test runs:** 2 passed, 1 failed

- :x: **Failed** | unrelated to this PR | <code>Dependency and test discovery preflight</code> | <code>printf &#x27;profile=%s\n&#x27; &quot;${AUTOMATION_SANDBOX_PROFILE:-}&quot;; printf &#x27;node=&#x27;; node --version; printf &#x27;pnpm=&#x27;; pnpm --version; test -d node_modules &amp;&amp; echo &#x27;root_node_modules=present&#x27; || echo &#x27;root_node_modules=missing&#x27;; test -d packages/pyright-internal/node_modules &amp;&amp; echo &#x27;package_node_modules=present&#x27; || echo &#x27;package_node_modules=missing&#x27;; test -f pnpm-lock.yaml &amp;&amp; echo &#x27;pnpm_lock=present&#x27; || echo &#x27;pnpm_lock=missing&#x27;; git diff --name-status HEAD^ HEAD; git diff --unified=0 HEAD^ HEAD -- packages/pyright-internal/src/tests/typeEvaluator6.test.ts packages/pyright-internal/src/tests/samples/constructor35.py</code>
- :white_check_mark: **Passed** | <code>typeEvaluator6 Jest suite</code> | <code>node ./node_modules/jest/bin/jest.js src/tests/typeEvaluator6.test.ts --runInBand</code>
- :white_check_mark: **Passed** | <code>Offline dependency bootstrap</code> | <code>pnpm install --offline --frozen-lockfile</code>

<details><summary>:x: <code>Dependency and test discovery preflight</code> diagnostic output</summary>

<pre>profile=typescript
node=v24.15.0
pnpm=11.21.0
root_node_modules=missing
package_node_modules=missing
pnpm_lock=present
error: Could not access &#x27;HEAD^&#x27;
warning: Not a git repository.</pre>
</details>
</details>

bschnurr (approved):

packages/pyright-internal/src/tests/samples/constructor35.py
Shubham-Padkonde: Added in `ce525b3` — `func4` uses an unannotated parameter as the `Unknown` source, and covers `ClassB` too so the non-ambiguous path is exercised for `Unknown` as well. It carries its own weight: reverting the source files gives `got 2` with `func4` versus `got 1` without it.

heejaechang:
### 🔴 Pyright CLI QA — `b23dc54b` — verdict: red

11 focused check(s) completed.

**`main.py`**

```python
from typing import Generic, TypeAlias, TypeVar, assert_type

T = TypeVar("T")
U = TypeVar("U")


class Pair(Generic[T, U]):
    def __init__(self: "Pair[int, str]", value: str) -> None: ...


IntPair: TypeAlias = Pair[int, U]

# The alias fixes T but leaves U available for constructor inference.
assert_type(Pair("text"), Pair[int, str])
result = IntPair("text")
reveal_type(result)
assert_type(result, Pair[int, str])
```

**`pyrightconfig.json`**

```json
{
  "include": [
    "*.py"
  ],
  "pythonVersion": "3.13",
}
```

```text
```

**Base `916d7dc62e063ba6342493e44fc55389d42d5120`**

```json
{
  "exit_code": 0,
  "diagnostics": [
    {
      "file": "main.py",
      "severity": "information",
      "rule": null,
      "message": "Type of \"result\" is \"Pair[int, str]\"",
      "range": {
        "start": [
          15,
          12
        ],
        "end": [
          15,
          18
        ]
      }
    }
  ]
}
```

**PR `b23dc54bc808438c8401b7652a09539641b0b249`**

```json
{
  "exit_code": 1,
  "diagnostics": [
    {
      "file": "main.py",
      "severity": "information",
      "rule": null,
      "message": "Type of \"result\" is \"Pair[int, Unknown]\"",
      "range": {
        "start": [
          15,
          12
        ],
        "end": [
          15,
          18
        ]
      }
    },
    {
      "file": "main.py",
      "severity": "error",
      "message": "\"assert_type\" mismatch: expected \"Pair[int, str]\" but received \"Pair[int, Unknown]\"",
      "range": {
        "start": [
          16,
          12
        ],
        "end": [
          16,
          18
        ]
      }
    }
  ]
}
```

</details>

Shubham-Padkonde:
Reproduced and fixed in c939aaf. With overloads constructing ClassC[str, int] and ClassC[str, str], an Any/Unknown argument previously produced ClassC[bytes, bool] from unrelated generic defaults. It now preserves the shared argument and returns ClassC[str, Unknown]. The unambiguous int and str calls retain their exact specializations.

The existing overload subsumption is preserved. When it leaves conflicting specializations of the same generic class, the common arguments are retained and only differing arguments become Unknown, instead of discarding every argument and applying defaults.

Validation:
- New regressions for Any and Unknown fail on the previous head with ClassC[bytes, bool] and pass after the change.
- All eight type-evaluator suites pass: 1,224 tests, including constructor, overload and union-expansion coverage.
- TypeScript noEmit, changed-source ESLint, Prettier and whitespace checks pass.

These are local results; hosted QA on the new head is still pending. Prepared with Codex assistance.

bschnurr (approved):

Shubham-Padkonde:
Reproduced the TypeVarTuple collapse and fixed it in 5d2bf2c. Conflicting variadic arguments now remain unpacked tuples: matching tuple shapes preserve shared element types and use Unknown only for differing elements; differing shapes become an unpacked variable-length Unknown tuple instead of a single Unknown element.

The new two-element regression fails on the previous head (four diagnostics, including indexing/unpacking) and passes after the correction. Added Any/Unknown input coverage, exact overload controls, and differing pack lengths. All 159 typeEvaluator6 tests pass; TypeScript noEmit, changed-source ESLint, Prettier and whitespace checks pass. Hosted QA on this update is pending.

Prepared with Codex assistance.

bschnurr (approved):

Shubham-Padkonde:
Addressed the concrete covariant-constructor reproduction in 7e2e056. The constructed type now preserves type arguments supplied by the caller instead of replacing them with a wider specialized __init__ self type. Added coverage for Any and Unknown arguments, explicit Box[Dog] and Box[Animal], and an unparameterized call that still infers Box[Animal]. The updated sample produced three errors before the fix and passes afterward. All 159 tests in typeEvaluator6 pass; TypeScript compilation, ESLint, Prettier and git diff --check pass. Prepared and tested with Codex assistance.

bschnurr (approved):

Shubham-Padkonde:
Fixed the generic-alias regression in b23dc54. An unspecialized generic alias represented with Unknown arguments can now infer its specialization from __init__ again. Explicit caller arguments and concrete aliases retain their declared types.

Added direct/alias construction and return-value checks plus concrete and explicitly specialized covariant-alias cases. The new regression fails against the previous implementation; all 159 typeEvaluator6 tests pass with the fix, along with TypeScript compilation, ESLint, Prettier and git diff --check.

Prepared and tested with Codex assistance.

PR #11786, opened 2026-09-20
Don't report abstract instantiation when __new__ returns other classes

### LINKED ISSUE #11255: reportAbstractUsage check ignores __new__ return type
**Background**

Some classes define `__new__` so that calling the class always instantiates some specific subclass; for example, instantiating `pathlib.Path` always produces a `PosixPath` or a `WindowsPath`.

Here is an [example](https://pyright-play.net/?pyrightVersion=1.1.408&pythonVersion=3.13&strict=true&code=GYJw9gtgBA%2BjwFcAuCQFM5QJYQA5hCSgEMA7UsJYpLMUgZwChRIokBPXLUgc2zwJF0ANzTEANjA640jHPkJQQZACaRGjAMbji9elABCutAC5GUC1BVpgUUsQhoAFPTTjgASigBaAHxR6JBAzS1ClNBQQUigAIgAjYxiNUOtbOFI0AHc4J216Lz8oADEwMCgAH0NiYPMwiyxbPOx9I1cQurCmgF4lVUgAOk0ACzAsTWcAbRKwABoqkABdD1q63VdFLD0EOO013PF6OadpuaMQD2WO8MjosDiAKzRNJH70rJy85a0dPWLSp1aaA87QsqTsDmcrncBX8gRqV3QN1iwFKSW%2Ba3mAOMwJWVhs4McLjcnh8sKCILCiNQ0Xi1TRcSgPUBTmWIjEkmkzjiHiAA).

Note that you can type `__new__` as `def __new__(cls) -> Foo | Bar`, and pyright correctly understands that if you write `b = Base()`, `b`'s type is `Foo | Bar`.

**Describe the bug**

However, if your base class is abstract, pyright reports a "Cannot instantiate abstract class" error, even if the class(es) it actually instantiates are concrete. [Example](https://pyright-play.net/?pyrightVersion=1.1.408&pythonVersion=3.13&strict=true&code=GYJw9gtgBA%2BjwFcAuCQFM5QJYQA5hCSgEMA7UsJYpLMUgZwChRISAjAY2zwKIEEAQgGEANO3pIQxDkghokACzAATZuGhIAnriykA5t3yEo6AG5piAGxhbcaRjiNEppZZEaMOl4vXpQBPmgAFIJCAJQAXIxQMVAAAsRsElIycooq0bHKaMBQpMRyQfRolsBhUAC0AHxQyRFQAHRNHrFQ2blwpGgA7nBBXvTl1VAAYmBgUAA%2B-sQgUa2tWLkD2H4BxfMLWysAvCZkbhANHEpYHMEA2mNgYgEgALphmVs%2BxcZYvgic3r79lvRiILXW6zMJPLYxdAoECkKBgNgAKzQMganR6fQGT08Pz81yC6zQkWebRyeQKwWKpSGNTqxNaUNQsIARMBxkyPF5XjMQPjAkTWu0yYVKWVKjTJJstgyYVAmWxZuzGGwoHsCUEnmYLNZbME2GEgA)

**Expected result**
I think the `reportAbstractUsage` check should base its decision on `__new__`'s return type. If it returns a concrete class, or union of concrete classes, it should not flag it as an attempt to instantiate an abstract class.

Shubham-Padkonde:
Fixes #11255

## Problem

A class can define `__new__` so that calling it always produces some other concrete class. `pathlib.Path` is the canonical example — instantiating it yields a `PosixPath` or a `WindowsPath`.

pyright already understands the return type: given `def __new__(cls) -> "Foo | Bar"`, `Base()` evaluates to `Foo | Bar`. But the abstract check in `validateCallForInstantiableClass` runs before the constructor is evaluated, so if `Base` has abstract members the call is still reported:

```
error: Cannot instantiate abstract class "Base"
    "Base.method" is not implemented
```

That makes the pattern unusable with an abstract base, even though no abstract instance is ever created.

## Fix

Before reporting `reportAbstractUsage`, check whether the class's `__new__` is declared to return only instances of *other* classes that are themselves concrete. Existing behaviour is kept everywhere else:

- `__new__` returning the class itself or `Self` → still reported;
- a returned class that is itself abstract → still reported;
- overloaded `__new__` → retains the existing tolerated behavior, as the characterization test confirms;
- synthesized `__new__` (including the default `object.__new__` signature) → still reported;
- `Never` → still reported;
- `Any` return types → retain the existing tolerated behavior.

## Test

The initial `abstractClass12.py` sample covered three cases: a `Base` whose `__new__` returns `Concrete1 | Concrete2` (no error, and `reveal_type` confirms the call still evaluates to that union), a class whose `__new__` returns an abstract class, and one whose `__new__` returns itself — the last two still error, so the initial sample expected exactly 2 errors. Review follow-up expanded it with `Self`, `Never`, `Any`, and overloaded `__new__` cases. The checked-in comments explicitly characterize `Any` and overloaded `__new__` as tolerated existing behavior.

For the initial three-case sample, `main` produced 3 errors:

```
✕ AbstractClass12
    Expected 2 errors, got 3
```

With the change:

```
$ npx jest checker
Tests: 75 passed, 75 total

$ npx jest typeEvaluator
Test Suites: 8 passed, 8 total
Tests:       1223 passed, 1223 total
```

`prettier --check` is clean.

---

Disclosure: this change was written by Claude Code (Claude Opus 5) working as my agent, at my direction. The test output quoted above comes from real runs in my local environment; I am accountable for what is submitted here and will follow up on review feedback.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


bschnurr:

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -29465,6 +29471,46 @@ export function createTypeEvaluator(
 
     // Returns a list of unimplemented abstract symbols (methods or variables) for
     // the specified class.
```
bschnurr: > **Info** · Optional note

The comment describing `getAbstractSymbols` now precedes `instantiatesOnlyOtherConcreteClasses`, where it incorrectly describes the following function. Move it immediately above `getAbstractSymbols`.

[verified]

bschnurr:
**Result:** ⚠️ `needs-more-tests`

**Verification:** Isolated verification observed failures whose relationship to this PR could not be determined\: Type evaluator tests; this review is not fully verified. Isolated verification observed failures that were not classified as caused by this PR\: Dependency and test discovery.

**Summary:** The targeted abstract-class checker tests passed\: all 12 \`AbstractClass\` cases, including the new \`AbstractClass12\`, succeeded. The broader type-evaluator run exceeded the 300-second limit without producing a result. Coverage is missing for explicitly preserved cases such as \`Self\`, \`Never\`, non-class, and overloaded \`\_\_new\_\_\` return types, so confidence is limited.

**Test runs:** 1 passed, 2 failed

- :x: **Failed** | relationship unknown | <code>Type evaluator tests</code> | <code>node ./node_modules/jest/bin/jest.js typeEvaluator --runInBand</code>
- :x: **Failed** | unrelated to this PR | <code>Dependency and test discovery</code> | <code>printf &#x27;SANDBOX_PROFILE=%s\n&#x27; &quot;${AUTOMATION_SANDBOX_PROFILE:-}&quot;; node --version; pnpm --version; if [ -d node_modules ] &amp;&amp; [ -d packages/pyright-internal/node_modules ]; then echo &#x27;node_modules=present&#x27;; else echo &#x27;node_modules=missing&#x27;; fi; test -f pnpm-lock.yaml &amp;&amp; echo &#x27;pnpm-lock.yaml=present&#x27;; git status --short; git log -1 --oneline; git diff --name-status HEAD^..HEAD; node packages/pyright-internal/node_modules/jest/bin/jest.js --listTests checker typeEvaluator | grep -E &#x27;(checker|typeEvaluator)\.test&#x27;</code>
- :white_check_mark: **Passed** | <code>Abstract class checker tests</code> | <code>node ./node_modules/jest/bin/jest.js src/tests/checker.test.ts --runInBand --testNamePattern=&#x27;AbstractClass&#x27;</code>

<details><summary>:x: <code>Type evaluator tests</code> diagnostic output</summary>

<pre>Command timed out after 300s; the sandbox container was destroyed.</pre>
</details>

<details><summary>:x: <code>Dependency and test discovery</code> diagnostic output</summary>

<pre>SANDBOX_PROFILE=typescript
v24.15.0
11.21.0
node_modules=missing
pnpm-lock.yaml=present
fatal: not a git repository
Error: Cannot find module &#x27;/workspace/packages/pyright-internal/node_modules/jest/bin/jest.js&#x27;</pre>
</details>
</details>

bschnurr (approved):
Approved via Review Center.

Shubham-Padkonde:
Both points addressed in `2235d59`.

**The comment** — you were exactly right: the new function landed between the `getAbstractSymbols` comment and the function it documents. Moved back, so each comment now sits above its own function:

```
29472     // Determines whether the class's `__new__` method is declared to return
...
29476     function instantiatesOnlyOtherConcreteClasses(...)
...
29512     // Returns a list of unimplemented abstract symbols (methods or variables) for
29513     // the specified class.
29514     function getAbstractSymbols(classType: ClassType): AbstractSymbol[] {
```

**The missing coverage** — added all four cases. Two behave as the check intends, and two turned out differently from what I first assumed, so I checked them against `main` rather than writing the sample to match my expectation:

| `__new__` return type | error reported? | same on `main`? |
|---|---|---|
| `Self` | yes | yes |
| `Never` | yes | yes |
| `Any` (non-class) | **no** | **yes — also no error on `main`** |
| overloaded `__new__` | **no** | **yes — also no error on `main`** |

`Self` and `Never` are the ones this check actively preserves, and they do still report. For the `Any` and overloaded cases the function returns `false` as designed, but no diagnostic appears — because the abstract check already tolerated those before this PR. I confirmed by reverting only `typeEvaluator.ts` to `main`: that run reports 5 errors on the extended sample (the 4 above plus `Base()`, which is the bug this PR fixes), and `ReturnsNonClass()` and `OverloadedNew(1)` are absent from both lists.

So I've written those two into the sample as *not* generating an error, with a comment saying the behavior predates this change, rather than asserting errors that never fire. Expected count moved 2 → 4.

The sample still pins the fix: reverting `typeEvaluator.ts` gives `Expected 4 errors, got 5`.

**On the run that timed out in your sandbox** — I ran it here without the 300s cap:

```
node ./node_modules/jest/bin/jest.js src/tests/typeEvaluator --runInBand

Test Suites: 8 passed, 8 total
Tests:       1223 passed, 1223 total
```

Also `checker.test.ts`: `75 passed, 75 total` (all 12 `AbstractClass` cases), `npx tsc --noEmit -p .` clean, and `ESLINT_USE_FLAT_CONFIG=false npx eslint src/analyzer/typeEvaluator.ts` clean.

---

Disclosure: this change was written by Claude Code (Claude Opus 5) working as my agent, at my direction. The test output quoted above comes from real runs in my local environment; I am accountable for this PR.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
Shubham-Padkonde: Moved back in `2235d59` — the `getAbstractSymbols` comment now sits directly above that function again, and the new function carries its own comment.

packages/pyright-internal/src/tests/samples/abstractClass12.py
```diff
@@ -0,0 +1,106 @@
...
+    def __new__(cls, x: Any) -> "Concrete1 | Concrete2": ...
+
+    @abstractmethod
+    def method(self) -> int: ...
+
+
+# This does not generate an error either. An overloaded `__new__` is left
+# to the normal abstract check, which already tolerates it. That behavior
+# also predates this change.
+OverloadedNew(1)
```
bschnurr, line 157: > **Info** · Optional note

📍 packages/pyright-internal/src/tests/samples/abstractClass12.py:103

This characterization confirms that overloaded `__new__` remains tolerated, while the PR description says it remains reported. Update the PR description to match the verified behavior.

[verified]

bschnurr (approved):

packages/pyright-internal/src/tests/samples/abstractClass12.py
Shubham-Padkonde: Corrected the description to match the checked-in characterization: overloaded `__new__` remains tolerated. I also corrected the analogous `Any` claim and distinguished the original three-case sample from the expanded review tests. No code behavior changed.

heejaechang:
### 🔴 Pyright CLI QA — `e573a843` — verdict: red

16 focused check(s) completed.

#### Needs review: Pyright exits with code 2

The base returns diagnostics for this example; the PR exits with code 2.

**`main.py`**

```python
from abc import ABC, abstractmethod
from typing import Callable


class Identity:
    def __call__[T](self, fn: Callable[..., T]) -> Callable[..., T]:
        return fn


class Factory(ABC):
    def __new__(cls) -> Identity | Decorated:
        return Identity()

    @abstractmethod
    def missing(self) -> None: ...


class Decorated(Identity):
    @Factory()
    def method(self) -> int:
        return 1
```

**`pyrightconfig.json`**

```json
{
  "include": [
    "."
  ],
  "pythonVersion": "3.13",
  "reportUnnecessaryTypeIgnoreComment": "error"
}
```

```text
```

**Base `916d7dc62e063ba6342493e44fc55389d42d5120`**

```json
{
  "exit_code": 1,
  "diagnostics": [
    {
      "file": "main.py",
      "severity": "error",
      "rule": "reportAbstractUsage",
      "message": "Cannot instantiate abstract class \"Factory\"\n  \"Factory.missing\" is not implemented",
      "range": {
        "start": [
          19,
          5
        ],
        "end": [
          19,
          14
        ]
      }
    }
  ]
}
```

**PR `e573a843a92c46956e1a018ac92af3759eb5e5f9`**

```json
{
  "exit_code": 2,
  "diagnostics": []
}
```

</details>

Shubham-Padkonde:
The new CLI QA report flags a potential behavior difference but does not identify a failing example or the expected and actual diagnostics. Could you share the reproducer and CLI output so I can investigate the specific difference and add regression coverage? I will not treat this report as resolved without reproducing it.

Prepared with Codex assistance.

rchiodo:
> The new CLI QA report flags a potential behavior difference but does not identify a failing example or the expected and actual diagnostics. Could you share the reproducer and CLI output so I can investigate the specific difference and add regression coverage? I will not treat this report as resolved without reproducing it.
> 
> Prepared with Codex assistance.

Expand the report. It shows the difference.

Shubham-Padkonde:
Fixed the reported inherited __new__ regression in a051e05. Returning a base class (including object) does not establish that the constructed instance is a different concrete class, so those return types no longer bypass the abstract check. Added direct and indirect inherited __new__ cases plus an object-return case, placed before the existing Never-returning call so they are reachable. The expanded sample reports only four of the seven expected errors on the original PR and all seven after the fix. All 75 checker tests pass; TypeScript compilation, ESLint, Prettier and git diff --check pass. Existing concrete-subclass factory cases remain passing. Prepared and tested with Codex assistance.

packages/pyright-internal/src/analyzer/typeEvaluator.ts
```diff
@@ -29463,6 +29469,49 @@ export function createTypeEvaluator(
...
+        doForEachSubtype(declaredReturnType, (subtype) => {
+            if (!isClassInstance(subtype)) {
+                instantiatesOthers = false;
+                return;
+            }
+
+            // The class itself or one of its bases can still describe an instance
+            // of this abstract class, so neither proves a concrete factory result.
+            if (
+                derivesFromClassRecursive(classType, ClassType.cloneAsInstantiable(subtype), /* ignoreUnknown */ false)
```
bschnurr: > **Issue** · Please address or respond

`derivesFromClassRecursive` covers only nominal bases, so a structurally satisfied concrete `Protocol` return annotation can still suppress the diagnostic even when `__new__` returns an instance of this abstract class. Also reject return types to which an instance of `classType` is assignable, and add a protocol regression case.

[unverified]

bschnurr:
**Result:** ⚠️ `partially-verified`

**Summary:** The new \`AbstractClass12\` characterization test passed, including the concrete factory-return and retained-diagnostic cases. All 11 pre-existing abstract-class checker tests also passed. The broader type-evaluator suite timed out after 300 seconds without producing a result, so verification is partial. Offline dependency installation succeeded.

**Test runs:** 4 passed, 1 not run

- :warning: **Not run** | <code>Type evaluator tests</code> | <code>node ./node_modules/jest/bin/jest.js typeEvaluator --runInBand</code>
- :white_check_mark: **Passed** | <code>Environment and dependency discovery</code> | <code>printf &#x27;PROFILE=%s\n&#x27; &quot;$AUTOMATION_SANDBOX_PROFILE&quot;; git diff --name-status HEAD^ HEAD; if [ -d node_modules ] &amp;&amp; [ -x packages/pyright-internal/node_modules/.bin/jest ]; then echo &#x27;DEPS=ready&#x27;; else echo &#x27;DEPS=missing&#x27;; fi; node --version; pnpm --version</code>
- :white_check_mark: **Passed** | <code>Offline dependency installation</code> | <code>pnpm install --offline --frozen-lockfile</code>
- :white_check_mark: **Passed** | <code>AbstractClass12 checker test</code> | <code>node ./node_modules/jest/bin/jest.js src/tests/checker.test.ts --runInBand --testNamePattern=&#x27;^AbstractClass12$&#x27;</code>
- :white_check_mark: **Passed** | <code>Existing abstract-class checker tests</code> | <code>node ./node_modules/jest/bin/jest.js src/tests/checker.test.ts --runInBand --testNamePattern=&#x27;^AbstractClass(1|2|3|4|5|6|7|8|9|10|11)$&#x27;</code>

<details><summary>:warning: <code>Type evaluator tests</code> diagnostic output</summary>

<pre>Command timed out after 300 seconds; the sandbox container was destroyed.</pre>
</details>
</details>

Shubham-Padkonde:
The current Windows job failed on three 5-second timeouts in evaluationTrace.viewer.test.ts (viewer server) and mypyPrimerMcp.test.ts (native discovery and protocol-error). It completed 3,134 other tests successfully. These test files and the MCP preflight implementation are unchanged relative to the run's base d2a1d009.

I reran both complete test modules locally on Windows with their existing timeout settings: 31 passed, 3 skipped. Command: node ./node_modules/jest/bin/jest.js src/tests/evaluationTrace.viewer.test.ts src/tests/mypyPrimerMcp.test.ts --runInBand (with the Jest cache on D:). This does not reproduce the hosted parallel-run timeouts, so I have not increased timeouts or modified unrelated code.

Could a maintainer rerun Windows job 106851973224 in run 35758680899? The rerun API returns 403 requiring repository admin rights. Prepared and tested with Codex assistance.

Shubham-Padkonde:
Following up on the partially-verified result: the full type-evaluator suite completed locally on Windows at the unchanged PR head a051e050c1ba5fc58038ed04182de683e2a0d460.

- 8 suites passed; 1,223 tests passed; exit code 0.
- Elapsed test time: 387.996 seconds, exceeding the isolated review's 300-second limit.
- Command (from packages/pyright-internal): `node ./node_modules/jest/bin/jest.js typeEvaluator --runInBand --cacheDirectory=D:/codex-oss-work/pyright-jest-cache`.

No code or test timeout settings were changed. This completes the missing local suite verification; it does not resolve the separately reported hosted Windows timeouts or establish that the SymPy mypy_primer diagnostic changes are benign.

Verification performed with Codex assistance.

Shubham-Padkonde:
Fixed the protocol-return case in e573a84. The helper now also checks whether the declared return type can accept an instance of the abstract class, so structural compatibility cannot suppress the abstract-instantiation diagnostic. Added the supplied Readable/AbstractReader regression before the Never-return call.

The new regression fails against the prior implementation (7 errors instead of 8); the fixed implementation passes all 75 checker tests, TypeScript compilation, ESLint and formatting checks. The separate example returning integer literals is an intended concrete factory result: those integers cannot be instances of the abstract class.

Prepared and tested with Codex assistance.

PR #11787, opened 2026-09-21
Benchmark PR revisions on the same runner

bschnurr:
## Summary
- benchmark the PR base and candidate revisions sequentially on the same GitHub runner
- use the base revision's benchmark harness for both runs, with one warmup and three measured runs
- compare median time and memory from paired artifacts instead of the stale checked-in baseline
- render comments from trusted default-branch comparison code

## Testing
- `python -m unittest discover -p 'test_*.py'` from `build/benchmark`
- `pnpm run check`


bschnurr:
Superseded by #11765. The multi-run median comparison and shared base-harness improvements have been moved there in efc516edd2f8c3b8d9bf8dad459a8fe416f6aa38.

PR #11791, opened 2026-09-21
Fix class-scope PEP 695 type alias resolution under postponed annotations

### LINKED ISSUE #11789: Different scope handling of type aliases compared to other type checkers
**Describe the bug**

If same-named type alias exists in both class and module scope, references from class scope resolve to the one in the global scope.
mypy and ty use the one in class scope.

**Code or Screenshots**

```python

type A = str

class C:
    type A = int
    a: A

reveal_type(C.a)
```

|    | pyright | mypy | ty |
|--|--|--|--|
| `C.a` | str | int | int |

```
$ npx pyright qr.py 
qr.py
  qr.py:9:13 - information: Type of "C.a" is "str"
0 errors, 0 warnings, 1 information
$ mypy qr.py 
qr.py:9: note: Revealed type is "int"
$ ty check qr.py 
info[revealed-type]: Revealed type
 --> qr.py:9:13
  |
9 | reveal_type(C.a)
  |             ^^^ `int`

Found 1 diagnostic
```

command-line tool, 1.1.414

(mypy 2.3.1, ty 0.0.82)

lukiod:
Fixes #11789.

## Problem

A `type` statement declared inside a class body is not visible to an
annotation in that same class body once annotation evaluation is
postponed (PEP 649, the default starting in Python 3.14):

```python

type A = str

class C:
    type A = int
    a: A

reveal_type(C.a)  # pyright: "str" -- mypy and ty: "int"
```

At Python 3.13 and earlier this resolves correctly; the bug is specific
to postponed evaluation.

## Cause

`getTypeOfExpressionExpectingType` sets `EvalFlags.ForwardRefs` on every
annotation once `isAnnotationEvaluationPostponed` is true. That flag
makes `getTypeOfName` pass `preferGlobalScope: true` into
`lookUpSymbolRecursive`, whose PEP 563 heuristic ("if a forward
reference can be resolved in module scope, prefer that") then walks
past the correctly-found class-scope symbol to the module-scope one
with the same name.

That heuristic exists for genuine forward references, where a name
used in a deferred annotation may not be bound in an intervening scope
yet. A `type` statement's target isn't one of those: per the PEP 695
grammar it binds directly and synchronously in the scope where it
appears, the same way a normal assignment does, so it's already the
correct resolution by the time the annotation using it is evaluated.

## Fix

In the `preferGlobalScope` walk inside `lookUpSymbolRecursive`, skip it
when the already-resolved symbol has a `TypeAlias` declaration.

## Testing

Added `typeAliasStatement6.py` (Python 3.14) covering the reported
case and confirmed it fails without the fix with the same mismatch the
issue reports, and passes with it.

Ran the full test suite (`pnpm run webpack:testserver && jest`,
including the pre-existing `TypeAliasStatement1`-`5` cases and the
language-server suite, 3148 tests total): 3148 passed, 0 failed.

Branched from `a75d42d60`.


bschnurr:

bschnurr (approved):
Approved via Review Center.

heejaechang:
### ✅ Pyright CLI QA — `6ca19d4e` — verdict: **green**

20 focused check(s) completed.

rchiodo (approved):

